PR #2, opened 2022-09-08
Change the type of "Invalid VM name" to a warning.

bkhizgiy:
This PR changes the concerned type to warning as part of the controller flow, this validation is no longer blocking the migration,


ahadas:
cherry picked the changes to #3

PR #4, opened 2022-09-09
ovirt: use access token with raw http client

bennyz:
This would make the client usable with setups that have keycloak enabled, and is the standard way to access the API.

Since we cannot rely on the "exp" field return from oVirt, the current approach refreshes the client every 30 minutes (default session timeout) by requesting a new token or when the token is expired (in the event it doesn't use the default Long.MAX_VALUE).

Signed-off-by: Benny Zlotnik <bzlotnik@redhat.com>

### DIFF
--- a/pkg/controller/provider/container/ovirt/client.go
+++ b/pkg/controller/provider/container/ovirt/client.go
@@ -3,7 +3,6 @@ package ovirt
 import (
 	"crypto/tls"
 	"crypto/x509"
-	"encoding/base64"
 	"net"
 	"net/http"
 	liburl "net/url"
@@ -11,6 +10,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/go-logr/logr"
 	liberr "github.com/konveyor/controller/pkg/error"
 	libweb "github.com/konveyor/controller/pkg/inventory/web"
 	core "k8s.io/api/core/v1"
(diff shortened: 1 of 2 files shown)


pkg/controller/provider/container/ovirt/client.go
```diff
@@ -33,14 +33,34 @@ type Client struct {
 	// Raw client.
 	client *libweb.Client
 	// Secret.
-	secret *core.Secret
+	secret                *core.Secret
+	created               time.Time
+	clientTimeout         time.Duration
```
bennyz, line 38: @ahadas the question in the original PR was about clientTimeout and accessTokenExpiration?

ahadas: no, about 'created' and 'clientTimeout'

bennyz: ah ok, so I'm not sure how to unify them, I need `created` to know when the "timer" was started

ahadas: it could well be that I'm missing something - both `created` and `clientTimeout` are set in collector::New - can't we set an expiration field that is set there to Now() + clientTimeout instead?

bennyz: ah, I forgot about it, I only set it there because I thought the zero value of time.Time is not useful but it doesn't matter actually, the actual creation time is set in line 135

I changed `created` to `clientExpiration` and made it a calculated field so now the check looks nicer

pkg/controller/provider/container/ovirt/client.go
```diff
@@ -33,14 +33,34 @@ type Client struct {
...
+	AccessToken string `json:"access_token"`
+	Expiration  string `json:"exp"`
 }
 
 //
 // Connect.
 func (r *Client) connect() (err error) {
 	var TLSClientConfig *tls.Config
 
+	if time.Now().After(r.clientExpiration) {
```
ahadas, line 53: we can check here if client != nil to avoid printing the log about timeout exceeded / token expired the first time we connect

ahadas (commented):
minor comment, besides that lgtm

pkg/controller/provider/container/ovirt/client.go
bennyz: Added `!isZero` check, it should only be zero when the client was connected for the first time

ahadas: ok, also works :)

ahadas (approved):

ahadas (approved):

PR #6, opened 2022-09-13
E2E sanity ovirt test

liranr23:
The test will perform a single VM migration from existing ovirt
provider.
An existing k8s cluster with forklift installed, and a storage class
are required.
To use the CI in kind using fakeovirt, the test should be executed as
is. Otherwise the user shold provided enviromental variables:
`OVIRT_CUSTOM_ENV`, set `true` if you wish to use existing ovirt
environment.
`OVIRT_USERNAME`, the username in ovirt such as `admin@internal`.
`OVIRT_PASSWORD`, the password in ovirt.
`OVIRT_URL`, the API endpoint `https://<engine-fqdn>/ovirt-engine/api`.
`OVIRT_CACERT`, the CA certificate, located locally in a file:
`/path/to/cert`.
STORAGE_CLASS, the storage class to map the disks to.
`OVIRT_VM_ID`, the VM ID in ovirt.
`MIGRATION_TIMEOUT`, set the number of minutes to the migration completion
- by default it will be 3 minutes.

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -34,6 +34,25 @@ jobs:
         run: |
           curl -k "$CLUSTER/apis/forklift.konveyor.io/v1beta1/namespaces/konveyor-forklift/providers" --header "Authorization: Bearer $TOKEN"
 
+      # Run e2e sanity
+      - name: Run e2e sanity suite
+        env:
+          OVIRT_USERNAME: admin@internal
+          OVIRT_PASSWORD: 123456
+          OVIRT_URL: https://fakeovirt.konveyor-forklift:30001/ovirt-engine/api
+          OVIRT_CACERT: /home/runner/work/_actions/kubev2v/forkliftci/main/ovirt/e2e_cacert.cer
+          STORAGE_CLASS: standard
+          OVIRT_VM_ID: 31573c08-717b-43e0-825f-69a36fb0e1a1
+        run: |
+          GOPATH=${GITHUB_WORKSPACE}/go make e2e-sanity
+
+      # Push code coverage using Codecov Action
+      - name: Push code coverage to Codecov
+        uses: codecov/codecov-action@v3
+        with:
+          fail_ci_if_error: false # see https://github.com/codecov/codecov-action/issues/598
+
+
   build:
     # The type of runner that the job will run on
     runs-on: ubuntu-latest
--- a/tests/base_test.go
+++ b/tests/base_test.go
@@ -1,0 +1,572 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"github.com/konveyor/forklift-controller/pkg/controller/provider/container/ovirt"
+	"os"
+	"strconv"
+	"testing"
+	"time"
+
+	"github.com/go-logr/logr"
+	"github.com/konveyor/forklift-controller/pkg/apis"
+	"github.com/konveyor/forklift-controller/pkg/apis/forklift/v1beta1"
+	"github.com/konveyor/forklift-controller/pkg/apis/forklift/v1beta1/plan"
+	"github.com/konveyor/forklift-controller/pkg/apis/forklift/v1beta1/provider"
+	"github.com/konveyor/forklift-controller/pkg/apis/forklift/v1beta1/ref"
+	ovirtsdk "github.com/ovirt/go-ovirt"
+	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/kubernetes/scheme"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
+)
+
+func TestSanityOvirtProvider(t *testing.T) {
+	var log logr.Logger
+
+	conf, err := config.GetConfig()
+	if err != nil {
+		fmt.Println("unable to set up client config")
+		os.Exit(1)
+	}
+
+	logf.SetLogger(
+		logf.ZapLogger(false))
+	log = logf.Log.WithName("entrypoint")
+
+	customEnvVar := os.Getenv("OVIRT_CUSTOM_ENV")
+	customEnv := false
+	migrationTimeout := "3"
+	if customEnvVar == "true" {
+		customEnv = true
+	}
+	username := os.Getenv("OVIRT_USERNAME")
+	if username == "" {
+		t.Fatal("OVIRT_USERNAME is not set")
+	}
+	password := os.Getenv("OVIRT_PASSWORD")
+	if password == "" {
+		t.Fatal("OVIRT_PASSWORD is not set")
+	}
+
+	ovirtURL := os.Getenv("OVIRT_URL")
+	if ovirtURL == "" {
+		t.Fatal("OVIRT_URL is not set")
+	}
+
+	cacertFile := os.Getenv("OVIRT_CACERT")
+	if cacertFile == "" {
+		t.Fatal("OVIRT_CACERT is not set")
+	}
+
+	fileinput, err := os.ReadFile(cacertFile)
+	if err != nil {
+		t.Fatalf("Could not read %s", cacertFile)
+	}
+	cacert := fmt.Sprintf("%s", fileinput)
+
+	sc := os.Getenv("STORAGE_CLASS")
+	if sc == "" {
+		t.Fatal("STORAGE_CLASS is not set")
+	}
+
+	vmId := os.Getenv("OVIRT_VM_ID")
+	if vmId == "" {
+		t.Fatal("OVIRT_VM is not set")
+	}
+	timeoutInput := os.Getenv("MIGRATION_TIMEOUT")
+	if timeoutInput != "" {
+		migrationTimeout = timeoutInput
+	}
+
+	// Register
+	err = v1beta1.SchemeBuilder.AddToScheme(scheme.Scheme)
+	if err != nil {
+		t.Error(err, "Failed to build scheme")
+	}
+	err = apis.AddToScheme(scheme.Scheme)
+	if err != nil {
+		t.Error(err, "Failed to add scheme")
+	}
+
+	cl, err := client.New(conf, client.Options{Scheme: scheme.Scheme})
+	if err != nil {
+		t.Error(err, "Failed to create client")
+	}
+
+	localProviderNamespace := "konveyor-forklift"
+
+	log.Info("Creating secret...")
+
+	secret := &corev1.Secret{
+		TypeMeta: v1.TypeMeta{
+			Kind:       "Secret",
+			APIVersion: "v1",
+		},
+		ObjectMeta: v1.ObjectMeta{
+			Namespace: localProviderNamespace,
+			Name:      "ovirt-provider-test-secret",
+			Labels: map[string]string{
+				"createdForResource":     "ovirt-provider",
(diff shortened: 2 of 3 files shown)


tests/base_test.go
```diff
@@ -0,0 +1,606 @@
...
+	done = make(chan struct{})
+
+	// Wait for provider to be ready
+	statusCheck = func() {
+		err = cl.Get(context.TODO(), localProviderIdentifier, returnedLocalProvider)
+		if err != nil {
+			t.Error(err, "Could not get provider")
+		}
+
+		if returnedProvider.Status.Conditions.IsReady() {
```
liranr23, line 225: returnedProvider -> returnedLocalProvider

tests/base_test.go
```diff
@@ -0,0 +1,606 @@
...
+
+	err = cl.Create(context.TODO(), plan, &client.CreateOptions{})
+	if err != nil {
+		t.Fatal(err, "Failed to create the plan")
+	}
+	done = make(chan struct{})
+
+	returnedPlan := &v1beta1.Plan{}
+	planIdentifier := types.NamespacedName{Namespace: plan.Namespace, Name: plan.Name}
+	// Wait for provider to be ready
```
liranr23, line 439: provider -> plan

tests/base_test.go
```diff
@@ -0,0 +1,606 @@
...
+	defer cl.Delete(context.TODO(), migration, &client.DeleteOptions{})
+
+	err = cl.Create(context.TODO(), migration, &client.CreateOptions{})
+	if err != nil {
+		t.Fatal(err, "Failed to create the migration")
+	}
+
+	returnedMigration := &v1beta1.Migration{}
+	migrationIdentifier := types.NamespacedName{Namespace: migration.Namespace, Name: migration.Name}
+	// Wait for provider to be ready
```
liranr23, line 489: provider -> migration

tests/base_test.go
```diff
@@ -0,0 +1,606 @@
...
+	if err != nil {
+		t.Fatalf("Could not read %s", cacertFile)
+	}
+
+	sc := os.Getenv("STORAGE_CLASS")
+	if sc == "" {
+		t.Fatal("STORAGE_CLASS is not set")
+	}
+
+	vmId := os.Getenv("OVIRT_VM")
```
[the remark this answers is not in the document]
liranr23: yes, i wrote it in the commit itself in a TODO. also to clean the code, and maybe check other validations (on the plan status + check the vm existence in the namespace).

tests/base_test.go
```diff
@@ -0,0 +1,606 @@
...
+			Namespace: namespaceName,
+			Name:      "ovirt-provider-test-secret",
+		},
+		Data: map[string][]byte{
+			"user":     []byte(username),
+			"password": []byte(password),
+			"cacert":   cacert,
+		},
+		Type: corev1.SecretTypeOpaque,
+	}
```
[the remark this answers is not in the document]
liranr23: you mean:
migration, plan and vmID?
why would we need them here?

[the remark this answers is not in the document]
ahadas: no, I mean create for providers, and created for provider type - just create a provider through the UI and you'll see them in the created secret

tests/base_test.go
```diff
@@ -0,0 +1,606 @@
...
+	err = cl.Create(context.TODO(), secret, &client.CreateOptions{})
+	if err != nil {
+		t.Fatal(err, "Failed to create secret")
+	}
```
ahadas, line 127: it would be nice to replace those with `Expect(err).ToNot(HaveOccurred())` and in general, use gomega constructs to get to more readable tests - see [this](https://github.com/kubevirt/kubevirt/blob/main/tests/vmipreset_test.go#L109-L138) for instance

tests/base_test.go
```diff
@@ -0,0 +1,606 @@
...
+	migrationIdentifier := types.NamespacedName{Namespace: migration.Namespace, Name: migration.Name}
+	// Wait for provider to be ready
+	statusCheck = func() {
+		err = cl.Get(context.TODO(), migrationIdentifier, returnedMigration)
+		if err != nil {
+			t.Error(err, "Could not get migration")
+		}
+
+		if returnedMigration.Status.Conditions.IsReady() {
+			log.Info("Migration is ready")
```
rokkbert, line 497: In my KinD-tests this is the last log message, nothing visible happens afterwards until the test gets terminated by the test-timeout (even if that is set to 30 minutes).

tests/base_test.go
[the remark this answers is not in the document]
ahadas: > Why is it better to get a template? In any case something needs to be there already?

either a template or an image - it would save the time to install the vm (by using a pre-installed image in both cases)

tests/base_test.go
```diff
@@ -0,0 +1,649 @@
...
+		TypeMeta: v1.TypeMeta{
+			Kind:       "Namespace",
+			APIVersion: "v1",
+		},
+		ObjectMeta: v1.ObjectMeta{
+			Name: namespaceName,
+		},
+	}
+
+	defer cl.Delete(context.TODO(), namespace, &client.DeleteOptions{})
```
ahadas, line 131: it seems like it's a preparation for running several tests on the same k8s instance but we create the instance per test, right? it's not the cleanup part that bothers me but the fact we're diverging from a typical flow in which providers for instance are created on the same namepsace that the controller and other forklift components run on

tests/base_test.go
```diff
@@ -0,0 +1,649 @@
...
+			if !ok {
+				t.Fatal("Failed to get network id")
+			}
+			pair := v1beta1.NetworkPair{
+				Source: ref.Ref{ID: networkId},
+				Destination: v1beta1.DestinationNetwork{
+					Type: "pod",
+				},
+			}
+			nicPairs = append(nicPairs, pair)
```
ahadas, line 367: so we'll map all networks to the pod network? it doesn't seem to reflect a real world scenario..
but I see we also map all storages to the specified one, so we can keep it this way for now and change when we'll design the other cases to test

tests/base_test.go
```diff
@@ -0,0 +1,649 @@
...
+		},
+	}
+
+	log.Info("Creating Network Map...")
+	err = cl.Create(context.TODO(), networkMap, &client.CreateOptions{})
+	if err != nil {
+		t.Fatal(err, "Failed to create networkmap")
+	}
+
+	planCustomResource := &v1beta1.Plan{
```
ahadas, line 429:
```suggestion
	plan := &v1beta1.Plan{
```
I don't think the CustomResource postfix adds any value here

ahadas (commented):
nice
using ginkgo, gomega would be done separately?

tests/base_test.go
liranr23: the initial thought was - we should create a vm, migrate and then cleanup everything. currently we do create instance per test, so it's fine to be removed. later on, we may think of re-usage and therefore some cleanup.

.github/workflows/pull_request.yml
```diff
@@ -34,6 +34,18 @@ jobs:
         run: |
           curl -k "$CLUSTER/apis/forklift.konveyor.io/v1beta1/namespaces/konveyor-forklift/providers" --header "Authorization: Bearer $TOKEN"
 
+      # Run e2e
+      - name: Run e2e suite
+        run: |
+          GOPATH=${GITHUB_WORKSPACE}/go make e2e
+
+      # Push code coverage using Codecov Action
+      - name: Push code coverage to Codecov
```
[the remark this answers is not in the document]
liranr23: i think it's needed since those are different jobs

liranr23:
> nice using ginkgo, gomega would be done separately?

yes

tests/base_test.go
ahadas: I see, so let's take it step by step then - if we don't need that extra namespace now, we can use the namespace of the forklift components. when we get to the point we execute multiple tests on the same k8s instance, we can discuss if that would be the right way to go - I think the benefit of imitating what happens in real deployment would be great than removing the namespace

tests/base_test.go
ahadas: agreed we'll to that separately

tests/base_test.go
liranr23: i actually prefer to have the extra namespace and remove the namespace cleanup. it's easier to follow afterwards, if you wish to run locally you won't need to remove every resource, you will just remove the namespace.

liranr23: because you will know you have everything (mappings, plan, migration, vm) under the new namespace. if you wish to do another test on the environment just delete the namespace and run the test again.

ahadas: ah I see, the comma characters made it look like you list several reasons and I thought you meant that it would be easier to follow the result of the tests after their execution while what you meant was to follow up with execution the same test (or other test that may conflict with the previous one) on the same environment..

I see your point but I see much greater value in being aligned with how the system works in production. I think we can address the use cases you mentioned differently, e.g., if the test creates a provider then I'd expect that all those entities would be removed when removing the provider - that way we exercise real-world scenarios rather than removing a namespace of plans/migrations/mapping/etc - a scenario that shouldn't really happen..

tests/base_test.go
```diff
@@ -0,0 +1,649 @@
...
+		if vmId == "" {
+			t.Fatal("OVIRT_VM is not set")
+		}
+		timeoutInput := os.Getenv("MIGRATION_TIMEOUT")
+		if timeoutInput != "" {
+			migrationTimeout = timeoutInput
+		}
+	} else {
+		username = "admin@internal"
+		password = "123456"
```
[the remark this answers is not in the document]
liranr23: done, we only miss the nic and sd IDs from ovirtSDK. we need to resolve the secure connection first and then we can drop it.

.github/workflows/pull-request.yml
```diff
@@ -34,6 +34,25 @@ jobs:
...
+          STORAGE_CLASS: standard
+          OVIRT_VM_ID: 31573c08-717b-43e0-825f-69a36fb0e1a1
+        run: |
+          GOPATH=${GITHUB_WORKSPACE}/go make e2e-sanity
+
+      # Push code coverage using Codecov Action
+      - name: Push code coverage to Codecov
+        uses: codecov/codecov-action@v3
+        with:
+          fail_ci_if_error: false # see https://github.com/codecov/codecov-action/issues/598
```
ahadas, line 53: I don't see out previous thread on this part but I remember seeing that you say it's on a different job - still, so why do we need to post this twice? maybe I'm missing something but this part is under ci_setup although it's not about setup, should we replace `Run test suite` below with this?

tests/base_test.go
ahadas: @liranr23 doesn't happen anymore, right?

ahadas (commented):
looks good!
commented on one thread that should be resolved if it's not reproduced anymore and a question about the configuration of the github action

tests/base_test.go
liranr23: when executing you won't see live logging. if the test passes, you will see `ok` and that's it. in case of failure (error, timeout), you will see the whole logs and the error.

.github/workflows/pull-request.yml
liranr23: the `Run test suite` is under `build` job. you can see the steps we are doing there. we *don't* deploy kind of anything like that so i don't think it's suitable to the test. maybe i'm missing something but this is what i understand.

tests/base_test.go
ahadas: ok so based on that, the test doesn't time out anymore for us and we can consider this worksforme

.github/workflows/pull-request.yml
ahadas: I see, that makes sense - so how about having a better differentiation between unit tests and integration (e2e) tests here and in the makefile? it would make it a bit more clear..

ahadas (approved):
ok, let's get this one in and post follow up patches to improve it if we want to

PR #9, opened 2022-09-14
migrate host validation: change vmware fail error msg

mnecas:
Migration of: https://github.com/konveyor/forklift-controller/pull/445
Bz: https://bugzilla.redhat.com/show_bug.cgi?id=1958826
Issue:
When selecting the migration network on the VMware host and entering the wrong credentials the thrown error is unclear.

### DIFF
--- a/pkg/controller/host/validation.go
+++ b/pkg/controller/host/validation.go
@@ -86,7 +86,7 @@ func (r *Reconciler) validate(host *api.Host) error {
 			Status:   True,
 			Reason:   Completed,
 			Category: Advisory,
-			Message:  "The host has been validated.",
+			Message:  "The host inventory reference, IP address, and credentials have been validated.",
 		})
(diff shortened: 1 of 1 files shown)


nyoxi (approved):
Typo `throwen` -> `thrown` in the commit message. With that LGTM.

pkg/controller/host/validation.go
```diff
@@ -86,7 +86,7 @@ func (r *Reconciler) validate(host *api.Host) error {
 			Status:   True,
 			Reason:   Completed,
 			Category: Advisory,
-			Message:  "The host has been validated.",
+			Message:  "The host inventory reference, IP address, and credentials have been validated.",
```
ahadas, line 89: (most of) those should be validated using an admission webhook so we won't need this condition..

ahadas (approved):

PR #10, opened 2022-09-14
Bazel: Add push-forklift-validation

mnecas:
The bazel is missing the push of forklift-validation image.


BUILD.bazel
```diff
@@ -58,3 +58,23 @@ container_push(
     repository = "openshift/forklift-controller",
     tag = "devel",
 )
+
+container_push(
+    name = "push-forklift-validation",
+    format = "Docker",
+    image = "//validation:forklift-validation-image",
+    registry = "image-registry.openshift-image-registry.svc:5000",
```
ahadas, line 66: we talked offline that it didn't work for me when I tried it with openshift's image registry. I didn't investigate it too much and seems I merged `container_push` for forklift-controller by mistake. please check that, if it doesn't work then if we don't parameterize the registry name, we should default it to something different

mnecas: How about defaulting it to quay as most of the developers use it or the local registry which is for every devel diffrent?

nyoxi: Could you please elaborate on "we talked offline that it didn't work for me"? Which one didn't work for you, the internal (`image-registry...`) or external (`default-route...`) URL?

I think we either want targets for maintainers so they can easily push to upstream repos (do we have any?) and these can be fixed. Or we want something that could be easily parametrized (quay or openshift registry? which namespace? ...).

ahadas: > Could you please elaborate on "we talked offline that it didn't work for me"? Which one didn't work for you, the internal (`image-registry...`) or external (`default-route...`) URL?

the internal registry

> 
> I think we either want targets for maintainers so they can easily push to upstream repos (do we have any?) and these can be fixed. Or we want something that could be easily parametrized (quay or openshift registry? which namespace? ...).

sure, if we want these targets to push to quay that's ok, but need to check it works

mnecas: Updated the PR with the env variables with default quay and that the user needs to specify the account name so somebody does not by accident pushes to the konveyor or openshift organization

rokkbert (approved):
LGTM

PR #13, opened 2022-09-20
Bazel: Set opa version in WORKSPACE

mnecas:
Once the opa releases new version the latest binary will not have matching hash and it would cause ci failure.


rokkbert (approved):
Yes, that is the new hash and the current version.

PR #14, opened 2022-09-21
Automation: Add kind_setup using forkliftci repo

mnecas:
Integration of forkliftci repositry to the github action of forklift repo with bazel.


rokkbert (changes requested):

.github/workflows/pull_request.yml
```diff
@@ -14,6 +14,51 @@ on:
 
 # A workflow run is made up of one or more jobs that can run sequentially or in parallel
 jobs:
+  kind_setup:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out forklift repository
+        uses: actions/checkout@v3
+
+      - name: Checkout forkliftci
```
michalskrivanek, line 23: this eventually moves here, right? I do not see a need for separate repo. For now that's ok

.github/workflows/pull_request.yml
```diff
@@ -14,6 +14,51 @@ on:
...
+          key: ${{ runner.os }}-bazel-cache-${{ hashFiles('WORKSPACE') }}
+
+      - name: Add cwd to path for kubectl.
+        run: echo `pwd` >> $GITHUB_PATH
+
+      - name: Get kubectl
+        run: curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl"
+      - run: chmod u+x kubectl
+
+      - name: Build and setup everything with bazel
```
michalskrivanek, line 43: this would be best to move to its own action so it's not so ugly here:)

michalskrivanek: in that case a separate repo would be appropriate, not to check out but as a home for this action

.github/workflows/pull_request.yml
mnecas: Either the custom action or move the ci

rokkbert: I see the forkliftci-repo as more of an experimental playground which has stuff which we might not need at the end.

.github/workflows/pull_request.yml
```diff
@@ -14,6 +14,51 @@ on:
 
 # A workflow run is made up of one or more jobs that can run sequentially or in parallel
```
rokkbert, line 15: In our case it has to be sequentially. We can't test things until the are running.

.github/workflows/pull_request.yml
rokkbert: The action doesn't have a nice name, but it does most of the work, including building forklift. Not sure where else it should go and why?

.github/workflows/pull_request.yml
```diff
@@ -14,6 +14,51 @@ on:
...
+      - run: kubectl get pods -n konveyor-forklift
+
+      - run: echo "CLUSTER=`kind get kubeconfig | grep server | cut -d ' ' -f6`" >> $GITHUB_ENV
+      - run: echo "TOKEN=`kubectl get secrets -n kube-system -o jsonpath='{.items[0].data.token-id}' | base64 -d`.`kubectl get secrets -n kube-system -o jsonpath='{.items[0].data.token-secret}' | base64 -d`" >> $GITHUB_ENV
+
+      - name: Test call to Forklift
+        run: |
+          curl -k "$CLUSTER/apis/forklift.konveyor.io/v1beta1/namespaces/konveyor-forklift/providers" --header "Authorization: Bearer $TOKEN"
+
   # This workflow contains a single job called "build"
```
rokkbert, line 62: This is going to "build" only the test, and run it, so this name is a bit misleading. (Also, the whole line is probably copy&paste, because _this_ workflow was 2 jobs :) )

.github/workflows/pull_request.yml
```diff
@@ -14,6 +14,51 @@ on:
 
 # A workflow run is made up of one or more jobs that can run sequentially or in parallel
 jobs:
+  kind_setup:
```
rokkbert, line 17: It's hard to find a good name, but this seems to "small". It's not just about the kind cluster but about building + deploying forklift...

.github/workflows/pull_request.yml
michalskrivanek: "action" - as in e.g. https://github.com/oVirt/upload-rpms-action. That would be a best place to isolate and hide the complexity of bringing up kind environment with local repo.
then you just use it here

.github/workflows/pull_request.yml
```diff
@@ -14,6 +14,51 @@ on:
...
+        run: |
+          cd forkliftci
+          FORKLIFT_DIR=${{ github.workspace }} ./build_and_setup_everything_bazel.sh
+          cd ..
+
+      - run: kubectl version
+
+      - run: kubectl wait deployment -n konveyor-forklift forklift-controller --for condition=Available=True --timeout=180s
+
+      - run: kubectl get pods -n konveyor-forklift
```
rokkbert, line 53: This is an optional debug-step, we might not need it, once things are known to generally work. (It's not expensive, of course, but causes clutter.)

.github/workflows/pull_request.yml
mnecas: sooo just `ci_setup` or `setup` to make it more obscure? Don't want to list everything this name should be general

.github/workflows/pull_request.yml
rokkbert: I see, yes, we could split it up into reusable parts.

.github/workflows/pull_request.yml
rokkbert: Yes. That's why I named it "do it" before ;-)

.github/workflows/pull_request.yml
liranr23: if it shows the output somewhere, it won't harm.
what will happen if we drop this line and we just get the timeout in the line above? or better say - how we fail if the deployment fails?

rokkbert: Good question! I don't know. It depends on the failure. The "get pods" can't fail, AFAIK, it can just return no pods.

liranr23: OK, than it also grants us a hint in such case. Can we leverage it to know if it's empty, fail with error?

rokkbert: Yes, we can. For example:
     - run: echo "RUNNING_FORKLIFT_PODS=$(kubectl get pods -A | grep forklift | grep Running | wc | tr -s ' ' | cut -f 2 -d ' ')" >> $GITHUB_ENV
      - name: verify_four_pods
        if: ${{ env.RUNNING_FORKLIFT_PODS }} != "4"
        uses: actions/github-script@v3
        with:
          script: |
            core.setFailed('Wrong number of running forklift pods.')

.github/workflows/pull_request.yml
rokkbert: Or we just use https://github.com/marketplace/actions/kubernetes-kind-cluster which looks like it already does what we need.

mnecas:
I have updated the changes according to @rokkbert.
The rest are suggestions for the future. I would merge this PR so we can start testing as much as possible and the rest (like adding the custom action) we can do later.

.github/workflows/pull_request.yml
```diff
@@ -12,9 +12,30 @@ on:
...
 jobs:
-  # This workflow contains a single job called "build"
+  ci_setup:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out forklift repository
+        uses: actions/checkout@v3
+
+      - name: Build and setup everything with baze
+        uses: kubev2v/forkliftci@main
```
mnecas, line 24: I have created a custom action for the setup of forklift ci https://github.com/kubev2v/forkliftci/pull/6/.

.github/workflows/pull_request.yml
```diff
@@ -12,9 +12,28 @@ on:
...
+# A workflow run is made up of one or more jobs that can run sequentially
 jobs:
-  # This workflow contains a single job called "build"
+  ci_setup:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out forklift repository
+        uses: actions/checkout@v3
+
+      - name: Build and setup everything with baze
```
rokkbert, line 23: baze -> bazel

rokkbert (approved):

liranr23:
> I have updated the changes according to @rokkbert. The rest are suggestions for the future. I would merge this PR so we can start testing as much as possible and the rest (like adding the custom action) we can do later.

ok, hopefully nobody will forget them :)

liranr23 (approved):

PR #15, opened 2022-09-22
Add phase status

yaacov:
Add phase status to Ptoviders CRD.

K8S resources uses `status.phase` fields to help users follow the resources life cycle from creation to deletion.
the `phase` helps users to easily catch resource that stray into sad paths or get stuck.
`phase`s do not replace `conditions`, `conditions` role is to report the state of the resource while `phase` report the place of the resource during it's life voyage.
Ref: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/

Issue:
The UI want to show the resource stage in it's life cycle as a way to help users easily glance the happiness of there cluster. Currently providers only report it's `conditions`, and require the users to guess the stage of the providers using this conditions.

Proposed solution:
Add `phase` field to the providers status.

Screenshot:
mtv-phase

### DIFF
--- a/pkg/apis/forklift/v1beta1/provider.go
+++ b/pkg/apis/forklift/v1beta1/provider.go
@@ -75,6 +75,9 @@ type ProviderSpec struct {
 //
 // ProviderStatus defines the observed state of Provider
 type ProviderStatus struct {
+	// Current life cycle phase of the provider.
+	// +optional
+	Phase string `json:"phase,omitempty"`
 	// Conditions.
 	libcnd.Conditions `json:",inline"`
 	// The most recent generation observed by the controller.
@@ -88,6 +91,7 @@ type ProviderStatus struct {
 // +k8s:openapi-gen=true
 // +kubebuilder:subresource:status
 // +kubebuilder:printcolumn:name="TYPE",type="string",JSONPath=".spec.type"
+// +kubebuilder:printcolumn:name="STATUS",type="string",JSONPath=".status.phase"
 // +kubebuilder:printcolumn:name="READY",type=string,JSONPath=".status.conditions[?(@.type=='Ready')].status"
 // +kubebuilder:printcolumn:name="CONNECTED",type=string,JSONPath=".status.conditions[?(@.type=='ConnectionTestSucceeded')].status"
 // +kubebuilder:printcolumn:name="INVENTORY",type=string,JSONPath=".status.conditions[?(@.type=='InventoryCreated')].status"
--- a/pkg/controller/provider/validation.go
+++ b/pkg/controller/provider/validation.go
@@ -56,6 +56,15 @@ const (
 )
 
 //
+// Phases
+const (
+	ValidationFailed = "ValidationFailed"
+	ConnectionFailed = "ConnectionFailed"
+	Ready            = "Ready"
+	Staging          = "Staging"
+)
+
+//
 // Statuses
 const (
 	True  = libcnd.True
@@ -112,6 +121,7 @@ func (r *Reconciler) validateType(provider *api.Provider) error {
 		}
 	}
 
+	provider.Status.Phase = ValidationFailed
 	provider.Status.SetCondition(
 		libcnd.Condition{
 			Type:     TypeNotSupported,
@@ -131,6 +141,7 @@ func (r *Reconciler) validateURL(provider *api.Provider) error {
 		return nil
 	}
 	if provider.Spec.URL == "" {
+		provider.Status.Phase = ValidationFailed
 		provider.Status.SetCondition(
 			libcnd.Condition{
 				Type:     UrlNotValid,
@@ -142,6 +153,7 @@ func (r *Reconciler) validateURL(provider *api.Provider) error {
 	}
 	_, err := url.Parse(provider.Spec.URL)
 	if err != nil {
+		provider.Status.Phase = ValidationFailed
 		provider.Status.SetCondition(
 			libcnd.Condition{
 				Type:     UrlNotValid,
@@ -174,6 +186,7 @@ func (r *Reconciler) validateSecret(provider *api.Provider) (secret *core.Secret
 	}
 	ref := provider.Spec.Secret
 	if !libref.RefSet(&ref) {
+		provider.Status.Phase = ValidationFailed
 		provider.Status.SetCondition(newCnd)
 		return
 	}
@@ -187,6 +200,7 @@ func (r *Reconciler) validateSecret(provider *api.Provider) (secret *core.Secret
 	if errors.IsNotFound(err) {
 		err = nil
 		newCnd.Reason = NotFound
+		provider.Status.Phase = ValidationFailed
 		provider.Status.SetCondition(newCnd)
 		return
 	}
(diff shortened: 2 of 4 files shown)


yaacov:
@sjd78 @rszwajko @bennyz @mnecas @bkhizgiy @ahadas please review

yaacov:
cc:// @mturley @vconzola

mnecas (changes requested):
Please remove the `autogenerated mainifests` commit in favour of https://github.com/kubev2v/forklift/pull/16

yaacov:
@mnecas thanks, I will rebase after #16 is merged

yaacov:
@mnecas rebased, please re-review

pkg/apis/forklift/v1beta1/provider.go
```diff
@@ -75,6 +75,9 @@ type ProviderSpec struct {
 //
 // ProviderStatus defines the observed state of Provider
 type ProviderStatus struct {
+	// Current life cycle phase of the provider.
+	// +optional
+	Phase string `json:"phase,omitempty"`
```
mnecas, line 80: Are you able to access this value also from `kubectl`? If I remember correctly you might need to add the `kubebuilder:printcolumn:name` right above the `Provider struct`
Something like 
```
// +kubebuilder:printcolumn:name="PHASE",type="string",JSONPath=".status.phase"
```

mnecas (changes requested):

pkg/apis/forklift/v1beta1/provider.go
yaacov: Thanks, missed that ( didn't know you can do that :-) )
Added the phase column before the conditions

```bash
oc get providers
NAME   TYPE    PHASE   READY   CONNECTED   INVENTORY   URL                                                          AGE
t1     ovirt   Ready   True    True        True        https://vm-10-173.lab.eng.tlv2.redhat.com/ovirt-engine/api   6m48s
```

yaacov:
@mnecas does it makes sense to try and remove the formatting stuff from this PR ?

pkg/controller/provider/validation.go
```diff
@@ -55,6 +55,15 @@ const (
...
 )
 
+//
+// Phases
+const (
+	ValidationFailed = "ValidationFailed"
+	ConnectionFailed = "ConnectionFailed"
+	Ready            = "Ready"
+	Staging          = "Staging"
+)
```
rszwajko, line 65: I'm missing here 2 states that are used in the UI:
1. `pending`  - displayed as a spinner with message "Loading the inventory"
2. `running` (in the sense that there are running migrations) - used indirectly in the UI i.e. you cannot remove or edit a  provider that is being used by a running migration. This information is not available from CLI which may lead to problems. For details see https://github.com/konveyor/forklift-ui/blob/main/pkg/web/src/app/Providers/components/ProvidersTable/ProviderActionsDropdown.tsx#L40

Some questions to the provider lifecycle:
1.  does the provider need to stay connected all the time to be useful? if the connection is used to synchronize then perhaps we could rephrase the condition to reflect that we have (potentially) outdated/obsolete inventory data
2.  is it possible to have an invalid provider in the system? if not then validation is part of initialization flow (`staging`)

pkg/apis/forklift/v1beta1/provider.go
```diff
@@ -88,6 +91,7 @@ type ProviderStatus struct {
 // +k8s:openapi-gen=true
 // +kubebuilder:subresource:status
 // +kubebuilder:printcolumn:name="TYPE",type="string",JSONPath=".spec.type"
+// +kubebuilder:printcolumn:name="PHASE",type="string",JSONPath=".status.phase"
```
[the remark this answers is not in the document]
yaacov: Thanks, makes sense if pods use it, we can too

pkg/controller/provider/validation.go
```diff
@@ -112,6 +121,7 @@ func (r *Reconciler) validateType(provider *api.Provider) error {
 		}
 	}
 
+	provider.Status.Phase = ValidationFailed
```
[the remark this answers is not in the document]
yaacov: `phase` is independent than `conditions`, it's ok for `phase` to change when `conditions` are stable, and it's ok for `phase` to stay the same when `conditions` change, `phase` show information about the resource life cycle, while `conditions` show information about the resource state.
https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/

the resource life cycle can be:
"positive" life cycle:
`... -> Staging -> Ready -> ...`
"Negative" life cycle:
`... -> Staging -> ValidationFailed / ConnectionFailed -> ...`

It should not flicker, @rszwajko in what cases do you see a potential for flickering, am I missing some flow ?

pkg/controller/provider/validation.go
yaacov: `phase` is not state, it is not replacing debugging by looking up the conditions when the resource is on a "negative" flow,
it can be used to quickly filter the resources that are on a "negative" flow, and then fetch there conditions for more information.

> `pending` - displayed as a spinner with message "Loading the inventory"

`pending` can be roughly approximated by `Staging` as it indicate a transitional state that should turn to `Ready` or some fail phase, a `spinner` can be a good icon for this phase, because it should be temporary state.

> `running` (in the sense that there are running migrations) - used indirectly in the UI 

`running` in the meaning of "we currently have a running plan that use this provider", is more problematic, the provider resource does not store this information, (e.g. plans do not tag/un-tag a provider when they start/stop), I'm ok with not adding this life phase of a provider (e.g. `currently used by a running plan` phase), we may want to add in some future planning adding a feature for plans, that they will tag the provider somehow when running.
@rszwajko  we can continue the current behavior of fetching this information from the plans list, like it's done in the example you provided, in the example we do not use the provider state, we fetch all plans and then filter them to find the once that are running and using this provider.
https://github.com/konveyor/forklift-ui/blob/main/pkg/web/src/app/Providers/components/ProvidersTable/ProviderActionsDropdown.tsx#L40

yaacov: > does the provider need to stay connected all the time to be useful? if the connection is used to synchronize then perhaps > we could rephrase the condition to reflect that we have (potentially) outdated/obsolete inventory data

At this moment a provider is always connected, it is currently debated whether it's a good design feature:
pros: 
a user can know the provider is connected before starting a plan (and not just that it was connected at some point)
a UI user can have an up to date information about the cluster
cons:
lots of unneeded net traffic

(*) AFAIK, no work on that
 
> is it possible to have an invalid provider in the system? if not then validation is part of initialization flow (staging)

At this moment yes, as demonstrated in this PR description, it is currently debated whether it's a good design feature:
pros:
a user can create providers that are currently not reachable, and then connect them just when starting migration.
cons:
a user can't confirm the information (e.g. passwords, certs) are correct and may end up redoing the providers creation once the provider clusters come online.

(*) AFAIK, we do have PR, creating a web hook that will change this behavior, and will not allow creation of invalid providers.

cc:// @ahadas @michalskrivanek

yaacov:
@mnecas @rszwajko, hi,

a. I removed the formatting changes
b. changed the column title from `PHASE` to `STATUS` as suggested by @rszwajko , reasoning is to be similar to the way pods show the `status.phase` under `STATUS`

please re-review

rszwajko (approved):

yaacov:
@mnecas can we merge this ?

mnecas:
Lets merge it after the CI stuff https://github.com/kubev2v/forklift/pull/17 or please revet the formatting

PR #17, opened 2022-09-29
Run make vet and make fmt

yaacov:
When running `make run` on current code, go will re-format the code.
This PR, re set the code to conform to go format 1.18.

Running formatting with go v1.18

``` bash
go version
go version go1.18.6 linux/amd64

make vet
make fmt
```

Signed-off-by: yzamir <yzamir@redhat.com>


yaacov:
@mnecas @ahadas @bennyz @bkhizgiy @liranr23 please review

pkg/forklift-api/webhooks/validating-webhook/admitters/secret-admitter.go
```diff
@@ -51,27 +51,27 @@ func (admitter *SecretAdmitter) Admit(ar *admissionv1.AdmissionReview) *admissio
 	secret := &core.Secret{}
 	err := json.Unmarshal(raw, secret)
 	if err != nil {
-//		reviewResponse := admissionv1.AdmissionResponse{}
-//		reviewResponse.Allowed = false
-//		return &reviewResponse
+		//		reviewResponse := admissionv1.AdmissionResponse{}
+		//		reviewResponse.Allowed = false
+		//		return &reviewResponse
```
nyoxi, line 56: This looks little bit weird now and could be reformatted or removed (what do you think @mnecas?) in a separate patch.

nyoxi (approved):

yaacov:
> I don't see any good solution here, but having a consistent code formatting has its benefits. We should at least try to summarize our position on this.

I also would like some rules I can follow. or some tests that will fail when a PR not formated, ( e.g. format test fail, format using .... ) , 
p.s.
@ahadas, bazel is using go 1.6  ?
(*) https://github.com/kubev2v/forklift/blob/main/WORKSPACE#L3357

mnecas:
How about to use some tool like https://github.com/golangci/golangci-lint in the ci?

mnecas (approved):
LGTM, this PR just reformats so we don't have this mess in other PRs.
The other lint stuff can be added later in other PRs.

PR #18, opened 2022-10-05
Fix digest ansible-operator-image

mnecas:
Previously we used the digest of v1.22 which got changed with every new release of 1.22.Z. Now we will use the v1.22.0 which hopefully will no longer be overwritten.


mnecas:
Fixes issue with the CI https://github.com/kubev2v/forkliftci/pull/5

PR #20, opened 2022-10-24
Add bazel requirements

liranr23:
Adding to the readme the necassary dependencies we need to install.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,12 @@ Konveyor Forklift controller.
 
 For the build, the forklift uses [Bazel](https://bazel.build/).
 
+Install the following system components:
+
+- bazel >= 5
+- gcc
+- glibc-static
+
 ### Configuration
 
 The environment which you can set across all projects.


README.md
```diff
@@ -10,6 +10,14 @@ Konveyor Forklift controller.
 
 For the build, the forklift uses [Bazel](https://bazel.build/).
 
+Install the following system components:
+
+- bazel >= 5
+- docker
```
liranr23, line 16: actually not sure about docker, it is required for CI. but now sure if bazel needs it.

nyoxi: It is not necessary for build. It is only needed if you want to run the final container.

README.md
```diff
@@ -10,6 +10,14 @@ Konveyor Forklift controller.
...
+- gcc
+- go
```
nyoxi, line 18: Are you sure? What for? As far as I can tell bazel pulls in go toolset for the project. Specifically version 1.16.5 that we specify in `WORKSPACE`.

README.md
ahadas: it shouldn't be needed for CI as well but maybe [some tweaks are required](https://github.com/kubevirt/kubevirtci/blob/main/PODMAN.md)

liranr23: using podman-docker i didn't try those tweaks. but i had an error about insecure connection to the localhost, when editing the relevant configuration to allow insecure, it still didn't work. with docker it worked out of the box for me. in any case, this is unrelated to the bazel.

ahadas (approved):

PR #21, opened 2022-10-25
Prevent data volume deletion

liranr23:
With the new [CDI change](https://github.com/kubevirt/containerized-data-importer/pull/2421), the data volumes being deleted when the import completes. However, the controller lookup for the data volumes in order to know when the disk transfer is complete and continue with the migration flow. In order to prevent it, an annotation is added to the data volumes.

### DIFF
--- a/pkg/controller/plan/kubevirt.go
+++ b/pkg/controller/plan/kubevirt.go
@@ -54,6 +54,8 @@ const (
 	AnnOriginalName = "original-name"
 	//  Original VM name on source (value=vmOriginalID)
 	AnnOriginalID = "original-ID"
+	// DV deletion on completion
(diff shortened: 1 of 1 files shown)


rokkbert:
Nice! But do we need to delete it later, instead? Or can it just remain there?

liranr23:
> Nice! But do we need to delete it later, instead? Or can it just remain there?

Before the change it just remained there. AFAIK, it will act normally as it previously did.
This PR is a fast fix as it breaks forklift for any migration. A later resolution will be discussed, I assume we might switch the controller from looking up from the DV to PVC.

nyoxi (approved):

ahadas (approved):
@liranr23 please file a bug on that

PR #24, opened 2022-11-02
Switch to EFI without secure boot

liranr23:
Since we can't transfer NVRAM or TPM data, it is better to have EFI VMs without secure boot.

### DIFF
--- a/pkg/controller/plan/adapter/ovirt/builder.go
+++ b/pkg/controller/plan/adapter/ovirt/builder.go
@@ -23,6 +23,7 @@ import (
 const (
 	ClusterDefault = "cluster_default"
 	Q35Ovmf        = "q35_ovmf"
+	Q35SecureBoot  = "q35_secure_boot"
 )
 
 // Bus types
@@ -379,12 +380,12 @@ func (r *Builder) mapFirmware(vm *model.Workload, cluster *model.Cluster, object
 		Serial: serial,
 	}
 	switch biosType {
-	case Q35Ovmf:
-		smmEnabled := true
-		features.SMM = &cnv.FeatureState{
-			Enabled: &smmEnabled,
-		}
-		firmware.Bootloader = &cnv.Bootloader{EFI: &cnv.EFI{}}
+	case Q35Ovmf, Q35SecureBoot:
+		secureBootEnabled := false
+		firmware.Bootloader = &cnv.Bootloader{
+			EFI: &cnv.EFI{
+				SecureBoot: &secureBootEnabled,
+			}}
 	default:
 		firmware.Bootloader = &cnv.Bootloader{BIOS: &cnv.BIOS{}}
 	}
--- a/validation/policies/io/konveyor/forklift/vmware/uefi_boot.rego
+++ b/validation/policies/io/konveyor/forklift/vmware/uefi_boot.rego
@@ -8,7 +8,7 @@ concerns[flag] {
     has_uefi_boot
     flag := {
         "category": "Warning",
-        "label": "UEFI secure boot detected",
-        "assessment": "UEFI secure boot is enabled. NVRAM data is not copied during the migration and thus the VM might not boot on OpenShift Virtualization."
+        "label": "UEFI detected",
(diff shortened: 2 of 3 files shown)


nyoxi (approved):

pkg/controller/plan/adapter/ovirt/builder.go
```diff
@@ -380,11 +380,11 @@ func (r *Builder) mapFirmware(vm *model.Workload, cluster *model.Cluster, object
...
-		smmEnabled := true
-		features.SMM = &cnv.FeatureState{
-			Enabled: &smmEnabled,
-		}
-		firmware.Bootloader = &cnv.Bootloader{EFI: &cnv.EFI{}}
+		secureBootEnabled := false
+		firmware.Bootloader = &cnv.Bootloader{
+			EFI: &cnv.EFI{
+				SecureBoot: &secureBootEnabled,
+			}}
```
ahadas, line 387: the changes look good but how about documenting (in the code) why we do that?

ahadas: hmm this part is actually self-explained, I meant after adding this handling to the other 'bios-type' that represents uefi+secure boot

validation/policies/io/konveyor/forklift/vmware/uefi_boot.rego
```diff
@@ -8,7 +8,7 @@ concerns[flag] {
     has_uefi_boot
     flag := {
         "category": "Warning",
-        "label": "UEFI secure boot detected",
-        "assessment": "UEFI secure boot is enabled. NVRAM data is not copied during the migration and thus the VM might not boot on OpenShift Virtualization."
+        "label": "UEFI detected",
+        "assessment": "If UEFI secure boot is enabled, NVRAM data is not copied during the migration and thus the VM might not boot on OpenShift Virtualization."
```
nyoxi, line 12: I would explicitly mention that the migtrated VM will have secure boot disabled.

nyoxi (changes requested):
Actually, I don't understand the sentence in commit message: `As for some VMs it may the boot process of it.`

validation/policies/io/konveyor/forklift/vmware/uefi_boot.rego
ahadas: yeah, maybe something like:
```suggestion
        "assessment": "UEFI secure boot will be disabled on Openshift Vrtualization. If the VM was set with UEFI secure boot, manual steps within the guest would be needed in order to boot the guest operating system."
```

nyoxi (approved):

pkg/controller/plan/adapter/ovirt/builder.go
```diff
@@ -380,11 +381,12 @@ func (r *Builder) mapFirmware(vm *model.Workload, cluster *model.Cluster, object
 	}
 	switch biosType {
 	case Q35Ovmf:
-		smmEnabled := true
-		features.SMM = &cnv.FeatureState{
-			Enabled: &smmEnabled,
-		}
-		firmware.Bootloader = &cnv.Bootloader{EFI: &cnv.EFI{}}
```
ahadas, line 387: I think you'll need to add `fallthrough` here if you want to switch to the next block or do `case Q35Ovmf, Q35SecureBoot:`

validation/policies/io/konveyor/forklift/vmware/uefi_boot.rego
```diff
@@ -8,7 +8,7 @@ concerns[flag] {
     has_uefi_boot
     flag := {
         "category": "Warning",
-        "label": "UEFI secure boot detected",
-        "assessment": "UEFI secure boot is enabled. NVRAM data is not copied during the migration and thus the VM might not boot on OpenShift Virtualization."
+        "label": "UEFI detected",
+        "assessment": "UEFI secure boot will be disabled on Openshift Vrtualization. If the VM was set with UEFI secure boot, manual steps within the guest would be needed for the guest operating system to boot."
```
ahadas, line 12:
```suggestion
        "assessment": "UEFI secure boot will be disabled on Openshift Virtualization. If the VM was set with UEFI secure boot, manual steps within the guest would be needed for the guest operating system to boot."
```

ahadas (approved):

PR #26, opened 2022-11-07
Document why we always disable secure boot

ahadas:


liranr23 (approved):

pkg/controller/plan/adapter/ovirt/builder.go
```diff
@@ -381,6 +381,9 @@ func (r *Builder) mapFirmware(vm *model.Workload, cluster *model.Cluster, object
 	}
 	switch biosType {
 	case Q35Ovmf, Q35SecureBoot:
+		// we disable secure boot even if it was enabled on the source because the guest OS won't
+		// be able to boot without getting the NVRAM data. so we start the VM without secure boot
+		// to ease the procedure users need to do in order to make the guest OS to boot
```
nyoxi, line 386: Capital letters at the beginning of the sentence and period at the end please. Otherwise it sounds fine. I assume we want the same comment for VSphere too, right?

nyoxi (changes requested):

pkg/controller/plan/adapter/ovirt/builder.go
ahadas: > I assume we want the same comment for VSphere too, right?

yeah but for vsphere we also don't have any other option since we can't identify (at the moment?) whether secure boot is supposed to be enabled on the source, anyway - added there as well

nyoxi (approved):

PR #28, opened 2022-11-07
operator: restructure the templates

mnecas:
There are too many operator templates which makes it hard to find the proper file. This patch restructures the templates by the component which they affect like UI, controller etc.


nyoxi (approved):
lgtm

PR #29, opened 2022-11-07
Add controller library to forklift main repo

mnecas:
Migrate [controller](https://github.com/konveyor/controller) library to forklift repo.

We have migrated most of the konveyor project to one repo but we still had some parts separated. This patch adds main library around the controller. This will help us in future development since the developers will not need to create multiple patches.

The migration was made by following steps:
- Clone the controller and move the pkg content to pkg/lib
- Replace the old controller path name with the new lib path (form `github.com/konveyor/controller/pkg` to `github.com/konveyor/forklift-controller/pkg/lib`
```bash
sed -i 's/github.com\/konveyor\/controller\/pkg/github.com\/konveyor\/forklift-controller\/pkg\/lib/g' **/*.go
```
- Cleanup the unused go packages and add them to vendor
```bash
go mod tidy
go mod vendor
```
- Add bazel builds with `gazzel`


.github/workflows/pull_request.yml
```diff
@@ -58,7 +58,7 @@ jobs:
       - name: Setup Golang
         uses: actions/setup-go@v2
         with:
-          go-version: '1.19.1'
+          go-version: '1.18.7'
```
mnecas, line 61: @liranr23 @bennyz @ahadas 
This raises the question of which go we would use/test.
The `go.mod` has defined the 1.14 but the CI was running on 1.19.1 on which the migration failed with:
```
go vet ./pkg/... ./cmd/...
# github.com/konveyor/forklift-controller/pkg/lib/inventory/model
Error: pkg/lib/inventory/model/client.go:557:6: second argument to errors.As should not be *error
Error: pkg/lib/inventory/model/client.go:650:6: second argument to errors.As should not be *error
make: *** [Makefile:55: vet] Error 2
Error: Process completed with exit code 2.
```
I was running the migration on my env with 1.18.7 and the ci passed so that's why I changed it. It is possible that I might just need to bump my version of go to 1.19.1 but we should decide which version to use.

liranr23: IMO we do need to *try* and have the latest go version. however, if it breaks something it's ok to have it with older version and have a task to upgrade golang version? @ahadas

nyoxi: I agree with Liran. It's fine to merge this with older version but we should at least have a task to try to fix the problem and upgrade to latest Go.

Also, note that Bazel `WORKFLOW` has: `go_register_toolchains(version = "1.16.5")`
This is the version that we're actually using. I believe that this should be in sync with what you have for CI.

I quickly tried to build with bazel for both 1.18.7 and 1.19.1 the first builds fine but for the later it fails with:

```
2022/11/08 10:19:14 link: unlinkable object (from package github.com/konveyor/forklift-controller/vendor/github.com/cespare/xxhash/v2) - compiler requires -p flag
```

So we have another problem to solve. As for the problem with `errors.As` from a quick look we're using it wrong in the code. It should either be used in the form:
```
var notFound *NotFound;
if errors.As(err, notFound) {
```

or, which is probably preferable in our case since we don't need the value, use `errors.Is` instead.

nyoxi: Btw 1.19.1 is not the latest anymore. There are already 1.19.2 (released in October) and 1.19.3 (released in November).

pkg/apis/forklift/v1beta1/hook.go
```diff
@@ -1,7 +1,7 @@
 package v1beta1
 
 import (
-	libcnd "github.com/konveyor/controller/pkg/condition"
+	libcnd "github.com/konveyor/forklift-controller/pkg/lib/condition"
```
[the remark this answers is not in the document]
mnecas: On this note, I already have a PR where we would kind of split from the konveyor.
The PR [1] is related to the stop of reconciling with wrong auth credentials. 
[1] https://github.com/konveyor/controller/pull/104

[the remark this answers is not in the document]
ahadas: I had something a bit different in mind - we are still redirected from github.com/konveyor/forklift-controller(or validation or virt-v2v) to the corresponding repos in github.com/kubev2v, which is good, but we can start pointing to the repos on kubev2v directly

.github/workflows/pull_request.yml
mnecas: The suggestion to split it to two lines did not help.
So I updated it according to the docs as in [1]
```
if err == NotFound{
```
[1] http://go-database-sql.org/errors.html

pkg/apis/forklift/v1beta1/hook.go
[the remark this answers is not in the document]
ahadas: irrelevant to this change

pkg/lib/inventory/model/client.go
```diff
@@ -29,6 +29,8 @@ type DB interface {
 	Count(Model, Predicate) (int64, error)
 	// Begin a transaction.
 	Begin(...string) (*Tx, error)
+	// With transaction.
+	With(fn func(*Tx) error, labels ...string) error
```
ahadas, line 33: did you upgrade the library in the process or where did this change came from?

ahadas (commented):
managed to look at all changes :) one question inside, besides that it lgtm

pkg/lib/inventory/model/client.go
ahadas: in that case, it can be handy to document the git hash we "forked" from

mnecas: ahh yeah great point I took it from the main branch which is v0.11.0. Added README

ahadas (approved):

PR #30, opened 2022-11-14
Add build and release GitHub action job

mnecas:
- Migrate all build images to kubev2v quay.io repo
- Add `.github/workflows/build-push-images.yml` which is triggered with push to master.
- Add `hack/release-images.sh` to build and push all Bazel repos
- Cleanup the unused `.github/workflows/`


.github/workflows/build-push-images.yml
```diff
@@ -0,0 +1,37 @@
+name: Build and push images
+
+on:
+  push:
+    branches:
+      - main
```
ahadas, line 6: we probably don't want to update the 'latest' image on every PR that gets to the master branch but rather to update from stable branches, no?

mnecas: For the stable branches, I would suggest doing something like
`stable-2.3` pushes images to `2.3` tag and manually we would create the Z versions.
As for the main branch, I would put it either to the latest or some dev tag like `unstable`. If we will have some larger e2e tests it would be best to use the latest code.

ahadas: > For the stable branches, I would suggest doing something like `stable-2.3` pushes images to `2.3` tag and manually we would create the Z versions.

the more it is automated, the better it is.
what's wrong with the current approach that pushes images from a stable branch `release-v2.3.4` to tag `release-v2.3.4`?

> As for the main branch, I would put it either to the latest or some dev tag like `unstable`. If we will have some larger e2e tests it would be best to use the latest code. And the "latest" would be just the latest release.

looking at [konveyor/forklift-controller on quay.io](https://quay.io/repository/konveyor/forklift-controller?tab=tags) I see it was done like what you wrote above - just the tag is 'main' and the 'latest' is probably updated automatically so that's fine

hack/release-images.sh
```diff
@@ -0,0 +1,33 @@
...
+    echo "Please set all REGISTRY, REGISTRY_TAG and REGISTRY_ACCOUNT environment variables!" 
+    exit 1
+fi
+
+CONTROLLER_IMAGE=${REGISTRY}/${REGISTRY_ACCOUNT}/forklift-controller:${REGISTRY_TAG}
+OPERATOR_IMAGE=${REGISTRY}/${REGISTRY_ACCOUNT}/forklift-operator:${REGISTRY_TAG}
+MUST_GATHER_IMAGE=${REGISTRY}/${REGISTRY_ACCOUNT}/forklift-must-gather:${REGISTRY_TAG}
+MUST_GATHER_API_IMAGE=${REGISTRY}/${REGISTRY_ACCOUNT}/forklift-must-gather-api:${REGISTRY_TAG}
+UI_IMAGE=${REGISTRY}/${REGISTRY_ACCOUNT}/forklift-ui:${REGISTRY_TAG}
+UI_PLUGIN_IMAGE=${REGISTRY}/${REGISTRY_ACCOUNT}/forklift-console-plugin:${REGISTRY_TAG}
```
ahadas, line 13: we won't have corresponding tags for UI images, no?

ahadas: ok, REGISTRY_TAG gets the branch name, not the last commit hash, and that should be the same

ahadas (approved):

PR #31, opened 2022-11-15
Add ConnectionAuthFailed for one try auth check

mnecas:
Issue: when a user enters the correct username for the provider but an incorrect password it will try to connect multiple times to the engine. Which locks the user account.
Fix: Adding a special case of condition ConnectionAuthFailed and finishing the reconciliation.

Migrated patches from: https://github.com/konveyor/controller/pull/104 and https://github.com/kubev2v/forklift-controller/pull/447

### DIFF
--- a/pkg/controller/provider/container/vsphere/collector.go
+++ b/pkg/controller/provider/container/vsphere/collector.go
@@ -2,12 +2,18 @@ package vsphere
 
 import (
 	"context"
+	"net/http"
+	liburl "net/url"
+	"path"
+	"strings"
+	"time"
+
 	"github.com/go-logr/logr"
+	api "github.com/konveyor/forklift-controller/pkg/apis/forklift/v1beta1"
+	model "github.com/konveyor/forklift-controller/pkg/controller/provider/model/vsphere"
 	liberr "github.com/konveyor/forklift-controller/pkg/lib/error"
 	libmodel "github.com/konveyor/forklift-controller/pkg/lib/inventory/model"
 	"github.com/konveyor/forklift-controller/pkg/lib/logging"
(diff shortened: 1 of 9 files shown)


pkg/controller/provider/container/vsphere/collector.go
```diff
@@ -495,11 +498,11 @@ func (r *Collector) watch() (list []*libmodel.Watch) {
 
 //
 // Build the client.
-func (r *Collector) connect(ctx context.Context) error {
+func (r *Collector) connect(ctx context.Context) (status int, err error) {
 	r.close()
 	url, err := liburl.Parse(r.url)
 	if err != nil {
-		return liberr.Wrap(err)
+		return 0, liberr.Wrap(err)
```
ahadas, line 505: wouldn't it be better to set `err` and return? I mean, there's no http error code 0 but still look a bit awkward..

pkg/controller/provider/controller.go
```diff
@@ -172,6 +172,15 @@ func (r Reconciler) Reconcile(request reconcile.Request) (result reconcile.Resul
 		r.Log.V(2).Info("Conditions.", "all", provider.Status.Conditions)
 	}()
 
+	defer func() {
+		// Stop reconciliation when auth fails
+		if provider.Status.HasCondition(ConnectionAuthFailed) {
+			result.RequeueAfter = 0
+			err = nil
+			return
```
ahadas, line 180: how do we get out of this "loop" then, does the reconciliation continue when the url or credentials change?

ahadas: ack

ahadas (approved):

PR #33, opened 2022-11-21
Enable non-admin users to query providers from inventory

ahadas:
Previously, querying '/providers' by users that don't have cluster-admin
permissions failed because the request doesn't specify a UID of a
provider to query and thus, we ended up checking if the user has
permissions on all resources.
    
With this change, we'd always query 'providers.forklift.konveyor.io' for
'get' when a provider is specified and for 'list' otherwise. That way,
all users with ClusterRole on providers.forklift.konveyor.io can access
the data of providers from the inventory.

### DIFF
--- a/pkg/apis/forklift/v1beta1/register.go
+++ b/pkg/apis/forklift/v1beta1/register.go
@@ -23,6 +23,9 @@ limitations under the License.
 package v1beta1
 
 import (
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"sigs.k8s.io/controller-runtime/pkg/runtime/scheme"
 )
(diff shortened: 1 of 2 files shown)


bennyz (approved):

pkg/apis/forklift/v1beta1/register.go
```diff
@@ -33,3 +36,19 @@ var SchemeGroupVersion = schema.GroupVersion{
...
 var SchemeBuilder = &scheme.Builder{GroupVersion: SchemeGroupVersion}
+
+// TODO: find a better place for this, it would be nice to use it also to
+// determine the plural configuration of a resource for the operator
+func GetGroupResource(required runtime.Object) (group string, resource string, err error) {
+
+	switch required.(type) {
+	case *Provider:
+		group = SchemeGroupVersion.Group
+		resource = "providers"
```
mnecas, line 47: This might not be related to this patch directly but...
What should the users be able to do in a forklift?
This checks if they can list/get the providers, what about the mappings?
Should all the users be able also to create all those CRs?

ahadas: it depends on the permissions the users have, users with create/update permissions on providers, plans, migrations, mappings should be able to do everything an admin can do with forklift :)

pkg/apis/forklift/v1beta1/register.go
```diff
@@ -33,3 +36,19 @@ var SchemeGroupVersion = schema.GroupVersion{
...
+	default:
```
liranr23, line 48: shouldn't we add more types? user should query only providers? what about other resources?

ahadas: there's no need to add more at the moment - the inventory handles only 'providers' (and from those it takes data such as mappings), other endpoints are handled by the 'infrastructure'

liranr23 (approved):

PR #36, opened 2022-11-23
making kvm device for the virt-v2v pod optional

eslutsky:
we want to be able to run virt-v2v on nodes without KVM capabilities (ci)

### DIFF
--- a/operator/roles/forkliftcontroller/defaults/main.yml
+++ b/operator/roles/forkliftcontroller/defaults/main.yml
@@ -118,3 +118,4 @@ must_gather_api_state: absent
 
 must_gather_image_fqin: "{{ lookup( 'env', 'MUST_GATHER_IMAGE') or lookup( 'env', 'RELATED_IMAGE_MUST_GATHER') }}"
 virt_v2v_image_fqin: "{{ lookup( 'env', 'VIRT_V2V_IMAGE') or lookup( 'env', 'RELATED_IMAGE_VIRT_V2V') }}"
+virt_v2v_dont_request_kvm: "{{ lookup( 'env', 'VIRT_V2V_DONT_REQUEST_KVM') }}"
--- a/operator/roles/forkliftcontroller/templates/controller/deployment-controller.yml.j2
+++ b/operator/roles/forkliftcontroller/templates/controller/deployment-controller.yml.j2
@@ -167,6 +167,11 @@ spec:
         - name: PROFILE_DURATION
           value: "{{ controller_profile_duration }}"
 {% endif %}
+{% if virt_v2v_dont_request_kvm|bool %}
+        - name: VIRT_V2V_DONT_REQUEST_KVM
+          value: "true"
+{% endif %}
+
         envFrom:
         - configMapRef:
             name: {{ controller_configmap_name }}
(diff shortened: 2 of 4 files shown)


eslutsky:
with this PR in-place managed to run migration against fake vmwhere and converter pod https://github.com/eslutsky/forkliftci/actions/runs/3532779911/jobs/5927561806

eslutsky:
as discussed  offline,  we need to add new  parameter which control this setting, and turn it on by default

eslutsky:
> We merged #38 few days ago, now we delete it?

no we are going to create a cascading parameter which will disable it for CI , so it will be ON by default.

operator/roles/forkliftcontroller/defaults/main.yml
```diff
@@ -118,3 +118,4 @@ must_gather_api_state: absent
 
 must_gather_image_fqin: "{{ lookup( 'env', 'MUST_GATHER_IMAGE') or lookup( 'env', 'RELATED_IMAGE_MUST_GATHER') }}"
 virt_v2v_image_fqin: "{{ lookup( 'env', 'VIRT_V2V_IMAGE') or lookup( 'env', 'RELATED_IMAGE_VIRT_V2V') }}"
+virt_v2v_require_kvm: "{{ lookup( 'env', 'VIRT_V2V_REQUIRE_KVM')"
```
nyoxi, line 121: Missing terminating `}}`.

nyoxi (changes requested):

pkg/controller/plan/kubevirt.go
```diff
@@ -983,6 +983,18 @@ func (r *KubeVirt) findTemplate(vm *plan.VMStatus) (tmpl *template.Template, err
...
+	resourceReq := core.ResourceRequirements{}
+
+	// Request access to /dev/kvm via Kubevirt's Device Manager
+	// That is to ensure the appliance virt-v2v uses would not
+	// run in emulation mode, which is significantly slower
+	if Settings.VirtV2vRequireKVM {
+		resourceReq = core.ResourceRequirements{
+			Limits: core.ResourceList{
+				"devices.kubevirt.io/kvm": resource.MustParse("1"),
+			},
```
ahadas, line 995: why not to just set the Limits of resourceReq here then (but to override the entire struct)?

operator/roles/forkliftcontroller/templates/controller/deployment-controller.yml.j2
```diff
@@ -167,6 +167,11 @@ spec:
         - name: PROFILE_DURATION
           value: "{{ controller_profile_duration }}"
 {% endif %}
+{% if virt_v2v_require_kvm|bool %}
+        - name: VIRT_V2V_REQUIRE_KVM
+          value: "true"
+{% endif %}
```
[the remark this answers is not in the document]
eslutsky: its probably better to rename it to `virt_v2v_dont_require_kvm` , since the default and the recommended way is with KVM.

pkg/controller/plan/kubevirt.go
eslutsky: agreed, done

ahadas (approved):

nyoxi:
SonarCloud seems to be confused again.

ahadas:
> SonarCloud seems to be confused again.

yeah, I've started looking at that and then got distracted by codecov..

PR #37, opened 2022-11-23
Limit required permissions to the konveyor-forklift namespace

ahadas:
This is a follow up on #33 that limits the permissions users need to have in order to access the inventory API to the `konveyor-forklift` namespace and generates a better error message when the user is not allowed to access the inventory API


pkg/controller/provider/web/base/auth.go
```diff
@@ -142,11 +144,17 @@ func (r *Auth) permit(token string, p *api.Provider) (allowed bool, err error) {
 		return
 	}
 
-	allowed = review.Status.Allowed
+	if allowed = review.Status.Allowed; !allowed {
+		groupResource := &schema.GroupResource{
+			Resource: resource,
+			Group:    group,
+		}
+		err = fmt.Errorf("%s is forbidden: User %q cannot %s resource %q in API group %q in the namespace %q",
```
liranr23, line 152: why %q? isn't %q is for a char? don't you need more than a char?

ahadas: I want this error to match the one that is returned by k8s so I copied it from [here](https://github.com/kubernetes/kubernetes/blob/v1.27.0-alpha.0/staging/src/k8s.io/apiserver/pkg/endpoints/handlers/responsewriters/errors.go#L67)

liranr23: I looked into the integer part :man_facepalming:

liranr23 (approved):

bkhizgiy (approved):

PR #42, opened 2022-11-29
Use different images for cold and for warm migration in importer pod

nyoxi:
We would like to use C9S in importer POD as this should allow us to migrate also guests with EL9. Here is some preliminary work to make this happen. While at the moment the commits could be merged right now without breaking anything they do not yet contain the switch to C9S. For the switch to happen we need to use virt-v2v to copy the data instead of CDI -- a change I am still working on.

However, regarding this PR, there are some things I would like to open for discussion. The problematic part is that we cannot use C9S for warm migration because virt-v2v is missing the necessary features. That means we will need to have two separate images for import pod for some time -- one for cold migration and one for warm migration. So bearing in mind that the two images are (hopefully) only temporary solution and we may change back to single image some time in the future my questions are:

* *How to pass image to Forklift controller:* The image is configured in `VIRT_V2V_IMAGE` environment variable. Included patch changes the semantic of the variable in that the value can be two images separated by `|`, i.e. `my/image-cold|my/image-warm`. I understand this may be confusing or error prone and having two different variables (e.g. `VIRT_V2V_IMAGE_COLD` and `VIRT_V2V_IMAGE_WARM`) would be nicer. But it seems to me that adding new variables to the operator and CRs is rather complex so I wanted to know the general opinion before attempting this.

* *How to name the images:* At the moment I keep the name `forklift-virt-v2v` for the cold migration image and introduce new `forklift-virt-v2v-warm` for warm migration image. Similar to above, it would be nice to have two images `*-cold` and `*-warm` and deprecate use of `forklift-virt-v2v`. This however brings up a question of administration cost this introduces both u/s and d/s -- instead of adding one new image we add two images plus a deprecation.

Feel free to share your opinion on any of the above.

PS: The PR depends on #41.


ahadas (commented):
> We would like to use C9S in importer POD as this should allow us to migrate also guests with EL9.

right, and also to pick up fixes in recent versions of virt-v2v

> How to pass image to Forklift controller

I like the proposed approach in this PR - it minimized the changes as you wrote, and it will also make it easier to switch back to a single image later on

> How to name the images

I'd go with `forklift-virt-v2v` (rather than deprecating it) and `forklift-virt-v2v-warm` because the way I think of it is: we should use a single image of virt-v2v but due to limitations we have `forklift-virt-v2v-warm` as a temporary exception for warm migration

so in short, the PR looks good to me, maybe except the change of DefaultVirtV2vImage - we need to replace other occurrences (of quay.io/konveyor) as well so it can be part of that change

nyoxi:
Rebased and verified.

ahadas (approved):

PR #43, opened 2022-11-29
plan: add VDDK side car

nyoxi:
The VDDK library is not yet used by virt-v2v. This is a preliminary change that will be used later once we start using virt-v2v for copying data from source instead of CDI.


liranr23:
> The VDDK library is not yet used by virt-v2v. This is a preliminary change that will be used later once we start using virt-v2v for copying data from source instead of CDI.
> 
> Signed-off-by: Tomáš Golembiovský [tgolembi@redhat.com](mailto:tgolembi@redhat.com)

Overall LGTM, can you elaborate how virt-v2v is going to use it? or why?(replacing the CDI required VDDK image from outer source to a sidecar?)
Will the VDDK be provided out of the box? or need to be provided by the user?

ahadas (approved):
lgtm
it reminds me that we also thought about making the vddk part optional (knowing the transfer will be much slower without it..)

nyoxi:
> Overall LGTM, can you elaborate how virt-v2v is going to use it? or why?(replacing the CDI required VDDK image from outer source to a sidecar?) Will the VDDK be provided out of the box? or need to be provided by the user?

There's some confusion in terminology perhaps? We're going to use the same VDDK image and in the same way CDI does -- we attach an initial container that makes sure the VDDK library is usable by another process from the container (in our case virt-v2v).

> lgtm it reminds me that we also thought about making the vddk part optional (knowing the transfer will be much slower without it..)

True. We but I don't recall if we discussed the details. Is this something we want to configure per provider or per migration plan?

ahadas:
> True. We but I don't recall if we discussed the details. Is this something we want to configure per provider or per migration plan?

yeah, we didn't get to this level of details - I think that per-provider would make more sense, if you were able to set VDDK for one migration, I don't see a reason for not using it in other migrations.. I think it's merely for skipping the initial setting of the VDDK image

PR #44, opened 2022-11-29
plan: store disk source in annotation

nyoxi:
Instead of relying on disk source being part of DataVolume.Spec we store the source also in the annotation. This is a preliminary change that will ease our transition from CDI.

### DIFF
--- a/pkg/controller/plan/adapter/base/doc.go
+++ b/pkg/controller/plan/adapter/base/doc.go
@@ -10,6 +10,13 @@ import (
 	cdi "kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1"
 )
 
+// Annotations
+const (
+	// Used on DataVolume, contains disk source -- e.g. backing file in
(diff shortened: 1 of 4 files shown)


ahadas (commented):
for ovirt we should store that on the new CRD - I guess it would be an overkill for vsphere.. @nyoxi did you consider that?

nyoxi:
Just a rebase.

> for ovirt we should store that on the new CRD - I guess it would be an overkill for vsphere.. @nyoxi did you consider that?

What CRD are you talking about? But yes, I don't think adding a CRD just for that for VSphare is worth the complexity. I see in #44 that disk sources are now stored also in PVC annotation. It seems to me we should be able to use that instead of adding new annotation to DV, but I'll have to investigate a little bit more about that.

ahadas:
another rebase, CI should pass now

ahadas (approved):

ahadas (approved):

PR #47, opened 2022-12-04
Add github.com/xi2/xz as a dependency

ahadas:
It is needed for building the virt-v2v image.


eslutsky (approved):
looks good, fixes the issue at my gh-action test run https://github.com/eslutsky/forkliftci/actions/runs/3613431774/jobs/6089178439

PR #54, opened 2022-12-14
Fail requests to add providers with invalid credentials

ahadas:
- [x] Implement proper validation logic (was already implemented by @mnecas - just polished it a bit)
- [x] Add certification manager for K8s
- [x] Requests from the UI should be validated by default, requests during 'disaster recovery' should not be validated (should be fine as we validate the secret rather than the provider)


pkg/forklift-api/webhooks/webhooks.go
```diff
@@ -35,8 +35,17 @@ func RegisterValidatingWebhooks() {
...
 	server := http.Server{
-		Addr:    ":9876",
+		Addr:    ":8443",
 		Handler: mux,
 	}
-	server.ListenAndServeTLS("/var/run/secrets/forklift-api/tls.crt", "/var/run/secrets/forklift-api/tls.key")
+	log.Info("start listening")
+	errors := make(chan error)
+	go func() {
+		errors <- server.ListenAndServeTLS("/var/run/secrets/forklift-api-serving-cert/tls.crt", "/var/run/secrets/forklift-api-serving-cert/tls.key")
```
bennyz, line 44: I don't understand this part, why not `err = server.ListenAndServeTLS...`?

ahadas: yeah, there's no need for the go routine in this code but that's a work in progress, I think we will eventually want this code to shutdown the pod on certain signals [like kubevirt's virt-api does](https://github.com/kubevirt/kubevirt/blob/ee70cefd68aa3df0c9926e677c852e295e2229b8/pkg/virt-api/api.go#L917-L921)

ahadas:
that's very strange - CI works on github with the main branch, it works locally with the changes in this PR but fails on github with the changes in this PR..

ahadas:
I think I understand what's going on here now - the imageio container probably fails to start on github hosted runner due to memory/cpu constraints due to the forklift-api deployment..

ahadas:
@bennyz @bkhizgiy @liranr23 ping

pkg/forklift-api/webhooks/validating-webhook/admitters/secret-admitter.go
```diff
@@ -62,46 +39,50 @@ func (admitter *SecretAdmitter) Admit(ar *admissionv1.AdmissionReview) *admissio
 			},
 		}
 	}
-	createdForProviderType, ok := secret.GetLabels()["createdForProviderType"]
-	// TODO: need to change this, secret without labels should be allowed
```
liranr23, line 66: do we need to take care about dealing non-labeled secret? if so (in the future), i would leave the comment.

ahadas: I removed this TODO because it's handled now - secrets without this label won't be processed by the validation webhook

ahadas: see the ObjectSelector in ValidatingWebhookConfiguration

liranr23 (approved):

PR #57, opened 2022-12-15
Fix operator bundle crd bases

mnecas:
Issue: 
When running the opm to create the index from bundle, it reads the base files.
A few of the base files started with empty space on top and `---` right after it.
Which the OPM understood as the end of an empty file.
 
```
WARN[0005] found more than one document inside forklift.konveyor.io_storagemaps.yaml,
...
error checking provided apis in bundle mtv-operator.v2.3.5: couldn't find forklift.konveyor.io/v1beta1/StorageMap (storagemaps) in bundle. found: map[forklift.konveyor.io/v1beta1/ForkliftController (forkliftcontrollers):{} forklift.konveyor.io/v1beta1/Hook 
```

### DIFF
--- a/operator/config/crd/bases/forklift.konveyor.io_hooks.yaml
+++ b/operator/config/crd/bases/forklift.konveyor.io_hooks.yaml
@@ -1,4 +1,3 @@
-
 ---
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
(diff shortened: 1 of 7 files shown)


ahadas (commented):
aren't these files auto-generated?

ahadas:
> aren't these files auto-generated?

posted #58 - an upgraded version of controller-gen fixes this

PR #64, opened 2022-12-27
Populator

liranr23:


hack/rpm-deps.sh
```diff
@@ -20,6 +20,17 @@ virt_v2v="
   virtio-win
 "
 
+ovirt-imageio="
+  gcc
+  py3-pip
+  python3-dev
```
ahadas, line 26:
```suggestion
  python3-devel
```

hack/rpm-deps.sh
```diff
@@ -20,6 +20,17 @@ virt_v2v="
   virtio-win
 "
 
+ovirt-imageio="
+  gcc
+  py3-pip
```
ahadas, line 25:
```suggestion
  python3-pip
```

bennyz: I think that's the name of the package in alpine, perhaps we should use some UBI go tools image instead

hack/rpm-deps.sh
ahadas: how about using el9? then we can take it from https://mirror.stream.centos.org/9-stream/AppStream/x86_64/os/Packages/

ovirt-imageio/main.go
```diff
@@ -0,0 +1,199 @@
+package main
```
bennyz, line 1: I forgot to merge https://github.com/kubev2v/ovirt-imageio-populator/pull/8 previously
so this file looks different now, as well as the CRD

Makefile
```diff
@@ -87,6 +87,15 @@ push-operator: build-operator
 	$(CONTAINER_CMD) tag operator:forklift-operator-image ${OPERATOR_IMAGE}
 	$(CONTAINER_CMD) push ${OPERATOR_IMAGE}
 
+# Build the docker image
+build-imageio:
+	bazel run ovirt-imageio/ovirt-imageio-image
+
+# Push the docker image
+push-imageio: build-operator
```
ahadas, line 95: the term `imageio` is not that good in this context - how about `ovirt-populator` instead?

WORKSPACE
```diff
@@ -3468,11 +3476,11 @@ http_file(
...
     name = "bazeldnf",
-    sha256 = "c37709d05ad7eae4d32d7a525f098fd026483ada5e11cdf84d47028222796605",
-    strip_prefix = "bazeldnf-0.5.2",
+    sha256 = "404fc34e6bd3b568a7ca6fbcde70267d43830d0171d3192e3ecd83c14c320cfc",
+    strip_prefix = "bazeldnf-0.5.4",
     urls = [
-        "https://github.com/rmohr/bazeldnf/archive/v0.5.2.tar.gz",
-        "https://storage.googleapis.com/builddeps/c37709d05ad7eae4d32d7a525f098fd026483ada5e11cdf84d47028222796605",
+        "https://github.com/rmohr/bazeldnf/archive/v0.5.4.tar.gz",
+        "https://storage.googleapis.com/builddeps/404fc34e6bd3b568a7ca6fbcde70267d43830d0171d3192e3ecd83c14c320cfc",
```
ahadas, line 3483: how about separating this (upgrade of bazeldnf) out to a different PR?

WORKSPACE
```diff
@@ -3502,2335 +3510,2407 @@ rules_proto_dependencies()
...
+    urls = ["https://mirror.stream.centos.org/9-stream/BaseOS/x86_64/os/Packages/zip-3.0-33.el9.x86_64.rpm"],
 )
 
 rpm(
-    name = "zlib-0__1.2.11-19.el8.x86_64",
-    sha256 = "439833454c91b662c1ed99eff65e4726d765e974f65faadaf1c29eb1281f28f9",
-    urls = ["http://mirror.centos.org/centos/8-stream/BaseOS/x86_64/os/Packages/zlib-1.2.11-19.el8.x86_64.rpm"],
+    name = "zlib-0__1.2.11-35.el9.x86_64",
+    sha256 = "80df42ac4be2c057e332647ca98d65b1548d4e4adf52beb75d79e79fc8e48aa7",
+    urls = ["https://mirror.stream.centos.org/9-stream/BaseOS/x86_64/os/Packages/zlib-1.2.11-35.el9.x86_64.rpm"],
```
ahadas, line 5915: how come that everything is changed from el8 to el9? I'd expect the el9 dependencies to be added in addition to the el8 ones as we keep building virt-v2v on el8, no?

rpm/repo.yaml
```diff
@@ -1,9 +1,12 @@
...
-  baseurl: http://mirror.centos.org/centos/8-stream/BaseOS/x86_64/os/
-  name: centos/stream8-baseos-x86_64
+  baseurl: https://mirror.stream.centos.org/9-stream/BaseOS/x86_64/os/
+  name: centos/stream9-baseos-x86_64
   gpgkey: https://www.centos.org/keys/RPM-GPG-KEY-CentOS-Official
 - arch: x86_64
-  baseurl: http://mirror.centos.org/centos/8-stream/AppStream/x86_64/os/
-  name: centos/stream8-appstream-x86_64
+  baseurl: https://mirror.stream.centos.org/9-stream/AppStream/x86_64/os/
+  name: centos/stream9-appstream-x86_64
```
ahadas, line 8: ah ok, that answers my previous question (about el8 dependencies being replaced) - need to keep the el8 ones for the time being (even when we upgrade virt-v2v to el9, seems we'll still keep the el8 version for warm migration..), maybe a separate file?

operator/BUILD.bazel
```diff
@@ -67,6 +67,7 @@ genrule(
         "bundle/manifests/forklift.konveyor.io_plans.yaml",
         "bundle/manifests/forklift.konveyor.io_providers.yaml",
         "bundle/manifests/forklift.konveyor.io_storagemaps.yaml",
+        "bundle/manifests/forklift.konveyor.io_ovirtimageiopopulators.yaml",
```
ahadas, line 70: I know this comes from the other repo but ovirtimageiopopulators is not the best name, how about ovirtpopulators?

operator/config/imageio_populator/deployment.yaml
```diff
@@ -0,0 +1,35 @@
+kind: Deployment
+apiVersion: apps/v1
+metadata:
+  name: imageio-populator
```
ahadas, line 4: same here and below - how about 'ovirt-populator'?

operator/config/rbac/forklift-imageio-populator_service_account.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: imageio-populator-account
+  namespace: konveyor-forklift
```
ahadas, line 5: would it make sense to set this role also to openstack populator and to the new virt-v2v pod later on? if so, let's generalize it. and the namespace should not be `konveyor-forklift` explicitly

operator/forklift-operator-catalog.yaml
```diff
@@ -4,7 +4,10 @@ metadata:
   name: konveyor-forklift
   namespace: openshift-marketplace
 spec:
-  displayName: Forklift Operator
+  displayName: Liran Rotenberg
```
ahadas, line 7: :)

operator/forklift-operator-catalog.yaml
```diff
@@ -4,7 +4,10 @@ metadata:
...
   publisher: Konveyor
   sourceType: grpc
-  image: quay.io/konveyor/forklift-operator-index:latest
+  image: quay.io/lrotenbe/forklift-operator-index:latest
+  updateStrategy:
+    registryPoll:
+      interval: 10m0s
```
ahadas, line 13: I realize the changes above are temporary but also this one?

ovirt-imageio/main.go
```diff
@@ -0,0 +1,262 @@
...
```
ahadas, line 1: this package should move to the `cmd` folder

ovirt-imageio/pkg/v1beta1/BUILD.bazel
```diff
@@ -0,0 +1,16 @@
+load("@io_bazel_rules_go//go:def.bzl", "go_library")
```
ahadas, line 1: and the sources within `ovirt-imageio/pkg` should move to `pkg/X` (pkg/ovirt-populator?)

pkg/apis/forklift/v1beta1/BUILD.bazel
```diff
@@ -30,6 +30,6 @@ go_library(
         "//vendor/k8s.io/client-go/rest",
         "//vendor/sigs.k8s.io/controller-runtime/pkg/client",
         "//vendor/sigs.k8s.io/controller-runtime/pkg/client/config",
-        "//vendor/sigs.k8s.io/controller-runtime/pkg/runtime/scheme",
+        "//vendor/sigs.k8s.io/controller-runtime/pkg/scheme",
```
ahadas, line 33: it can simplify the review if we extract the change of the dependencies to a separate PR

ahadas (commented):
this would be a great addition but it's a big - I've commented inside with some suggestions of what we can do separately (and get in faster) in order to simplify this PR

WORKSPACE
liranr23: i thought you wish to have el9 also for virt-v2v, i'll add back el8.

operator/forklift-operator-catalog.yaml
liranr23: yup

pkg/apis/forklift/v1beta1/BUILD.bazel
bennyz: yeah, given the amount of code changes needed along with the version bumps it's probably better done separately

WORKSPACE
ahadas: separated out to #77

liranr23:
moved to other PRs.

PR #66, opened 2022-12-27
Use pointers for ListOpts

bennyz:


pkg/controller/provider/container/openstack/client.go
```diff
@@ -264,13 +265,13 @@ func (r *Client) list(object interface{}, listopts interface{}) (err error) {
 
 	case *[]Image:
 		switch listopts.(type) {
-		case ImageListOpts:
+		case *ImageListOpts:
 		default:
-			err = liberr.New("Wrong image list opts")
+			err = liberr.New("Wrong image list opts", "type", reflect.TypeOf(listopts))
```
ahadas, line 270: talked offline - need to drop

ahadas (approved):

PR #68, opened 2022-12-28
WIP: add autimatic retrival of missing CA certificate

bkhizgiy:


ahadas:
@bkhizgiy does it mean that it (download) actually works for ovf_store or does it still need to be tested?

pkg/controller/provider/container/ovirt/client.go
```diff
@@ -226,3 +234,365 @@ func GetInsecureSkipVerifyFlag(secret *core.Secret) bool {
...
+
+	itURL, ok := currImageTransfer.ProxyUrl()
+
+	// creat regex expresions to filter the proxy URL
+	reInit := regexp.MustCompile(`https://`)
+	reUrl := regexp.MustCompile(`^[^/]+`)
+
+	transferURL := reInit.ReplaceAllString(itURL, "")
+	transferURL = reUrl.FindString(transferURL)
+	transferURL = fmt.Sprint("https://", transferURL, "/info/")
```
bennyz, line 364: I think you can use `url.Parse` to manipulate the URL, should be easier

bkhizgiy:
@ahadas yes, it's working:) and preformed as a regular disk transfer

pkg/controller/provider/container/ovirt/client.go
```diff
@@ -226,3 +234,365 @@ func GetInsecureSkipVerifyFlag(secret *core.Secret) bool {
...
+			return
+		}
+
+		data, errFile := ioutil.ReadFile("ca.pem")
+		if err != nil {
+			err = liberr.Wrap(errFile)
+			return
+		}
+
+		updatedCa := fmt.Sprint(string(secret.Data["caCert"]), "/n", string(data))
```
bennyz, line 401:
```suggestion
		updatedCa := fmt.Sprint(string(secret.Data["caCert"]), "\n", string(data))
```

pkg/controller/provider/container/ovirt/client.go
```diff
@@ -226,3 +234,365 @@ func GetInsecureSkipVerifyFlag(secret *core.Secret) bool {
...
+		ok = false
+		return
+	}
+
+	ok = true
+	return
+
+}
+
+// func uploadDiskTestConection() bool {
```
bkhizgiy, line 441: this is currently marked as a comment but this is the code for the uploaded flow, keeping it here for now just in case.

pkg/controller/provider/container/ovirt/client.go
```diff
@@ -226,3 +234,365 @@ func GetInsecureSkipVerifyFlag(secret *core.Secret) bool {
...
+	_, err = conn.SystemService().ImageTransfersService().ImageTransferService(imageTransferId).Cancel().Send()
+	if err != nil {
+		err = liberr.Wrap(err)
+		return
+	}
+
+	// Check server response
+	if resp.StatusCode != http.StatusOK {
+		//retrive missing CA cert and update secrete
+		caUrl := fmt.Sprint(string(secret.Data["url"]), "ovirt-engine/services/pki-resource?resource=ca-certificate&format=X509-PEM-CA")
```
bkhizgiy, line 387: suggestions regarding the ca retrieval:
I think it would be better initially to retrieve the CA and check the transfer with it, and in case it works compare it to the secret one and add if missing because in the current implementation we won't have any indication in the case of a failure with that automated retrieve CA.
Also should we handle or address somehow a failure (if a separated cert is confined for imagIO)?

pkg/controller/provider/container/ovirt/client.go
```diff
@@ -226,3 +234,365 @@ func GetInsecureSkipVerifyFlag(secret *core.Secret) bool {
 
 	return insecureSkipVerify
 }
+
+func DownloadOVFTestConection(secret *core.Secret) (ok bool, err error) {
```
ahadas, line 238: how about TestDownloadOvfStore?

pkg/controller/provider/container/ovirt/client.go
```diff
@@ -226,3 +234,365 @@ func GetInsecureSkipVerifyFlag(secret *core.Secret) bool {
...
+
+	var diskIdOVF string
+
+	url := fmt.Sprint(string(secret.Data["url"]), "/ovirt-engine/api")
```
ahadas, line 242: I didn't find where we do that (appending 'ovirt-engine/api') elsewhere, does the UI send a different URL within the secret and within the provider?

pkg/controller/provider/container/ovirt/client.go
```diff
@@ -226,3 +234,365 @@ func GetInsecureSkipVerifyFlag(secret *core.Secret) bool {
 
 	return insecureSkipVerify
 }
+
+func DownloadOVFTestConection(secret *core.Secret) (ok bool, err error) {
+
+	var diskIdOVF string
```
ahadas, line 240: how about placing it closer to where it's used?

pkg/controller/provider/container/ovirt/client.go
```diff
@@ -226,3 +234,365 @@ func GetInsecureSkipVerifyFlag(secret *core.Secret) bool {
...
+		Compress(true).
+		Timeout(time.Second * 10).
+		Build()
+	if err != nil {
+		err = liberr.Wrap(err)
+		return
+	}
+	defer conn.Close()
+
+	//Find OVF disk ID
```
ahadas, line 259: OVF STORE

pkg/controller/provider/container/ovirt/client.go
```diff
@@ -226,3 +234,365 @@ func GetInsecureSkipVerifyFlag(secret *core.Secret) bool {
...
+	disksService := conn.SystemService().DisksService()
+	diskResponse, err := disksService.List().Send()
+	if err != nil {
+		err = liberr.Wrap(err)
+		return
+	}
+
+	disks, ok := diskResponse.Disks()
+
+	if ok {
```
ahadas, line 269: need to improve the error handling here and below

pkg/controller/provider/container/ovirt/client.go
```diff
@@ -226,3 +234,365 @@ func GetInsecureSkipVerifyFlag(secret *core.Secret) bool {
...
+
+	// Get a disk identified by uuid
+	diskService := disksService.DiskService(diskIdOVF)
+	diskRequest := diskService.Get()
+	diskRequestSent, err := diskRequest.Send()
+	if err != nil {
+		err = liberr.Wrap(err)
+		return
+	}
+	disk, ok := diskRequestSent.Disk()
```
ahadas, line 291: do we have to call get-disk here or can we use the disk we got from the 'list' call above?

pkg/controller/provider/container/ovirt/client.go
```diff
@@ -226,3 +234,365 @@ func GetInsecureSkipVerifyFlag(secret *core.Secret) bool {
...
+
+	diskId, ok := disk.Id()
+
+	//in case the OVF disk is locked wait fo OK status befor stating the image transfer
+	conn.WaitForDisk(diskId, ovirtsdk.DISKSTATUS_OK, 90*time.Second)
```
ahadas, line 296: 90 sec would be too much for a webhook, need to lower this

pkg/controller/provider/container/ovirt/client.go
```diff
@@ -226,3 +234,365 @@ func GetInsecureSkipVerifyFlag(secret *core.Secret) bool {
...
+	}
+	currImageTransfer, ok := transferReq.ImageTransfer()
+
+	imageTransferId, ok := currImageTransfer.Id()
+
+	for {
+		if currImageTransfer.MustPhase() == ovirtsdk.IMAGETRANSFERPHASE_INITIALIZING {
+			time.Sleep(1 * time.Second)
+			currImageTransferReq, errIT := conn.SystemService().ImageTransfersService().ImageTransferService(imageTransferId).Get().Send()
+			if err != nil {
```
ahadas, line 331: you probably meant errIT here, but why having two different variables?

pkg/controller/provider/container/ovirt/client.go
ahadas: yeah

pkg/controller/provider/container/ovirt/client.go
ahadas: yeah, makes sense and when we think about the webhook we can to the following in the mutating webhook:
if there's connectivity to the source ovirt/rhv environment:
1. get the certificate
2. compare it against the one provided
3. if they are different, append the one that was retrieved to the specified one
and in the validation webhook:
if there's connectivity to the source ovirt/rhv environment:
1. test the ability to download the OVF store

pkg/controller/provider/container/ovirt/client.go
```diff
@@ -226,3 +234,365 @@ func GetInsecureSkipVerifyFlag(secret *core.Secret) bool {
...
+		secret := &core.Secret{
+			Data: map[string][]byte{
+				"caCert":   []byte(updatedCa),
+				"url":      []byte(string(secret.Data["url"])),
+				"user":     []byte(string(secret.Data["user"])),
+				"password": []byte(string(secret.Data["password"])),
+			},
+			Type: "Opaque",
+		}
+		_, err = secretsClient.Update(context.TODO(), secret, metav1.UpdateOptions{})
```
ahadas, line 427: with the mutating webhook we won't need to update the secret this way
it's an open question whether we really need that for the less likely scenario that the source ovirt/rhv environment is not available when the provider is created..

pkg/controller/provider/container/ovirt/client.go
bkhizgiy: I didn't find the saving place but it's sent with '/ovirt-engine/api', removing

pkg/controller/provider/container/ovirt/client.go
bkhizgiy: you're right, no need removed the entire section.

pkg/controller/provider/container/ovirt/client.go
bkhizgiy: I've changed it to 20 sec, the webhook timeout is 30 sec so leaving some time for the rest of the flow...

pkg/controller/provider/container/ovirt/client.go
bkhizgiy: for some reason the compiler having issue with using the original parameter so I had to create a new one.

pkg/controller/provider/container/ovirt/client.go
bkhizgiy: ack, done.

pkg/forklift-api/webhooks/validating-webhook/admitters/secret-admitter.go
```diff
@@ -71,7 +73,22 @@ func (admitter *SecretAdmitter) Admit(ar *admissionv1.AdmissionReview) *admissio
 			if err != nil {
 				log.Info("Connection test failed, yet passing", "status", status, "error", err.Error())
 			} else {
-				log.Info("Test succeeded, passing")
+				// Perform validation for the provided engine CA certificate,
+				// initiate dummy image transfer to make sure the certificate is correct and that the migration will work in a later stage.
+				if isCaOK, err := ovirt.TestDownloadOvfStore(secret); isCaOK {
```
ahadas, line 78: yeah, this part needs to be executed specifically for ovirt as you suggested offline

pkg/controller/provider/container/ovirt/client.go
```diff
@@ -219,3 +221,138 @@ func GetInsecureSkipVerifyFlag(secret *core.Secret) bool {
...
+						diskIdOVF = diskId
+					}
+					break
+				}
+			}
+		}
+	}
+
+	//in case the OVF STORE is locked wait fo OK status befor stating the image transfer
+	conn.WaitForDisk(diskIdOVF, ovirtsdk.DISKSTATUS_OK, 20*time.Second)
```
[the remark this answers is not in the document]
bkhizgiy: yes, it assumes that it will always exist, I can add the check for that but is a situation like that can happen at all?

pkg/controller/provider/container/ovirt/client.go
```diff
@@ -219,3 +221,138 @@ func GetInsecureSkipVerifyFlag(secret *core.Secret) bool {
...
+	}
+
+	// Create http client
+	tlsConfig := &tls.Config{
+		RootCAs: caCertPool,
+	}
+	transport := &http.Transport{TLSClientConfig: tlsConfig}
+	client := &http.Client{Transport: transport}
+
+	// Parse the url for the CA test http request using the ProxyUrl from the image transfer
```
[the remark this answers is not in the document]
bkhizgiy: I used the same endpoint as configured in the ovirt engine for the same test, IIRC I originally tried to use the transfer URL but it didn't return the desired response.

pkg/controller/provider/container/ovirt/client.go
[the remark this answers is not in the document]
ahadas: it's unlikely - when adding a storage domain, it comes without an OVF store and ovirt-engine periodically updates the OVF store or creates it if it's missing. But otherwise I believe static analysis of this code would raise errors

pkg/controller/provider/container/ovirt/client.go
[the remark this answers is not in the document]
ahadas: I see
@bennyz do we still configure the proxy in newer oVirt versions?..

pkg/controller/provider/container/ovirt/client.go
```diff
@@ -219,3 +221,136 @@ func GetInsecureSkipVerifyFlag(secret *core.Secret) bool {
...
+					if diskId, ok := disk.Id(); ok {
+						ovfStoreId = diskId
+					}
+					break
+				}
+			}
+		}
+	}
+
+	//in case the OVF STORE is locked wait fo OK status befor stating the image transfer
```
ahadas, line 265: befor -> before

pkg/forklift-api/webhooks/validating-webhook/admitters/secret-admitter.go
```diff
@@ -71,7 +73,32 @@ func (admitter *SecretAdmitter) Admit(ar *admissionv1.AdmissionReview) *admissio
...
+								Result: &metav1.Status{
+									Code:    http.StatusBadRequest,
+									Message: err.Error(),
+								},
+							}
+						}
+					}
+					log.Info("Test credentials and CA certificat succeeded, passing")
+				} else {
+					log.Info("Test credentials succeeded but engine CA certificate is not valid")
```
ahadas, line 92: but in this case, when it's not ovirt, `engine CA certificate` is irrelevant, right?

ahadas (commented):
besides the open question and minor things, it looks good - but the failure in CI seems related, @bkhizgiy can you please check that?

pkg/controller/provider/container/ovirt/client.go
bkhizgiy: logs were added in the client function for all the possible errors.

pkg/controller/provider/container/ovirt/client.go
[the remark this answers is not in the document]
bkhizgiy: After testing with the lab env yesterday seems that this check is really not the best (returning OK in cases it should be false)  so it does not cover all cases. 
I switched the proxy URL to the transfer URL and now it will directly test on the host, in case the certificate is missing the request will be failed to send and the error message will give us an indication about that.

pkg/forklift-api/webhooks/validating-webhook/admitters/secret-admitter.go
bkhizgiy: yes, I've split the cases between ovirt and other provider's types, and also added good flow and bad flow specific to ovirt.

pkg/controller/provider/container/ovirt/client.go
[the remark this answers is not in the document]
bkhizgiy: Added a check to see that the OVF id is not empty, and returned false in the case it is.
btw I think that the CI failed due to that, since the fake ovirt doesn't have OVF STORE so the rest of the flow failed and the secret didn't create.

pkg/forklift-api/webhooks/validating-webhook/admitters/secret-admitter.go
```diff
@@ -83,6 +101,9 @@ func (admitter *SecretAdmitter) Admit(ar *admissionv1.AdmissionReview) *admissio
 			Allowed: true,
 		}
 	}
+	return &admissionv1.AdmissionResponse{
+		Allowed: true,
+	}
```
ahadas, line 106: how about moving it to be at the of the first 'if' section?

pkg/forklift-api/webhooks/validating-webhook/admitters/secret-admitter.go
ahadas: positive and negative flows :)

pkg/controller/provider/container/ovirt/client.go
```diff
@@ -219,3 +219,156 @@ func GetInsecureSkipVerifyFlag(secret *core.Secret) bool {
...
+					}
+					break
+				}
+			}
+		}
+	}
+	// In case no OVF store was found in the cluster
+	if ovfStoreId == "" {
+		log.Error(err, "OVF Store does not exist on cluster, CA certificate test can't performed")
+		ok = false
```
ahadas, line 267: I find this a bit confusing - if we pass this line then `ok` is set to true (line 251) , right?
how about renaming the output argument `ok` to something like `validCA` and then only set it to true in line 369?

pkg/controller/provider/container/ovirt/client.go
```diff
@@ -219,3 +219,156 @@ func GetInsecureSkipVerifyFlag(secret *core.Secret) bool {
...
+	).Build()
+	if err != nil {
+		log.Error(err, "Failed to build image transfer")
+		err = liberr.Wrap(err)
+		return
+	}
+
+	// Initialize image transfer and lock the disk
+	transfer.ImageTransfer(imageTransfer)
+	transferReq, err := transfer.Send()
```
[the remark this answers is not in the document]
bkhizgiy: no need, changing into one.
regarding the name I've used req because it's the request to initiate image transfer, I saw they were using similar names in the ovirt SDK flow, but I can change it to `response` if you think it's more suitable.

pkg/controller/provider/container/ovirt/client.go
bkhizgiy: yes, I agree I've been thinking about that as well, it was supposed to be also an error handler (the way the SDK is set) but in any critical errors we just return err so it's redundant...

pkg/controller/provider/container/ovirt/client.go
```diff
@@ -219,3 +219,156 @@ func GetInsecureSkipVerifyFlag(secret *core.Secret) bool {
...
+				if diskName == "OVF_STORE" {
+					if diskId, ok := disk.Id(); ok {
+						ovfStoreId = diskId
+					}
+					break
+				}
+			}
+		}
+	}
+	// In case no OVF store was found in the cluster
```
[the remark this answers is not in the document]
ahadas: sorry, I was wrong here - I see that it doesn't check this on a particular data center but system-wide, right? so let's rather use 'system' in this case?

pkg/controller/provider/container/ovirt/client.go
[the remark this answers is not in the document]
ahadas: I see, but here it seems to be a response of the request to start a transfer, no? if so then yeah, let's change it so it will better reflect what it represents

pkg/forklift-api/webhooks/validating-webhook/admitters/secret-admitter.go
```diff
@@ -71,12 +68,36 @@ func (admitter *SecretAdmitter) Admit(ar *admissionv1.AdmissionReview) *admissio
...
+								Message: "Test credentials succeeded but engine CA certificate is not valid",
+							},
+						}
+					}
+				} else {
+					log.Info("Test credentials succeeded, passing")
+				}
+				return &admissionv1.AdmissionResponse{
+					Allowed: true,
+				}
```
ahadas, line 95: now we don't need this, right?

pkg/forklift-api/webhooks/validating-webhook/admitters/secret-admitter.go
```diff
@@ -71,12 +68,36 @@ func (admitter *SecretAdmitter) Admit(ar *admissionv1.AdmissionReview) *admissio
...
+						return &admissionv1.AdmissionResponse{
+							Allowed: true,
+							Result: &metav1.Status{
+								Code:    http.StatusForbidden,
+								Message: "Test credentials succeeded but engine CA certificate is not valid",
+							},
+						}
+					}
+				} else {
+					log.Info("Test credentials succeeded, passing")
```
ahadas, line 91: well, that's a bit misleading for non-ovirt providers since we don't test their credentials (yet?), no?

ahadas (commented):
looks better now, thanks @bkhizgiy

pkg/forklift-api/webhooks/validating-webhook/admitters/secret-admitter.go
bkhizgiy: yes, too many if/else was a bit messy:)
I've moved the statement to be at the end of the big first else (this is the positive flow that means the credentials test passed)  so any of the cases there should return true.

pkg/forklift-api/webhooks/validating-webhook/admitters/secret-admitter.go
bkhizgiy: this is part of the original admitter (not related to my changes), because we only check the label `createdForProviderType` exists and not the provider type itself for all of the providers, and the matching Test() function should be called (IIUC), if it's not yet supported then we should switch the check in the beginning to be only for ovirt.

ahadas: ah you're right

ahadas (commented):
now the code looks good but I have second thoughts about adding this test..
it was needed when we didn't gather the engine's CA - so we thought about testing download-image in order to see if something is missing, but now that we add that certificate.. there are three options:
1. we get the engine's CA, download-image succeeds - good (this check passes)
2. we get the engine's CA, download-image fails because imageio is set with a different certificate - we should fail but we also don't really care about this case (it's not a supported configuration)
3. we get the engine's CA,  download-image fails for other reason - I'm not sure if we always want to fail here (let's say the user doesn't have permission on the OVF store that we picked)

looking at the code, it differentiates (2) and (3) above (`Test credentials succeeded but engine CA certificate is not valid, passing` and `CA certificate test failed, yet passing`, respectively) but can it be that isCaOK would be true and err != nil?

pkg/forklift-api/webhooks/validating-webhook/admitters/secret-admitter.go
```diff
@@ -71,7 +68,28 @@ func (admitter *SecretAdmitter) Admit(ar *admissionv1.AdmissionReview) *admissio
...
+					if isCaOK, err := ovirt.TestDownloadOvfStore(secret, log); isCaOK {
+						if err != nil {
+							log.Info("CA certificate test failed, yet passing", "error", err.Error())
+						} else {
+							log.Info("Test credentials and CA certificate succeeded, passing")
+						}
+					} else {
+						log.Info("Test credentials succeeded but engine CA certificate is not valid, passing")
+						return &admissionv1.AdmissionResponse{
+							Allowed: true,
```
ahadas, line 83: false, right?

ahadas:
> now the code looks good but I have second thoughts about adding this test.. it was needed when we didn't gather the engine's CA - so we thought about testing download-image in order to see if something is missing, but now that we add that certificate.. there are three options:
> 
> 1. we get the engine's CA, download-image succeeds - good (this check passes)
> 2. we get the engine's CA, download-image fails because imageio is set with a different certificate - we should fail but we also don't really care about this case (it's not a supported configuration)
> 3. we get the engine's CA,  download-image fails for other reason - I'm not sure if we always want to fail here (let's say the user doesn't have permission on the OVF store that we picked)
> 
> looking at the code, it differentiates (2) and (3) above (`Test credentials succeeded but engine CA certificate is not valid, passing` and `CA certificate test failed, yet passing`, respectively) but can it be that isCaOK would be true and err != nil?

actually, now I see that we always return 'Allowed: true' - I thought it's a typo but the log message also says "passing".. so maybe we really need to think about what we try to achieve by this

bennyz:
> > looking at the code, it differentiates (2) and (3) above (`Test credentials succeeded but engine CA certificate is not valid, passing` and `CA certificate test failed, yet passing`, respectively) but can it be that isCaOK would be true and err != nil?
> 
> actually, now I see that we always return 'Allowed: true' - I thought it's a typo but the log message also says "passing".. so maybe we really need to think about what we try to achieve by this

I suppose we do not want to accept the provider if we cannot perform the transfer. But there is a question of what is the scope of webhook, if the credentials work and the required certificates were added, which should always be true, since we fetch them automatically (except the third party). So at this point there is no reason to reject the provider, because the transfer should not fail because of certificate issues.
So yes, unless we do want to check if a third party certificate is needed, we don't really care about the result of the transfer...

PR #76, opened 2023-01-04
Upgrade virt-v2v (el8)

ahadas:
Using bazeldnf v5.4.0 for the first time


liranr23:
That't pretty strange that RPMs went to a lower version, isn't it?

ahadas:
> That't pretty strange that RPMs went to a lower version, isn't it?

yep, which ones?

liranr23:
> > That't pretty strange that RPMs went to a lower version, isn't it?
> 
> yep, which ones?

I could swear that I saw lower versions (I looked into workspace, not the rpm list), but looking again it's OK. Maybe I didn't look on the date for some of them, which did update.

liranr23 (approved):

PR #79, opened 2023-01-08
openstack: Add error checks for GetUpdates methods

bennyz:


ahadas (commented):
the error checks are fine but the other resources seem like a WIP, no?

bennyz:
> the error checks are fine but the other resources seem like a WIP, no?

yeah it was not supposed to go in, I'll remove it

ahadas (approved):

PR #94, opened 2023-01-09
added e2e tests for vsphere migration

eslutsky:


eslutsky:
we need this  https://github.com/kubev2v/forklift/pull/36  merged before the migration flow can complete.

tests/suit/migration_test.go
```diff
@@ -57,5 +57,12 @@ var _ = Describe("[level:component]Migration tests for vSphere provider", func()
 		err = utils.WaitForPlanReadyWithTimeout(f.CrClient, f.Namespace.Name, "plan-test", 15*time.Second)
 		Expect(err).ToNot(HaveOccurred())
 
+		By("Creating migration")
+		migrationDef := utils.NewMigration(provider.Namespace, "migration-test", "plan-test")
```
ahadas, line 61: I see the test keeps failing but unrelated to that - please define a constant instead of duplicating "plan-test" as suggested by SonarCloud

eslutsky: moved the managed objects names to constants, 
i think its failing because  of the conversion v2v Pod  still requests  kvm caps.

eslutsky: it seems we were missing this part https://github.com/kubev2v/forklift/pull/106

.github/workflows/pull-request.yml
```diff
@@ -51,6 +51,12 @@ jobs:
           kind get kubeconfig > /tmp/kubeconfig
           GOPATH=${GITHUB_WORKSPACE}/go make e2e-sanity
 
+      - name: show abnormal events and all objects in konveyor-forklift
+        if: always()
```
ahadas, line 55: so why do we need this statement? wouldn't it make more sense to run this only in some 'debug' mode?

.github/workflows/pull-request.yml
```diff
@@ -51,6 +51,12 @@ jobs:
...
+        run: |
+          kubectl get events --field-selector type!=Normal -A --sort-by='.lastTimestamp'
+          oc get all -n konveyor-forklift -o yaml
```
ahadas, line 58: I don't the last part in [the output of job 6721780653](https://github.com/kubev2v/forklift/actions/runs/3931038305/jobs/6721780653)  and we don't assume 'oc' is installed..

ahadas (commented):
looks good besides the `show abnormal events and all objects in konveyor-forklift` step

.github/workflows/pull-request.yml
eslutsky: well it will execute the step even if the previous steps fails.

.github/workflows/pull-request.yml
eslutsky: well the output of the `oc get all` command was really helpful to show all the  pods env variables ie , which is very useful for debugging.  it can be replaced to `kubectl get all`

.github/workflows/pull-request.yml
ahadas: ack

ahadas (approved):
the `show abnormal events and all objects in konveyor-forklift` is useful in specific cases but as the output is hidden by default and it takes about a second to generate it, I'm ok with having it this way. the other parts look great

PR #95, opened 2023-01-10
vsphere: Allow insecure connection to the server

nyoxi:
This is a preparation for future changes. Virt-v2v and libvirt cannot handle SSL certificates that it cannot verify (e.g. self-signed) and need a flag that tell it to disable checks. This does not mean we don't need the thumbprint (certificate fingerprint). Thumbprint is still needed by controller and will remain necessary for conversion pod when VDDK is used.

### DIFF
--- a/pkg/controller/provider/BUILD.bazel
+++ b/pkg/controller/provider/BUILD.bazel
@@ -14,6 +14,7 @@ go_library(
         "//pkg/controller/base",
         "//pkg/controller/provider/container",
         "//pkg/controller/provider/container/ovirt",
+        "//pkg/controller/provider/container/vsphere",
         "//pkg/controller/provider/model",
         "//pkg/controller/provider/web",
         "//pkg/controller/validation/policy",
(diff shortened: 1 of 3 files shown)


nyoxi:
We will need some UI changes for this to be useful. @yaacov, @rszwajko

yaacov:
@nyoxi greate thanks !

@sjd78 , you just asked about how to solve the missing ysphere CAs :tada:

nyoxi:
> @sjd78 , you just asked about how to solve the missing ysphere CAs tada

Missing in what context?

yaacov:
> Missing in what context?

yes, in the UI we will need to let users manually enter the CA manually, because we are dropping the service that got it automatically, letting users that can use insecure connection check in "use insecure" instead of manually fetching the CA will be very helpful for the UI.

liranr23 (approved):

nyoxi:
> because we are dropping the service that got it automatically

I see, then this is a problem. We still require fingerprint of the certificate even if the 'Insecure' checkbox is checked so there either has to be manual way to provide the certificate or at least the fingerprint.

It is not done in the patch, but it should be possible to modify the controller to blindly trust the certificate and compute the fingerprint when needed (e.g. when connecting to the provider or starting importer pod).

ahadas (commented):
minor comment, other than that lgtm

ahadas (approved):

PR #102, opened 2023-01-11
Bumping dependencies

liranr23:
Sign-off-by: Liran Rotenberg <lrotenbe@redhat.com>


cmd/forklift-controller/main.go
```diff
@@ -19,6 +19,8 @@ package main
 import (
 	"net/http"
 	"os"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
```
ahadas, line 23: the location of these imports looks wrong - did the IDE placed them there?

go.mod
```diff
@@ -115,15 +115,15 @@ replace github.com/openshift/client-go => github.com/openshift/client-go v0.0.0-
 // k8s deps pinning
 replace k8s.io/cluster-bootstrap => k8s.io/cluster-bootstrap v0.16.4
 
-replace k8s.io/api => k8s.io/api v0.19.3
+replace k8s.io/api => k8s.io/api v0.22.0
```
ahadas, line 118: I'm ok with limiting it this way if needed, but why is it needed?

cmd/forklift-controller/main.go
liranr23: i think so

go.mod
liranr23: without it, it will move towards newer version. i seeked for a smallest bump.

cmd/forklift-controller/main.go
ahadas: ok, that's a bit not standard and doesn't align with the rest of our code - standard imports are grouped together before the other imports. I've changed it but please take a look at the settings of the IDE then to see how to change it

go.mod
ahadas: sure, that's clear that without limiting the version, it will go with a higher version :)
my point was what blocked you from taking a higher version - but that's ok, we can bump them separately if there's no reason not to

PR #108, opened 2023-01-16
Cleanup in validation of provider secrets

ahadas:
This is a cleanup for #54 :

- Add debug information to logs in secret admitter
- Return proper error from secret-admitter


nyoxi (approved):
Probably missing word in commit message in `to let the user what
needs to change.` Otherwise LGTM.

PR #109, opened 2023-01-16
openstack: development checkpoint 

bennyz:
This PR is a checkpoint as a plan can currently be added and reconciled without crashing

### DIFF
--- a/pkg/controller/map/network/handler/openstack/handler.go
+++ b/pkg/controller/map/network/handler/openstack/handler.go
@@ -1,0 +1,116 @@
+package openstack
+
+import (
+	"path"
+	"strings"
+
+	api "github.com/konveyor/forklift-controller/pkg/apis/forklift/v1beta1"
+	"github.com/konveyor/forklift-controller/pkg/controller/provider/web/openstack"
+	"github.com/konveyor/forklift-controller/pkg/controller/watch/handler"
+	liberr "github.com/konveyor/forklift-controller/pkg/lib/error"
+	libweb "github.com/konveyor/forklift-controller/pkg/lib/inventory/web"
+	"github.com/konveyor/forklift-controller/pkg/lib/logging"
+	"golang.org/x/net/context"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+)
+
+// Package logger.
+var log = logging.WithName("networkMap|openstack")
+
+// Provider watch event handler.
+type Handler struct {
+	*handler.Handler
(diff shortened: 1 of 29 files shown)


pkg/controller/map/network/handler/openstack/handler.go
```diff
@@ -0,0 +1,116 @@
...
+	"github.com/konveyor/forklift-controller/pkg/controller/watch/handler"
+	liberr "github.com/konveyor/forklift-controller/pkg/lib/error"
+	libweb "github.com/konveyor/forklift-controller/pkg/lib/inventory/web"
+	"github.com/konveyor/forklift-controller/pkg/lib/logging"
+	"golang.org/x/net/context"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+)
+
+// Package logger.
+var log = logging.WithName("networkMap|ovirt")
```
ahadas, line 18:
```suggestion
var log = logging.WithName("networkMap|openstack")
```

pkg/controller/map/network/handler/openstack/handler.go
```diff
@@ -0,0 +1,116 @@
...
+			r.Provider().Name),
+		"watch",
+		w.ID())
+
+	return
+}
+
+// Resource created.
+func (r *Handler) Created(e libweb.Event) {
+	if network, cast := e.Resource.(*ovirt.Network); cast {
```
ahadas, line 49: do we really set it to ovirt.Network or how is it that it doesn't fail?

pkg/controller/provider/container/openstack/client.go
```diff
@@ -46,8 +46,8 @@ const (
 
 // Client struct
 type Client struct {
-	url                 string
-	secret              *core.Secret
+	URL                 string
+	Secret              *core.Secret
```
ahadas, line 50: for visibility, right?

pkg/controller/map/network/handler/openstack/handler.go
bennyz: I suppose it isn't called yet 😅

pkg/controller/provider/container/openstack/client.go
bennyz: yes, to make it usable in the plan

ahadas (commented):
lgtm but need to fix that issue that is reported by bazel

ahadas (approved):

PR #111, opened 2023-01-17
Add opinionated script for pushing to CRC

ahadas:


hack/deploy-to-crc.sh
```diff
@@ -0,0 +1,37 @@
...
+[[ -z "${REGISTRY_ACCOUNT}" ]] && export REGISTRY_ACCOUNT=openshift
+[[ -z "${REGISTRY_TAG}" ]] && export REGISTRY_TAG=devel
+
+CERT_PATH=/etc/pki/ca-trust/source/anchors/${REGISTRY}.crt
+if [[ ! -f ${CERT_PATH} ]]; then
+	oc get secret -n openshift-ingress router-certs-default -o go-template='{{index .data "tls.crt"}}' | base64 -d | sudo tee ${CERT_PATH}  > /dev/null
+	sudo update-ca-trust enable
+fi
+
+podman login -u kubeadmin -p $(oc whoami -t) $REGISTRY
```
mmartinv, line 22: Shouldn't it be `docker login...`?

nyoxi: Yes, well... the question of this already came up in other PR. My opinion is that we should prefer and advocate the use of Podman as an open source alternative to Docker. Using `docker` command everywhere gives a wrong image about our project and we should not get forced to this just because Basel Docker plugin needs it. And the fact that it makes us depend on `podman-docker` shouldn't be used as an excuse here either. The way I see it, we should prefer Podman, and if we want to give developers/users the freedom of choice between those we need to write the scripts in a configurable manner (despite `podman-docker`).

mmartinv: I agree with you on almost everything, but in that case there should be some kind of container runtime detection/selection because if the user chooses to use docker only,  the script will fail. We also need to enhance the documentation stating that either `podman-docker` or `docker` are required for the build. If not, the user might have `docker` and `podman` installed (and not `podman-docker`) and the build may work but this script will fail because the container image does not exist, and the push targets will fail the same way.

bkhizgiy:
Looking good, just a small suggestion - maybe it is worth adding a flag or variable that when set a 'bazel clean' action will be performed (or even just checking diffs on the brunch and if some files other than '.go' exists run it), just if we're already going for something entirely automated I think its worth covering all common cases, especially for the operator builds...

hack/deploy-to-crc.sh
ahadas: @mmartinv you're right but look at the comment at the top of the script - it's an opinionated script that assumes that 'podman' is installed. many projects don't provide such utilities to pushing images and logging in to the container registry, and we don't have to supply such either, but I think this script can serve us - and if we think it's better to use 'podman', we can go with podman here. developers that think differently can change the script and have their opinionated version :)

ahadas:
> Looking good, just a small suggestion - maybe it is worth adding a flag or variable that when set a 'bazel clean' action will be performed (or even just checking diffs on the brunch and if some files other than '.go' exists run it), just if we're already going for something entirely automated I think its worth covering all common cases, especially for the operator builds...

Thanks @bkhizgiy, about your suggestion - the problem I see with that is when the changes have already been committed in git (and then we'll need to compare the changes compared to the main branch / the branch we forked from), I wouldn't like to introduce partial solution that people would start relying on..

hack/deploy-to-crc.sh
```diff
@@ -0,0 +1,37 @@
...
+#
+# This is an opinionated script for deploying the images of all components within
+# the repository to an CRC instance. It is assumed that 'oc' is already configured
+# to connect to the CRC instance, 'podman' is installed and that the script is
+# executed from the root folder of this repository.
+# After executing this script there should be an operator named Forklift Operator
+# with 'Forklift (devel)' label available in the operator hub.
+
+set -e 
+
```
mmartinv, line 11:
```suggestion

CONTAINER_CMD=$(command -v podman)
[ ! -z "CONTAINER_CMD" ] || CONTAINER_CMD=$(command -v docker)

```

hack/deploy-to-crc.sh
mmartinv, line 22:
```suggestion
$CONTAINER_CMD login -u kubeadmin -p $(oc whoami -t) $REGISTRY
```

mmartinv: Detecting the container runtime and preferring `podman` over `docker` if both are installed is easy enough and I don't see a reason for not doing it within the script.

ahadas: accepted the changes you proposed - setting this as resolved

nyoxi (approved):

PR #113, opened 2023-01-17
Make the Makefile build fully customizable

mmartinv:
- Prefer `podman` over `docker` in Makefile builds
- Make the build fully customizable through environment
variables.
- Use `CONTROLLER_GEN` var to specify which `controller-gen`
binary must be used and make sure the correct version is
installed when using the default one.
- Make consistent use of parentheses instead of curly braces
for variable delimitation.
- Add sandbox writeable paths, needed after `bazel clean`.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -1,3 +1,6 @@
+# Container Runtime
+build --action_env=CONTAINER_CMD
+build --action_env=XDG_RUNTIME_DIR
 # Operator
 ## Default operator config
 build --action_env=VERSION=99.0.0
--- a/Makefile
+++ b/Makefile
@@ -1,23 +1,45 @@
+GOOS ?= `go env GOOS`
+GOPATH ?= `go env GOPATH`
+GOBIN ?= $(GOPATH)/bin
+GO111MODULE = auto
+
+CONTAINER_CMD ?= $(shell command -v podman)
+ifeq ($(CONTAINER_CMD),)
+CONTAINER_CMD := $(shell command -v docker)
+endif
+
 REGISTRY ?= quay.io
 REGISTRY_ACCOUNT ?= kubev2v
 REGISTRY_TAG ?= devel
-CONTROLLER_IMAGE ?= ${REGISTRY}/${REGISTRY_ACCOUNT}/forklift-controller:${REGISTRY_TAG}
-API_IMAGE ?= ${REGISTRY}/${REGISTRY_ACCOUNT}/forklift-api:${REGISTRY_TAG}
-VALIDATION_IMAGE ?= ${REGISTRY}/${REGISTRY_ACCOUNT}/forklift-validation:${REGISTRY_TAG}
-VIRT_V2V_IMAGE_COLD ?= ${REGISTRY}/${REGISTRY_ACCOUNT}/forklift-virt-v2v:${REGISTRY_TAG}
-VIRT_V2V_IMAGE_WARM ?= ${REGISTRY}/${REGISTRY_ACCOUNT}/forklift-virt-v2v-warm:${REGISTRY_TAG}
-OPERATOR_IMAGE ?= ${REGISTRY}/${REGISTRY_ACCOUNT}/forklift-operator:${REGISTRY_TAG}
-OPERATOR_BUNDLE_IMAGE ?= ${REGISTRY}/${REGISTRY_ACCOUNT}/forklift-operator-bundle:${REGISTRY_TAG}
-OPERATOR_INDEX_IMAGE ?= ${REGISTRY}/${REGISTRY_ACCOUNT}/forklift-operator-index:${REGISTRY_TAG}
-GOOS ?= `go env GOOS`
-GOBIN ?= ${GOPATH}/bin
-GO111MODULE = auto
 
-ifeq (, $(shell which docker))
-    CONTAINER_CMD = podman
-else
-    CONTAINER_CMD = docker
-endif
+VERSION ?= 99.0.0
+NAMESPACE ?= konveyor-forklift
+CHANNELS ?= development
+DEFAULT_CHANNEL ?= development
+
+# Use OPM_OPTS="--use-http" when using a non HTTPS registry
+OPM_OPTS ?=
+
+# By default use the controller gen installed by the
+# 'controller-gen' target
+DEFAULT_CONTROLLER_GEN = $(GOBIN)/controller-gen
+CONTROLLER_GEN ?= $(DEFAULT_CONTROLLER_GEN)
+
+# Image URLs to use all building/pushing image targets
+CONTROLLER_IMAGE ?= $(REGISTRY)/$(REGISTRY_ACCOUNT)/forklift-controller:$(REGISTRY_TAG)
+API_IMAGE ?= $(REGISTRY)/$(REGISTRY_ACCOUNT)/forklift-api:$(REGISTRY_TAG)
+VALIDATION_IMAGE ?= $(REGISTRY)/$(REGISTRY_ACCOUNT)/forklift-validation:$(REGISTRY_TAG)
+VIRT_V2V_IMAGE_COLD ?= $(REGISTRY)/$(REGISTRY_ACCOUNT)/forklift-virt-v2v:$(REGISTRY_TAG)
+VIRT_V2V_IMAGE_WARM ?= $(REGISTRY)/$(REGISTRY_ACCOUNT)/forklift-virt-v2v-warm:$(REGISTRY_TAG)
+OPERATOR_IMAGE ?= $(REGISTRY)/$(REGISTRY_ACCOUNT)/forklift-operator:$(REGISTRY_TAG)
+OPERATOR_BUNDLE_IMAGE ?= $(REGISTRY)/$(REGISTRY_ACCOUNT)/forklift-operator-bundle:$(REGISTRY_TAG)
+OPERATOR_INDEX_IMAGE ?= $(REGISTRY)/$(REGISTRY_ACCOUNT)/forklift-operator-index:$(REGISTRY_TAG)
+
+### External images
+MUST_GATHER_IMAGE ?= quay.io/kubev2v/forklift-must-gather:latest
+MUST_GATHER_API_IMAGE ?= quay.io/kubev2v/forklift-must-gather-api:latest
+UI_IMAGE ?= quay.io/kubev2v/forklift-ui:latest
+UI_PLUGIN_IMAGE ?= quay.io/kubev2v/forklift-console-plugin:latest
 
 ci: all
 
@@ -57,7 +79,7 @@ deploy: manifests
 
 # Generate manifests e.g. CRD, Webhooks
 manifests: controller-gen
-	${CONTROLLER_GEN} crd rbac:roleName=manager-role webhook paths="./pkg/apis/..." output:dir=operator/config/crd/bases
+	$(CONTROLLER_GEN) crd rbac:roleName=manager-role webhook paths="./pkg/apis/..." output:dir=operator/config/crd/bases
 
 # Run go fmt against code
 fmt:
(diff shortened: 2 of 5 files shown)


Makefile
```diff
@@ -68,56 +91,79 @@ vet:
...
-	${CONTROLLER_GEN} object:headerFile="./hack/boilerplate.go.txt" paths="./pkg/apis/..."
+	$(CONTROLLER_GEN) object:headerFile="./hack/boilerplate.go.txt" paths="./pkg/apis/..."
 
 build-controller-image:
 	bazel run cmd/forklift-controller:forklift-controller-image
 
 push-controller-image: build-controller-image
-	$(CONTAINER_CMD) tag cmd/forklift-controller:forklift-controller-image ${CONTROLLER_IMAGE}
-	$(CONTAINER_CMD) push ${CONTROLLER_IMAGE}
+	$(CONTAINER_CMD) tag bazel/cmd/forklift-controller:forklift-controller-image $(CONTROLLER_IMAGE)
```
nyoxi, line 100: If we want to be precise it should be `localhost/bazel/...`. Or at least that's how it looks on my env. Or is that perhaps different when Docker is used instead of Podman?

This also applies to all the other occurrences in Makefile.

nyoxi (changes requested):

Makefile
ahadas: > If we want to be precise it should be `localhost/bazel/...`. Or at least that's how it looks on my env. Or is that perhaps different when Docker is used instead of Podman?

yeah, Miguel and I also suspected it depends on Docker vs. Podman - since I got the same image tag as you with Podman

mmartinv: Prepending `localhost` to the image name is something specific to podman, docker does not do that at all. On the other hand, podman is smart enough when a "hostless" image is specified and it prepends `localhost` to it.

mmartinv: I also realized that the `CONTAINER_CMD` variable is not necessary, we should use `docker` command always as  it's the only runtime supported by bazel's container rules. To use `podman` under the hood `podman-docker` package is necessary so the `docker` command exists to be used by bazel.

ahadas: > On the other hand, podman is smart enough when a "hostless" image is specified and it prepends `localhost` to it.

right, I just verified that these changes work for me with podman

ahadas (approved):

Makefile
```diff
@@ -127,11 +165,9 @@ bazel-generate:
...
 controller-gen:
-ifeq (, $(shell which controller-gen))
+ifeq (, $(CONTROLLER_GEN))
 	@{ \
 	go install sigs.k8s.io/controller-tools/cmd/controller-gen@v0.10.0 ;\
 	}
-CONTROLLER_GEN=$(GOBIN)/controller-gen
-else
-CONTROLLER_GEN=$(shell which controller-gen)
+CONTROLLER_GEN = $(shell command -v controller-gen)
```
nyoxi, line 172: Could you please elaborate on why you chose to switch from `which` to `command -v`? While the commands are not exactly the same it shouldn't cause trouble in this context. It just may seem inconsistent to causal reader.

mmartinv: I thought it would minimize dependencies as `command` is part of the `bash` package and the `which` command has his own `which` package. Of course I am assuming everybody is using `bash` and maybe it's a big assumption. I think this change is not important so I can revert it if you want.

nyoxi: What I really meant is that on some places (e.g. `CONTAINER_CMD`) there is `which` and on other places (e.g. `CONTROLLER_GEN`) there is `command -v`. I don't mind using either of those, but we should try to make it consistent unless there are some reasons why not to.

mmartinv: Yeah, you are right, I realized yesterday when I was going to revert that change.  Other shells like `zsh` also have the built-in command `command -v`  so I will keep this change and make it consistent within the Makefile

Makefile
```diff
@@ -127,11 +171,9 @@ bazel-generate:
 # find or download controller-gen
 # download controller-gen if necessary
 controller-gen:
-ifeq (, $(shell which controller-gen))
+ifeq (, $(CONTROLLER_GEN))
 	@{ \
 	go install sigs.k8s.io/controller-tools/cmd/controller-gen@v0.10.0 ;\
 	}
-CONTROLLER_GEN=$(GOBIN)/controller-gen
-else
```
nyoxi, line 135: I don't think you can skip the `if-else` here. GOBIN path is not in PATH so you need to hard set the `CONTROLLER_GEN` when you install it with `go install`.

Related to this, please have a look at the PR here: https://github.com/kubev2v/forklift-controller/pull/463. One of the thing it does is it fixes the `controller-gen` target. What we have here is essentially broken and does not work people would think (hint: `controller-gen` is empty target and does not do anything). I closed it because with our transition to Basel I believed it won't be needed and we will drop the Makefile completely. But since it is not the case the PR is still relevant and should be applied.

mmartinv: Again, you are right, I took the following approach: use `$(GOBIN)/controller-gen` as the default command, and if not changed by setting the `CONTROLLER_GEN` variable, make sure the required version is installed. what do you think?

nyoxi: I like the approach. Now fixing the target will be even easier since we won't need `if-else` anymore (make will figure it on it's own). Something like this should be enough:

```
.PHONY: controller-gen
controller-gen: $(CONTROLLER_GEN)
$(DEFAULT_CONTROLLER_GEN):
	go install sigs.k8s.io/controller-tools/cmd/controller-gen@v0.10.0
```

Makefile
```diff
@@ -1,22 +1,46 @@
...
+# That is to ensure the appliance virt-v2v uses would not
+# run in emulation mode, which is significantly slower
+VIRT_V2V_DONT_REQUEST_KVM ?= false
+
+# Use OPM_OPTS="--use-http" when using a non HTTPS registry
+OPM_OPTS ?=
+
+# By default use the controller gen installed by the 
+# 'controller-gen' target 
+DEFAULT_CONTROLLER_GEN ?= $(GOBIN)/controller-gen
```
nyoxi, line 25: I suggest using `:=` here. I don't think we want users to change the default.

nyoxi (changes requested):

Makefile
mmartinv: Great! I just updated the code to reflect that.

Makefile
mmartinv: Done

nyoxi (approved):
Looks good. The only questionable part (from my POV) is the use of `shell which docker` versus `shell which podman`. I would prefer to have Podman as default but I am not going to block the PR on this.

Makefile
```diff
@@ -68,70 +92,92 @@ vet:
...
-	 $(CONTAINER_CMD) tag operator:forklift-operator-bundle-image ${OPERATOR_BUNDLE_IMAGE}
-	 $(CONTAINER_CMD) push ${OPERATOR_BUNDLE_IMAGE}
+	 $(CONTAINER_CMD) tag bazel/operator:forklift-operator-bundle-image $(OPERATOR_BUNDLE_IMAGE)
+	 $(CONTAINER_CMD) push $(OPERATOR_BUNDLE_IMAGE)
 
 build-operator-index-image:
-	bazel run operator:forklift-operator-index-image --action_env REGISTRY=${REGISTRY} --action_env REGISTRY_ACCOUNT=${REGISTRY_ACCOUNT} --action_env REGISTRY_TAG=${REGISTRY_TAG}
+	bazel run operator:forklift-operator-index-image \
+		--action_env VERSION=$(VERSION) \
+		--action_env NAMESPACE=$(NAMESPACE) \
```
ahadas, line 160: the namespace is not used by the operator-index, no?

Makefile
```diff
@@ -68,70 +92,92 @@ vet:
...
+		--action_env OPERATOR_IMAGE=$(OPERATOR_IMAGE) \
+		--action_env MUST_GATHER_IMAGE=$(MUST_GATHER_IMAGE) \
+		--action_env MUST_GATHER_API_IMAGE=$(MUST_GATHER_API_IMAGE) \
+		--action_env UI_IMAGE=$(UI_IMAGE) \
+		--action_env UI_PLUGIN_IMAGE=$(UI_PLUGIN_IMAGE) \
+		--action_env VALIDATION_IMAGE=$(VALIDATION_IMAGE) \
+		--action_env VIRT_V2V_IMAGE=$(VIRT_V2V_IMAGE) \
+		--action_env CONTROLLER_IMAGE=$(CONTROLLER_IMAGE) \
+		--action_env API_IMAGE=$(API_IMAGE) \
+		--action_env VIRT_V2V_DONT_REQUEST_KVM=$(VIRT_V2V_DONT_REQUEST_KVM)
```
ahadas, line 151: does it mean that it would be added to ForkliftController or why do we pass it here? we didn't want this option to be so visible

Makefile
```diff
@@ -1,22 +1,46 @@
...
-OPERATOR_INDEX_IMAGE ?= ${REGISTRY}/${REGISTRY_ACCOUNT}/forklift-operator-index:${REGISTRY_TAG}
-GOOS ?= `go env GOOS`
-GOBIN ?= ${GOPATH}/bin
-GO111MODULE = auto
 
-ifeq (, $(shell which docker))
-    CONTAINER_CMD = podman
-else
-    CONTAINER_CMD = docker
-endif
```
ahadas, line 19: this small part is really catching the eye - what do we get out of dropping this? I agree with @nyoxi that it would be better not to require users to install docker or podman-docker but to default to podman

Makefile
mmartinv: True, I will fix it in the next push.

Makefile
mmartinv: Yes, the idea was to have everything customizable from the Makefile so we don't need to run any `bazel` command by hand, but I can delete this var if you don't want it to be visible within the Makefile.

Makefile
mmartinv: I might be wrong but AFAIK `docker` command is currently mandatory for `bazel` to be able to build the images (https://github.com/bazelbuild/rules_docker#setup), so either `docker` or `podman-docker` is mandatory in the environment. My thought was: if `docker` command is going to be required anyway what's the point in having this kind of logic?

On the other hand, we could try  to make `docker_toolchain_configure` in bazel's image builds to point `podman` cmd instead of `docker`, this way we would be able to prefer `podman` and we wouldn't need to install `podman-docker`. Not sure if this will work though.

ahadas: I think I used podman, without docker or podman-docker, after we switched to bazel but I may not remember it correctly (although I also see the path to docker is optional and the statement: `These rules do not require / use Docker for pulling, building, or pushing images` in the link above)

mmartinv: In my env the build fails if `docker` command is not available:
```
bazel clean ; make build-operator-index-image
INFO: Starting clean (this may take a while). Consider using --async if the clean takes more than several minutes.
bazel run operator:forklift-operator-index-image \
	--verbose_failures \
	--action_env VERSION=99.0.0 \
	--action_env CHANNELS=development \
	--action_env DEFAULT_CHANNEL=development \
	--action_env OPT_OPTS= \
	--action_env REGISTRY=default-route-openshift-image-registry.apps.ocp4.example.com \
	--action_env REGISTRY_TAG=devel \
	--action_env REGISTRY_ACCOUNT=mmartinv
INFO: Analyzed target //operator:forklift-operator-index-image (104 packages loaded, 9145 targets configured).
INFO: Found 1 target...
Target //operator:forklift-operator-index-image up-to-date:
  bazel-out/k8-fastbuild-ST-4a519fd6d3e4/bin/operator/forklift-operator-index-image-layer.tar
INFO: Elapsed time: 3.277s, Critical Path: 1.92s
INFO: 63 processes: 17 internal, 46 linux-sandbox.
INFO: Build completed successfully, 63 total actions
INFO: Build completed successfully, 63 total actions
error: docker not found; do you need to manually configure the docker toolchain?
make: *** [Makefile:161: build-operator-index-image] Error 1
```

Adding the following to the `WORKSPACE` file:
```
load("@io_bazel_rules_docker//toolchains/docker:toolchain.bzl",
    docker_toolchain_configure="toolchain_configure"
)

docker_toolchain_configure(
  name = "docker_config",
  docker_path="/usr/bin/podman",
  docker_flags = [
    "--log-level=info",
  ],
)
```

seems to fix the issue:
```
bazel clean ; make build-operator-index-image
INFO: Starting clean (this may take a while). Consider using --async if the clean takes more than several minutes.
bazel run operator:forklift-operator-index-image \
	--verbose_failures \
	--action_env VERSION=99.0.0 \
	--action_env CHANNELS=development \
	--action_env DEFAULT_CHANNEL=development \
	--action_env OPT_OPTS= \
	--action_env REGISTRY=default-route-openshift-image-registry.apps.ocp4.example.com \
	--action_env REGISTRY_TAG=devel \
	--action_env REGISTRY_ACCOUNT=mmartinv
INFO: Analyzed target //operator:forklift-operator-index-image (104 packages loaded, 9145 targets configured).
INFO: Found 1 target...
Target //operator:forklift-operator-index-image up-to-date:
  bazel-out/k8-fastbuild-ST-4a519fd6d3e4/bin/operator/forklift-operator-index-image-layer.tar
INFO: Elapsed time: 3.366s, Critical Path: 2.02s
INFO: 63 processes: 17 internal, 46 linux-sandbox.
INFO: Build completed successfully, 63 total actions
INFO: Build completed successfully, 63 total actions
INFO[0000] /usr/bin/podman filtering at log level info  
INFO[0000] Setting parallel job count to 25             
Getting image source signatures
Copying blob 3dddd78e9f35 skipped: already exists  
Copying blob 1baf2ef4355a skipped: already exists  
Copying blob c8ff0b214684 skipped: already exists  
Copying blob 03dea58e7c9a skipped: already exists  
Copying blob cc6e9fe26463 skipped: already exists  
Copying blob 84ff92691f90 skipped: already exists  
Copying blob c456571abc85 skipped: already exists  
Copying config 013957b72a done  
Writing manifest to image destination
Storing signatures
Loaded image: sha256:013957b72a77501557c2925b1053cba6b51c34e94ab3e90a6ef0a94ecc9b0e28
Tagging 013957b72a77501557c2925b1053cba6b51c34e94ab3e90a6ef0a94ecc9b0e28 as bazel/operator:forklift-operator-index-image
INFO[0000] /usr/bin/podman filtering at log level info  
INFO[0000] Setting parallel job count to 25             
```
I also found other problems with `opm render` command using `~/.docker/config.json` by default (https://github.com/operator-framework/operator-registry/issues/919) but I think we could try to configure the bazel toolchain to use podman instead of docker and pass `DOCKER_CONFIG` env var to `opm` depending on which runtime is in use.

ahadas: ack, but then we'll need to install podman on the github-hosted runners?

Makefile
mmartinv: Removed from the Makefile.

.bazelrc
```diff
@@ -1,3 +1,6 @@
+# Container Runtime
+build --action_env=CONTAINER_CMD=/usr/bin/docker
```
ahadas, line 2: why do we need to set this?

ahadas (commented):
lgtm, just one comment about setting CONTAINER_CMD in .bazelrc

.bazelrc
mmartinv: In case both docker and podman are installed this is used to force the use of one of them. This could be useful to test builds with both container runtimes without having to remove podman to test docker.

Makefile
```diff
@@ -69,77 +96,126 @@ vet:
...
 
 push-all-images: push-api-image push-controller-image push-validation-image push-operator-image push-virt-v2v-image push-virt-v2v-warm-image push-operator-bundle-image push-operator-index-image
 
+.PHONY: check_container_runtmime
+check_container_runtmime:
+	  @if [ ! -x "$(CONTAINER_CMD)" ]; then \
+        echo "Container runtime was not automatically detected"; \
+        echo "Please install podman or docker and make sure it's available in PATH"; \
+        exit 1; \
+    fi
```
ahadas, line 213: the indentation here seems wrong, @mmartinv can you please check why?

ahadas (commented):
minor comment about indentation, other than that - lgtm
@mnecas can you please take a look at the operator part?

ahadas (approved):

.bazelrc
```diff
@@ -1,3 +1,6 @@
+# Container Runtime
+build --action_env=CONTAINER_CMD
+build --action_env=XDG_RUNTIME_DIR
```
mnecas, line 3: Do we need to set these variables?

Please add them to the README.md.
https://github.com/kubev2v/forklift/blob/main/README.md

mmartinv: No, those variables are taken from the environment, if `CONTAINER_CMD` is not set, the container runtime is detected automatically. `XDG_RUNTIME_DIR` is used by podman so it must exist in the environment, if it doesn't it means the runtime is docker and the variable is not needed/used. I will add `CONTAINER_CMD` to the README.

Makefile
mmartinv: Done

mnecas (approved):

PR #114, opened 2023-01-17
build: Build and push proper virt-v2v image

nyoxi:
We need to push the image with our entrypoint and not just the bare virt-v2v image.


ahadas (approved):
one can tell that I didn't run migrations from vsphere with that code :)

PR #115, opened 2023-01-17
Add ovirt populator deployment

liranr23:


.bazelrc
```diff
@@ -14,4 +14,6 @@ build --action_env=UI_IMAGE=quay.io/kubev2v/forklift-ui:latest
 build --action_env=UI_PLUGIN_IMAGE=quay.io/kubev2v/forklift-console-plugin:latest
 build --action_env=VALIDATION_IMAGE=quay.io/kubev2v/forklift-validation:latest
 build --action_env=VIRT_V2V_IMAGE=quay.io/kubev2v/forklift-virt-v2v:latest
+# build --action_env=OVIRT_POPULATOR_IMAGE=quay.io/kubev2v/ovirt-populator:latest
```
liranr23, line 17: this will be uncommented and the line below dropped once we post the image to `kubev2v`

operator/config/rbac/kustomization.yaml
```diff
@@ -15,6 +15,11 @@ resources:
 - forklift-controller_role.yaml
 - forklift-controller_role_binding.yaml
 
+# volume-data-source-validator roles
+- volume-data-source-validator/service_account.yaml
+- volume-data-source-validator/role.yaml
+- volume-data-source-validator/role_binding.yaml
```
liranr23, line 21: these for some reason don't apply.. @mnecas thoughts?

operator/config/crd/kustomization.yaml
```diff
@@ -10,4 +10,5 @@ resources:
 - bases/forklift.konveyor.io_plans.yaml
 - bases/forklift.konveyor.io_providers.yaml
 - bases/forklift.konveyor.io_storagemaps.yaml
+- bases/forklift.konveyor.io_ovirtimageiopopulators.yaml
```
liranr23, line 13: I need to add somehow the CRD of `VolumePopulator` (from [volume-data-source-validator](https://github.com/kubernetes-csi/volume-data-source-validator)). but it seems it requires also to add it into the pkg/api, is it? i would like to somehow apply it without it. @mnecas

pkg/apis/forklift/v1beta1/ovirtpopulator.go
```diff
@@ -0,0 +1,33 @@
+package v1beta1
+
+import (
+	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +k8s:openapi-gen=true
+type OvirtImageIOPopulator struct {
```
bennyz, line 10: we don't want to change the name to OvirtVolumePopulator?

README.md
```diff
@@ -44,9 +44,10 @@ Another option to override the default values can use `--action_env` as in the e
 | MUST_GATHER_IMAGE     | quay.io/kubev2v/forklift-must-gather:latest      | The forklift must gather an image.                          |
 | MUST_GATHER_API_IMAGE | quay.io/kubev2v/forklift-must-gather-api:latest  | The forklift must gather image api.                         |
 | UI_IMAGE              | quay.io/kubev2v/forklift-ui:latest               | The forklift UI image.                                      |
-| UI_PLUGIN_IMAGE       | quay.io/kubev2v/forklift-console-plugin:latest   | The forklift OKD/OpenShift UI plugin image.                                      |
+| UI_PLUGIN_IMAGE       | quay.io/kubev2v/forklift-console-plugin:latest   | The forklift OKD/OpenShift UI plugin image.                 |
 | VALIDATION_IMAGE      | quay.io/kubev2v/forklift-validation:latest       | The forklift validation image.                              |
 | VIRT_V2V_IMAGE        | quay.io/kubev2v/forklift-virt-v2v:latest         | The forklift virt v2v image.                                |
+| OVIRT_IMAGEIO_IMAGE   | quay.io/kubev2v/ovirt-populator:latest           | The forklift ovirt populator image.                         |
```
ahadas, line 50: there's another image for the ovirt-imageio container (in [quay.io/ovirt](https://quay.io/repository/ovirt/ovirt-imageio)), how about OVIRT_POPULATOR_IMAGE?

operator/config/crd/kustomization.yaml
ahadas: that's one option we talked about yesterday, or to extend the path that controller-gen looks at

operator/config/manifests/bases/forklift-operator.clusterserviceversion.yaml
```diff
@@ -73,6 +73,11 @@ spec:
       kind: Hook
       name: hooks.forklift.konveyor.io
       version: v1beta1
+    - description: Ovirt ImageIO Populator
```
ahadas, line 76:
```suggestion
    - description: oVirt ImageIO Populator
```

operator/config/rbac/role.yaml
```diff
@@ -60,3 +60,19 @@ rules:
...
   verbs:
   - '*'
+- apiGroups:
+  - apps
+  resources:
+  - statefulsets
+  verbs:
+  - get
+  - list
+  - create
```
ahadas, line 70: why do we need to add these for statefulsets?

operator/roles/forkliftcontroller/tasks/main.yml
```diff
@@ -80,6 +80,16 @@
...
 
+  - name: "Setup volume-data-source-validator deployment"
+    k8s:
+      state: present
+      definition: "{{ lookup('template', 'ovirt-populator/deployment-data-source-validator.yml.j2') }}"
+
+  - name: "Setup ovirt-populator deployment"
+    k8s:
+      state: present
+      definition: "{{ lookup('template', 'ovirt-populator/deployment-ovirt-populator.yml.j2') }}"
```
ahadas, line 91: shouldn't the populator be like the conversion pods we have with virt-v2v, why do we need a deployment?

pkg/apis/forklift/v1beta1/ovirtpopulator.go
ahadas: yeah, it also crossed my mind when seeing 'imageio' in other places before

operator/config/rbac/role.yaml
liranr23: to create the statefulset of the volume validator

operator/roles/forkliftcontroller/tasks/main.yml
liranr23: so we should create a pod each time we wish to transfer disk and remove it afterwards? there is already a pod from populator that doing the actual "job" and removed once the transfer finishes..

operator/config/rbac/role.yaml
ahadas: and why do we need a statefulset for the volume validator?

operator/roles/forkliftcontroller/tasks/main.yml
ahadas: > there is already a pod from populator that doing the actual "job" and removed once the transfer finishes..

ok, I thought you referred to this one - so you mean the controller?

operator/config/rbac/role.yaml
liranr23: To deploy it (basing on their repo). i think we might skip it and we should care about having the `VolumePopulator`. But it should verify the source. I didn't test it without having it

operator/roles/forkliftcontroller/tasks/main.yml
liranr23: we have a pod that listen and once triggered it calls the volume populator - which creates another pod (as we had importer pod per disk).

i'm not sure what you exactly mean

ahadas: ok, what you described above is the controller (so we can call it populator-controller to distinguish it from 'populators' which corresponds to the "other pod" you mentioned above) - that needs a deployment, ok

operator/config/rbac/role.yaml
ahadas: needs to see if it really needs to scale, if not - there's no need for a statefulset, a regular deployment would be alright

operator/config/rbac/kustomization.yaml
liranr23: we can do without them. the user may deploy it. we only need the `VolumePopulator` CRD.

operator/config/rbac/role.yaml
liranr23: dropped

operator/roles/forkliftcontroller/tasks/main.yml
liranr23: @bennyz this does act as controller. do you plan to use the same pod for OSP? or it should be distinguished to ovirt?

bennyz: openstack will have a separate pod

operator/config/rbac/role.yaml
```diff
@@ -60,3 +60,12 @@ rules:
   - certificates
   verbs:
   - '*'
+- apiGroups:
```
liranr23, line 63: probably needs to drop now.

operator/roles/forkliftcontroller/defaults/main.yml
```diff
@@ -101,6 +101,10 @@ api_selfsigned_issuer_name: "{{ api_service_name }}-selfsigned-issuer"
 api_issuer_name: "{{ api_service_name }}-issuer"
 api_certificate_name: "{{ api_service_name }}-selfsigned-ca"
 
+ovirt_populator_image_fqin: "{{ lookup( 'env', 'OVIRT_POPULATOR_IMAGE') or lookup( 'env', 'RELATED_IMAGE_OVIRT_POPULATOR') }}"
+ovirt_populator_deployment_name: "{{ app_name }}-ovirt-populator"
```
ahadas, line 105: this should also refer to the controller

operator/config/crd/bases/forklift.konveyor.io_ovirtvolumepopulators.yaml
```diff
@@ -0,0 +1,59 @@
+---
+apiVersion: apiextensions.k8s.io/v1
```
ahadas, line 2: can we make the CRD deployment optional? I think about the case that CDI would also deploy that CRD (I don't think that the flow of deploying forklift and then kubevirt is that likely to happen - it's more likely that kubevirt will be there already when forklift is deployed - would the operator just "skip" that or fail?)

operator/roles/forkliftcontroller/defaults/main.yml
```diff
@@ -101,6 +101,10 @@ api_selfsigned_issuer_name: "{{ api_service_name }}-selfsigned-issuer"
 api_issuer_name: "{{ api_service_name }}-issuer"
 api_certificate_name: "{{ api_service_name }}-selfsigned-ca"
 
+ovirt_populator_image_fqin: "{{ lookup( 'env', 'OVIRT_POPULATOR_IMAGE') or lookup( 'env', 'RELATED_IMAGE_OVIRT_POPULATOR') }}"
```
ahadas, line 104: there's a mixture here between the populator and the controller since they use the same image - as we talked offline, I think it wouldn't make sense in our case to have a controller per provider (ovirt, openstack and maybe 'libvirt/kvm' later on), I think we should have one controller, at least until we are able to push that to CDI (to simplify deployments on remote clusters)
but anyway, I think it's better to define two variables here - one for the populator image and one for the controller

operator/config/crd/bases/forklift.konveyor.io_ovirtvolumepopulators.yaml
liranr23: i don't think so. @mnecas am i right?

operator/roles/forkliftcontroller/tasks/main.yml
```diff
@@ -80,6 +80,11 @@
       definition: "{{ lookup('template', 'controller/provider-host.yml.j2') }}"
     when: "'kubevirt.io' in api_groups"
 
+  - name: "Setup ovirt-populator deployment"
```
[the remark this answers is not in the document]
liranr23: we need to align with them.
if they are going to use their own controller - we don't need to deploy here. we might need to add a `when` to know if we wish to deploy or not.
this might be tricky - let's say it's a version that CDI doesn't support vol-pop yet - we are good. but once the user upgrade to CDI that does support - we will somehow need to remove the deployment on our side (or CDI should do it for us) and vice versa.

about CRDs - depend on what they are planning - if it's a generic CRD for a global import usage, we will probably need to change the controller code to support their CRD and drop our CRD. if they are planning to use multiple CRDs (per source), we might tell them to use ours. i don't think it will matter much if we both deploy the same CRD(override?).

operator/roles/forkliftcontroller/defaults/main.yml
liranr23: as we talked offline - having the current state and then split them?

operator/config/crd/bases/forklift.konveyor.io_ovirtvolumepopulators.yaml
mnecas: This is our custom CRD, so this is fine.
Plus, if we would use some other CRD, like from the volume-data-source-validator. We should define it in our `owned` CRDs [1]. Multiple operators can own the CRD. Once some operator wants to change/update the CRD it will not let them until all operators allow it [2].

---

Dependency resolution of multiple operators [2]
```
OLM will upgrade CRD right away if it is owned by singular CSV. If CRD is owned by multiple CSVs, then CRD is upgraded when it is satisfied all of the following backward compatible conditions...
```

The design documentation philosophy [3].
```
An operator should declare the CRDs it owns and manages, as well as those that it expects to exist (but be managed by other operators).
```

[1] https://github.com/kubev2v/forklift/blob/main/operator/config/manifests/bases/forklift-operator.clusterserviceversion.yaml#L34-L75
[2] https://github.com/operator-framework/operator-lifecycle-manager/blob/master/doc/design/dependency-resolution.md
[3] https://github.com/operator-framework/operator-lifecycle-manager/blob/master/doc/design/philosophy.md

liranr23: the v olume-data-source-validator dropped. it works without it, although it's a good thing to verify the source ref. if a user wish to have it - he can deploy it.
as for the OvirtVolumePopulator - then i think it's fine.

README.md
```diff
@@ -34,21 +34,22 @@ The environment which you can set across all projects.
...
+| CONTROLLER_IMAGE           | quay.io/kubev2v/forklift-controller:latest      | The forklift controller image.                              |
+| MUST_GATHER_IMAGE          | quay.io/kubev2v/forklift-must-gather:latest     | The forklift must gather an image.                          |
+| MUST_GATHER_API_IMAGE      | quay.io/kubev2v/forklift-must-gather-api:latest | The forklift must gather image api.                         |
+| UI_IMAGE                   | quay.io/kubev2v/forklift-ui:latest              | The forklift UI image.                                      |
+| UI_PLUGIN_IMAGE            | quay.io/kubev2v/forklift-console-plugin:latest  | The forklift OKD/OpenShift UI plugin image.                 |
+| VALIDATION_IMAGE           | quay.io/kubev2v/forklift-validation:latest      | The forklift validation image.                              |
+| VIRT_V2V_IMAGE             | quay.io/kubev2v/forklift-virt-v2v:latest        | The forklift virt v2v image. See note below.                |
+| POPULATOR_CONTROLLER_IMAGE | quay.io/kubev2v/populator-controller:latest     | The forklift populator controller image.                    |
```
ahadas, line 52:
```suggestion
| POPULATOR_CONTROLLER_IMAGE | quay.io/kubev2v/populator-controller:latest     | The forklift volume-populator controller image.                    |
```

operator/config/manager/manager.yaml
```diff
@@ -56,6 +56,8 @@ spec:
           value: ${VIRT_V2V_IMAGE}
         - name: VIRT_V2V_DONT_REQUEST_KVM
           value: ${VIRT_V2V_DONT_REQUEST_KVM}
+        - name: OVIRT_POPULATOR_IMAGE
+          value: ${OVIRT_POPULATOR_IMAGE}
```
ahadas, line 60: so POPULATOR_CONTROLLER_IMAGE?

operator/roles/forkliftcontroller/defaults/main.yml
```diff
@@ -101,6 +101,10 @@ api_selfsigned_issuer_name: "{{ api_service_name }}-selfsigned-issuer"
 api_issuer_name: "{{ api_service_name }}-issuer"
 api_certificate_name: "{{ api_service_name }}-selfsigned-ca"
 
+populator_controller_image_fqin: "{{ lookup( 'env', 'POPULATOR_CONTROLLER_IMAGE') or lookup( 'env', 'RELATED_IMAGE_POPULATOR_CONTROLLER') }}"
+populator_controller_deployment_name: "{{ app_name }}-populator-controller"
```
ahadas, line 105: how about:
```suggestion
populator_controller_deployment_name: "{{ app_name }}-volume-populator-controller"
```

operator/roles/forkliftcontroller/tasks/main.yml
```diff
@@ -80,6 +80,11 @@
       definition: "{{ lookup('template', 'controller/provider-host.yml.j2') }}"
     when: "'kubevirt.io' in api_groups"
 
+  - name: "Setup populator deployment"
+    k8s:
+      state: present
+      definition: "{{ lookup('template', 'populator/deployment-populator.yml.j2') }}"
```
ahadas, line 86: would it make sense to check if the CRD is defined first to cope with a situation in which CDI deploys it first (assuming they'll use the same CRD - that's something we'll need to figure out)?

operator/config/crd/bases/forklift.konveyor.io_ovirtvolumepopulators.yaml
ahadas: > This is our custom CRD, so this is fine. 

Yes, but the context is that we'd like to let CDI deploy it so it would be available on remote clusters

operator/roles/forkliftcontroller/tasks/main.yml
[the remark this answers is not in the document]
ahadas: ok, so how about starting with making the deployment of the controller optional?

operator/roles/forkliftcontroller/templates/populator/deployment-populator.yml.j2
```diff
@@ -0,0 +1,27 @@
...
+      labels:
+        app: {{ app_name }}
+    spec:
+      serviceAccountName: forklift-controller
+      containers:
+        - name: {{ populator_controller_container_name }}
+          image: {{ populator_controller_image_fqin }}
+          imagePullPolicy: Always
+          args:
+            - --image-name={{ populator_ovirt_image_fqin }}
```
ahadas, line 22: should be 
```suggestion
            - --image-name={{ populator_controller_image_fqin }}
```
right?

ahadas (commented):
overall looks good but please also check that the operator wouldn't fail or override an existing populator CRD - besides that, the comments inside are pretty minor

operator/roles/forkliftcontroller/templates/populator/deployment-populator.yml.j2
liranr23: nope, in the args we will send the populator image to use (ovirt populator). in the future - with OSP. we will need 2 args here (e.g. `--ovirt-image-name` & `--osp-image-name`).

liranr23:
> overall looks good but please also check that the operator wouldn't fail or override an existing populator CRD - besides that, the comments inside are pretty minor

needs some more testing. i did check if the CRD exists(but with the forklift group) and the deployment override it.

liranr23:
> > overall looks good but please also check that the operator wouldn't fail or override an existing populator CRD - besides that, the comments inside are pretty minor
> 
> needs some more testing. i did check if the CRD exists(but with the forklift group) and the deployment override it.

if there is another CRD with another group (e.g `cdi.kubevirt.io`), then we will have 2 CRDs, one for each group. If it's the same group, the later deployment seems to override. however, i didn't implement it in the CDI operator, i played with CRDs manually.

operator/roles/forkliftcontroller/templates/populator/deployment-populator.yml.j2
ahadas: ack, it was confusing because it's different than the convention of passing-in those configurable image urls using environment variables - I change it to comply with this convention

PR #116, opened 2023-01-19
Openstack enhancements

mmartinv:


pkg/controller/map/storage/handler/openstack/handler.go
```diff
@@ -92,9 +130,42 @@ func (r *Handler) changed(models ...*openstack.Storage) {
 		for _, pair := range mp.Spec.Map {
 			ref := pair.Source
 			for _, ds := range models {
-				if ref.ID == ds.ID || strings.HasSuffix(ds.Path, ref.Name) {
```
ahadas, line 95: it's unrelated to this particular PR but how about renaming `mp` and `ds` to something more meaningful?

ahadas (commented):
looks good, minor comments

pkg/controller/provider/container/openstack/model.go
```diff
@@ -164,17 +177,89 @@ func (r *ProjectAdapter) GetUpdates(ctx *Context, lastSync time.Time) (updates [
...
+			updates = append(updates, updater)
+
+		default:
+			updater := func(tx *libmodel.Tx) (err error) {
+				m := &model.Image{
+					Base: model.Base{ID: image.ID},
+				}
+				err = tx.Get(m)
+				if err != nil {
+					if errors.Is(err, &NotFound{}) {
```
bennyz, line 231: should be
```suggestion
					if errors.Is(err, libmodel.NotFound) {
```
no? (same for the other models)

mmartinv: You are right, I guess this is part was copied from another provider. I will change it in the next push

bennyz (approved):

pkg/controller/map/storage/handler/openstack/handler.go
mmartinv: Done

pkg/controller/provider/container/openstack/client.go
```diff
@@ -201,12 +187,13 @@ func (r *Client) insecure() bool {
 }
 
 // List Servers.
-func (r *Client) list(resourceType ResourceType, listopts interface{}) (obj interface{}, err error) {
```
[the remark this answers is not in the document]
ahadas: ack, I find the previous way clearer but that's subjective of course

ahadas (approved):

ahadas:
@mmartinv there's an unsigned commit - please fix that and rebase

PR #118, opened 2023-01-19
Autoenable console plugin

mnecas:


ahadas (approved):

ahadas:
I know @yaacov also agrees with this so let's get it in

yaacov:
> I know @yaacov also agrees with this so let's get it in

Sorry to be late :-) for future ref, yes I very agree :tada:

PR #121, opened 2023-01-23
Introduce ovirt volume populator image

liranr23:
Current, the image is built using the dockerfile. Later it should migrate to bazel.


cmd/ovirt-populator/main.go
```diff
@@ -0,0 +1,262 @@
...
+	args = append(args, "--secret-name="+ovirtVolumePopulator.Spec.EngineSecretName)
+	args = append(args, "--disk-id="+ovirtVolumePopulator.Spec.DiskID)
+	args = append(args, "--engine-url="+ovirtVolumePopulator.Spec.EngineURL)
+	args = append(args, "--cr-name="+ovirtVolumePopulator.Name)
+	args = append(args, "--cr-namespace="+ovirtVolumePopulator.Namespace)
+
+	return args, nil
+}
+
+func populate(masterURL, kubeconfig, crName, engineURL, secretName, diskID, fileName, namespace, crNamespace string) {
```
eslutsky, line 160: can we add a timeout setting for disk transfers?

eslutsky:
do we have any functionality tests for this change ?

liranr23:
> do we have any functionality tests for this change ?

It's just an image. The test is the migration itself, once all the 3 related PRs will merge.
This image will be part of the forklift deployment, once we will migrate from oVirt to the local host it will use this image.

cmd/ovirt-populator/main.go
liranr23: TBH, i'm not sure that will be good. if the transfer being executed - would you like to suddenly hit a timeout? might be 1 byte left to transfer. also, it depends on network and size.
@bennyz what do you think?

bennyz: we have a 1 minute timeout (for idle requests) in imageio already, and ovirt-engine's limit is 50 hours which is fine as the transfer of a very large disk can theoretically take over day, so I don't think we need to add a timeout on our end as well, unless we find an actual issue

cmd/ovirt-populator/example/crd.yaml
```diff
@@ -0,0 +1,59 @@
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.10.0
+  creationTimestamp: null
+  name: ovirtimageiopopulators.forklift.konveyor.io
```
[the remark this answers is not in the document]
bennyz: now I see this the example, I guess we can remove this directory from here?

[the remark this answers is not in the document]
bennyz: yeah, we can keep the separate repo if anyone wants to use it as a standalone tool

cmd/ovirt-populator/Makefile
```diff
@@ -0,0 +1,8 @@
+GOBIN ?= ${GOPATH}/bin
+CONTROLLER_GEN=$(GOBIN)/controller-gen
+
+manifests:
+	${CONTROLLER_GEN} crd rbac:roleName=manager-role webhook paths="./pkg/..." output:crd:artifacts:config=config/crds/bases output:crd:dir=config/crds
+
+generate:
```
ahadas, line 7: all but `main.go` should move out of `cmd` - `ovirtpopulator.go` should be placed under `pkg/apis/..` and then it would be picked up by the existing targets in the main Makefile

cmd/ovirt-populator/go.mod
```diff
@@ -0,0 +1,73 @@
+module forklift.konveyor.io/ovirt-populator
+
```
ahadas, line 2: the build stuff should move to the `hack` folder, how about `hack/ovirt-populator`?

cmd/ovirt-populator/main.go
```diff
@@ -0,0 +1,262 @@
...
+		crName       string
+		crNamespace  string
+		httpEndpoint string
+		metricsPath  string
+		masterURL    string
+		kubeconfig   string
+		imageName    string
+		showVersion  bool
+		namespace    string
+	)
```
ahadas, line 55: that's a bit unusual to see.. if most of the variables are of the same type and there's no assignment, why not a single or multiple line(s) of
```golang
var x,y,z string
var showVersion bool
```

cmd/ovirt-populator/example/crd.yaml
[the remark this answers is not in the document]
ahadas: > yeah, we can keep the separate repo if anyone wants to use it as a standalone tool

or in the [operator's samples folder](https://github.com/kubev2v/forklift/tree/main/operator/config/samples)

cmd/ovirt-populator/main.go
```diff
@@ -0,0 +1,262 @@
+package main
+
+import (
```
ahadas, line 3: how about renaming the file `main.go` to `ovirt-populator.go`?

cmd/ovirt-populator/Containerfile
```diff
@@ -0,0 +1,18 @@
+FROM golang:1.19.3-alpine3.16
```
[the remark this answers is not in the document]
ahadas: you may want to do it in steps - building with this base image and then basing the container with the binary on centos-stream9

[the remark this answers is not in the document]
ahadas: sorry, not 'steps' but 'stages - [multi-stage builds](https://docs.docker.com/build/building/multi-stage/)

cmd/ovirt-populator/ovirt-populator.go
```diff
@@ -0,0 +1,252 @@
...
+
+	// Main arg
+	flag.StringVar(&mode, "mode", "", "Mode to run in (controller, populate)")
+	// Populate args
+	flag.StringVar(&engineUrl, "engine-url", "", "ovirt-engine url (https//engine.fqdn)")
+	flag.StringVar(&secretName, "secret-name", "", "secret containing oVirt credentials")
+	flag.StringVar(&diskID, "disk-id", "", "ovirt-engine disk id")
+	flag.StringVar(&fileName, "file-name", "", "File name to populate")
+	flag.StringVar(&crName, "cr-name", "", "Custom Resource instance name")
+	flag.StringVar(&crNamespace, "cr-namespace", "", "Custom Resource instancce namespace")
```
ahadas, line 51:
```suggestion
	flag.StringVar(&crNamespace, "cr-namespace", "", "Custom Resource instance namespace")
```

cmd/ovirt-populator/ovirt-populator.go
```diff
@@ -0,0 +1,252 @@
...
+	// Write credentials to files
+	ovirtPass, err := os.Create("/tmp/ovirt.pass")
+	if err != nil {
+		klog.Fatalf("Failed to create ovirt.pass %s", err)
+	}
+
+	defer ovirtPass.Close()
+	if err != nil {
+		klog.Fatalf("Failed to create file %s", err)
+	}
```
ahadas, line 162: what did you mean to check in this block?

cmd/ovirt-populator/ovirt-populator.go
```diff
@@ -0,0 +1,252 @@
...
+
+	cert, err := os.Create("/tmp/ca.pem")
+	if err != nil {
+		klog.Fatalf("Failed to create ca.pem %s", err)
+	}
+
+	defer cert.Close()
+	if err != nil {
+		klog.Fatalf("Failed to create file %s", err)
+	}
```
ahadas, line 173: and what did you mean to check here?

cmd/ovirt-populator/ovirt-populator.go
```diff
@@ -0,0 +1,252 @@
...
+	return args, nil
+}
+
+func populate(masterURL, kubeconfig, crName, engineURL, secretName, diskID, fileName, namespace, crNamespace string) {
+	engineConfig := getSecret(secretName, engineURL, namespace)
+
+	// Write credentials to files
+	ovirtPass, err := os.Create("/tmp/ovirt.pass")
+	if err != nil {
+		klog.Fatalf("Failed to create ovirt.pass %s", err)
```
ahadas, line 156: shouldn't it be %v?

cmd/ovirt-populator/ovirt-populator.go
```diff
@@ -0,0 +1,252 @@
...
+	defer cert.Close()
+	if err != nil {
+		klog.Fatalf("Failed to create file %s", err)
+	}
+
+	cert.Write([]byte(engineConfig.ca))
+
+	args := []string{
+		"download-disk",
+		"--log-level", "debug",
```
ahadas, line 179: hard-coded debug level?..

cmd/ovirt-populator/ovirt-populator.go
```diff
@@ -0,0 +1,252 @@
...
+		"--password-file=/tmp/ovirt.pass",
+		"--cafile=" + "/tmp/ca.pem",
+		"-f", "raw",
+		diskID,
+		fileName,
+	}
+
+	if err != nil {
+		klog.Fatal(err.Error())
+	}
```
ahadas, line 192: and what did you mean to check in this block?

cmd/ovirt-populator/ovirt-populator.go
```diff
@@ -0,0 +1,252 @@
...
+	var mode, engineUrl, secretName, diskID, fileName, crName, crNamespace, httpEndpoint, metricsPath, masterURL, kubeconfig, imageName, namespace string
+	var showVersion bool
+
+	// Main arg
+	flag.StringVar(&mode, "mode", "", "Mode to run in (controller, populate)")
+	// Populate args
+	flag.StringVar(&engineUrl, "engine-url", "", "ovirt-engine url (https//engine.fqdn)")
+	flag.StringVar(&secretName, "secret-name", "", "secret containing oVirt credentials")
+	flag.StringVar(&diskID, "disk-id", "", "ovirt-engine disk id")
+	flag.StringVar(&fileName, "file-name", "", "File name to populate")
```
ahadas, line 49: it's not a filename but rather a path to volume (file or block), no?

cmd/ovirt-populator/ovirt-populator.go
```diff
@@ -0,0 +1,252 @@
...
+	flag.StringVar(&metricsPath, "metrics-path", "/metrics", "The HTTP path where prometheus metrics will be exposed. Default is `/metrics`.")
+	// Other args
+	flag.BoolVar(&showVersion, "version", false, "display the version string")
+	flag.StringVar(&namespace, "namespace", "konveyor-forklift", "Namespace to deploy controller")
+	flag.Parse()
+
+	if showVersion {
+		fmt.Println(os.Args[0], version)
+		os.Exit(0)
+	}
```
ahadas, line 72: why do we need this?

cmd/ovirt-populator/ovirt-populator.go
```diff
@@ -0,0 +1,252 @@
...
+}
+
+type TransferProgress struct {
+	Transferred uint64  `json:"transferred"`
+	Description string  `json:"description"`
+	Size        *uint64 `json:"size,omitempty"`
+	Elapsed     float64 `json:"elapsed"`
+}
+
+func getSecret(secretName, engineURL, namespace string) engineConfig {
```
ahadas, line 103: it doesn't 'get' a secret but an engine config, right? how about `loadEngineConfig`?

ahadas (commented):
this image serves as both controller and populator, and as we talked offline, if we go this way we'll have a controller per-provider type that uses volume-populator which is not that great - but we can continue to think about that in the context of the other PR that handles the deployment

hack/ovirt-populator/Containerfile
```diff
@@ -0,0 +1,17 @@
...
+
+COPY . ./
+WORKDIR /app/cmd/ovirt-populator
+
+# When debugging by replacing the lib-volume-populator with a local modified copy
+# COPY . ./
+
+RUN CGO_ENABLED=0 go build -o /ovirt-populator
+
+FROM quay.io/centos/centos:stream9
```
ahadas, line 12: and I see there's stream9-minimal in [quay.io/centos/centos](https://quay.io/repository/centos/centos?tab=tags&tag=latest) - how about trying it?

cmd/ovirt-populator/ovirt-populator.go
liranr23: i guess if you wish to run the image as a standalone and just do version check. @bennyz fix me if i'm wrong?

bennyz: yeah it's copied from the example

ahadas: the example of a volume populator? unless I'm missing something - it is always be "unknown" here, right? I don't see an added value in having this..

cmd/ovirt-populator/ovirt-populator.go
ahadas: still there's no assignment to `err` after it was checked above

ahadas (commented):
minor comment, besides that lgtm

cmd/ovirt-populator/ovirt-populator.go
liranr23: correct, i forgot the assignment.

ahadas (approved):

PR #124, opened 2023-01-25
build: fix virt-v2v image builds

nyoxi:
The build targets for virt-v2v warm image and operator bundle were broken after commit a9043160.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,8 @@ REGISTRY_TAG ?= devel
 CONTROLLER_IMAGE ?= ${REGISTRY}/${REGISTRY_ACCOUNT}/forklift-controller:${REGISTRY_TAG}
 API_IMAGE ?= ${REGISTRY}/${REGISTRY_ACCOUNT}/forklift-api:${REGISTRY_TAG}
 VALIDATION_IMAGE ?= ${REGISTRY}/${REGISTRY_ACCOUNT}/forklift-validation:${REGISTRY_TAG}
-VIRT_V2V_IMAGE ?= ${REGISTRY}/${REGISTRY_ACCOUNT}/forklift-virt-v2v:${REGISTRY_TAG}
+VIRT_V2V_IMAGE_COLD ?= ${REGISTRY}/${REGISTRY_ACCOUNT}/forklift-virt-v2v:${REGISTRY_TAG}
+VIRT_V2V_IMAGE_WARM ?= ${REGISTRY}/${REGISTRY_ACCOUNT}/forklift-virt-v2v-warm:${REGISTRY_TAG}
 OPERATOR_IMAGE ?= ${REGISTRY}/${REGISTRY_ACCOUNT}/forklift-operator:${REGISTRY_TAG}
 OPERATOR_BUNDLE_IMAGE ?= ${REGISTRY}/${REGISTRY_ACCOUNT}/forklift-operator-bundle:${REGISTRY_TAG}
 OPERATOR_INDEX_IMAGE ?= ${REGISTRY}/${REGISTRY_ACCOUNT}/forklift-operator-index:${REGISTRY_TAG}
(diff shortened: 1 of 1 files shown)


ahadas (approved):

Makefile
```diff
@@ -102,15 +103,15 @@ build-virt-v2v-image:
...
 	bazel run virt-v2v:forklift-virt-v2v-warm
 
 push-virt-v2v-warm-image: build-virt-v2v-warm-image
-	$(CONTAINER_CMD) tag virt-v2v:forklift-virt-v2v-warm ${VIRT_V2V_IMAGE}
-	$(CONTAINER_CMD) push ${VIRT_V2V_IMAGE}
+	$(CONTAINER_CMD) tag virt-v2v:forklift-virt-v2v-warm ${VIRT_V2V_IMAGE_WARM}
+	$(CONTAINER_CMD) push ${VIRT_V2V_IMAGE_WARM}
 
 build-operator-bundle-image:
 	bazel run operator:forklift-operator-bundle-image --action_env CONTROLLER_IMAGE=${CONTROLLER_IMAGE} --action_env VALIDATION_IMAGE=${VALIDATION_IMAGE} --action_env OPERATOR_IMAGE=${OPERATOR_IMAGE} --action_env VIRT_V2V_IMAGE=${VIRT_V2V_IMAGE_COLD} --action_env API_IMAGE=${API_IMAGE}
```
mmartinv, line 117:
```suggestion
	bazel run operator:forklift-operator-bundle-image --action_env CONTROLLER_IMAGE=${CONTROLLER_IMAGE} --action_env VALIDATION_IMAGE=${VALIDATION_IMAGE} --action_env OPERATOR_IMAGE=${OPERATOR_IMAGE} --action_env VIRT_V2V_IMAGE="${VIRT_V2V_IMAGE_COLD}|${VIRT_V2V_IMAGE_WARM}" --action_env API_IMAGE=${API_IMAGE}
```

Makefile
```diff
@@ -4,7 +4,8 @@ REGISTRY_TAG ?= devel
 CONTROLLER_IMAGE ?= ${REGISTRY}/${REGISTRY_ACCOUNT}/forklift-controller:${REGISTRY_TAG}
 API_IMAGE ?= ${REGISTRY}/${REGISTRY_ACCOUNT}/forklift-api:${REGISTRY_TAG}
 VALIDATION_IMAGE ?= ${REGISTRY}/${REGISTRY_ACCOUNT}/forklift-validation:${REGISTRY_TAG}
-VIRT_V2V_IMAGE ?= ${REGISTRY}/${REGISTRY_ACCOUNT}/forklift-virt-v2v:${REGISTRY_TAG}
+VIRT_V2V_IMAGE_COLD ?= ${REGISTRY}/${REGISTRY_ACCOUNT}/forklift-virt-v2v:${REGISTRY_TAG}
```
mmartinv, line 7:
```suggestion
VIRT_V2V_IMAGE ?= ${REGISTRY}/${REGISTRY_ACCOUNT}/forklift-virt-v2v:${REGISTRY_TAG}
```

Makefile
```diff
@@ -102,15 +103,15 @@ build-virt-v2v-image:
 	bazel run virt-v2v:forklift-virt-v2v
 
 push-virt-v2v-image: build-virt-v2v-image
-	$(CONTAINER_CMD) tag virt-v2v:forklift-virt-v2v ${VIRT_V2V_IMAGE}
-	$(CONTAINER_CMD) push ${VIRT_V2V_IMAGE}
+	$(CONTAINER_CMD) tag virt-v2v:forklift-virt-v2v ${VIRT_V2V_IMAGE_COLD}
```
mmartinv, line 106:
```suggestion
	$(CONTAINER_CMD) tag virt-v2v:forklift-virt-v2v ${VIRT_V2V_IMAGE}
```

Makefile
nyoxi: I didn't want to make this change before we actually have the image in Quay. If we want to change this right now I will need to modify also `hack/release-images.sh`.

Makefile
```diff
@@ -102,15 +103,15 @@ build-virt-v2v-image:
...
+	$(CONTAINER_CMD) push ${VIRT_V2V_IMAGE_COLD}
```
mmartinv, line 107:
```suggestion
	$(CONTAINER_CMD) push ${VIRT_V2V_IMAGE}
```

Makefile
nyoxi: I would like to have a clear distinction what is the meaning of a variable so they cannot be confused. Where there is `VIRT_V2V_IMAGE_COLD` it refers to a single image that is used for cold migration. Where there is `VIRT_V2V_IMAGE` it refers to our internal variable that can have a combination of two images (cold and warm).

Makefile
nyoxi: Same comment as above.

mmartinv (changes requested):
I would use `VIRT_V2V_IMAGE` variable name for cold migration and `VIRT_V2V_IMAGE_WARM` for warm migration. We also need to update the value of `VIRT_V2V_IMAGE` with the combination of the two: `"${VIRT_V2V_IMAGE}|${VIRT_V2V_IMAGE_WARM}"`

Makefile
mmartinv: I am thinking that this change should be temporary, so I was suggesting the minimal changes :).

mmartinv: I find the way the "VIRT_V2V_IMAGE" variable is used already confusing (being able to put two different images separated by '|') and I think nobody will understand it without reading the documentation first. So I would add the warm target with the warm variable and leave the rest of the code untouched, this is, with `--action_env VIRT_V2V_IMAGE=${VIRT_V2V_IMAGE}` in the `build-operator-bundle-image` target.

mmartinv: After discussing it offline I am ok with the changes the way they are

mmartinv (approved):

Makefile
nyoxi: `VIRT_V2V_IMAGE` is for internal use so the requirement on reading the documentation is not a bad thing. I understand it is already confusing to some extent, but we should find a good balance and try to reduce the confusion if it means just a few more (simple) changes.

PR #125, opened 2023-01-25
Ease execution of forkliftci locally

ahadas:
1. By default use 'kubectl', assuming it's within PATH rather than `/home/eslutsky/bin/kubectl`
2. Do not store kubeconfig in /tmp before executing the end-to-end tests, there's no need for this anymore (see https://github.com/kubev2v/forkliftci/pull/55)


eslutsky:
we dont  use methods that require kubectl cli  , its  used in  the framework to watch for specific events, ie: to block execution until certain events met.

eslutsky (approved):

PR #128, opened 2023-01-29
Set version to 2.4.0

ahadas:
Previously, the version was 99.0.0 although the upcoming version is 2.4.0.
This PR also includes two more changes, in a separate commit, that I had to do in order to deploy to CRC using the `deploy-to-crc.sh` script


mnecas (changes requested):
Please change the version also in README

ahadas:
> Please change the version also in README

done

mnecas (approved):

PR #131, opened 2023-01-31
Improving automated VM rename handling

bkhizgiy:
Follow up the automated VM name renaming feature, improving some flows:

1. [MTV-404](https://issues.redhat.com/browse/MTV-404): Fixing long name handling, in case the name exists in the system and its length is already 63 chars after adding the VM ID as a unique identifier the name will exceed the allowed length limit and the migration will fail.
2. [MTV-480](https://issues.redhat.com/browse/MTV-480): change the check scope for the already existing VM name from the cluster to the plan namespace.
3. [MTV-382](https://issues.redhat.com/browse/MTV-382): Fix VM length validation to be a maximum of 63 chars and not 64.

### DIFF
--- a/pkg/controller/plan/kubevirt.go
+++ b/pkg/controller/plan/kubevirt.go
@@ -56,6 +56,8 @@ const (
 	AnnOriginalID = "original-ID"
 	// DV deletion on completion
 	AnnDeleteAfterCompletion = "cdi.kubevirt.io/storage.deleteAfterCompletion"
+	// Max Length for vm name
+	NameMaxLength = 63
 )
 
 // Labels
@@ -776,12 +778,17 @@ func (r *KubeVirt) virtualMachine(vm *plan.VMStatus) (object *cnv.VirtualMachine
 		originalName = vm.Name
 
 		generatedName := changeVmName(vm.Name, vm.ID)
-		nameExist, errName := r.checkIfVmNameExist(generatedName)
+		nameExist, errName := r.checkIfVmNameExistsInNamespace(generatedName, r.Plan.Spec.TargetNamespace)
 		if errName != nil {
 			err = liberr.Wrap(errName)
 			return
 		}
 		if nameExist {
+			// If the name exists and it's at max allowed length, remove 5 chars from the end
+			// so we won't reach the limit after appending vmId
+			if len(generatedName) == NameMaxLength {
+				generatedName = generatedName[0 : NameMaxLength-5]
+			}
 			generatedName = generatedName + "-" + vm.ID[:4]
 		}
 		vm.Name = generatedName
@@ -1025,7 +1032,7 @@ func (r *KubeVirt) guestConversionPod(vm *plan.VMStatus, vmVolumes []cnv.Volume,
 			},
 			Containers: []core.Container{
 				{
-					Name:            "virt-v2v",
+					Name: "virt-v2v",
 					EnvFrom: []core.EnvFromSource{
 						{
 							Prefix: "V2V_",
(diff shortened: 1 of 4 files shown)


pkg/controller/plan/kubevirt.go
```diff
@@ -1379,11 +1385,15 @@ func (r *KubeVirt) vmLabels(vmRef ref.Ref) (labels map[string]string) {
 }
 
 // Checks if VM with the newly generated name exists on the destination
-func (r *KubeVirt) checkIfVmNameExist(name string) (nameExist bool, err error) {
+func (r *KubeVirt) checkIfVmNameExistInNameSpace(name string, nameSpace string) (nameExist bool, err error) {
```
ahadas, line 1388:
```suggestion
func (r *KubeVirt) checkIfVmNameExistInNameSpace(name string, namespace string) (nameExist bool, err error) {
```

pkg/controller/plan/kubevirt.go
```diff
@@ -1379,11 +1385,15 @@ func (r *KubeVirt) vmLabels(vmRef ref.Ref) (labels map[string]string) {
...
 	list := &cnv.VirtualMachineList{}
 	nameFiled := "metadata.name"
+	nameSpaceFiled := "metadata.namespace"
```
ahadas, line 1391:
```suggestion
	namespaceFiled := "metadata.namespace"
```

pkg/controller/plan/kubevirt.go
```diff
@@ -1379,11 +1385,15 @@ func (r *KubeVirt) vmLabels(vmRef ref.Ref) (labels map[string]string) {
 }
 
 // Checks if VM with the newly generated name exists on the destination
-func (r *KubeVirt) checkIfVmNameExist(name string) (nameExist bool, err error) {
+func (r *KubeVirt) checkIfVmNameExistInNameSpace(name string, nameSpace string) (nameExist bool, err error) {
 	list := &cnv.VirtualMachineList{}
 	nameFiled := "metadata.name"
```
ahadas, line 1390: did "field" here and below?

ahadas (commented):
adding the Jira tickets to the description of the PR - great
I'm not sure I understand how the second part of this PR addresses https://issues.redhat.com/browse/MTV-404 though - shouldn't we cut 4 more chars from the name?

pkg/controller/plan/kubevirt.go
```diff
@@ -776,12 +778,16 @@ func (r *KubeVirt) virtualMachine(vm *plan.VMStatus) (object *cnv.VirtualMachine
...
 
 		generatedName := changeVmName(vm.Name, vm.ID)
-		nameExist, errName := r.checkIfVmNameExist(generatedName)
+		nameExist, errName := r.checkIfVmNameExistInNamespace(generatedName, r.Plan.Spec.TargetNamespace)
 		if errName != nil {
 			err = liberr.Wrap(errName)
 			return
 		}
 		if nameExist {
+			// If the name exists and it's at max allowed length, remove 5 chars from the end to not reach the limit after appending vmId
```
ahadas, line 787:
```suggestion
			// If the name exists and it's at max allowed length, remove 5 chars from the end so we won't reach the limit after appending vmId
```

pkg/controller/plan/kubevirt.go
```diff
@@ -1379,11 +1385,15 @@ func (r *KubeVirt) vmLabels(vmRef ref.Ref) (labels map[string]string) {
 }
 
 // Checks if VM with the newly generated name exists on the destination
-func (r *KubeVirt) checkIfVmNameExist(name string) (nameExist bool, err error) {
+func (r *KubeVirt) checkIfVmNameExistInNamespace(name string, namespace string) (nameExist bool, err error) {
```
ahadas, line 1388:
```suggestion
func (r *KubeVirt) checkIfVmNameExistsInNamespace(name string, namespace string) (nameExist bool, err error) {
```

pkg/controller/plan/kubevirt.go
```diff
@@ -776,12 +778,16 @@ func (r *KubeVirt) virtualMachine(vm *plan.VMStatus) (object *cnv.VirtualMachine
...
 		if errName != nil {
 			err = liberr.Wrap(errName)
 			return
 		}
 		if nameExist {
+			// If the name exists and it's at max allowed length, remove 5 chars from the end to not reach the limit after appending vmId
+			if len(generatedName) == NameMaxLength {
+				generatedName = generatedName[0 : NameMaxLength-5]
+			}
 			generatedName = generatedName + "-" + vm.ID[:4]
```
ahadas, line 791: that's subjective maybe - but wouldn't it be nicer to append the suffix and then to take a substring from index 5?

bkhizgiy: umm it probably doesn't make much difference but I think it would be better to take it off from the end rather than the beginning, just from the point of the name mining, I think the beginning is more meaningful than the end so butter not cutting it...

PR #135, opened 2023-01-31
Split the populator image

liranr23:
This patch will split the populator image.
One image acts as a controller to the populator - 'populator-controller' and the actual populator action will be executed by the 'ovirt-populator' image.

This will allow a separation in logic, allowing to add more source providers images and use the same controller for all.

### DIFF
--- a/cmd/populator-controller/BUILD.bazel
+++ b/cmd/populator-controller/BUILD.bazel
@@ -1,0 +1,37 @@
+load("@io_bazel_rules_go//go:def.bzl", "go_binary", "go_library")
+
+go_binary(
+    name = "populator-controller",
+    embed = [":populator-controller_lib"],
+    visibility = ["//visibility:public"],
+)
+
+go_library(
+    name = "populator-controller_lib",
+    srcs = ["populator-controller.go"],
+    importpath = "github.com/konveyor/forklift-controller/cmd/populator-controller",
+    visibility = ["//visibility:private"],
+    deps = [
+        "//pkg/apis/forklift/v1beta1",
+        "//vendor/github.com/kubev2v/lib-volume-populator/populator-machinery",
+        "//vendor/k8s.io/apimachinery/pkg/apis/meta/v1/unstructured",
+        "//vendor/k8s.io/apimachinery/pkg/runtime",
+        "//vendor/k8s.io/apimachinery/pkg/runtime/schema",
+    ],
+)
+
+load(
+    "@io_bazel_rules_docker//container:container.bzl",
+    "container_image",
+)
+
+container_image(
+    name = "populator-controller-image",
+    base = "@ubi9-minimal//image",
+    directory = "/usr/local/bin/",
+    entrypoint = ["/usr/local/bin/populator-controller"],
+    # workaround for github.com/bazelbuild/rules_go/issues/1706
+    env = {"GODEBUG": "netdns=go"},
+    files = [":populator-controller"],
+    visibility = ["//visibility:public"],
+)
--- a/cmd/populator-controller/populator-controller.go
+++ b/cmd/populator-controller/populator-controller.go
@@ -1,0 +1,72 @@
+package main
+
+import (
+	"flag"
+	"github.com/konveyor/forklift-controller/pkg/apis/forklift/v1beta1"
+	populator_machinery "github.com/kubev2v/lib-volume-populator/populator-machinery"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+const (
+	prefix     = "forklift.konveyor.io"
+	mountPath  = "/mnt/"
+	devicePath = "/dev/block"
+)
+
+const (
+	groupName  = "forklift.konveyor.io"
+	apiVersion = "v1beta1"
+	kind       = "OvirtVolumePopulator"
+	resource   = "ovirtvolumepopulators"
+)
+
+func main() {
+	var httpEndpoint, metricsPath, masterURL, kubeconfig, imageName, namespace string
+
+	// Controller args
+	if f := flag.Lookup("kubeconfig"); f != nil {
+		kubeconfig = f.Value.String()
+	} else {
+		flag.StringVar(&kubeconfig, "kubeconfig", "", "Path to a kubeconfig. Only required if out-of-cluster.")
+	}
+	flag.StringVar(&masterURL, "master", "", "The address of the Kubernetes API server. Overrides any value in kubeconfig. Only required if out-of-cluster.")
+	flag.StringVar(&imageName, "image-name", "", "Image to use for populating")
+	// Metrics args
+	flag.StringVar(&httpEndpoint, "http-endpoint", "", "The TCP network address where the HTTP server for diagnostics, including metrics and leader election health check, will listen (example: `:8080`). The default is empty string, which means the server is disabled.")
+	flag.StringVar(&metricsPath, "metrics-path", "/metrics", "The HTTP path where prometheus metrics will be exposed. Default is `/metrics`.")
+	// Other args
+	flag.StringVar(&namespace, "namespace", "konveyor-forklift", "Namespace to deploy controller")
+	flag.Parse()
+
+	var (
+		gk  = schema.GroupKind{Group: groupName, Kind: kind}
+		gvr = schema.GroupVersionResource{Group: groupName, Version: apiVersion, Resource: resource}
+	)
+	populator_machinery.RunController(masterURL, kubeconfig, imageName, httpEndpoint, metricsPath,
+		namespace, prefix, gk, gvr, mountPath, devicePath, getPopulatorPodArgs)
+}
+
+func getPopulatorPodArgs(rawBlock bool, u *unstructured.Unstructured) ([]string, error) {
+	var ovirtVolumePopulator v1beta1.OvirtVolumePopulator
+	err := runtime.DefaultUnstructuredConverter.FromUnstructured(u.UnstructuredContent(), &ovirtVolumePopulator)
+	var args []string
+	if nil != err {
+		return nil, err
+	}
+
+	if rawBlock {
+		args = append(args, "--volume-path="+devicePath)
+	} else {
+		args = append(args, "--volume-path="+mountPath+"disk.img")
+	}
+
+	args = append(args, "--secret-name="+ovirtVolumePopulator.Spec.EngineSecretName)
+	args = append(args, "--disk-id="+ovirtVolumePopulator.Spec.DiskID)
+	args = append(args, "--engine-url="+ovirtVolumePopulator.Spec.EngineURL)
+	args = append(args, "--cr-name="+ovirtVolumePopulator.Name)
+	args = append(args, "--cr-namespace="+ovirtVolumePopulator.Namespace)
+
+	return args, nil
+}
--- a/.github/workflows/build-push-images.yml
+++ b/.github/workflows/build-push-images.yml
@@ -38,4 +38,4 @@ jobs:
         run: |
           cd ${GITHUB_WORKSPACE}
           ./hack/release-images.sh
-          make push-populator-image
+          make push-ovirt-populator-image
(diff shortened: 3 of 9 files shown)


liranr23:
Building the `ovirt-populator` image works. Locally it also runs, but on my OCP it doesn't - `exec /ovirt-populator: exec format error`.
There are some bugs that seems to relate the issue I see https://github.com/golang/go/issues/53116 and https://github.com/golang/go/issues/53804 (Basing on https://github.com/golang/go/issues/53804#issuecomment-1182345944, it relates to `CGO_ENABLED=0`. There is also a kernel fix in v6.1 and above. But the images used to compile is using 5.14.0-162.6.1.el9_1.x86_64. Still, it does seem to work locally.

cmd/populator-controller/populator-controller.go
```diff
@@ -0,0 +1,73 @@
...
+	flag.StringVar(&metricsPath, "metrics-path", "/metrics", "The HTTP path where prometheus metrics will be exposed. Default is `/metrics`.")
+	// Other args
+	flag.StringVar(&namespace, "namespace", "konveyor-forklift", "Namespace to deploy controller")
+	flag.Parse()
+
+	var (
+		gk  = schema.GroupKind{Group: groupName, Kind: kind}
+		gvr = schema.GroupVersionResource{Group: groupName, Version: apiVersion, Resource: resource}
+	)
+	imageName = "quay.io/lrotenbe/ovirt-populator:test"
```
liranr23, line 47: will need to change into `kubev2v/ovirt-populator:latest`

ovirt-populator/ovirt-populator.go
```diff
@@ -9,133 +9,54 @@ import (
...
-	kind       = "OvirtVolumePopulator"
 	resource   = "ovirtvolumepopulators"
 )
 
 func main() {
-	var mode, engineUrl, secretName, diskID, volPath, crName, crNamespace, httpEndpoint, metricsPath, masterURL, kubeconfig, imageName, namespace string
-
-	// Main arg
-	flag.StringVar(&mode, "mode", "", "Mode to run in (controller, populate)")
+	var engineUrl, secretName, diskID, volPath, crName, crNamespace, masterURL, kubeconfig, namespace string
```
liranr23, line 42: `masterURL` and `kubeconfig` not used. need to check whether it is needed once the image works.

cmd/populator-controller/BUILD.bazel
```diff
@@ -0,0 +1,37 @@
...
+)
+
+load(
+    "@io_bazel_rules_docker//container:container.bzl",
+    "container_image",
+)
+
+container_image(
+    name = "populator-controller-image",
+    base = "@ubi8-minimal//image",
```
mnecas, line 30: Does it need the ubi8? 
(I have just bumped all bazel images to ubi9)

liranr23: i don't think it needs ubi8. ubi9 should fit.

cmd/populator-controller/populator-controller.go
```diff
@@ -0,0 +1,73 @@
...
+func getPopulatorPodArgs(rawBlock bool, u *unstructured.Unstructured) ([]string, error) {
+	var ovirtVolumePopulator v1beta1.OvirtVolumePopulator
+	err := runtime.DefaultUnstructuredConverter.FromUnstructured(u.UnstructuredContent(), &ovirtVolumePopulator)
+	var args []string
+	if nil != err {
+		return nil, err
+	}
+
+	if rawBlock {
+		args = append(args, "--file-name="+devicePath)
```
ahadas, line 61: didn't you change the file-name to path?

ovirt-populator/ovirt-populator.go
```diff
@@ -9,133 +9,54 @@ import (
...
+}
 
 const (
 	groupName  = "forklift.konveyor.io"
 	apiVersion = "v1beta1"
-	kind       = "OvirtVolumePopulator"
 	resource   = "ovirtvolumepopulators"
 )
 
 func main() {
```
ahadas, line 41: why moving this file out of `cmd` then?

cmd/populator-controller/populator-controller.go
ahadas: ok, you've handled it [elsewhere](https://github.com/kubev2v/forklift/pull/141)

cmd/populator-controller/populator-controller.go
liranr23: i think that will be wrong. we create this pod as controller and it should (in the future) be for multiple providers using different images. generally thinking - we will have a switch case here, based on the provider(we do need to think however how it will know which to use). or, we do need to think how we provide the image name. this is env var. so it will be wrong (in my opinion) to do it that way.

liranr23:
> Building the `ovirt-populator` image works. Locally it also runs, but on my OCP it doesn't - `exec /ovirt-populator: exec format error`. There are some bugs that seems to relate the issue I see [golang/go#53116](https://github.com/golang/go/issues/53116) and [golang/go#53804](https://github.com/golang/go/issues/53804) (Basing on [golang/go#53804 (comment)](https://github.com/golang/go/issues/53804#issuecomment-1182345944), it relates to `CGO_ENABLED=0`. There is also a kernel fix in v6.1 and above. But the images used to compile is using 5.14.0-162.6.1.el9_1.x86_64. Still, it does seem to work locally.

works now (somehow, my guess it's either luck or changing it into cmd directory) :)

cmd/populator-controller/populator-controller.go
ahadas: @bennyz correct me if I'm wrong here but we thought about adding the image to use on the CR rather than using an environment variable, right?

bennyz: yes, but it doesn't look like the library can actually handle different populators without us modifying it. It seems like we need to have getPopulatorPodArgs select the image according the CR (assuming it will have an imageName field)
https://github.com/kubernetes-csi/lib-volume-populator/blob/4adb975aba60aa689893d452d33806befb977efd/populator-machinery/controller.go#L540

and only then set the image https://github.com/kubernetes-csi/lib-volume-populator/blob/4adb975aba60aa689893d452d33806befb977efd/populator-machinery/controller.go#L555

liranr23: if it's in the CR, we can use the populator-controller image to take it in provide it to the the library, without modifying the library further.

bennyz:
Right I forgot we have the group and kind

liranr23:
> Right I forgot we have the group and kind
>> if it's in the CR, we can use the populator-controller image to take it in provide it to the the library, without modifying the library further.

So, I think we don't really need to change a thing? once we have other kinds we may change the image based on that? or we should providing an image anyway in the CR?

ahadas:
@liranr23 you mean another provider type that uses volume-populator? if so I don't see why not to do it before - even if ovirt is the only provider, I'd still expect the image of the populator to be configurable (and propagated from the controller that takes the forklift or mtv image that is set in the operator)

liranr23:
> @liranr23 you mean another provider type that uses volume-populator? if so I don't see why not to do it before - even if ovirt is the only provider, I'd still expect the image of the populator to be configurable (and propagated from the controller that takes the forklift or mtv image that is set in the operator)

benny and i discussed offline, we think a configmap might be a solution for this. i'm still bothered with how the controller will know which kind he will be looking for.

cmd/populator-controller/populator-controller.go
```diff
@@ -0,0 +1,97 @@
...
+	if err != nil {
+		klog.Fatal(err.Error())
+	}
+	// creates the clientset
+	clientset, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		klog.Fatal(err.Error())
+	}
+
+	configMap, err := clientset.CoreV1().ConfigMaps(namespace).Get(context.TODO(), configName, metav1.GetOptions{})
```
ahadas, line 69: who is expected to create this configmap and what's the benefit compared to specifying it on the CR?

operator/config/crd/bases/forklift.konveyor.io_ovirtvolumepopulators.yaml
```diff
@@ -39,10 +39,13 @@ spec:
                 type: string
               engineUrl:
                 type: string
+              imageName:
```
ahadas, line 42: ah so it's still here?

cmd/populator-controller/populator-controller.go
bennyz: specifying in the CR is tricky, because the CR is available only during the migration and the volpop controller is always running so there will be nothing to read most of the time

liranr23: we should deploy the configMap.

operator/config/crd/bases/forklift.konveyor.io_ovirtvolumepopulators.yaml
liranr23: forgotten :)

cmd/populator-controller/populator-controller.go
ahadas: > we should deploy the configMap.

yeah, I understand it's forklift - but do you expect the operator to do that?

> specifying in the CR is tricky, because the CR is available only during the migration and the volpop controller is always running so there will be nothing to read most of the time

I don't understand that part, I envisioned it as:
1. forklift-controller posts the CR, including the image that it gets from its env that was populated from ForkliftController
2. populator-controller reads the CR, including the image
3. populator-controller starts ovirt/osp-populator with the right image

do I understand this correctly that the problem with this is that it requires changing the library that k8s post as an example/utility for using volume-populator?

operator/config/crd/bases/forklift.konveyor.io_ovirtvolumepopulators.yaml
ahadas: hold your horses! it doesn't sounds so bad..

cmd/populator-controller/populator-controller.go
bennyz: > I don't understand that part, I envisioned it as:
> 
> 1. forklift-controller posts the CR, including the image that it gets from its env that was populated from ForkliftController
> 2. populator-controller reads the CR, including the image
> 3. populator-controller starts ovirt/osp-populator with the right image
> 
> do I understand this correctly that the problem with this is that it requires changing the library that k8s post as an example/utility for using volume-populator?

yes, the controller starts watching for PVCs with a specific `dataSourceRef` that matches the `gk`+`gvr` values we pass to `RunController`, so it needs know which CRs it's watching and which image to use when starting, but we can add this logic in our code and start multiple controllers, one for each kind with the correct image. We can probably optimize it in the future and only start a certain controller when the relevant provider is added

bennyz: >well yeah but this just exchanges listening to the populator CRs with listening to providers, no?
yeah... very optional in my opinion

cmd/populator-controller/populator-controller.go
liranr23: as we talked offline, we may run the `RunController` twice, once for ovirt and once for OSP. currently, we will get the `image-name` for ovirt. it will be part of the deployment.

operator/config/crd/bases/forklift.konveyor.io_ovirtvolumepopulators.yaml
liranr23: dropped

cmd/populator-controller/populator-controller.go
```diff
@@ -0,0 +1,97 @@
...
+	if err != nil {
+		klog.Fatal(err.Error())
+	}
+	return configMap.Data[kind]
+}
+
+func getPopulatorPodArgs(rawBlock bool, u *unstructured.Unstructured) ([]string, error) {
```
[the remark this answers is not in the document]
liranr23: discussed offline, we may have 2 functions and run `RunController` per provider.

cmd/populator-controller/populator-controller.go
liranr23: configmap idea dropped.

ahadas (approved):

ahadas:
@liranr23 please check the failure in:
https://github.com/kubev2v/forklift/actions/runs/4117193517/jobs/7108193103

PR #136, opened 2023-01-31
Let virt-v2v copy the data

nyoxi:
This is a set of patches that modify Forklift controller and forklift-virt-v2v image to make virt-v2v do the data copying instead of CDI.

The gap we will have with this is that progress for data copying is not reported as it used to. That is still work to be done and a topic for a separate PR.

### DIFF
--- a/pkg/controller/plan/adapter/vsphere/builder.go
+++ b/pkg/controller/plan/adapter/vsphere/builder.go
@@ -244,8 +244,10 @@ func (r *Builder) DataVolumes(vmRef ref.Ref, secret *core.Secret, _ *core.Config
 		for _, disk := range vm.Disks {
 			if disk.Datastore.ID == ds.ID {
 				storageClass := mapped.Destination.StorageClass
-				dvSpec := cdi.DataVolumeSpec{
-					Source: &cdi.DataVolumeSource{
+				var dvSource cdi.DataVolumeSource
+				if r.Plan.Spec.Warm {
+					// Let CDI do the copying
+					dvSource = cdi.DataVolumeSource{
 						VDDK: &cdi.DataVolumeSourceVDDK{
 							BackingFile:  trimBackingFileName(disk.File),
 							UUID:         vm.UUID,
@@ -254,7 +256,15 @@ func (r *Builder) DataVolumes(vmRef ref.Ref, secret *core.Secret, _ *core.Config
 							Thumbprint:   thumbprint,
 							InitImageURL: r.Source.Provider.Spec.Settings["vddkInitImage"],
 						},
-					},
+					}
+				} else {
+					// Let virt-v2v do the copying
+					dvSource = cdi.DataVolumeSource{
+						Blank: &cdi.DataVolumeBlankImage{},
+					}
+				}
+				dvSpec := cdi.DataVolumeSpec{
+					Source: &dvSource,
 					Storage: &cdi.StorageSpec{
 						Resources: core.ResourceRequirements{
 							Requests: core.ResourceList{
@@ -491,7 +501,11 @@ func (r *Builder) mapDisks(vm *model.VM, persistentVolumeClaims []core.Persisten
 	for i := range persistentVolumeClaims {
 		pvc := &persistentVolumeClaims[i]
 		// the PVC BackingFile value has already been trimmed.
-		pvcMap[pvc.Annotations[AnnImportBackingFile]] = pvc
+		if source, ok := pvc.Annotations[planbase.AnnDiskSource]; ok {
+			pvcMap[source] = pvc
+		} else {
+			pvcMap[pvc.Annotations[AnnImportBackingFile]] = pvc
+		}
 	}
 	for i, disk := range disks {
 		pvc := pvcMap[trimBackingFileName(disk.File)]
--- a/pkg/controller/plan/adapter/vsphere/builder.go
+++ b/pkg/controller/plan/adapter/vsphere/builder.go
@@ -9,6 +9,7 @@ import (
 	"github.com/konveyor/forklift-controller/pkg/apis/forklift/v1beta1/ref"
 	planbase "github.com/konveyor/forklift-controller/pkg/controller/plan/adapter/base"
 	plancontext "github.com/konveyor/forklift-controller/pkg/controller/plan/context"
+	container "github.com/konveyor/forklift-controller/pkg/controller/provider/container/vsphere"
 	"github.com/konveyor/forklift-controller/pkg/controller/provider/model/vsphere"
 	"github.com/konveyor/forklift-controller/pkg/controller/provider/web"
 	"github.com/konveyor/forklift-controller/pkg/controller/provider/web/base"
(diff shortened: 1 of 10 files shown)


nyoxi:
I am still struggling with building the forklift-virt-v2v so that I can test it end-to-end, but it is in good enough shape for review.

virt-v2v/cold/entrypoint
```diff
@@ -1,23 +1,66 @@
...
-		echo Unable to complete import!
-		exit 1
-	fi
+echo "Preparing virt-v2v"
+
+# This variable is used to build the list of arguments for virt-v2v.
+args=()
+
+# Temporary storage location
+DIR="/tmp/v2v"
```
liranr23, line 10: i think we had troubles with just `/tmp` in the past. maybe worth to changing it to `/var/tmp`?

pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -704,6 +742,7 @@ func (r *Builder) host(hostID string) (host *model.Host, err error) {
 	return
 }
 
+//
```
liranr23, line 745: i think we are trying to get rid from those. don't we?

virt-v2v/cold/entrypoint
ahadas: we did but with storing kubernetes stuff there (https://github.com/kubev2v/forkliftci/pull/37) - here it's within the container image

liranr23: regardless we also saw it with ovirt using OVAs. i think it's better to be safe than sorry.

pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -167,9 +168,46 @@ func (r *Builder) macConflicts(vm *model.VM) (conflictingVMs []string, err error
...
+	if err != nil {
+		err = liberr.Wrap(
+			err,
+			"VM lookup failed.",
+			"vm",
+			vmRef.String())
+		return
+	}
+
+	h, err := r.host(vm.Host)
```
ahadas, line 185: how about 'host' instead of 'h'?

pkg/controller/plan/adapter/vsphere/builder.go
ahadas: +1
yeah, 'go fmt' should drop those so need to avoid that

pkg/controller/plan/kubevirt.go
```diff
@@ -1027,6 +1027,14 @@ func (r *KubeVirt) guestConversionPod(vm *plan.VMStatus, vmVolumes []cnv.Volume,
...
 					Name:            "virt-v2v",
 					EnvFrom: []core.EnvFromSource{
+						{
+							Prefix: "V2V_",
+							ConfigMapRef: &core.ConfigMapEnvSource{
+								LocalObjectReference: core.LocalObjectReference{
+									Name: configMap.Name,
+								},
+							},
+						},
```
ahadas, line 1037: what's the reason for using a config map here - to simplify the code or is there more to it? if there's a good reason for that, does the config map configured with the migration as its owner-reference?

virt-v2v/cold/entrypoint
ahadas: ah that's what you had in mind.. I don't think what virt-v2v does with that space is as complicated as inspecting an OVA source that is placed there so it would be fine but anyway, +1 to move it elsewhere if it doesn't work :)

virt-v2v/cold/entrypoint
```diff
@@ -1,23 +1,66 @@
...
+
+# Disks on filesystem storage.
+# e.g.: /mnt/disks/disk0/disk.img -> vmName-sda
+for disk in /mnt/disks/disk[0-9]* ; do
+	num="${disk:15}"
+	ln -s "$disk/disk.img" "$DIR/$V2V_vmName-sd$(gen_name "$((num+1))")"
+done
+# Disks on block storage.
+# e.g.: /dev/block0 -> vmName-sda
+for disk in /dev/block[0-9] ; do
```
ahadas, line 39: I don't think we'll have more than 10 volumes but shouldn't it be /dev/block[0-9]* as above?

ahadas:
CC @rwmjones

rwmjones:
We have good progress data internally, I'm just not sure where to send that progress data to.  I asked a question related to this in December on RH `tech-list` subject line `Progress bars in Kubernetes/OpenShift pods?`

virt-v2v/cold/entrypoint
nyoxi: It definitely should!

pkg/controller/plan/kubevirt.go
nyoxi: You mean why don't we specify the environment variables directly in POD specification instead of using ConfigMap? I suppose we could do that too.

nyoxi:
> We have good progress data internally, I'm just not sure where to send that progress data to. I asked a question related to this in December on RH `tech-list` subject line `Progress bars in Kubernetes/OpenShift pods?`

What we used to do in IMS and what I plan to do now too is to store the information in POD annotation. That will be done by small tool that will read virt-v2v output and use k8s API to update the annotation. Whether the tool will be simple bash, Python, or Go I am still undecided. Nicer solution would be to store the annotation on PVC of each particular disk, but that would make the logic much more complex. If virt-v2v could aid with the task in the long term it would make our life simpler.

nyoxi:
> > We have good progress data internally, I'm just not sure where to send that progress data to. I asked a question related to this in December on RH `tech-list` subject line `Progress bars in Kubernetes/OpenShift pods?`
> 
> What we used to do in IMS and what I plan to do now too is to store the information in POD annotation. [...] Nicer solution would be to store the annotation on PVC of each particular disk, but that would make the logic much more complex.

I forgot to mention that this is also what CDI does. It stores progress in PVC/DataVolume annotation.

rwmjones:
Is there any information about pod annotations?  The only thing I can find is information about adding static annotations to the yaml `metadata` section.

nyoxi:
> Is there any information about pod annotations? The only thing I can find is information about adding static annotations to the yaml `metadata` section.

You can modify some of the resource information when the resource already exists. See the kubectl [example](https://kubernetes.io/docs/tasks/manage-kubernetes-objects/update-api-object-kubectl-patch/) for the general idea how that can work. The implementation will vary depending on the tooling/library you use to communicate with the cluster. To give another example you can see how we explointed the k8s REST API to do that in IMS:

https://github.com/ManageIQ/manageiq-v2v-conversion_host/blob/master/wrapper/hosts.py#L141
https://github.com/ManageIQ/manageiq-v2v-conversion_host/blob/master/wrapper/hosts.py#L179

ahadas:
yeah, there's also ['kubectl annotate' command](https://kubernetes.io/blog/2021/04/20/annotating-k8s-for-humans/) to update annotations - the fact the annotations are part of the metadata section doesn't mean they cannot be added/updated after the k8s entity is created

pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -173,6 +174,53 @@ func (r *Builder) ConfigMap(_ ref.Ref, _ *core.Secret, _ *core.ConfigMap) (err e
...
+	if err != nil {
+		err = liberr.Wrap(
+			err,
+			"Host lookup failed.",
+			"host",
+			vm.Host)
+		return
+	}
+
+	sslVerify := ""
```
ahadas, line 199: can also be `var sslVerify string` which I like more but that's subjective

ahadas (approved):
ok, just one minor comment - so ack from my side, I'll check it again with #149

ahadas:
> I am still struggling with building the forklift-virt-v2v so that I can test it end-to-end, but it is in good enough shape for review.

I managed to verify that migrations work with this change (on our mtv07 env) but it's really slower compared to CDI + 'in-place' conversion - @nyoxi can you please check to see if you see something wrong there, I've expected migrations to be faster with this change..

rwmjones:
What's the exact virt-v2v command line which is slow?  Also what version of virt-v2v is this?  In RHEL 9.1 we spent a lot of time optimizing various v2v scenarios so they performed much better, and if using an older version you wouldn't have those optimizations.

Also CDI + in-place conversion is not a supportable use case.

ahadas:
@rwmjones  I tagged you in the relevant places of #149

nyoxi:
I did not make any extensive benchmark yet, but from what I recall the situation from my experiments last year it is as you say for smaller VMs where CDI + in-place is much faster. For larger VMs the difference evens out. Also el9 is slightly faster than el8.

nyoxi:
I have added a commit which changes the default value of `VIRT_V2V_IMAGE` to have both cold and warm image. We need that before we try to merge #149.

ahadas (approved):
we've noticed that with this mechanism the migration is much slower than the existing, not supported, mechanism with CDI + in-place conversion, however, when comparing that with imports from vSphere to RHV, it's still slower but the difference is lower.

we want to have it in order to have a supported mechanism that later enables conversion of el9 guests and installation of qemu-guest-agent - so let's get it in, that will also simplify the investigation of the performance we've seen

PR #137, opened 2023-02-01
Fix using non-up-to-date image of forklift-virt-v2v

ahadas:
Drop the settings of ImagePullPolicy for forklift-virt-v2v's image so the default policy will be set and the image would always be pulled for the 'latest' tag.

Also, refer to the libguestfs-appliance image by digest rather than by tag.


nyoxi:
To quote the [documentation](https://kubernetes.io/docs/concepts/containers/images/#imagepullpolicy-defaulting):

> When you (or a controller) submit a new Pod to the API server, your cluster sets the imagePullPolicy field when specific conditions are met:
>
>-    if you omit the imagePullPolicy field, and the tag for the container image is :latest, imagePullPolicy is automatically set to Always;
>-    if you omit the imagePullPolicy field, and you don't specify the tag for the container image, imagePullPolicy is automatically set to Always;
>-    if you omit the imagePullPolicy field, and you specify the tag for the container image that isn't :latest, the imagePullPolicy is automatically set to IfNotPresent.

This is useful for the development where we want to use the `latest` tag. On release we use the release tag for images so users will not negatively be affected by the change.

nyoxi (approved):

PR #140, opened 2023-02-01
Use centos-stream8 based libguestfs-appliance

ahadas:
The converision of forklift-virt-v2v to ubi doesn't make much sense at this point - let's keep using centos-stream8 and we'll reconsider this for el9


ahadas:
closing, @nyoxi found the issue with the ubi8

PR #144, opened 2023-02-02
Push the populator image

liranr23:
This patch will push the populator image that contains both populator controller and the populator itself to quay. Later on we will split the images.


mnecas (approved):
LGTM, but we need to get in https://github.com/kubev2v/forklift/pull/143 before this so we have working CI.

PR #152, opened 2023-02-07
Introducing random suffix generator for VM name in case of an already used name.

bkhizgiy:
Currently, during the migration, if the VM has an invalid name and the newly generated name exists in the namespace, a name of the format `vm-vmId[0:4]` is picked as a unique identifier.
For oVirt, this solution works well, but for vmware, the vm Id is in the following format: `vm-#vm_number_counter`. Because the counter can be any number and is not sufficiently unique, attaching it to the new name can cause issues.
To address this issue, this PR introduces a tool that generates a new random string of six lowercase letters and digits.

Also, this PR separates all VM name changing flow into a new file.

### DIFF
--- a/pkg/controller/plan/kubevirt.go
+++ b/pkg/controller/plan/kubevirt.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"math/rand"
 	"path"
-	"regexp"
 	"sort"
 	"strconv"
 	"strings"
@@ -33,7 +32,6 @@ import (
 	core "k8s.io/api/core/v1"
 	k8serr "k8s.io/apimachinery/pkg/api/errors"
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/labels"
 	cdi "kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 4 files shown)


pkg/controller/plan/kubevirt.go
```diff
@@ -1542,14 +1542,22 @@ func changeVmName(currName string, vmID string) string {
...
 	return newName
 }
+
+// Generates a random string of length six, consisting of lowercase letters and digits.
+func generateRandVmNameSuffix() string {
+	const charset = "abcdefghijklmnopqrstuvwxyz" + "0123456789"
+	source := rand.NewSource(time.Now().UTC().UnixNano())
+	seededRand := rand.New(source)
+
+	b := make([]byte, 6)
```
ahadas, line 1558: shouldn't it be of length 4 as otherwise we may exceed the max length?

pkg/controller/plan/kubevirt.go
```diff
@@ -1542,14 +1542,22 @@ func changeVmName(currName string, vmID string) string {
...
+	for i := range b {
+		b[i] = charset[seededRand.Intn(len(charset))]
+	}
+	return string(b)
+}
```
ahadas, line 1563: as we're adding more logic to this, how about separating the name handling to a different file in this package?

pkg/controller/plan/kubevirt.go
bkhizgiy: yes, you're right, forgot about it (I can change it over there)
I've extended it to six for more unique naming, not sure if it's really needed...

pkg/controller/plan/kubevirt.go
bkhizgiy: good suggestion, that will be more organized and easy to deal with that way, looking into that.

pkg/controller/plan/kubevirt.go
ahadas: practically, I think it's fine to go with a random string of all those characters with length 4

pkg/controller/plan/vmNameHandler.go
```diff
@@ -0,0 +1,104 @@
...
+			generatedName = generatedName[0 : NameMaxLength-5]
+		}
+		generatedName = generatedName + "-" + generateRandVmNameSuffix()
+	}
+	return
+}
+
+// changes VM name to match DNS1123 RFC convention.
+func changeVmName(currName string) string {
+
```
ahadas, line 36: please remove this empty line

pkg/controller/plan/vmNameHandler_test.go
```diff
@@ -8,15 +8,14 @@ import (
 
 func TestKubevirt(t *testing.T) {
```
ahadas, line 9: and then this file should also be renamed to vm_name_handler_test.go

pkg/controller/plan/kubevirt.go
ahadas: thanks, looks better but the filenames don't comply with the convention in golang (commented elsewhere)

PR #155, opened 2023-02-14
Openstack provider fixes

mmartinv:


pkg/controller/provider/container/openstack/model.go
```diff
@@ -138,22 +136,17 @@ func (r *RegionAdapter) DeleteUnexisting(ctx *Context) (updates []Updater, err e
...
+				m := &model.Region{
+					Base: model.Base{ID: region.ID},
 				}
-				updates = append(updates, updater)
-			default:
-				return
+				return tx.Delete(m)
 			}
+			updates = append(updates, updater)
+			err = nil
```
ahadas, line 147: what's the expectation from this method?
if we fail to get the last region due to an error that is different than NotFound, we'll return err != nil
if we fail to get a region due to an error that is different than NotFound and then got another region that is NotFound, we won't return an err = nil
I'm a bit confused here (and below :) )

mmartinv: Yeah, you are right, in case we got an error in an intermediate step no error is returned... :(
I will fix it in the next push

ahadas (approved):

bennyz (approved):

PR #163, opened 2023-02-20
CI - save k8s logs in gh-action workflow artifacts

eslutsky:


.github/workflows/pull-request.yml
```diff
@@ -62,11 +62,22 @@ jobs:
...
-          kubectl get events --field-selector type!=Normal -A --sort-by='.lastTimestamp'
-          kubectl get all -n konveyor-forklift -o yaml
+          mkdir /tmp/artifacts/
+          kubectl get pods -n konveyor-forklift >> /tmp/artifacts/k8s-pods.log
+          kubectl get events --field-selector type!=Normal -A --sort-by='.lastTimestamp' >> /tmp/artifacts/k8s_abnormal_events.log
+          kubectl get all -n konveyor-forklift -o yaml >> /tmp/artifacts/k8s-all-forklift-objects.log
+          kubectl logs -n konveyor-forklift $(kubectl get po -n konveyor-forklift  -o=name | grep forklift-controller) >> /tmp/artifacts/forklift-controller-inventory.log
+
+      - uses: actions/upload-artifact@master
+        if: always()
```
ahadas, line 75: don't we want this only on failure?

eslutsky: eventually,but for now we have many false positives so its useful to have it always.

ahadas: ok, so how about adding TODO?

ahadas (approved):

PR #164, opened 2023-02-20
Authenticate requests for the currect namespace

yaacov:
Issue: currently the authentication code checks if user can list providers on all namespaces (or on "POD_NAMESPACE" env variable if available) but list the providers using a URL param, this may allow users to list providers the do not have permission to list.

Fix:
  - [x] use the context `Request` param instead of the context param to get the namespace, the  `Request` param was never defined, and is empty always.
  - [x] fix the authentication code to use the real namespace, preventing a user that can list providers on "POD_NAMESPACE" from listing providers on other namespaces.

example of use:
``` bash
export TOKEN=[secret-token]
curl -k -H "Authorization: Bearer ${TOKEN}" http://localhost:8080/providers?namespace=forklift | jq
```

### DIFF
--- a/pkg/controller/provider/web/base/auth.go
+++ b/pkg/controller/provider/web/base/auth.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"net/http"
-	"os"
 	"path"
 	"strings"
 	"sync"
@@ -42,6 +41,7 @@ type Auth struct {
 // Authenticate token.
 func (r *Auth) Permit(ctx *gin.Context, p *api.Provider) (status int, err error) {
 	r.mutex.Lock()
+	ns := ""
 	defer r.mutex.Unlock()
 	status = http.StatusOK
 	if r.cache == nil {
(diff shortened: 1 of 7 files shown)


pkg/controller/provider/web/base/auth.go
```diff
@@ -59,7 +59,11 @@ func (r *Auth) Permit(ctx *gin.Context, p *api.Provider) (status int, err error)
 			return
 		}
 	}
-	allowed, err := r.permit(token, p)
+	if p.ObjectMeta.UID != "" {
```
ahadas, line 62: don't you mean
```suggestion
	if p.ObjectMeta.UID == "" {
```

yaacov: wow, thanks, fixing :+1:

yaacov: fixed, 
( now we don't need to fix in the test, but it's a good fix, leaving :-) )

ahadas (approved):

PR #165, opened 2023-02-20
Use right namespaces in tests

ahadas:


ahadas:
nah, that's incorrect..

eslutsky (approved):
looks good , maybe we  can we add  labels to the created CRDs  so it will be able to perform cleanup (in different PR) .

PR #166, opened 2023-02-21
CI - add more logs and separate artifacts per provider

eslutsky:


ahadas (approved):
that's a good step forward but what would really ease debugging issues and maybe accelerate execution time a bit is not to delete the namespace by default when the execution completes

eslutsky:
> that's a good step forward but what would really ease debugging issues and maybe accelerate execution time a bit is not to delete the namespace by default when the execution completes



> that's a good step forward but what would really ease debugging issues and maybe accelerate execution time a bit is not to delete the namespace by default when the execution completes

for openstack-tests its  not using the ephemeral namespace, everything is `konveyor-forklift`  which is not deleted .

ahadas:
@eslutsky ok, I didn't try openstack yet, I wrote the previous comment based on what I see for oVirt

eslutsky:
@ahadas ,also added 2 CRs to the openstack artifacts: pvs and openstackvolumepopulator,
imo we can merge it

ahadas (approved):

PR #167, opened 2023-02-21
Adding insecure migration from RHV

bkhizgiy:
This PR adds an option to support an insecure migration from RHV.
When enabled from the UI, the user will not be required to provide a CA cert at the provided adding stage,  and the disk transfer will instead be held in an insecure flow.

### DIFF
--- a/cmd/ovirt-populator/ovirt-populator.go
+++ b/cmd/ovirt-populator/ovirt-populator.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"strconv"
 
 	"k8s.io/klog/v2"
 )
@@ -16,6 +17,7 @@ type engineConfig struct {
 	username string
 	password string
 	ca       string
+	insecure bool
 }
 
 type TransferProgress struct {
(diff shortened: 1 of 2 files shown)


pkg/forklift-api/webhooks/mutating-webhook/mutators/ovirt-cert-mutator.go
```diff
@@ -34,7 +36,15 @@ func (mutator *OvirtCertMutator) Mutate(ar *admissionv1.AdmissionReview) *admiss
 		return util.ToAdmissionResponseError(err)
 	}
 
-	if providerType, ok := secret.GetLabels()["createdForProviderType"]; ok && providerType == "ovirt" {
+	if _, ok := secret.Data["insecureSkipVerify"]; ok {
+		insecure, err = strconv.ParseBool(string(secret.Data["insecureSkipVerify"]))
```
ahadas, line 40: so why not to take the the value of secret.Data["insecureSkipVerify"] from the line above?

pkg/forklift-api/webhooks/mutating-webhook/mutators/ovirt-cert-mutator.go
```diff
@@ -25,6 +26,7 @@ type OvirtCertMutator struct {
 }
 
 func (mutator *OvirtCertMutator) Mutate(ar *admissionv1.AdmissionReview) *admissionv1.AdmissionResponse {
+	var insecure = false
```
ahadas, line 29: please move it closer to the check of 'insecureSkipVerify'

pkg/forklift-api/webhooks/mutating-webhook/mutators/ovirt-cert-mutator.go
```diff
@@ -76,6 +86,10 @@ func (mutator *OvirtCertMutator) Mutate(ar *admissionv1.AdmissionReview) *admiss
 			log.Info("Engine CA certificate was missing, updating the secret")
 		}
 
+		if !insecure {
+			secret.Data["insecureSkipVerify"] = []byte("false")
+		}
```
ahadas, line 91: we certainly don't need to check `if !insecure` here (since it was checked before) but why setting the secret - it should already be set this way, no?

pkg/forklift-api/webhooks/mutating-webhook/mutators/ovirt-cert-mutator.go
```diff
@@ -102,11 +116,11 @@ func (mutator *OvirtCertMutator) Mutate(ar *admissionv1.AdmissionReview) *admiss
 		}
 	}
 
-	// Response for other providers type
+	// Response for other providers type or insecure mode
 	return &admissionv1.AdmissionResponse{
 		Allowed: true,
 		Result: &metav1.Status{
-			Message: "Provider type diffretn from oVirt, no changes required",
+			Message: "Certificate retrieval is not required, passing ",
```
ahadas, line 123: and fixed a typo, good :)

cmd/ovirt-populator/ovirt-populator.go
```diff
@@ -168,10 +187,16 @@ func loadEngineConfig(secretName, engineURL, namespace string) engineConfig {
 		klog.Fatal(err.Error())
 	}
 
+	insecure, err := strconv.ParseBool(string(secret.Data["insecureSkipVerify"]))
```
liranr23, line 190: that's going to be change after the populator PR, it changed to use a mounted volume with the secret. it will be required to read the relevant file within the pod.

pkg/forklift-api/webhooks/mutating-webhook/mutators/ovirt-cert-mutator.go
bkhizgiy: because later on  we are relaying on this value to set the transfer,  we don't want to assume that its already there.
I've changed the code so it only be added if not existing at all.

cmd/ovirt-populator/ovirt-populator.go
bkhizgiy: waiting for the new changes, I"ll adjust it accordingly

ahadas (approved):

liranr23 (approved):

PR #171, opened 2023-02-27
Fix race condition on inventory synchronization

mmartinv:
Currently we save the time where the last sync has been
successful for all the objects.

If any object is updated in that sync interval it's possible
to loose updates.

The proposed changes try to fix this issue.

### DIFF
--- a/pkg/controller/provider/container/openstack/model.go
+++ b/pkg/controller/provider/container/openstack/model.go
@@ -62,7 +62,7 @@ type Adapter interface {
 	// List REST collections.
 	List(ctx *Context) (itr fb.Iterator, err error)
 	// Get object updates
-	GetUpdates(ctx *Context, lastSync time.Time) (updates []Updater, err error)
+	GetUpdates(ctx *Context) (updates []Updater, err error)
 	// Clean unexisting objects within the database
 	DeleteUnexisting(ctx *Context) (updates []Updater, err error)
 }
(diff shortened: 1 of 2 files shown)


pkg/controller/provider/container/openstack/model.go
```diff
@@ -258,14 +260,15 @@ func (r *ImageAdapter) List(ctx *Context) (itr fb.Iterator, err error) {
...
 	return
 }
 
-func (r *ImageAdapter) GetUpdates(ctx *Context, lastSync time.Time) (updates []Updater, err error) {
+func (r *ImageAdapter) GetUpdates(ctx *Context) (updates []Updater, err error) {
 	opts := &ImageListOpts{}
-	opts.setUpdateAtQueryFilterGTE(lastSync)
+	opts.setUpdateAtQueryFilterGTE(r.lastSync)
 	imageList := []Image{}
+	now := time.Now()
```
bennyz, line 271: I think we still have a small race between lines 269-271, maybe it would make sense to move it to line 268?

ahadas: yeah, it's better to be on the "safe side" and initiate a redundant update later on than missing an update

mmartinv: Sorry but I don't follow, `time.Now()` is taken before doing the actual query, probably some microseconds before. So  if an update is done while updating the `Images`, it will be refreshed in the next iteration, right?

ahadas: ack, I read it as if setUpdateAtQueryFilterGTE triggers the call

ahadas (approved):

PR #172, opened 2023-02-27
Fix OpenStack StorageMap watch

mmartinv:
The storage map for OpenStack maps VolumeType objects to
StorageClass so no need to watch for Image, Volume or
Snapshot. We just need to watch for VolumeType instead, just
as oVirt watches for StorageDomains or vSphere for Datastores.

### DIFF
--- a/pkg/controller/map/storage/handler/openstack/handler.go
+++ b/pkg/controller/map/storage/handler/openstack/handler.go
@@ -22,98 +22,58 @@ type Handler struct {
 	*handler.Handler
 }
 
-// Ensure watch on Images, Snapshots and Volumes.
+// Ensure watch on VolumeType.
 func (r *Handler) Watch(watch *handler.WatchManager) (err error) {
 	w, err := watch.Ensure(
 		r.Provider(),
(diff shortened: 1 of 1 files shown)


pkg/controller/map/storage/handler/openstack/handler.go
```diff
@@ -22,98 +22,58 @@ type Handler struct {
...
 // Storage changed.
-// Find all of the StorageMap CRs referencing both the
-// provider and the changed storage and enqueue reconcile events.
-func (r *Handler) changed(models ...interface{}) {
-
+// Find all of the StorageMap CRs the reference both the
+// provider and the changed volume type and enqueue reconcile events.
+func (r *Handler) changed(models ...*openstack.VolumeType) {
+	log.V(3).Info(
+		"Storage domain changed.",
```
ahadas, line 74: volume type?

mmartinv: Fixed

bennyz (approved):

PR #178, opened 2023-03-01
Implement the tree endpoint

mmartinv:
Implement the "tree/project" endpoint showing
the list of projects with their associated VMs
in a tree like view

### DIFF
--- a/pkg/controller/provider/web/openstack/base.go
+++ b/pkg/controller/provider/web/openstack/base.go
@@ -56,21 +56,33 @@ type PathBuilder struct {
 	// Database.
 	DB libmodel.DB
 	// Cached resources.
-	cache map[string]string
+	cache map[string]interface{}
 }
 
 // Build.
 func (r *PathBuilder) Path(m model.Model) (path string) {
 	var err error
 	if r.cache == nil {
-		r.cache = map[string]string{}
+		r.cache = map[string]interface{}{}
 	}
 	switch m.(type) {
 	case *model.Project:
-		path = m.(*model.Project).Name
+		project := m.(*model.Project)
+		path = project.Name
+		if project.IsDomain {
+			return
+		}
+		if project.ParentID != project.DomainID {
+			var parentProject *model.Project
+			parentProject, err = r.getProject(project.ParentID)
+			path = pathlib.Join(r.Path(parentProject), path)
+		}
+
 	case *model.VM:
-		object := m.(*model.VM)
-		path, err = r.forProject(object.TenantID, object.Name)
+		vm := m.(*model.VM)
+		var project *model.Project
+		project, err = r.getProject(vm.TenantID)
+		path = pathlib.Join(r.Path(project), vm.Name)
 	}
 
 	if err != nil {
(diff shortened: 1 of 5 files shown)


mmartinv:
cc:// @yaacov

pkg/controller/provider/web/openstack/base.go
```diff
@@ -79,3 +83,23 @@ func (r *PathBuilder) Path(m model.Model) (path string) {
 
 	return
 }
+
+// Path based on Project.
+func (r *PathBuilder) forProject(id, leaf string) (path string, err error) {
```
ahadas, line 88: can it be that id is different than tenantId or leaf is different than a vm name? it's written in a generic way

mmartinv: Could be, for example, in the future we might want to add the project networks to the tree:
```go
case *model.Network:
		path = m.(*model.Network).Name
		object := m.(*model.Network)
		path, err = r.forProject(object.ProjectID, object.Name)
```
In this case `id` is `ProjectID` and `leaf` is `Name` but it could be any other object field like object.ID.
Not sure if I am answering your question.

yaacov:
works for me in the UI :+1:

pkg/controller/provider/web/openstack/base.go
ahadas: sort of :)
so if the 'id' is that of the project then let's call it projectId and would subresource be a better fit for the second argument?

ahadas (approved):
looks good, I'll post another PR with a suggestion for different naming of those arguments - that's minor..

pkg/controller/provider/web/openstack/base.go
mmartinv: What about `childName` instead of `subresource`?

ahadas: sounds good

mmartinv: Hold on, After reviewing @yaacov's PR I realized the projects are organized in a tree hierarchy.

ahadas:
@mmartinv what's the status of the update of this PR? if it works for Kobi, even if it's incomplete I'd like to get it merged

pkg/controller/provider/web/openstack/base.go
```diff
@@ -55,18 +56,33 @@ type PathBuilder struct {
...
+			parentProject, err = r.getProject(project.ParentID)
+			path = pathlib.Join(r.Path(parentProject), path)
+		}
+
 	case *model.VM:
-		path = m.(*model.VM).Name
+		vm := m.(*model.VM)
+		var project *model.Project
+		project, err = r.getProject(vm.TenantID)
+		path = pathlib.Join(r.Path(project), vm.Name)
```
ahadas, line 85: if we get an error above (err != nil) we'll probably fail here - is so, we'll need to handle the error here

ahadas (approved):
one comment inside
@mmartinv, I believe it's something we can address later but if not and we need to have it in the build, please reach out to @mnecas

PR #185, opened 2023-03-02
Fix the access modes of the PVC

liranr23:
When passing on the storage profile it can have multiple accessmodes and multiple volume modes. We prefer the block volume mode and set the access modes accordingly.

### DIFF
--- a/pkg/controller/plan/kubevirt.go
+++ b/pkg/controller/plan/kubevirt.go
@@ -693,8 +693,16 @@ func (r *KubeVirt) getStorageProfileModes(storageName string) (accessModes []cor
 		storageProfile := &storageProfileList.Items[i]
 		if storageProfile.Name == storageName {
 			for _, claimProperty := range storageProfile.Status.ClaimPropertySets {
-				accessModes = append(accessModes, claimProperty.AccessModes...)
 				volumeMode = claimProperty.VolumeMode
+				if *volumeMode == core.PersistentVolumeBlock {
+					// Preferring Block volume mode
+					break
+				}
+			}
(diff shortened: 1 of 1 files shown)


pkg/controller/plan/kubevirt.go
```diff
@@ -693,8 +693,17 @@ func (r *KubeVirt) getStorageProfileModes(storageName string) (accessModes []cor
 		storageProfile := &storageProfileList.Items[i]
 		if storageProfile.Name == storageName {
 			for _, claimProperty := range storageProfile.Status.ClaimPropertySets {
-				accessModes = append(accessModes, claimProperty.AccessModes...)
 				volumeMode = claimProperty.VolumeMode
+				// Preferring Block volume mode
+				if *claimProperty.VolumeMode == core.PersistentVolumeBlock && *volumeMode != core.PersistentVolumeBlock {
+					volumeMode = claimProperty.VolumeMode
+					break
+				}
```
ahadas, line 701: isn't it over complicated? why not to just check if *volumeMode == core.PersistentVolumeBlock after the assignment and then break?

liranr23: yes, done

ahadas: would have been nicer to have:
```
if volumeMode = claimProperty.VolumeMode; *volumeMode == core.PersistentVolumeBlock {
        // Preferring Block volume mode
	break
}
```
in my opinion but that's a matter of taste

ahadas (approved):

bennyz (approved):

PR #187, opened 2023-03-02
Monitor virt-v2v progress in EL9 image

nyoxi:


pkg/controller/plan/context/migration.go
```diff
@@ -103,6 +103,10 @@ func (r *Context) SetMigration(migration *api.Migration) {
 			migration.Name))
 }
 
+func (r *Context) UseEl9VirtV2v() bool {
+	return !r.Plan.Spec.Warm
```
ahadas, line 107: should be !warm and not to the 'host' target provider, right?

pkg/controller/plan/migration.go
```diff
@@ -1405,6 +1449,62 @@ func (r *Migration) updateConversionProgress(vm *plan.VMStatus, step *plan.Step)
...
+	if pod.Status.PodIP == "" {
+		return
+	}
+
+	var disk_re = regexp.MustCompile(`v2v_disk_transfers\{disk_id="(\d+)"\} (\d{1,3}\.?\d*)`)
+	url := fmt.Sprintf("http://%s:2112/metrics", pod.Status.PodIP)
+	resp, err := http.Get(url)
+	if err != nil {
+		if strings.Contains(err.Error(), "connection refused") {
+			return nil
```
ahadas, line 1462: why do we return with err = nil here? is it because we may get here before the metrics  endpoint is initialied?

pkg/controller/plan/migration.go
```diff
@@ -1405,6 +1449,62 @@ func (r *Migration) updateConversionProgress(vm *plan.VMStatus, step *plan.Step)
...
 }
 
+func (r *Migration) updateConversionProgressEl9(pod *core.Pod, step *plan.Step) (err error) {
+	if pod.Status.PodIP == "" {
+		return
+	}
+
+	var disk_re = regexp.MustCompile(`v2v_disk_transfers\{disk_id="(\d+)"\} (\d{1,3}\.?\d*)`)
+	url := fmt.Sprintf("http://%s:2112/metrics", pod.Status.PodIP)
+	resp, err := http.Get(url)
```
ahadas, line 1459: I think that's ok for the time being but we need to think what's the API we'd like to have between forklift-controller and the populator controller/pod - I think it would eventually make sense to let forklift-controller post the CR (e.g., OvirtVolumePopulator) and then only look at it, so the populator controller would fetch the progress from the populator pod and update the CR (or PVC).

we don't have a CR for virt-v2v so we anyway can't do it this way now, but if that's what we have in mind (and I'm not saying we do, we barely talked about it yet) then we should comment that this is a "temporary" solution that is meant to change

ahadas: this should also be handled separately, resolving for now

ahadas (approved):

liranr23 (approved):

PR #189, opened 2023-03-05
Adding security context to volume populator

liranr23:


vendor/github.com/liranr23/lib-volume-populator/populator-machinery/controller.go
```diff
@@ -722,12 +722,30 @@ func (c *controller) syncPvc(ctx context.Context, key, pvcNamespace, pvcName str
 }
 
 func makePopulatePodSpec(pvcPrimeName, secretName string) corev1.PodSpec {
+	nonRoot := true
+	allowPrivilageEscalation := false
+	// QEMU sub GID
+	user := int64(107)
```
ahadas, line 728: as it also appears [here](https://github.com/kubev2v/forklift/blob/main/pkg/controller/plan/kubevirt.go#L1149) for the virt-v2v container, how about making it a reusable const?

liranr23: it's within the lib, but we can make a const there.

tests/suit/framework/framework.go
```diff
@@ -171,7 +171,7 @@ func (f *Framework) CreateNamespace(prefix string, labels map[string]string) (*v
 	}
 	// TODO: restore the pod-security label once https://github.com/kubev2v/forklift/issues/173 is fixed
 	// pod-security.kubernetes.io/<MODE>: <LEVEL>
-	//labels["pod-security.kubernetes.io/enforce"] = "restricted"
+	labels["pod-security.kubernetes.io/enforce"] = "restricted"
```
ahadas, line 174: as we spoke offline - this annotation can be reintroduced after changing the conversion pod (virt-v2v)

ahadas:
@liranr23 I didn't see, did it pass with that annotation?

liranr23:
> @liranr23 I didn't see, did it pass with that annotation?

vsphere failed, the rest passed.

ahadas:
> > @liranr23 I didn't see, did it pass with that annotation?
> 
> vsphere failed, the rest passed.

ack, good (and yeah, the vsphere test would probably also fail now.. something is weird with that test but that's unrelated - see #192)

PR #191, opened 2023-03-05
Enable UI plugin by default

ahadas:
Previously the 'feature_ui_plugin' var wasn't set with a default value
and thus triggered the following error when it wasn't define in
ForkliftController:

The conditional check 'feature_ui_plugin|bool and not k8s_cluster|bool'
failed. The error was: error while evaluating conditional
(feature_ui_plugin|bool and not k8s_cluster|bool): 'feature_ui_plugin'
is undefined

Now this variable is defined and set to 'true' in order to deploy the
console plugin by default.


yaacov (approved):
lgtm

PR #192, opened 2023-03-05
Do not set 'kubevirt.io/schedulable' when VirtV2vDontRequestKVM is set

ahadas:
When we don't require KVM, we shouldn't filter hosts that have kubevirt.io/schedulable=true

### DIFF
--- a/pkg/controller/plan/kubevirt.go
+++ b/pkg/controller/plan/kubevirt.go
@@ -1135,6 +1135,7 @@ func (r *KubeVirt) findTemplate(vm *plan.VMStatus) (tmpl *template.Template, err
 func (r *KubeVirt) guestConversionPod(vm *plan.VMStatus, vmVolumes []cnv.Volume, configMap *core.ConfigMap, pvcs *[]core.PersistentVolumeClaim, v2vSecret *core.Secret) (pod *core.Pod, err error) {
 	volumes, volumeMounts, volumeDevices := r.podVolumeMounts(vmVolumes, configMap, pvcs)
 	resourceReq := core.ResourceRequirements{}
+	var nodeSelector map[string]string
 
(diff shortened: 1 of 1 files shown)


tests/suit/framework/framework.go
```diff
@@ -202,7 +202,7 @@ func (f *Framework) CreateNamespace(prefix string, labels map[string]string) (*v
 
 // AddNamespaceToDelete provides a wrapper around the go append function
 func (f *Framework) AddNamespaceToDelete(ns *v1.Namespace) {
-	f.namespacesToDelete = append(f.namespacesToDelete, ns)
+	//f.namespacesToDelete = append(f.namespacesToDelete, ns)
```
liranr23, line 205: why? let's not call it and leave the function as is - or delete it as a whole if we are not going to use that.

tests/suit/vsphere_test.go
```diff
@@ -12,7 +12,7 @@ import (
 
 const (
 	vsphereProviderName = "vsphere-provider"
-	vsphereStorageClass = "standard"
+	vsphereStorageClass = "nfs-csi"
```
liranr23, line 15: just to align with other suits? any other reason?

tests/suit/framework/framework.go
ahadas: @liranr23 that's part of the commit that is used for debugging.. there's another task on @eslutsky 's backlog to change this

tests/suit/vsphere_test.go
ahadas: that's also part of the 'debug' commit :)
but now that the test passes without the other commit - it seems that this is the reason for the failures of this test (the standard storage class)

ahadas:
okok, now it also reproduces locally - the reason for the failure is that a new version of CDI (1.56.0) was released three days ago..

PR #158, opened 2023-02-15
adding gingo osp provider test

eslutsky:
this PR  handle e2e tests for OpenStack migrations using [openstack-volume-populator](https://github.com/kubev2v/openstack-volume-populator) and [csi-driver-nfs](https://github.com/kubernetes-csi/csi-driver-nfs)  storage class 

it  creates and tests the following:
- [x]  Provider
- [x] network map
- [x] storage map
- [x] Plans
- [x] Migration

PR #195, opened 2023-03-06
Create the relevant secret for populator

liranr23:
When migrating using the populator, a secret is required. It differs in the data and name compared to DataVolumes. The secret is required in the target namespace. This patch ensure the secret will be there with the relevant data and name.

### DIFF
--- a/pkg/controller/plan/kubevirt.go
+++ b/pkg/controller/plan/kubevirt.go
@@ -471,7 +471,7 @@ func (r *KubeVirt) SetRunning(vmCr *VirtualMachine, running bool) (err error) {
 }
 
 func (r *KubeVirt) DataVolumes(vm *plan.VMStatus) (dataVolumes []cdi.DataVolume, err error) {
-	secret, err := r.ensureSecret(vm.Ref)
+	secret, err := r.ensureSecret(vm.Ref, false)
 	if err != nil {
 		return
 	}
@@ -600,7 +600,7 @@ func (r *KubeVirt) getPVCs(vm *plan.VMStatus) (pvcs []core.PersistentVolumeClaim
 }
 
 func (r *KubeVirt) createVolumesForOvirt(vm ref.Ref) (err error) {
-	_, err = r.ensureSecret(vm)
+	secret, err := r.ensureSecret(vm, true)
 	if err != nil {
 		return
 	}
@@ -616,7 +616,7 @@ func (r *KubeVirt) createVolumesForOvirt(vm ref.Ref) (err error) {
 
 	storageName := &r.Context.Map.Storage.Spec.Map[0].Destination.StorageClass
 	for _, da := range ovirtVm.DiskAttachments {
-		populatorCr := r.OvirtVolumePopulator(da, sourceUrl)
+		populatorCr := r.OvirtVolumePopulator(da, sourceUrl, secret.Name)
 		failure := r.Client.Create(context.Background(), populatorCr, &client.CreateOptions{})
 		if failure != nil && !k8serr.IsAlreadyExists(failure) {
 			return failure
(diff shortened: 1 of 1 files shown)


eslutsky:
this PR fixed the issue when running the migration in different target Namespace (ie `default`),
https://github.com/kubev2v/forklift/pull/158

pkg/controller/plan/kubevirt.go
```diff
@@ -1519,6 +1519,10 @@ func (r *KubeVirt) secret(vmRef ref.Ref) (object *core.Secret, err error) {
 				"-") + "-",
 		},
 	}
+	if populator {
+		object.Data = r.Source.Secret.Data
+		return
+	}
```
ahadas, line 1525: how did CDI use the secret without having the data set?

ahadas (commented):
@liranr23 the description of this PR provides some information, yet, it doesn't explain why it's different than what we have for DVs when transferring the data using importer pods (CDI) - can you please explain?

pkg/controller/plan/kubevirt.go
liranr23: CDI has it with different key names (few lines below it calls the secret builder):
```
object.StringData = map[string]string{
		"accessKeyId": string(in.Data["user"]),
		"secretKey":   string(in.Data["password"]),
}
```
they also have configmap that holds the CA.

ahadas: ok, so let's be more explicit here - it's not the populator being used or not but whether we expect the secret to be a clone of the provider's secret or to have the information spread across the secret and the config map?

ahadas: > another thing is that we always provided the provider secret name instead the new one we create.

yes, @yaacov saw that when trying to migrate to a different namespace
we need to change the CI to exercise this flow

liranr23: #158 without the changes in the namespaces (using `default`) checks it out. this is how evgeny saw it and reached out.

pkg/controller/plan/kubevirt.go
```diff
@@ -1448,12 +1448,12 @@ func (r *KubeVirt) configMap(vmRef ref.Ref) (object *core.ConfigMap, err error)
 }
 
 // Ensure the DatVolume credential secret exists on the destination.
-func (r *KubeVirt) ensureSecret(vmRef ref.Ref) (secret *core.Secret, err error) {
+func (r *KubeVirt) ensureSecret(vmRef ref.Ref, populator bool) (secret *core.Secret, err error) {
```
ahadas, line 1451: ok, so following the thread below, how about renaming this to something like `cloneProviderSecret`?

pkg/controller/plan/kubevirt.go
```diff
@@ -1507,7 +1507,7 @@ func (r *KubeVirt) ensureSecret(vmRef ref.Ref) (secret *core.Secret, err error)
 }
 
 // Build the DataVolume credential secret.
-func (r *KubeVirt) secret(vmRef ref.Ref) (object *core.Secret, err error) {
+func (r *KubeVirt) secret(vmRef ref.Ref, populator bool) (object *core.Secret, err error) {
```
ahadas, line 1510: wouldn't it be nicer to pass a function that sets the data instead of propagating the arguments all the way down here?

liranr23: i think it will just complex it. it's ~1 liner at the moment. instead of propagating this boolean we will propagate the function...

ahadas: I think differently - it would be better to generalize this method with dependency injection rather than adding more code flows inside it, I'll post a separate PR to talk about the concrete changes I have in mind

PR #199, opened 2023-03-07
Cleanup in creation of secret for data transfers

ahadas:
follow-up on #195

### DIFF
--- a/pkg/controller/plan/adapter/openstack/builder.go
+++ b/pkg/controller/plan/adapter/openstack/builder.go
@@ -218,14 +218,7 @@ func (r *Builder) ResolvePersistentVolumeClaimIdentifier(pvc *core.PersistentVol
 
 // Build credential secret.
 func (r *Builder) Secret(_ ref.Ref, in, secret *core.Secret) (err error) {
-	secret.StringData = map[string]string{
-		"username":           string(in.Data["username"]),
-		"password":           string(in.Data["password"]),
-		"domainName":         string(in.Data["domainName"]),
-		"projectName":        string(in.Data["projectName"]),
-		"region":             string(in.Data["region"]),
-		"insecureSkipVerify": string(in.Data["insecureSkipVerify"]),
-	}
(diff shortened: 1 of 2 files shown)


pkg/controller/plan/kubevirt.go
```diff
@@ -1644,7 +1650,7 @@ func vmOwnerReference(vm *cnv.VirtualMachine) (ref meta.OwnerReference) {
 
 // TODO move elsewhere
 func (r *KubeVirt) createOpenStackVolumes(vm ref.Ref) (err error) {
-	secret, err := r.ensureSecret(vm, false)
+	secret, err := r.ensureSecret(vm, r.secretDataSetterForCDI(vm))
```
liranr23, line 1653: CDI doesn't support OSP. but we have singular usage of the builder. that's why it was ok with false.

ahadas, line 1653: for openstack it doesn't make much sense as you clone the data and we don't use CDI.. I'll change that

ahadas: sure, it's not that it didn't work - but it was misleading, we actually do want to clone the provider secret for openstack as well, now that the code is more explicit we can see that

pkg/controller/plan/adapter/openstack/builder.go
```diff
@@ -216,19 +216,6 @@ func (r *Builder) ResolvePersistentVolumeClaimIdentifier(pvc *core.PersistentVol
 	return ""
 }
 
-// Build credential secret.
-func (r *Builder) Secret(_ ref.Ref, in, secret *core.Secret) (err error) {
```
liranr23, line 220: you need to have this function and just return for the builder API.

ahadas: yeah, already done

liranr23 (approved):

bennyz (approved):

PR #204, opened 2023-03-07
progress reporting via metrics server

bennyz:
- [x] make controller play nice with both ovirt and openstack
- [x] cleanup


cmd/openstack-populator/openstack-populator.go
```diff
@@ -97,8 +104,25 @@ func loadConfig(secretName, endpoint, namespace string) openstackConfig {
...
+func populate(fileName, endpoint, secretName, imageID, controllerIp string) {
+	http.Handle("/metrics", promhttp.Handler())
+	go http.ListenAndServe(":2112", nil)
+	progressGague := prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Subsystem: "volume_populators",
+			Name:      "openstack_volume_populator",
+			Help:      "Amount of data transferred",
+		},
+		[]string{"image_id"},
```
liranr23, line 116: here it's image_id, ovirt will use something else and v2v already using disk_id (or planning to be). it would be nice if we could have a something single to align in the forklift-controller in order to read.

pkg/lib-volume-populator/populator-machinery/controller.go
```diff
@@ -723,14 +734,63 @@ func (c *controller) syncPvc(ctx context.Context, key, pvcNamespace, pvcName str
...
+		}
+		c.recorder.Eventf(pod, corev1.EventTypeWarning, reasonPopulatorProgress, "Failed to get url: %s, err: %w", url, err)
+
+		return nil
+	}
+	defer resp.Body.Close()
+	body, err := ioutil.ReadAll(resp.Body)
+
+	if err != nil {
+		c.recorder.Eventf(pod, corev1.EventTypeWarning, reasonPopulatorProgress, "Failed to read bosy: %s, err: %w", url, err)
```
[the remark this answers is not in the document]
bennyz: debugging, will be removed...

pkg/lib-volume-populator/populator-machinery/controller.go
```diff
@@ -723,14 +734,63 @@ func (c *controller) syncPvc(ctx context.Context, key, pvcNamespace, pvcName str
...
+	if matches == nil {
+		c.recorder.Eventf(pod, corev1.EventTypeWarning, reasonPopulatorProgress, "Failed to find matchs: %s", matches)
+		return nil
+	}
+
+	for _, match := range matches {
+		imageID := match[1]
+		progress, _ := strconv.ParseFloat(string(match[2]), 64)
+		c.recorder.Eventf(pod, corev1.EventTypeNormal, reasonPopulatorProgress, "disk_nr: %s, progress %d", imageID, int64(progress))
+		gvr := schema.GroupVersionResource{Group: *pvc.Spec.DataSourceRef.APIGroup, Version: "v1beta1", Resource: "openstackvolumepopulators"}
```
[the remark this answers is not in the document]
bennyz: of course

pkg/lib-volume-populator/populator-machinery/controller.go
```diff
@@ -723,14 +734,63 @@ func (c *controller) syncPvc(ctx context.Context, key, pvcNamespace, pvcName str
...
 func makePopulatePodSpec(pvcPrimeName, secretName string) corev1.PodSpec {
 	nonRoot := true
 	allowPrivilageEscalation := false
 	user := int64(qemuGroup)
 	return corev1.PodSpec{
 		Containers: []corev1.Container{
 			{
-				Name: populatorContainerName,
+				Name:            populatorContainerName,
+				ImagePullPolicy: corev1.PullAlways,
```
ahadas, line 792: better to drop the image pull policy

bennyz: it's for debugging, must not forget :)

cmd/openstack-populator/openstack-populator.go
```diff
@@ -155,61 +179,28 @@ func (cr *CountingReader) Read(p []byte) (int, error) {
...
-	// 		klog.Fatal(err.Error())
-	// 	}
-
-	// 	client, err := dynamic.NewForConfig(config)
-	// 	if err != nil {
-	// 		klog.Fatal(err.Error())
-	// 	}
-	// 	gvr := schema.GroupVersionResource{Group: groupName, Version: apiVersion, Resource: resource}
-
+func writeData(reader io.ReadCloser, file *os.File, imageID, controllerIp string, progress *prometheus.GaugeVec) error {
```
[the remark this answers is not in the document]
bennyz: indeed

cmd/openstack-populator/openstack-populator.go
```diff
@@ -155,61 +179,28 @@ func (cr *CountingReader) Read(p []byte) (int, error) {
...
-	// }
-	// status := map[string]interface{}{"transferred": *countingReader.total}
-	// unstructured.SetNestedField(populatorCr.Object, status, "status")
-
-	// _, err = client.Resource(gvr).Namespace(crNamespace).Update(context.TODO(), populatorCr, metav1.UpdateOptions{})
-	// if err != nil {
-	// 	klog.Error(err)
-	// }
+	done <- true
+	progress.WithLabelValues(imageID).Set(float64(*total))
```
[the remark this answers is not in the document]
bennyz: I can actually set to 100% percent here and it would be simpler, but it would be slightly harder for ovirt because we'll need to read the full size from disk or something similar

cmd/openstack-populator/openstack-populator.go
bennyz: disk_id doesn't really make sense for openstack, we can revisit this when it's actually unified

pkg/lib-volume-populator/populator-machinery/controller.go
```diff
@@ -614,6 +647,28 @@ func (c *controller) syncPvc(ctx context.Context, key, pvcNamespace, pvcName str
...
+							pod, err = c.podLister.Pods(populatorNamespace).Get(pod.Name)
+							if err != nil {
+								break
+							}
+							if pod.Status.Phase != corev1.PodRunning {
+								break
+							}
+
+							// TODO make configurable?
+							time.Sleep(5 * time.Second)
```
ahadas, line 667: this takes a different approach than the monitoring of virt-v2v as the latter initiates a call in each reconciliation cycle, right? wouldn't it be enough to do that when we reconcile the PVC?

cmd/openstack-populator/openstack-populator.go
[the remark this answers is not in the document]
ahadas: yeah, the actual/apparent size.. I think that for completeness we'll want to keep that "logic" in the controller that sets the transfer to 100% when the populator pod is completed successfully so it's not that important

pkg/lib-volume-populator/populator-machinery/controller.go
```diff
@@ -720,17 +775,126 @@ func (c *controller) syncPvc(ctx context.Context, key, pvcNamespace, pvcName str
...
+		if err != nil {
+			c.recorder.Eventf(pvc, corev1.EventTypeWarning, reasonPopulatorProgress, "Failed to get CR for kind: %s error: %w", populatorKind, err)
+			break
+		}
+
+		var updatedPopulator *unstructured.Unstructured
+		switch populatorKind {
+		case v1beta1.OpenstackVolumePopulatorKind:
+			updatedPopulator, err = createUpdatedOpenstackPopulatorProgress(int64(progress), latestPopulator)
+		// TODO add const
```
ahadas, line 837: +1

cmd/openstack-populator/openstack-populator.go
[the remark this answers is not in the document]
ahadas: yeah, the actual/apparent size.. I think that for completeness we'll want to keep that "logic" in the controller that sets the transfer to 100% when the populator pod is completed successfully so it's not that important

pkg/lib-volume-populator/populator-machinery/controller.go
ahadas: but that's the beauty of the default policy - that when the "latest" tag is used, it always pull the image - that's handy during development (we use 'devel' tag by default in our script, we can change it to 'latest')

pkg/lib-volume-populator/populator-machinery/controller.go
bennyz: Because the populator controller is callback based, it's not reconciled periodically, so without this we'll get only one read...

ahadas: ack

ahadas (approved):
lgtm

PR #207, opened 2023-03-08
Adding unit testing for certificate appending in the mutator

bkhizgiy:


pkg/forklift-api/webhooks/mutating-webhook/mutators/ovirt-cert-mutator.go
```diff
@@ -22,6 +22,8 @@ import (
 
 var log = logging.WithName("mutator")
 
+const NewLine = 0x0a
```
bkhizgiy, line 25: added a small cosmetic change so it will be easier to read:)

pkg/forklift-api/webhooks/mutating-webhook/mutators/ovirt-cert-mutator_test.go
```diff
@@ -0,0 +1,34 @@
...
+)
+
+func TestCertAppanding(t *testing.T) {
+	g := gomega.NewGomegaWithT(t)
+
+	providedCa, err := ioutil.ReadFile("originalCert.pem")
+	if err != nil {
+		panic(err)
+	}
+	newCa, err := ioutil.ReadFile("newCert.pem")
```
ahadas, line 17: the names 'providedCa' and 'newCa' is a bit confusing - maybe we should write here, even just for making things more clear, that we don't expect the provided Ca to contain the 'newCa' in practice? (because users are not supposed to provide both certificates) or change the 'providedCa' to include a single certificate?

pkg/forklift-api/webhooks/mutating-webhook/mutators/ovirt-cert-mutator_test.go
```diff
@@ -0,0 +1,34 @@
...
+
+	"github.com/onsi/gomega"
+)
+
+func TestCertAppanding(t *testing.T) {
+	g := gomega.NewGomegaWithT(t)
+
+	providedCa, err := ioutil.ReadFile("originalCert.pem")
+	if err != nil {
+		panic(err)
```
ahadas, line 15: if we use gomega (which is cool!) we can check that no error have-occurred

pkg/forklift-api/webhooks/mutating-webhook/mutators/ovirt-cert-mutator_test.go
bkhizgiy: good point, since the tests are only testing each part separately, we need to have the file with both of the certs to check that the contains function is working correctly for the flow that engineCa is parsed differently (diffrent lines count). In that case how about changing the names of the files to something like "completeCerts" and "engineerCert"?

pkg/forklift-api/webhooks/mutating-webhook/mutators/ovirt-cert-mutator_test.go
bkhizgiy: cool, wasn't aware of it, changing.

pkg/forklift-api/webhooks/mutating-webhook/mutators/ovirt-cert-mutator_test.go
```diff
@@ -0,0 +1,34 @@
...
+	}
+	//Test the case where two certificates are identical but have a different line count due to new lines.
+	g.Expect(contains(providedCa, newCa)).To(gomega.Equal(true))
+
+	//Test the case where the original certificate does not have a new line at the end.
+	g.Expect(appendCerts(newCa, newCa)).To(gomega.ContainSubstring("-----END CERTIFICATE-----\n-----BEGIN CERTIFICATE-----"))
+
+	//Test the case where the original certificate has a new line at the end and verify a redundant new line was not added.
+	newCa = append(newCa, 0x0a)
+	g.Expect(appendCerts(newCa, newCa)).To(gomega.ContainSubstring("-----END CERTIFICATE-----\n-----BEGIN CERTIFICATE-----"))
```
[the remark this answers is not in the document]
bkhizgiy: It checking that only one new line exists, the substring is the point where the original cert and the appended one merge, so it is supposed to contain only one "/n".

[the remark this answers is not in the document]
ahadas: ah I see, ok

pkg/forklift-api/webhooks/mutating-webhook/mutators/ovirt-cert-mutator_test.go
ahadas: changing the names to better reflect their purpose sound good but did you mean the second is 'engineCert'?

pkg/forklift-api/webhooks/mutating-webhook/mutators/ovirt-cert-mutator_test.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/onsi/gomega"
+)
+
+func TestCertAppending(t *testing.T) {
+	g := gomega.NewGomegaWithT(t)
+
+	providedCa, err := ioutil.ReadFile("completeCerts.pem")
+	if err != nil {
+		g.Fail("Failed to read certificate file")
+	}
```
ahadas, line 16: so how about replacing this with `Expect(err).ToNot(HaveOccurred())`?

pkg/forklift-api/webhooks/mutating-webhook/mutators/ovirt-cert-mutator_test.go
```diff
@@ -0,0 +1,36 @@
...
+
+	newCa, err := ioutil.ReadFile("engineCert.pem")
+	if err != nil {
+		g.Fail("Failed to read engine certificate file")
+	}
+
+	//Test the case where two certificates are identical but have a different line count due to new lines.
+	g.Expect(contains(providedCa, newCa)).To(gomega.Equal(true))
```
ahadas, line 24: one of the nice things about gomega is that you can write things in a readable way and having that `gomega.` everywhere makes it less readable. how about changing the import to `. "github.com/onsi/gomega"` and then change this to `g.Expect(contains(providedCa, newCa)).To(BeTrue()))` (and drop 'gomega.' below)

bkhizgiy: nice!

pkg/forklift-api/webhooks/mutating-webhook/mutators/ovirt-cert-mutator_test.go
bkhizgiy: I've used it initially and removed it because the error presented in that way was less nice, but I agree that its more readable and clean that way so switching back.

ahadas (approved):

ahadas:
@bkhizgiy the code changes look good but the test fails, please take a look at [the failure](https://github.com/kubev2v/forklift/actions/runs/4376548761/jobs/7658816739)

bkhizgiy (commented):
@ahadas  yes, had missing calls and accidentally pushed redundant files, should work now.

ahadas (approved):

PR #209, opened 2023-03-08
Create a consumer to PVC

liranr23:
When having a storage class with `WaitForConsumer` the disk transfer won't start until the PVC had a consumer and it has the annotation `volume.kubernetes.io/selected-node`. For that, we will create a simple pod associated with the PVC to let kubernetes schedule it - making a consumer to the PVC.

When executed by the volume populator, the PV won't create, letting the volume populator to be executed. As for vSphere, the data is overriden when transferred.

The pod will remain on destination namespace. Once the plan is archived it will be cleaned up.

### DIFF
--- a/pkg/controller/plan/kubevirt.go
+++ b/pkg/controller/plan/kubevirt.go
@@ -357,22 +357,9 @@ func (r *KubeVirt) DeleteSecret(vm *plan.VMStatus) (err error) {
 		return
 	}
 	for _, object := range list.Items {
-		err = r.Destination.Client.Delete(context.TODO(), &object)
+		err = r.DeleteObject(&object, vm, "Deleted secret.", "secret")
 		if err != nil {
-			if k8serr.IsNotFound(err) {
-				err = nil
-			} else {
-				return liberr.Wrap(err)
-			}
-		} else {
-			r.Log.Info(
-				"Deleted secret.",
-				"secret",
-				path.Join(
-					object.Namespace,
-					object.Name),
-				"vm",
-				vm.String())
+			return err
 		}
 	}
 	return
@@ -396,22 +383,9 @@ func (r *KubeVirt) DeleteConfigMap(vm *plan.VMStatus) (err error) {
 		return
 	}
 	for _, object := range list.Items {
-		err = r.Destination.Client.Delete(context.TODO(), &object)
+		err = r.DeleteObject(&object, vm, "Deleted configMap.", "configMap")
 		if err != nil {
-			if k8serr.IsNotFound(err) {
-				err = nil
-			} else {
-				return liberr.Wrap(err)
-			}
-		} else {
-			r.Log.Info(
-				"Deleted configMap.",
-				"configMap",
-				path.Join(
-					object.Namespace,
-					object.Name),
-				"vm",
-				vm.String())
+			return err
 		}
 	}
 	return
@@ -599,7 +573,7 @@ func (r *KubeVirt) getPVCs(vm *plan.VMStatus) (pvcs []core.PersistentVolumeClaim
 	return
 }
 
-func (r *KubeVirt) createVolumesForOvirt(vm ref.Ref) (err error) {
+func (r *KubeVirt) createVolumesForOvirt(vm ref.Ref) (pvcNames []string, err error) {
 	secret, err := r.ensureSecret(vm, r.copyDataFromProviderSecret)
 	if err != nil {
 		return
@@ -619,12 +593,12 @@ func (r *KubeVirt) createVolumesForOvirt(vm ref.Ref) (err error) {
 		populatorCr := r.OvirtVolumePopulator(da, sourceUrl, secret.Name)
 		failure := r.Client.Create(context.Background(), populatorCr, &client.CreateOptions{})
 		if failure != nil && !k8serr.IsAlreadyExists(failure) {
-			return failure
+			return nil, failure
 		}
 
 		accessModes, volumeMode, failure := r.getStorageProfileModes(*storageName)
 		if failure != nil {
-			return failure
+			return nil, failure
 		}
 
 		pvc := r.Builder.PersistentVolumeClaimWithSourceRef(da, storageName, populatorCr.Name, accessModes, volumeMode)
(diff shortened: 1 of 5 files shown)


pkg/controller/plan/kubevirt.go
```diff
@@ -636,11 +636,51 @@ func (r *KubeVirt) createVolumesForOvirt(vm ref.Ref) (err error) {
 		if err != nil {
 			return
 		}
+		err = r.createPodToBindPVC(vm, pvc.Name)
```
ahadas, line 639: it's not really part of the creation of the volumes.. maybe a better place would be `Execute` in migration.go?

pkg/controller/plan/kubevirt.go
```diff
@@ -636,11 +636,51 @@ func (r *KubeVirt) createVolumesForOvirt(vm ref.Ref) (err error) {
...
+		},
+		Spec: core.PodSpec{
+			RestartPolicy: core.RestartPolicyNever,
+			Containers: []core.Container{
+				{
+					Name:    name,
+					Image:   "ubi8",
+					Command: []string{"/bin/sh"},
+				},
+			},
```
ahadas, line 671: it's great that it works without mounting the volumes - that way we can be sure that we can 'populate' even before the execution of this pod stops

ahadas:
@liranr23 ` s: "error: [Pod \"plan-test-vm-55-pvcinitvlc9d\" is invalid: spec.volumes[0].persistentVolumeClaim.claimName: Required value]",` looks suspicious

ahadas:
> @liranr23 ` s: "error: [Pod \"plan-test-vm-55-pvcinitvlc9d\" is invalid: spec.volumes[0].persistentVolumeClaim.claimName: Required value]",` looks suspicious

ok, I see a similar issue without these changes so it seems like a consequence of an issue we have already

liranr23:
I do wonder about:
1. Cleanup - do we wish to clean the pod even without archive the plan?
2. skip the pod creation if the storage class is immediate? it will require to query the storage class.

pkg/controller/plan/kubevirt.go
```diff
@@ -599,13 +611,13 @@ func (r *KubeVirt) getPVCs(vm *plan.VMStatus) (pvcs []core.PersistentVolumeClaim
 	return
 }
 
-func (r *KubeVirt) createVolumesForOvirt(vm ref.Ref) (err error) {
-	secret, err := r.ensureSecret(vm, r.copyDataFromProviderSecret)
+func (r *KubeVirt) createVolumesForOvirt(vm *plan.VMStatus) (pvcNames []string, err error) {
```
ahadas, line 614: we can change the argument back to ref.Ref

pkg/controller/plan/kubevirt.go
```diff
@@ -1654,15 +1712,15 @@ func vmOwnerReference(vm *cnv.VirtualMachine) (ref meta.OwnerReference) {
 }
 
 // TODO move elsewhere
-func (r *KubeVirt) createOpenStackVolumes(vm ref.Ref) (err error) {
-	secret, err := r.ensureSecret(vm, r.copyDataFromProviderSecret)
+func (r *KubeVirt) createOpenStackVolumes(vm *plan.VMStatus) (pvcNames []string, err error) {
```
ahadas, line 1715: and also here

pkg/controller/plan/migration.go
```diff
@@ -599,15 +599,28 @@ func (r *Migration) execute(vm *plan.VMStatus) (err error) {
...
+				step.AddError(err.Error())
+				err = nil
+				break
+			}
 		}
 		if r.kubevirt.useOvirtPopulator(vm) {
-			err = r.kubevirt.createVolumesForOvirt(vm.Ref)
+			pvcNames, err = r.kubevirt.createVolumesForOvirt(vm)
+			if err != nil {
+				step.AddError(err.Error())
```
ahadas, line 619: why is it different than the error handling above (after the call to createOpenStackVolumes)?

liranr23: i went with the alignment we have of errors.
IMO doing it for step is failing the step while the wrapper just logs it and keep reconciling.

pkg/controller/plan/kubevirt.go
```diff
@@ -636,11 +636,51 @@ func (r *KubeVirt) createVolumesForOvirt(vm ref.Ref) (err error) {
...
+			PersistentVolumeClaim: &core.PersistentVolumeClaimVolumeSource{
+				ClaimName: pvcName,
+			},
+		},
+	})
+	pod := &core.Pod{
+		ObjectMeta: meta.ObjectMeta{
+			Namespace:    r.Plan.Spec.TargetNamespace,
+			Labels:       r.vmLabels(vm.Ref),
+			GenerateName: r.getGeneratedName(vm) + "pvcinit",
```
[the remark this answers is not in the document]
liranr23: now it should deal with multiple PVCs so i hope it's ok

pkg/controller/plan/migration.go
ahadas: @bennyz so what's the reason for doing it different for openstack above?

pkg/controller/plan/migration.go
```diff
@@ -675,6 +692,11 @@ func (r *Migration) execute(vm *plan.VMStatus) (err error) {
 			err = nil
 			break
 		}
+		err = r.kubevirt.DeletePVCConsumerPod(vm)
```
ahadas, line 695: why do we need to remove it here as well? isn't it enough to delete in the cleanup code?

pkg/controller/plan/kubevirt.go
```diff
@@ -809,41 +825,79 @@ func (r *KubeVirt) GetGuestConversionPod(vm *plan.VMStatus) (pod *core.Pod, err
...
 // Delete the guest conversion pod on the destination cluster.
 func (r *KubeVirt) DeleteGuestConversionPod(vm *plan.VMStatus) (err error) {
+	list, err := r.GetPods(vm)
+	if err != nil {
+		return liberr.Wrap(err)
+	}
+	for _, object := range list.Items {
+		if strings.Contains(object.GenerateName, "pvcinit") {
+			continue
+		}
```
ahadas, line 854: that's not that nice.. so why not to remove the consumer-pods first?

pkg/controller/plan/migration.go
bennyz: looks like a mistake on my part, not something intentional

pkg/controller/plan/migration.go
liranr23: cleanup is called when restarting/archiving the plan. here it's - we done with the migration, let's clean it.

pkg/controller/plan/migration.go
liranr23: changing the above error handling

pkg/controller/plan/migration.go
liranr23: oh, i thought we do based on the above discussion:

> > 1. Cleanup - do we wish to clean the pod even without archive the plan?

> it would be better, yes

ahadas (approved):

bennyz (approved):

ahadas (approved):

PR #210, opened 2023-03-09
Fix UseEl9VirtV2v

ahadas:
We should use virt-v2v on el9 only on cold-migration from vsphere to the same cluster that forklift runs on.


bennyz (approved):

liranr23 (approved):

ahadas:
it (import that doesn't involve virt-v2v on el9) still doesn't work but for other reasons - I'll post those changes separately

PR #212, opened 2023-03-09
add more logs to CI

eslutsky:


tests/suit/framework/framework.go
```diff
@@ -137,7 +137,9 @@ func (f *Framework) BeforeEach() {
 		})
 		gomega.Expect(err).NotTo(gomega.HaveOccurred())
 		f.Namespace = ns
-		f.AddNamespaceToDelete(ns)
+
+		//dont delete namespace after tests are done
```
ahadas, line 141: or to have a flag that is not set by default for deleting the generated namespace

.github/workflows/pull-request.yml
```diff
@@ -77,15 +77,42 @@ jobs:
...
-            kubectl get pvc -A >> /tmp/artifacts/k8s-pvc.log
             kubectl cp konveyor-forklift/$(kubectl get po -n konveyor-forklift  -o=name | grep packstack|cut -d/ -f2):/var/log/cinder/volume.log /tmp/artifacts/k8s-packstack-cinder-volume.log
+            kubectl cp konveyor-forklift/$(kubectl get po -n konveyor-forklift  -o=name | grep packstack|cut -d/ -f2):/var/log/nova/nova-compute.log /tmp/artifacts/k8s-packstack-nova-compute.log
+            kubectl cp konveyor-forklift/$(kubectl get po -n konveyor-forklift  -o=name | grep packstack|cut -d/ -f2):/var/log/glance/api.log /tmp/artifacts/k8s-packstack-glance-api.log
+            kubectl get openstackvolumepopulator -A >> /tmp/artifacts/k8s-openstackvolumepopulator.log
+            kubectl logs -n konveyor-forklift $(kubectl get po -n konveyor-forklift  -o=name | grep openstack-populator) >> /tmp/artifacts/k8s-forklift-openstack-populator.log
+            kubectl logs -n konveyor-forklift $(kubectl get po -n konveyor-forklift  -o=name | grep forklift-volume-populator) >> /tmp/artifacts/k8s-forklift-volume-populator.log
```
ahadas, line 111: it would be handy to get more complete logs (of all objects) from the generated namespace

eslutsky: what do you mean ? there is no pods on the generated NS , only CRs.. we can have logs from pods only

tests/suit/framework/framework.go
eslutsky: done, added flags `-delete-namespace` which can be added to the ginkgo execution parameter.

.github/workflows/pull-request.yml
eslutsky: done, see [commit](https://github.com/kubev2v/forklift/pull/212/commits/e022770f57d9b6cc130a22447ab7429b362d193f)

.github/workflows/pull-request.yml
```diff
@@ -77,15 +77,45 @@ jobs:
...
```
ahadas, line 114: we use the term populator pod for openstack/ovirt and populator-controller for the controller in the konveyor-forklift namespace so how about:
```suggestion
            kubectl logs -n konveyor-forklift $(kubectl get po -n konveyor-forklift  -o=name | grep forklift-volume-populator) >> /tmp/artifacts/k8s-forklift-volume-populator-controller.log
```

eslutsky: ack and done

ahadas (approved):

PR #215, opened 2023-03-12
Create a consumer for PVCs for vsphere

liranr23:
When importing a VM from vSphere and the storage class is set with binding mode "wait for consumer" the migration gets stuck, as the PVC and DV will wait for consumer.

This patch solves it by creating a dummy pod to initiate the PVC. It continues PR #209.

### DIFF
--- a/pkg/controller/plan/kubevirt.go
+++ b/pkg/controller/plan/kubevirt.go
@@ -74,6 +74,8 @@ const (
 	kPlan = "plan"
 	// VM label (value=vmID)
 	kVM = "vmID"
+	// App label
+	kApp = "forklift.app"
 )
 
 // Map of VirtualMachines keyed by vmID.
@@ -476,6 +478,7 @@ func (r *KubeVirt) EnsureDataVolumes(vm *plan.VMStatus, dataVolumes []cdi.DataVo
 		return
 	}
 
+	var pvcNames []string
 	for _, dv := range dataVolumes {
 		exists := false
 		for _, item := range list.Items {
@@ -498,6 +501,14 @@ func (r *KubeVirt) EnsureDataVolumes(vm *plan.VMStatus, dataVolumes []cdi.DataVo
 					dv.Name),
 				"vm",
 				vm.String())
+		}
+		// DataVolume and PVC names are the same
+		pvcNames = append(pvcNames, dv.Name)
+	}
+	if r.UseEl9VirtV2v() {
+		err = r.createPodToBindPVCs(vm, pvcNames)
+		if err != nil {
+			return err
 		}
 	}
 
@@ -633,15 +644,15 @@ func (r *KubeVirt) createPodToBindPVCs(vm *plan.VMStatus, pvcNames []string) err
 	pod := &core.Pod{
 		ObjectMeta: meta.ObjectMeta{
 			Namespace:    r.Plan.Spec.TargetNamespace,
-			Labels:       r.vmLabels(vm.Ref),
-			GenerateName: r.getGeneratedName(vm) + "pvcinit",
+			Labels:       r.consumerLabels(vm.Ref),
+			GenerateName: r.getGeneratedName(vm) + "pvcinit-",
 		},
 		Spec: core.PodSpec{
 			RestartPolicy: core.RestartPolicyNever,
 			Containers: []core.Container{
 				{
 					Name:    name,
-					Image:   "ubi9-minimal",
+					Image:   Settings.Migration.VirtV2vImageCold,
 					Command: []string{"/bin/sh"},
 				},
 			},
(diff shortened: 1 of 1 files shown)


pkg/controller/plan/kubevirt.go
```diff
@@ -499,6 +500,14 @@ func (r *KubeVirt) EnsureDataVolumes(vm *plan.VMStatus, dataVolumes []cdi.DataVo
 				"vm",
 				vm.String())
 		}
+		pvcNames = append(pvcNames, dv.Name)
+	}
+	// DataVolume and PVC name are the same
+	if r.Plan.Provider.Source.Type() == v1beta1.VSphere {
```
ahadas, line 506: this is needed only when CDI is not used so better to use `UseEl9VirtV2v`

pkg/controller/plan/kubevirt.go
```diff
@@ -784,8 +793,14 @@ func (r *KubeVirt) EnsureGuestConversionPod(vm *plan.VMStatus, vmCr *VirtualMach
...
 	}
+
+	for _, pod := range list.Items {
+		if !strings.Contains(pod.Name, "pvcinit") {
+			return
+		}
+	}
 	pod := &core.Pod{}
-	if len(list.Items) == 0 {
+	if len(list.Items) <= 1 {
```
ahadas, line 803: if we get here, it necessarily means that there's no conversion pod, no?
the logic in this block of changes seems correct but it's not really readable..

ahadas (commented):
I've made few phrasing changes in the description of this PR, please change the commit message as well

pkg/controller/plan/kubevirt.go
liranr23: yeah, removed the `if` and changed to it be more readable (i hope)

liranr23:
Left to figure why only in `vsphere` CI suit the `ubi9-minimal` image can't be found. @eslutsky

pkg/controller/plan/kubevirt.go
```diff
@@ -499,6 +500,14 @@ func (r *KubeVirt) EnsureDataVolumes(vm *plan.VMStatus, dataVolumes []cdi.DataVo
 				"vm",
 				vm.String())
 		}
+		pvcNames = append(pvcNames, dv.Name)
+	}
+	// DataVolume and PVC name are the same
```
ahadas, line 505: this comment is supposed to be next to `pvcNames = append(pvcNames, dv.Name)`, no?

pkg/controller/plan/kubevirt.go
```diff
@@ -784,24 +793,28 @@ func (r *KubeVirt) EnsureGuestConversionPod(vm *plan.VMStatus, vmCr *VirtualMach
...
-	pod := &core.Pod{}
-	if len(list.Items) == 0 {
-		pod = newPod
-		err = r.Destination.Client.Create(context.TODO(), pod)
-		if err != nil {
-			err = liberr.Wrap(err)
+
+	for _, pod := range list.Items {
+		if !strings.Contains(pod.Name, "pvcinit") {
+			// There is a conversion pod.
```
ahadas, line 799: as I wrote you offline, I think it would be better to set these pods with a label that reflects their purpose rather than having this logic - can we do that separately?

pkg/controller/plan/kubevirt.go
liranr23: yup :)

pkg/controller/plan/kubevirt.go
```diff
@@ -499,6 +502,14 @@ func (r *KubeVirt) EnsureDataVolumes(vm *plan.VMStatus, dataVolumes []cdi.DataVo
 				"vm",
 				vm.String())
 		}
+		// DataVolume and PVC name are the same
```
ahadas, line 505:
```suggestion
		// DataVolume and PVC names are the same
```

pkg/controller/plan/kubevirt.go
```diff
@@ -633,15 +644,15 @@ func (r *KubeVirt) createPodToBindPVCs(vm *plan.VMStatus, pvcNames []string) err
...
+			Labels:       r.consumerLabels(vm.Ref),
+			GenerateName: r.getGeneratedName(vm) + "pvcinit-",
 		},
 		Spec: core.PodSpec{
 			RestartPolicy: core.RestartPolicyNever,
 			Containers: []core.Container{
 				{
 					Name:    name,
-					Image:   "ubi9-minimal",
+					Image:   "registry.access.redhat.com/ubi9-minimal:latest",
```
ahadas, line 655: @liranr23 thinking about this again, the images are cached on the nodes so maybe it's actually better to use the image of the conversion pod here?

pkg/controller/plan/kubevirt.go
```diff
@@ -827,24 +830,22 @@ func (r *KubeVirt) GetGuestConversionPod(vm *plan.VMStatus) (pod *core.Pod, err
 
 // Delete the PVC consumer pod on the destination cluster.
 func (r *KubeVirt) DeletePVCConsumerPod(vm *plan.VMStatus) (err error) {
-	list, err := r.GetPods(vm)
+	list, err := r.GetPodsWithLabels(r.consumerLabels(vm.Ref))
```
ahadas, line 833: we don't need liberr.Wrap below, right?

pkg/controller/plan/kubevirt.go
ahadas: ok, that's better - filed #219 for using the right image, not that important though

ahadas (approved):

PR #218, opened 2023-03-13
operator: Add 2.4 support infomation

mnecas:


ahadas (approved):
we can drop forklift 2.2 from the doc next month once support for openshift 4.9 ends

PR #223, opened 2023-03-13
Set ownership and labels to the populator pod

liranr23:
This patch sets the PVC as the owner of the populator pod and the. It
also labels the OvirtVolumePopulator and the OpenstackVolumePopulator
CRs with the VM ID and the migration ID. This is useful when we will
wish to get the logs using must-gather or lookup for the relevant
populator resources.

The populator pod won't be deleted at the end of migration in order to
let us gather the logs.

Once cleanup is triggered by the controller, the CRs and populator pod
will be deleted.


pkg/controller/plan/kubevirt.go
```diff
@@ -31,7 +31,7 @@ import (
 	"github.com/konveyor/forklift-controller/pkg/apis/forklift/v1beta1/ref"
 	"github.com/konveyor/forklift-controller/pkg/controller/plan/adapter"
 	plancontext "github.com/konveyor/forklift-controller/pkg/controller/plan/context"
-	openstackutil "github.com/konveyor/forklift-controller/pkg/controller/plan/util"
+	util "github.com/konveyor/forklift-controller/pkg/controller/plan/util"
```
liranr23, line 34: redundant `util`

ahadas:
> This is useful when we will wish to get the logs using must-gather.

Indeed, not only when using must-gather - also when trying to investigate what's going on in a namespace that more than one vm is migrated to.

I've fixed the phrasing of the message, please update the commit message

pkg/controller/plan/util/openstack.go
```diff
@@ -9,11 +9,12 @@ import (
 	"github.com/konveyor/forklift-controller/pkg/controller/provider/web/openstack"
 )
 
-func OpenstackVolumePopulator(image *openstack.Image, sourceUrl *url.URL, targetNamespace, secretName, migrationName string) *api.OpenstackVolumePopulator {
+func OpenstackVolumePopulator(image *openstack.Image, sourceUrl *url.URL, targetNamespace, secretName, vmId string) *api.OpenstackVolumePopulator {
 	return &api.OpenstackVolumePopulator{
 		ObjectMeta: meta.ObjectMeta{
 			Name:      image.Name,
 			Namespace: targetNamespace,
+			Labels:    map[string]string{"vmID": vmId},
```
ahadas, line 17: the commit message says we use the migration name in the label to make it simpler to correlate - and here we set the vmId even though the VM doesn't exist at this point, so why not having the migration in-addition/instead?

pkg/lib-volume-populator/populator-machinery/controller.go
```diff
@@ -576,11 +578,23 @@ func (c *controller) syncPvc(ctx context.Context, key, pvcNamespace, pvcName str
...
 		// If the pod doesn't exist yet, create it
 		if pod == nil {
+			isController := false
+
 			// Make the pod
 			pod = &corev1.Pod{
 				ObjectMeta: metav1.ObjectMeta{
 					Name:      podName,
 					Namespace: populatorNamespace,
+					Labels:    map[string]string{"pvcID": pvc.Name},
```
ahadas, line 588: it's ok to set the PVC as the owner below - in case the pod wasn't removed along with the migration, but I think it would simplify things it we correlate the pod with the migration as well / instead

pkg/controller/plan/kubevirt.go
```diff
@@ -950,6 +931,116 @@ func (r *KubeVirt) DeleteHookJobs(vm *plan.VMStatus) (err error) {
...
+		return err
+	}
+	err = r.DeleteObject(&populatorCr, vm, "Deleted OvirtPopulator CR.", "OvirtVolumePopulator")
+	if err != nil {
+		return err
+	}
+	return
+}
+
+// Delete any populator pods that belong to a VM migration.
```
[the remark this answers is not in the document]
liranr23: yes as a label we can add it (a bit ugly due to the CRD changes), the problem we had is about making it as owner, which as you wrote, different namespaces are problematic.

pkg/controller/plan/kubevirt.go
```diff
@@ -950,6 +931,116 @@ func (r *KubeVirt) DeleteHookJobs(vm *plan.VMStatus) (err error) {
 	return
 }
 
+// Delete any populator resources that belong to a VM migration.
```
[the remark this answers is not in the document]
liranr23: I agree it's simplify the cleanup. But it will require additional changes in the populator controller and the CRDs. I don't know if we wish to have them (thinking of CDI usage). WDYT?

pkg/controller/plan/kubevirt.go
```diff
@@ -630,6 +615,17 @@ func (r *KubeVirt) createVolumesForOvirt(vm ref.Ref) (pvcNames []string, err err
 			return
 		}
 		pvcNames = append(pvcNames, pvc.Name)
+
+		err = k8sutil.SetOwnerReference(pvc, populatorCr, r.Scheme())
```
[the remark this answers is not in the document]
liranr23: who removes it when the migration is removed? i think nobody (unless they will archive the plan and it will clean it up). i think it's better to have an owner.

pkg/controller/plan/kubevirt.go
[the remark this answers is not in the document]
ahadas: why do we need to change the CRD to add a label?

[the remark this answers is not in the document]
liranr23: if you wish to have it when creating the populator pod, it need to be in the controller lib. then, to get the data of the migration id it needs to be propagated somehow, the populator-controller needs to provide it, using args. the args are provided from the CRD.

pkg/controller/plan/kubevirt.go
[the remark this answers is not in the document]
liranr23: the ownership is just another way of cleanup. to enter the CleanUp, you need to either archive or cancel the migration. many users may just delete it without these steps. in that case - if the migration passed and the plan deleted, we will have leftovers, but it still somehow relate to the PVC-VM. in case they are deleted we are really not interested/need these so the ownership will take care of it.

pkg/lib-volume-populator/populator-machinery/controller.go
```diff
@@ -749,17 +765,14 @@ func (c *controller) syncPvc(ctx context.Context, key, pvcNamespace, pvcName str
...
+		err = c.kubeClient.CoreV1().PersistentVolumeClaims(populatorNamespace).Delete(ctx, pvcPrime.Name, metav1.DeleteOptions{})
 		if err != nil {
 			return err
 		}
-	}
 
-	// If PVC' still exists, delete it
-	if pvcPrime != nil {
-		err = c.kubeClient.CoreV1().PersistentVolumeClaims(populatorNamespace).Delete(ctx, pvcPrime.Name, metav1.DeleteOptions{})
+		err = c.ensureFinalizer(ctx, pvcPrime, pvcKuberenetesFinalizer, false)
```
bennyz, line 775: I'm not sure about this, why is this needed? This is the kubernetes finalizer and it could lead to CSI not cleaning up the underlying storage

liranr23: it's the other way around, it unsure to remove the finalizer. without it we will get stuck with the prime pvc.

bennyz: Stuck because of what though? Since we are not the ones who set `kubernetes.io/pvc-protection` in theory we are not supposed to remove it

bennyz: maybe it's stuck because of the pvcinit pod?

pkg/controller/plan/kubevirt.go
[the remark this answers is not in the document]
ahadas: > the ownership is just another way of cleanup. 

does it imply that we would still initiate calls to remove the CRs when the migration is removed?

> to enter the CleanUp, you need to either archive or cancel the migration. many users may just delete it without these steps. in that case - if the migration passed and the plan deleted, we will have leftovers, but it still somehow relate to the PVC-VM.

while we don't have empirical data on the amount of users remove plans without archiving them, this should have been addressed already by the message that appears when deleting a plan without archiving it first - there's a bold statement that this would result in leaving behind leftovers.

> in case they are deleted we are really not interested/need these so the ownership will take care of it.

yes, I see why you went this way - my issue with that is that the pvc is not really the owner of the CR.
usually when you have true owner, you could specify it when creating a resource. for instance, when creating a migration you could say that the plan is its owner - the plan already exists, and the migration is derived by the plan.
here, the flow is awkward - we create the CR, we create the PVC that should point to this CR, and we patch the CR to set the PVC as its owner. this suggests that the PVC is not really its owner - and think about it, let's say that I migrate a VM and the migration completed successfully, then I remove the VM and its PVCs - does it necessarily mean that I want to remove all the resources that are associated with the migration that has not been removed yet? I don't know, maybe not - I would still expect to, e.g., 'Get logs' to return the logs, including the CR of the migration, even though the VM was removed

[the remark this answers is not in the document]
ahadas: > > to enter the CleanUp, you need to either archive or cancel the migration. many users may just delete it without these steps. in that case - if the migration passed and the plan deleted, we will have leftovers, but it still somehow relate to the PVC-VM.
> 
> while we don't have empirical data on the amount of users remove plans without archiving them, this should have been addressed already by the message that appears when deleting a plan without archiving it first - there's a bold statement that this would result in leaving behind leftovers.

BTW, if that's really something you'd like to address - why not to add a validation webhook that blocks removing a plan that has not been archived?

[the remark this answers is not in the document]
liranr23: > > the ownership is just another way of cleanup.
> 
> does it imply that we would still initiate calls to remove the CRs when the migration is removed?
> 
no, there are two ways with this PR - calling CleanUp(using cancel/archive) or removing the PVC(or VM+disks).
of course we can call it by ourselves but then we will miss the logs (and conflicting the below discussion).
> > to enter the CleanUp, you need to either archive or cancel the migration. many users may just delete it without these steps. in that case - if the migration passed and the plan deleted, we will have leftovers, but it still somehow relate to the PVC-VM.
> 
> while we don't have empirical data on the amount of users remove plans without archiving them, this should have been addressed already by the message that appears when deleting a plan without archiving it first - there's a bold statement that this would result in leaving behind leftovers.
> 
> > in case they are deleted we are really not interested/need these so the ownership will take care of it.
> 
> yes, I see why you went this way - my issue with that is that the pvc is not really the owner of the CR. usually when you have true owner, you could specify it when creating a resource. for instance, when creating a migration you could say that the plan is its owner - the plan already exists, and the migration is derived by the plan. here, the flow is awkward - we create the CR, we create the PVC that should point to this CR, and we patch the CR to set the PVC as its owner. this suggests that the PVC is not really its owner - and think about it, let's say that I migrate a VM and the migration completed successfully, then I remove the VM and its PVCs - does it necessarily mean that I want to remove all the resources that are associated with the migration that has not been removed yet? I don't know, maybe not - I would still expect to, e.g., 'Get logs' to return the logs, including the CR of the migration, even though the VM was removed

This is correct. If we look on that like this way - yes, it may make sense to leave it as is. Letting the user get logs while the VM deleted.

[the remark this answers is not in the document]
liranr23: > > > to enter the CleanUp, you need to either archive or cancel the migration. many users may just delete it without these steps. in that case - if the migration passed and the plan deleted, we will have leftovers, but it still somehow relate to the PVC-VM.
> > 
> > 
> > while we don't have empirical data on the amount of users remove plans without archiving them, this should have been addressed already by the message that appears when deleting a plan without archiving it first - there's a bold statement that this would result in leaving behind leftovers.
> 
> BTW, if that's really something you'd like to address - why not to add a validation webhook that blocks removing a plan that has not been archived?

I think it's too "hard" to limit the deletion of the plan by archive it. We can - on deletion trigger the archive as a step of deletion. or, give a proper message about leftovers.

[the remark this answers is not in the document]
ahadas: > I think it's too "hard" to limit the deletion of the plan by archive it. We can - on deletion trigger the archive as a step of deletion. or, give a proper message about leftovers.

you mean it's too "strict", right? it shouldn't be hard..
I wouldn't spend the time on triggering 'archive' before removing the plan - it could have been nice to implement it with finalizers instead of introducing the 'archive' operation, but past contributors faced issues with this approach and ended up adding the 'archive' operation. I just don't see much value in doing 'archive' before 'delete' compared to block delete of plans that are not archived

> > > the ownership is just another way of cleanup.
> > 
> > 
> > does it imply that we would still initiate calls to remove the CRs when the migration is removed?
> 
> no, there are two ways with this PR - calling CleanUp(using cancel/archive) or removing the PVC(or VM+disks). of course we can call it by ourselves but then we will miss the logs (and conflicting the below discussion).
> 
> > > to enter the CleanUp, you need to either archive or cancel the migration. many users may just delete it without these steps. in that case - if the migration passed and the plan deleted, we will have leftovers, but it still somehow relate to the PVC-VM.

I lost you here but it seems irrelevant if we agree that setting the PVC as the owner of the CR is not the right way to go - so we'll need to remove the CRs as well when cleaning after the migration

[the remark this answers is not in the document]
ahadas: > > or, give a proper message about leftovers.

@liranr23 we already present a message about leftovers when deleting a plan without archiving it, why isn't it a proper message?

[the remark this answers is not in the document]
liranr23: > > > or, give a proper message about leftovers.
> 
> @liranr23 we already present a message about leftovers when deleting a plan without archiving it, why isn't it a proper message?

i didn't notice we do :)
it is proper as long we have one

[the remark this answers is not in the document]
ahadas: > while we don't have empirical data on the amount of users remove plans without archiving them, this should have been addressed already by the message that appears when deleting a plan without archiving it first - there's a bold statement that this would result in leaving behind leftovers.

ah ok, that's the message I referred to here ^

pkg/lib-volume-populator/populator-machinery/controller.go
bennyz: yeah I wrote it when I thought we don't keep the populator after successful import...

pkg/controller/plan/kubevirt.go
```diff
@@ -969,6 +937,130 @@ func (r *KubeVirt) DeleteHookJobs(vm *plan.VMStatus) (err error) {
...
+
+// Get the OpenstackVolumePopulator CustomResource List.
+func (r *KubeVirt) getOpenstackPopulatorCrList() (populatorCrList v1beta1.OpenstackVolumePopulatorList, err error) {
+	populatorCrList = v1beta1.OpenstackVolumePopulatorList{}
+	err = r.Client.List(
+		context.TODO(),
+		&populatorCrList,
+		&client.ListOptions{
+			Namespace:     r.Plan.Spec.TargetNamespace,
+			LabelSelector: labels.SelectorFromSet(map[string]string{kMigration: string(r.Plan.Status.Migration.ActiveSnapshot().Migration.UID)}),
```
bennyz, line 974: why not `r.Migration.UID`?

liranr23: because `r.Migration.UID` isn't populated when the plan is not executed. meaning, if we try to clear it afterwards and query `r.Migration.UID` is nil.

pkg/controller/plan/kubevirt.go
```diff
@@ -969,6 +937,130 @@ func (r *KubeVirt) DeleteHookJobs(vm *plan.VMStatus) (err error) {
...
+		err = r.DeleteObject(&populatorCr, vm, "Deleted OvirtPopulator CR.", "OvirtVolumePopulator")
+		if err != nil {
+			return err
+		}
+	}
+	return
+}
+
+// Delete any populator pods that belong to a VM migration.
+func (r *KubeVirt) DeletePopulatorPods(vm *plan.VMStatus) (err error) {
```
ahadas, line 1050: can we say something like, if the CR has the migration label then get the pods by the migration label?

pkg/lib-volume-populator/populator-machinery/controller.go
```diff
@@ -586,13 +588,31 @@ func (c *controller) syncPvc(ctx context.Context, key, pvcNamespace, pvcName str
 				// Join the transfer network namespace and name
 				annotations[AnnDefaultNetwork] = fmt.Sprintf("%s/%s", transferNetwork["namespace"], transferNetwork["name"])
 			}
-
+			isController := false
+			migration, found, err := unstructured.NestedString(crInstance.Object, "metadata", "labels", "migration")
+			if err != nil {
+				return err
+			}
+			labels := map[string]string{"pvcID": pvc.Name}
```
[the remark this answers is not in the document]
ahadas: yeah, name and id are two different properties in the metadata and the name of the PVC might be an ID of another entity, but still - if the key is named `pvcID`, I would expect its value to be the ID of the pvc, or `pvcName` if the value is the name of the PVC

pkg/controller/plan/kubevirt.go
ahadas: ah yeah, sorry, I didn't provide the reason for this suggestion - I meant that in order to identify whether the populator pods are set with the label, we can check whether the CR is set with the migration label. I'm worried about cleanup of populator pods that were not set with the migration label (i.e., populator pods that were created by 2.4)

liranr23: yeah, understood. i did think of it as well when doing the "labels" update on the controller start. we don't need it on the populator pods. because - until this version, the populator pod is immediately deleted after completion. if it fails it loops forever. when this version will come in, if the user will archive the plan it will clean the populator CR, only then the populator pod will stop restarting - in that case he still may have leftovers such as PVCs. 
if we wish to overcome it, i think we may need to update the populator pods labels as well like this PR is doing to the CRs.

ahadas: I don't see why not to set the populator pods with a label that points to the migration - again, it eases debugging when we know the context that an entity was created in

so you rely on the populator pods being cleared in 2.4? one argument to separate out the limit on the number of restarts in case of failure is to backport it to 2.4 - let's not assume they are cleared

liranr23: done

pkg/controller/plan/migration.go
```diff
@@ -1595,6 +1621,143 @@ func (r *Migration) updateCopyProgressForOpenstack(vm *plan.VMStatus, step *plan
...
+func (r *Migration) setOpenStackPopulatorLabels(vm *plan.VMStatus, pvcs []core.PersistentVolumeClaim) (err error) {
+	openstackVm := &openstack.Workload{}
+	err = r.Source.Inventory.Find(openstackVm, vm.Ref)
+	if err != nil {
+		return
+	}
+	var images []*openstack.Image
+	for _, vol := range openstackVm.Volumes {
+		lookupName := fmt.Sprintf("%s-%s", r.Migration.Name, vol.ID)
+		image, err := r.getOpenstackImageByName(lookupName)
```
liranr23, line 1634: @bennyz can you confirm it's ok (getting the image name from inventory)?

pkg/lib-volume-populator/populator-machinery/controller.go
```diff
@@ -686,9 +706,11 @@ func (c *controller) syncPvc(ctx context.Context, key, pvcNamespace, pvcName str
...
 				c.recorder.Eventf(pvc, corev1.EventTypeWarning, reasonPodFailed, "Populator failed: %s", pod.Status.Message)
 				// Delete failed pods so we can try again
-				err = c.kubeClient.CoreV1().Pods(populatorNamespace).Delete(ctx, pod.Name, metav1.DeleteOptions{})
-				if err != nil {
-					return err
+				if !plan.RestartLimitExceeded(pod) {
+					err = c.kubeClient.CoreV1().Pods(populatorNamespace).Delete(ctx, pod.Name, metav1.DeleteOptions{})
+					if err != nil {
+						return err
+					}
```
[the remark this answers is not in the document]
liranr23: https://github.com/kubev2v/forklift/pull/350

pkg/controller/plan/migration.go
liranr23: thanks!

pkg/controller/plan/migration.go
```diff
@@ -341,6 +340,29 @@ func (r *Migration) Archive() {
...
+	for _, vm := range r.Plan.Status.Migration.VMs {
+		pvcs, err := r.kubevirt.getPVCs(vm)
+		if err != nil {
+			r.Log.Error(err,
+				"Couldn't get VM's PVCs.",
+				"vm",
+				vm.String())
+		}
+		if r.kubevirt.isOpenstack(vm) {
+			err = r.setOpenStackPopulatorLabels(vm, pvcs)
```
bennyz, line 359: do we want to log `err` here and below?

pkg/controller/plan/migration.go
```diff
@@ -1591,6 +1623,140 @@ func (r *Migration) updateCopyProgressForOpenstack(vm *plan.VMStatus, step *plan
...
+		return
+	}
+	for _, pod := range podList.Items {
+		if strings.HasPrefix(pod.Name, "populate-") {
+			// it's populator pod
+			if _, ok := pod.Labels["migration"]; !ok {
+				// un-labeled pod, we need to set it
+				args := pod.Spec.Containers[0].Args
+				for _, val := range args {
+					if strings.HasPrefix(val, "--disk-id=") {
```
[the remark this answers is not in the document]
bennyz: should be populate-< PVC  UID >, apparently stuff in <> are not shown...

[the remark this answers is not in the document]
liranr23: discussed offline - we can't depend on having the PVC available (after the migration the user can delete the VM, including the PVC). we can't also be only depend on the inventory (user can change the source). that's why i used the disk id which is available in both options and possible to do when using the populator args.

pkg/controller/plan/kubevirt.go
```diff
@@ -869,7 +832,7 @@ func (r *KubeVirt) GetGuestConversionPod(vm *plan.VMStatus) (pod *core.Pod, err
 
 // Delete the PVC consumer pod on the destination cluster.
 func (r *KubeVirt) DeletePVCConsumerPod(vm *plan.VMStatus) (err error) {
-	list, err := r.GetPodsWithLabels(r.consumerLabels(vm.Ref))
+	list, err := r.GetPodsWithLabels(r.consumerLabels(vm.Ref, true))
```
ahadas, line 835: it looks like we add all the labels, including the migration label, to the consume pod and we always drop the migration label when querying the consumer pods - why do we drop the migration label when querying?

pkg/controller/plan/kubevirt.go
```diff
@@ -945,8 +907,7 @@ func (r *KubeVirt) DeleteObject(object client.Object, vm *plan.VMStatus, message
 
 // Delete any hook jobs that belong to a VM migration.
 func (r *KubeVirt) DeleteHookJobs(vm *plan.VMStatus) (err error) {
-	vmLabels := r.vmLabels(vm.Ref)
-	delete(vmLabels, kMigration)
```
ahadas, line 949: actually, getting back to this PR after a while, I wanted to suggest to do something `filter(vmLabels(..))` which is basically equivalent to how it was before - isn't it more readable? maybe adding another function like `vmAllButMigrationLabels`?

ahadas (commented):
partial review

pkg/lib-volume-populator/populator-machinery/controller.go
```diff
@@ -761,17 +773,14 @@ func (c *controller) syncPvc(ctx context.Context, key, pvcNamespace, pvcName str
...
+		err = c.kubeClient.CoreV1().PersistentVolumeClaims(populatorNamespace).Delete(ctx, pvcPrime.Name, metav1.DeleteOptions{})
 		if err != nil {
 			return err
 		}
-	}
 
-	// If PVC' still exists, delete it
-	if pvcPrime != nil {
-		err = c.kubeClient.CoreV1().PersistentVolumeClaims(populatorNamespace).Delete(ctx, pvcPrime.Name, metav1.DeleteOptions{})
+		err = c.ensureFinalizer(ctx, pvcPrime, pvcKuberenetesFinalizer, false)
```
ahadas, line 783: @bennyz @liranr23 did we reach a consensus on whether we want/need to remove the prime pvc?

ahadas (commented):
there are quite a few good things in this PR:
1. setting the 'migration' label (also for further use, not just for cleanup)
2. cleaning up using the 'migration' label
3. limiting the populator pods to 3 attempts

but there are some aspects that I'm not completely sure about:
1. adding that boolean flag for filtering the migration label - it makes the code more clumsy and less explicit in my opinion
2. the removal of the finalizer from the prime pvc
3. and mainly, the code we introduce to add the labels to resources in existing migrations (those that were created before this change)
commented inside about the last three points

pkg/lib-volume-populator/populator-machinery/controller.go
liranr23: i think we did. both options are fine, since the status of the pvc prime is `Lost` and there is no intention to use it. i like the idea of deleting it without leaving it in terminating state. but if there is strong opinion to have it, that's fine as well.

ahadas: yes, that was the main concern - if we can identify when the PV switched back
but there was a case in which having the prime pvc can help us identifying the correlation between the populaor pod and the migration, @bennyz ?

pkg/controller/plan/kubevirt.go
```diff
@@ -969,14 +930,109 @@ func (r *KubeVirt) DeleteHookJobs(vm *plan.VMStatus) (err error) {
 	return
 }
 
+// Delete any populator resources that belong to a VM migration.
+func (r *KubeVirt) DeletePopulatorDataSource(vm *plan.VMStatus) (err error) {
+	if r.isOpenstack(vm) {
```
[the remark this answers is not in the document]
ahadas: posted #359

pkg/lib-volume-populator/populator-machinery/controller.go
bennyz: yes, I think we can put the prime PVCs to good use and find all the relevant pods, but I don't feel to strongly about this and I don't think we need to delay this PR for this optimization

bennyz: fair point

ahadas: yeah, I think we should to push this forward

pkg/controller/plan/kubevirt.go
```diff
@@ -969,14 +930,109 @@ func (r *KubeVirt) DeleteHookJobs(vm *plan.VMStatus) (err error) {
 	return
 }
 
+// Delete any populator resources that belong to a VM migration.
+func (r *KubeVirt) DeletePopulatorDataSource(vm *plan.VMStatus) (err error) {
```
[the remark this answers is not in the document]
liranr23: done

pkg/controller/plan/migration.go
```diff
@@ -415,6 +443,14 @@ func (r *Migration) CleanUp(vm *plan.VMStatus) (err error) {
 	if err != nil {
 		return
 	}
+	err = r.kubevirt.DeletePopulatorDataSource(vm)
+	if err != nil {
+		return
+	}
+	err = r.kubevirt.DeletePopulatorPods(vm)
```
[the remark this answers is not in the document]
ahadas: it doesn't matter that this code is called from three different places if we always check that the migration succeeded here before removing the importer pods.. I mentioned this because it could make sense to follow the same pattern for the populator pods

pkg/controller/plan/kubevirt.go
```diff
@@ -1821,20 +1879,6 @@ func (r *KubeVirt) ensureOpenStackVolumes(vm ref.Ref, ready bool) (pvcNames []st
...
-				err = liberr.Wrap(err)
-				return
-			}
-
-			err = r.Client.Update(context.TODO(), populatorCr, &client.UpdateOptions{})
-			if err != nil {
-				err = liberr.Wrap(err)
-				return
-			}
-
```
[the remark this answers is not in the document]
ahadas: ack, so we can keep it like that for such migrations

pkg/controller/plan/kubevirt.go
```diff
@@ -1885,3 +1929,33 @@ func (r *KubeVirt) openstackPVCsReady(vm ref.Ref, step *plan.Step) (ready bool,
...
+	populatorCr.Labels["migration"] = migrationId
+	err = r.Client.Update(context.TODO(), &populatorCr)
+	return
+}
+
+func (r *KubeVirt) setPopulatorPodLabels(pod core.Pod, pvcName, migrationId string) (err error) {
+	if pod.Labels == nil {
+		pod.Labels = make(map[string]string)
+	}
+	pod.Labels["pvcName"] = pvcName
```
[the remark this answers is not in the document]
liranr23: we still set it in the popoulator controller, but if it's older version we won't have it. do you wish to delete it also from the populator controller?

pkg/lib-volume-populator/populator-machinery/controller.go
```diff
@@ -586,13 +588,21 @@ func (c *controller) syncPvc(ctx context.Context, key, pvcNamespace, pvcName str
 				// Join the transfer network namespace and name
 				annotations[AnnDefaultNetwork] = fmt.Sprintf("%s/%s", transferNetwork["namespace"], transferNetwork["name"])
 			}
-
+			migration, found, err := unstructured.NestedString(crInstance.Object, "metadata", "labels", "migration")
+			if err != nil {
+				return err
+			}
+			labels := map[string]string{"pvcName": pvc.Name}
```
liranr23, line 595: @bennyz we can leave it or drop it. we don't set it once the pod exists and don't have it. i am in favor of leaving it.

pkg/controller/plan/adapter/openstack/destinationclient.go
```diff
@@ -0,0 +1,92 @@
...
+
+// Set the OpenstackVolumePopulator CustomResource Ownership.
+func (r *DestinationClient) SetPopulatorCrOwnership() (err error) {
+	populatorCrList, err := r.getPopulatorCrList()
+	if err != nil {
+		return
+	}
+	for _, populatorCr := range populatorCrList.Items {
+		pvc := core.PersistentVolumeClaim{}
+		err = r.Destination.Client.Get(context.TODO(), client.ObjectKey{Namespace: r.Plan.Spec.TargetNamespace, Name: populatorCr.Spec.ImageID}, &pvc)
```
liranr23, line 45: @bennyz please confirm the image id is the name of the created pvc.

ahadas:
100 comments!

pkg/controller/plan/adapter/openstack/destinationclient.go
```diff
@@ -0,0 +1,92 @@
...
+		if err != nil {
+			continue
+		}
+		err = k8sutil.SetOwnerReference(&pvc, &populatorCr, r.Scheme())
+		if err != nil {
+			continue
+		}
+		err = r.Destination.Client.Update(context.TODO(), &populatorCr)
```
ahadas, line 53: shouldn't we patch here as well?

pkg/controller/plan/adapter/ovirt/destinationclient.go
```diff
@@ -0,0 +1,88 @@
...
+		pvc := core.PersistentVolumeClaim{}
+		err = r.Destination.Client.Get(context.TODO(), client.ObjectKey{Namespace: r.Plan.Spec.TargetNamespace, Name: populatorCr.Spec.DiskID}, &pvc)
+		if err != nil {
+			continue
+		}
+		err = k8sutil.SetOwnerReference(&pvc, &populatorCr, r.Scheme())
+		if err != nil {
+			continue
+		}
+		err = r.Destination.Client.Update(context.TODO(), &populatorCr)
```
ahadas, line 50: if we change it above then also here

pkg/controller/plan/controller.go
```diff
@@ -262,6 +269,25 @@ func (r Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (r
 	return
 }
 
+func (r *Reconciler) setPopulatorDataSourceLabels(plan *api.Plan) {
+	if plan.Status.HasAnyCondition(Succeeded, Failed, Canceled) {
```
ahadas, line 273: why do we check this?

pkg/controller/plan/kubevirt.go
```diff
@@ -974,6 +936,68 @@ func (r *KubeVirt) DeleteHookJobs(vm *plan.VMStatus) (err error) {
...
+	}
+	for _, pod := range pods {
+		pvcId := strings.Split(pod.Name, "populate-")[1]
+		for _, pvc := range pvcs {
+			if string(pvc.UID) == pvcId {
+				err = k8sutil.SetOwnerReference(&pvc, &pod, r.Scheme())
+				if err != nil {
+					continue
+				}
+				err = r.Destination.Client.Update(context.TODO(), &pod)
```
ahadas, line 971: same question about patch vs update
and we can 'break' here

pkg/controller/plan/kubevirt.go
```diff
@@ -1893,3 +1917,32 @@ func (r *KubeVirt) openstackPVCsReady(vm ref.Ref, step *plan.Step) (ready bool,
...
 	return
 }
+
+func (r *KubeVirt) setOpenStackPopulatorLabels(populatorCr v1beta1.OpenstackVolumePopulator, vmId, migrationId string) (err error) {
+	if populatorCr.Labels == nil {
+		populatorCr.Labels = make(map[string]string)
+	}
+	populatorCr.Labels["vmID"] = vmId
+	populatorCr.Labels["migration"] = migrationId
+	err = r.Destination.Client.Update(context.TODO(), &populatorCr)
```
ahadas, line 1927: same question about patch vs update here and below

pkg/controller/plan/migration.go
```diff
@@ -1596,6 +1645,128 @@ func (r *Migration) updateCopyProgressForOpenstack(vm *plan.VMStatus, step *plan
...
+		lookupName := fmt.Sprintf("%s-%s", r.Migration.Name, vol.ID)
+		image, err := r.getOpenstackImageByName(lookupName)
+		if err != nil {
+			continue
+		}
+		images = append(images, image)
+	}
+	if len(images) != len(pvcs) {
+		// To be sure we have every disk based on what already migrated and what's not.
+		// e.g when initialization the plan and we have populator CR without PVC or when the source changed.
```
ahadas, line 1666: what does it mean "or when the source changed"?

pkg/lib-volume-populator/populator-machinery/controller.go
```diff
@@ -66,6 +67,7 @@ const (
 	populatorPvcPrefix      = "prime"
 	populatedFromAnnoSuffix = "populated-from"
 	pvcFinalizerSuffix      = "populate-target-protection"
+	pvcKuberenetesFinalizer = "kubernetes.io/pvc-protection"
```
ahadas, line 70: can be removed?

ahadas (commented):
looks much better now!

ahadas (commented):
looks much better now!

pkg/controller/plan/kubevirt.go
liranr23: i think it's best effort and if there are any more PVCs to update why not to try?

pkg/controller/plan/adapter/openstack/destinationclient.go
ahadas: even if we don't watch the resource, why overriding the entire configuration if we update a single field?
the classic example of this is when the definition you have at hand doesn't include all properties and then those that are set on the entity which you haven't set, will be overridden

pkg/controller/plan/migration.go
ahadas: ok so how about making it more clear:
```suggestion
		// e.g when initializing the plan and the PVC has not been created yet (but the populator CR is) or when the disks that are attached to the source VM change.
```

pkg/controller/plan/controller.go
liranr23: talked offline, the missing bit here is for new plans - probably to use the mutator to annotate it when a new plan is created. for existing plans it's ok, but it would be better to figure if the plan is new and didn't run to annotate it immediately.

ahadas: filed https://github.com/kubev2v/forklift/issues/366

ahadas (approved):
lgtm, [the gap we discussed internally](https://github.com/kubev2v/forklift/issues/366) can be fixed separately when with the mutator for plans

bennyz (approved):

bennyz (approved):

PR #225, opened 2023-03-14
Set security context to v2v

### LINKED ISSUE #173: Conversion pod fails in restricted namespaces
When migrating to a namespace that is labeled with `pod-security.kubernetes.io/enforce=restricted`, the migration fails because the conversion pod cannot start there with the following error:
`pods "..." is forbidden: violates PodSecurity "restricted:latest": allowPrivilegeEscalation != false (containers "vddk-side-car", "virt-v2v" must set securityContext.allowPrivilegeEscalation=false), unrestricted capabilities (containers "vddk-side-car", "virt-v2v" must set securityContext.capabilities.drop=["ALL"]), runAsNonRoot != true (pod or containers "vddk-side-car", "virt-v2v" must set securityContext.runAsNonRoot=true), seccompProfile (pod or containers "vddk-side-car", "virt-v2v" must set securityContext.seccompProfile.type to "RuntimeDefault" or "Localhost")`
I'm going to drop that security label from namespaces that are created by our testing framework and this change would need to be reverted when fixing this issue

liranr23:
The v2v container needs security context in order to run in restricted namespaces.

### DIFF
--- a/pkg/controller/plan/kubevirt.go
+++ b/pkg/controller/plan/kubevirt.go
@@ -78,6 +78,14 @@ const (
 	kApp = "forklift.app"
 )
 
+// User
+const (
+	// Qemu user
+	qemuUser = int64(107)
+	// Qemu group
+	qemuGroup = int64(107)
+)
+
 // Map of VirtualMachines keyed by vmID.
 type VirtualMachineMap map[string]VirtualMachine
 
@@ -641,6 +649,8 @@ func (r *KubeVirt) createPodToBindPVCs(vm *plan.VMStatus, pvcNames []string) err
 			},
 		})
 	}
+	nonRoot := true
+	allowPrivilageEscalation := false
 	pod := &core.Pod{
 		ObjectMeta: meta.ObjectMeta{
 			Namespace:    r.Plan.Spec.TargetNamespace,
@@ -654,6 +664,16 @@ func (r *KubeVirt) createPodToBindPVCs(vm *plan.VMStatus, pvcNames []string) err
 					Name:    name,
 					Image:   Settings.Migration.VirtV2vImageCold,
 					Command: []string{"/bin/sh"},
+					SecurityContext: &core.SecurityContext{
+						AllowPrivilegeEscalation: &allowPrivilageEscalation,
+						RunAsNonRoot:             &nonRoot,
+						Capabilities: &core.Capabilities{
+							Drop: []core.Capability{"ALL"},
+						},
+						SeccompProfile: &core.SeccompProfile{
+							Type: core.SeccompProfileTypeRuntimeDefault,
+						},
+					},
 				},
 			},
 			Volumes: volumes,
@@ -1202,7 +1222,8 @@ func (r *KubeVirt) guestConversionPod(vm *plan.VMStatus, vmVolumes []cnv.Volume,
 	}
 
 	// qemu group
-	qemuUser := int64(107)
+	fsGroup := qemuGroup
+	user := qemuUser
 	nonRoot := true
 	allowPrivilageEscalation := false
 	// virt-v2v image
@@ -1226,8 +1247,8 @@ func (r *KubeVirt) guestConversionPod(vm *plan.VMStatus, vmVolumes []cnv.Volume,
 		},
 		Spec: core.PodSpec{
 			SecurityContext: &core.PodSecurityContext{
-				FSGroup:      &qemuUser,
-				RunAsUser:    &qemuUser,
+				FSGroup:      &fsGroup,
+				RunAsUser:    &user,
 				RunAsNonRoot: &nonRoot,
 			},
 			RestartPolicy: core.RestartPolicyNever,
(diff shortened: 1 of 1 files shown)


pkg/controller/plan/kubevirt.go
```diff
@@ -1265,6 +1276,14 @@ func (r *KubeVirt) guestConversionPod(vm *plan.VMStatus, vmVolumes []cnv.Volume,
 							Protocol:      core.ProtocolTCP,
 						},
 					},
+					SecurityContext: &core.SecurityContext{
+						AllowPrivilegeEscalation: &allowPrivilageEscalation,
+						RunAsNonRoot:             &nonRoot,
+						RunAsUser:                &fsGroup,
```
nyoxi, line 1282: Why is this needed and why to set UID from GID?

pkg/controller/plan/kubevirt.go
```diff
@@ -1265,6 +1276,14 @@ func (r *KubeVirt) guestConversionPod(vm *plan.VMStatus, vmVolumes []cnv.Volume,
 							Protocol:      core.ProtocolTCP,
 						},
 					},
+					SecurityContext: &core.SecurityContext{
+						AllowPrivilegeEscalation: &allowPrivilageEscalation,
+						RunAsNonRoot:             &nonRoot,
```
nyoxi, line 1281: This can be moved to `Pod.SecurityContext` to avoid duplication.

nyoxi (changes requested):
You are not setting and `SeccompProfile` (can be set in `Pod.SecurityContext`) and you should also define the context for the init container with VDDK.

pkg/controller/plan/kubevirt.go
liranr23: qemu is 107:107, so it's the same user and group

liranr23:
> You are not setting `SeccompProfile` (can be set in `Pod.SecurityContext`) and you should also define the context for the init container with VDDK.

Please refer to https://github.com/kubev2v/forklift/pull/211

pkg/controller/plan/kubevirt.go
nyoxi: Sure but then we should be explicit about it and define `qemuUser` constant.

pkg/controller/plan/kubevirt.go
nyoxi: It worked for me, please re-check. The documentation states that:

>  May also be set in SecurityContext.  If set in both SecurityContext and PodSecurityContext, the value specified in SecurityContext takes precedence.

pkg/controller/plan/kubevirt.go
liranr23: i did, just the naming within the function is wrong, i'll change.

nyoxi:
> > You are not setting `SeccompProfile` (can be set in `Pod.SecurityContext`) and you should also define the context for the init container with VDDK.
> 
> Please refer to #211 for the `SeccompProfile`.

I wonder why I didn't have this issue.

> are you sure it's needed for the init container? it does pass the vsphere ci on restricted namespace.

Yes. If you look at the error provided in #173 it mentions both containers: `pod or containers "vddk-side-car", "virt-v2v" must set...`

pkg/controller/plan/kubevirt.go
nyoxi: Yes, I am not saying we don't. But we already have `PodSecurityContext` with some configuration so it makes sense to set the fields that can be set there instead of duplicating that per container.

pkg/controller/plan/kubevirt.go
```diff
@@ -1224,7 +1244,7 @@ func (r *KubeVirt) guestConversionPod(vm *plan.VMStatus, vmVolumes []cnv.Volume,
 		},
 		Spec: core.PodSpec{
 			SecurityContext: &core.PodSecurityContext{
-				FSGroup: &fsGroup,
+				FSGroup: &qemuUser,
 			},
```
nyoxi, line 1248: You added `SeccompProfile` to the PVC binding Pod but forgot to add it to the conversion Pod.

pkg/controller/plan/kubevirt.go
```diff
@@ -1202,7 +1220,9 @@ func (r *KubeVirt) guestConversionPod(vm *plan.VMStatus, vmVolumes []cnv.Volume,
 	}
 
 	// qemu group
-	fsGroup := int64(107)
+	qemuUser := int64(qemuGroup)
```
nyoxi, line 1223: That did not came out the way I have imagined it. :) Could we have two constants `qemuUser` and `qemuGroup` to avoid this confusing assignments? That way it will be clear that those have same value on purpose (not by coincidence) and it's not a bug or a some hack. Here you would have `fsGroup := qemuGroup` then.

nyoxi (changes requested):

pkg/controller/plan/kubevirt.go
liranr23: unfortunately i must assign it. it's problematic to use const as pointer.
and yes, i will do 2 consts with 107 and just assign here..

pkg/controller/plan/kubevirt.go
liranr23: i really don't like the mixup of `SeccompProfile` when setting user ID. let's not add it.

ahadas (approved):

PR #230, opened 2023-03-15
ovirt: only add 10% to PVC if volumeMode == Filesystem

bennyz:


pkg/controller/plan/adapter/ovirt/builder.go
```diff
@@ -525,7 +525,12 @@ func (r *Builder) PersistentVolumeClaimWithSourceRef(da interface{}, storageName
 	diskAttachment := da.(model.XDiskAttachment)
 
 	// We add 10% overhead because of the fsOverhead in CDI, around 5% to ext4 and 5% for root partition.
-	diskSize := int64(float64(diskAttachment.Disk.ProvisionedSize) * 1.1)
+	diskSize := diskAttachment.Disk.ProvisionedSize
+	// Accounting for fsOverhead is only required for `volumeMode: Filesystem`
+	if *volumeMode == core.PersistentVolumeFilesystem {
```
ahadas, line 530: can volumeMode be nil here?

liranr23: based on `getStorageProfileModes` which we get the volumeMode from if it's nil there, we set it with filesystem.

liranr23 (approved):

pkg/controller/plan/adapter/ovirt/builder.go
ahadas: @liranr23 thanks for checking

pkg/controller/plan/adapter/ovirt/builder.go
```diff
@@ -525,7 +525,12 @@ func (r *Builder) PersistentVolumeClaimWithSourceRef(da interface{}, storageName
 	diskAttachment := da.(model.XDiskAttachment)
 
 	// We add 10% overhead because of the fsOverhead in CDI, around 5% to ext4 and 5% for root partition.
-	diskSize := int64(float64(diskAttachment.Disk.ProvisionedSize) * 1.1)
+	diskSize := diskAttachment.Disk.ProvisionedSize
+	// Accounting for fsOverhead is only required for `volumeMode: Filesystem`
```
ahadas, line 529: that's not entirely accurate - only when the volumeMode is of type Filesystem and the PV is placed on block storage, right?

ahadas: (code-wise it's ok, but this comment is a bit confusing in that case)

ahadas: right, that's why I'm not asking for code changes but to make the comment more clear that it's not always required for volumeMode == Filesystem - just that it might be required only in this case

bennyz: ack

bennyz: let me know if it's better now

ahadas: yep, perfect - thanks

ahadas (approved):

PR #238, opened 2023-03-16
openstack vm configuration map

mmartinv:
- Retrieve Flavor ExtraSpecs
- Use Neutron API instead of Compute for networks
- Add Subnet type
- Move vm status definition to the model
- Make use of coherent field names and omit some fields when empty
- Check error when retrieving the project
- Workload enhancements
- Add openstack plan validations
- Remove duplicated import
- Map openstack vm configuration

### DIFF
--- a/pkg/controller/plan/adapter/openstack/builder.go
+++ b/pkg/controller/plan/adapter/openstack/builder.go
@@ -1046,16 +1046,22 @@ func (r *Builder) cleanup(c planbase.Client, imageName string) {
 	volume := &model.Volume{}
 	err := r.Source.Inventory.Find(volume, ref.Ref{Name: imageName})
 	if err != nil {
-		r.Log.Error(err, "Couldn't find volume", "name", imageName)
+		r.Log.Info("couldn't find volume for deletion, skipping...", "name", imageName)
+	} else {
+		err = volumes.Delete(osClient.BlockStorageService, volume.ID, volumes.DeleteOpts{Cascade: true}).ExtractErr()
+		if err != nil {
+			r.Log.Error(err, "error removing volume", "name", imageName)
+		}
 	}
-
-	volumes.Delete(osClient.BlockStorageService, volume.ID, volumes.DeleteOpts{Cascade: true})
 
 	snapshot := &model.Snapshot{}
 	err = r.Source.Inventory.Find(snapshot, ref.Ref{Name: imageName})
 	if err != nil {
-		r.Log.Error(err, "Couldn't find snapshot", "name", imageName)
+		r.Log.Info("couldn't find snapshot for deletion, skipping...", "name", imageName)
+	} else {
+		err = snapshots.Delete(osClient.BlockStorageService, snapshot.ID).ExtractErr()
+		if err != nil {
+			r.Log.Error(err, "error removing snapshot", "name", imageName)
+		}
 	}
-
-	snapshots.Delete(osClient.BlockStorageService, snapshot.ID)
 }


pkg/controller/plan/adapter/openstack/builder.go
```diff
@@ -33,6 +34,153 @@ const (
...
+	CPUThreadPolicyRequire = "require"
+)
+
+// Bus types
+const (
+	VirtioSCSI = "virtio_scsi"
+	VirtIO     = "virtio"
+	SATA       = "sata"
+	SCSI       = "scsi"
+	IDE        = "ide"
```
ahadas, line 81: interesting, if that is supported in openstack (which makes sense) then we need to introduce the same logic we've added for imports from ovirt that convert such vms to q35

pkg/controller/plan/adapter/openstack/builder.go
```diff
@@ -33,6 +34,153 @@ const (
...
+)
+
+// Default CPU parameters
+const (
+	DefaultCPUSockets = "1"
+	DefaultCPUCores   = "1"
+	DefaultCPUThreads = "1"
+)
+
+// Default properties
```
ahadas, line 166: default for the resources we get from openstack, right?

pkg/controller/plan/adapter/openstack/builder.go
```diff
@@ -61,17 +210,203 @@ func (r *Builder) VirtualMachine(vmRef ref.Ref, object *cnv.VirtualMachineSpec,
...
+	videoModel := DefaultProperties[VideoModel]
+	if imageVideoModel, ok := vm.Image.Properties[VideoModel]; ok {
+		videoModel = imageVideoModel.(string)
+	}
+	switch videoModel {
+	case None:
+	default:
+		autoAttachGraphicsDevice := false
+		object.Template.Spec.Domain.Devices.AutoattachGraphicsDevice = &autoAttachGraphicsDevice
+	}
```
ahadas, line 257: don't we miss setting the video model?

pkg/controller/plan/adapter/openstack/builder.go
```diff
@@ -61,17 +210,203 @@ func (r *Builder) VirtualMachine(vmRef ref.Ref, object *cnv.VirtualMachineSpec,
...
+	} else {
+		for _, volume := range vm.Volumes {
+			if volume.Bootable == "true" {
+				if volumeMachineType, ok := volume.VolumeImageMetadata[MachineType]; ok {
+					machineType = volumeMachineType
+				}
+			}
+		}
+	}
+	object.Template.Spec.Domain.Machine = &cnv.Machine{Type: machineType}
```
ahadas, line 285: note that the CI fails on:
spec.template.spec.domain.machine.type is not supported: pc (allowed values: [q35* pc-q35*])]

pkg/controller/plan/adapter/openstack/builder.go
```diff
@@ -61,17 +210,203 @@ func (r *Builder) VirtualMachine(vmRef ref.Ref, object *cnv.VirtualMachineSpec,
...
+	resourceRequests := map[core.ResourceName]resource.Quantity{}
+
+	if cpuCores != 1 || cpuThreads != 1 {
+		object.Template.Spec.Domain.CPU.Sockets = uint32(cpuSockets)
+		object.Template.Spec.Domain.CPU.Cores = uint32(cpuCores)
+		object.Template.Spec.Domain.CPU.Threads = uint32(cpuThreads)
+	} else {
+		cpuRequest := resource.NewQuantity(int64(cpuSockets*cpuCores*cpuThreads), resource.BinarySI)
+		resourceRequests[core.ResourceCPU] = *cpuRequest
+	}
```
ahadas, line 349: don't we need to set both the cpu topology in `object.Template.Spec.Domain.CPU` and `core.ResourceCPU` on the pod?

pkg/controller/plan/adapter/openstack/builder.go
```diff
@@ -61,17 +210,203 @@ func (r *Builder) VirtualMachine(vmRef ref.Ref, object *cnv.VirtualMachineSpec,
...
+
+	memoryRequest := resource.NewQuantity(int64(vm.Flavor.RAM*int(resource.Mega)), resource.BinarySI)
+	resourceRequests[core.ResourceMemory] = *memoryRequest
```
ahadas, line 352: note the other part of the failure in CI:
spec.template.spec.domain.resources.requests.memory '3Ki': must be greater than or equal to 1M.

pkg/controller/plan/adapter/openstack/builder.go
```diff
@@ -61,17 +210,203 @@ func (r *Builder) VirtualMachine(vmRef ref.Ref, object *cnv.VirtualMachineSpec,
...
+	}
+	if flavorMaxCPUThreads, ok := vm.Flavor.ExtraSpecs[FlavorMaxCPUThreads]; ok {
+		if threads, err := strconv.Atoi(flavorMaxCPUThreads); err == nil {
+			maxCPUThreads = threads
+		}
+	}
+
+	cpuLimit := resource.NewQuantity(int64(maxCPUSockets*maxCPUCores*maxCPUThreads), resource.BinarySI)
+	object.Template.Spec.Domain.Resources.Limits = map[core.ResourceName]resource.Quantity{
+		core.ResourceCPU: *cpuLimit,
```
ahadas, line 381: I don't think that's the right mapping - cpu-limit in k8s specifies the maximum consumption of cpu shares the pod can get while in openstack it seems like in ovirt - the (max) topology that defines how many cpus can be hot-plugged to the vm

pkg/controller/plan/adapter/openstack/builder.go
```diff
@@ -227,15 +585,15 @@ func (r *Builder) PodEnvironment(_ ref.Ref, _ *core.Secret) (env []core.EnvVar,
...
+	image := da.(*model.Image)
 	apiGroup := "forklift.konveyor.io"
 	virtualSize := image.VirtualSize
 	// virtual_size may not always be available
 	if virtualSize == 0 {
 		virtualSize = image.SizeBytes
 	}
 	if *volumeMode == core.PersistentVolumeFilesystem {
-		virtualSize = int64(float64(virtualSize) * 1.1)
+		virtualSize = int64(float64(virtualSize) * 2)
```
ahadas, line 596: why 2? that's the fsOverhead, no?

pkg/controller/plan/kubevirt.go
```diff
@@ -1767,8 +1767,8 @@ func (r *KubeVirt) createOpenStackVolumes(vm ref.Ref) (pvcNames []string, err er
 
 	storageName := r.Context.Map.Storage.Spec.Map[0].Destination.StorageClass
 
-	if len(openstackVm.AttachedVolumes) > 0 {
-		for _, vol := range openstackVm.AttachedVolumes {
+	if len(openstackVm.Volumes) > 0 {
```
ahadas, line 1770: does it mean that we'll transfer volumes that are not attached to the vm and attach them to the imported vm?

ahadas (commented):
didn't review all of the changes but most of them

pkg/controller/plan/adapter/openstack/builder.go
mmartinv: AFAIK Kubevirt API does not support setting the video model, the only way I [found](https://groups.google.com/g/kubevirt-dev/c/pEEjTIARJZc?pli=1) to do it is by creating a hook sidecar: https://github.com/kubevirt/kubevirt/tree/main/cmd/example-hook-sidecar.

pkg/controller/plan/adapter/openstack/builder.go
mmartinv: Yes, those are the observed defaults for VMs when the image they were created from has no properties configured.

pkg/controller/plan/adapter/openstack/builder.go
mmartinv: I think I misunderstood the meaning of `resource.Mega`. This will be fixed in the next push

pkg/controller/plan/adapter/openstack/builder.go
mmartinv: This change shouldn't be here :).
I got `no space left on device` errors in my env so I tried increasing this (which indeed fixed the issue).

pkg/controller/plan/kubevirt.go
mmartinv: No, the `Workload` field was renamed from `AttachedVolumes` to `Volumes` but it's the same thing.
This is because the `VM` definition already has an `AttachedVolumes` field which only contains the `IDs` of the volumes attached to it and having `Workload.XVM.AttachedVolumes`  and `Workload.AttachedVolumes` (containing all the attached volume details) is confusing:
```go
// Workload
type Workload struct {
	SelfLink string `json:"selfLink"`
	XVM
}

// Expanded: VM.
type XVM struct {
	VM
	Image       Image        `json:"image"`
	Flavor      Flavor       `json:"flavor"`
	Networks    []Network    `json:"networks"`
	Subnets     []Subnet     `json:"subnets"`
	Volumes     []Volume     `json:"volumes,omitempty"`
	VolumeTypes []VolumeType `json:"volumeTypes,omitempty"`
	Snapshots   []Snapshot   `json:"snapshots,omitempty"`
}
```

So  this way `Workload.AttachedVolumes` contains the attached volumes IDs and ` `Workload.Volumes` contain all the details of those attached volumes.

pkg/controller/plan/adapter/openstack/builder.go
ahadas: ack, I see we don't have a 'validation' for the video type when the source provider is of type 'ovirt' so I think that's ok to just override it

pkg/controller/plan/adapter/openstack/builder.go
ahadas: interesting
@bennyz ^

bennyz: yeah we discussed this, I'm still not clear on whether it's an issue specific to the iscsi backend of the ovirt-csi storage class. Additionally, block mode should be preferred for ovirt-csi as it supports it, and the ovirt csi driver uses ext4/xfs, neither of which should require extra 100% for the fsOverhead

pkg/controller/plan/adapter/openstack/builder.go
mmartinv: I have changed this to always use Q35 as it's the only supported machine type.

pkg/controller/plan/adapter/openstack/builder.go
mmartinv: According to this [documentation](https://kubevirt.io/user-guide/virtual_machines/dedicated_cpu_resources/) we should use either `Spec.Domain.CPU` or `Spec.Domain.Resources.[Requests,Limits], but not both at the same time.

pkg/controller/plan/adapter/openstack/builder.go
mmartinv: Fixed

pkg/controller/plan/adapter/openstack/builder.go
mmartinv: I have reverted this change but we need to figure out what's the actual size we need here to make sure the populator does not fail

pkg/controller/plan/adapter/openstack/builder.go
mmartinv: I have removed this part

pkg/controller/plan/adapter/openstack/builder.go
```diff
@@ -174,12 +631,11 @@ func (r *Builder) Tasks(vmRef ref.Ref) (list []*plan.Task, err error) {
 			vmRef.String())
 	}
 
-	for _, va := range vm.AttachedVolumes {
+	for _, va := range vm.Volumes {
 		gb := int64(va.Size)
 		list = append(
 			list,
 			&plan.Task{
-				Name: fmt.Sprintf("%s-%s", r.Migration.Name, va.ID),
```
bennyz, line 182: why?

mmartinv: Mmm, that line was removed by accident I am afraid... 😞

pkg/controller/plan/adapter/openstack/builder.go
ahadas: ack, but then it's better to use Spec.Domain.CPU as it should affect the cpu topology of the guest

pkg/controller/plan/adapter/openstack/builder.go
ahadas: ok, filed https://github.com/kubev2v/forklift/issues/257 for this

pkg/controller/provider/web/openstack/workload.go
```diff
@@ -95,36 +95,175 @@ func (r *Workload) Link(p *api.Provider) {
...
+			if volumeImageID, ok := volume.VolumeImageMetadata["image_id"]; ok {
+				imageID = volumeImageID
+			}
+		}
+
+	}
+
+	r.Volumes = volumes
+	r.VolumeTypes = volumeTypes
+	r.Snapshots = snapshots
```
ahadas, line 225: we don't support importing snapshots, right? in that case, we can add a validation warning that says snapshots won't exist for the migrated vm

pkg/controller/plan/adapter/openstack/builder.go
```diff
@@ -46,32 +234,252 @@ func (r *Builder) VirtualMachine(vmRef ref.Ref, object *cnv.VirtualMachineSpec,
...
+	bus := DefaultProperties[DiskBus]
+	if imageDiskBus, ok := vm.Image.Properties[DiskBus]; ok {
+		bus = imageDiskBus.(string)
+	} else {
+		for _, volume := range vm.Volumes {
+			if volume.Bootable == "true" {
+				if volumeDiskBus, ok := volume.VolumeImageMetadata[DiskBus]; ok {
+					bus = volumeDiskBus
+				}
+			}
```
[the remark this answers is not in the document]
mmartinv: AFAIK the disk bus type is specified in the `Image` properties (`hw_disk_bus` to be precise) and there's no way to have disks attached to the same VM with different bus types.

pkg/controller/plan/adapter/openstack/builder.go
```diff
@@ -203,7 +655,77 @@ func (r *Builder) DataVolumes(vmRef ref.Ref, secret *core.Secret, configMap *cor
...
+		}
+		if distro == CentOS && len(version) >= 1 && (version[:1] == "8" || version[:1] == "9") {
+			os = fmt.Sprintf("%s-stream%s", distro, version)
+		}
+	}
+
+	r.Log.Info("operative system detection", "OS", os)
+
+	labels = make(map[string]string)
+	labels[fmt.Sprintf(TemplateOSLabel, os)] = "true"
```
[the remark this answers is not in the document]
mmartinv: This is done the same way as in oVirt provider: the template labels set in this method are used by the caller to find a matching template. If none is found, no template will be applied to the VM

pkg/controller/plan/adapter/openstack/builder.go
```diff
@@ -46,32 +234,252 @@ func (r *Builder) VirtualMachine(vmRef ref.Ref, object *cnv.VirtualMachineSpec,
...
 	}
 
+	var conflicts []string
+	conflicts, err = r.macConflicts(vm)
+	if err != nil {
+		return
+	}
+	if len(conflicts) > 0 {
+		err = liberr.New(
+			fmt.Sprintf("Source VM has a mac address conflict with one or more destination VMs: %s", conflicts))
```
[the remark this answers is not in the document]
mmartinv: This is taken from the oVirt implementation and this is checked just before creating the openshift VM, If a conflicting MAC address is found the migration fails. Once we implement the validations for openstack we can add an additional check so the migration is not even started in that case.

ahadas:
ok, so two minor comments remain - let's address them and get this one in

pkg/controller/plan/adapter/openstack/builder.go
```diff
@@ -46,32 +243,262 @@ func (r *Builder) VirtualMachine(vmRef ref.Ref, object *cnv.VirtualMachineSpec,
...
+				}
+			}
+		}
+	}
+	// Only q35 machine type is supported in Kubevirt so we need to map
+	// openstack bus types to supported ones
+	switch bus {
+	case IdeBus:
+		bus = SataBus
+	case ScsiBus:
```
ahadas, line 492: so you mean IDE should be mapped to SATA and everything else besides SCSI should be mapped to VIRTIO? if so, how about specifying the SCSI->SCSI explicitly, for readability?

ahadas: or add `if bus != ScsiBus` below

yaacov:
for future ref:
https://github.com/kubev2v/forklift-console-plugin/pull/343#discussion_r1146163928

in the UI we use the VMs `attachedVolumes`, we would like to the the ` volume type ID` as well as the `volume ID`

pkg/controller/provider/web/openstack/workload.go
```diff
@@ -95,36 +95,175 @@ func (r *Workload) Link(p *api.Provider) {
...
+
+	if imageID != "" {
+		image := model.Image{Base: model.Base{
+			ID: imageID,
```
yaacov, line 229:
```suggestion
			ID: imageID,
			TypeID: volumeType,
```
this will be nice for the UI

mmartinv: The volume type is related to volumes, not images. If you are retrieving the workload information use `volumes` instead of `attachedVolumes` and you will get all the information about the volume, not only the `ID`.

pkg/controller/plan/adapter/openstack/builder.go
```diff
@@ -26,13 +30,206 @@ import (
...
+	// MAC addresses already in use on the destination cluster. k=mac, v=vmName
+	macConflictsMap map[string]string
 }
 
+// Template labels
+const (
+	TemplateOSLabel                 = "os.template.kubevirt.io/%s"
+	TemplateWorkloadLabel           = "workload.template.kubevirt.io/%s"
+	TemplateWorkloadServer          = "server"
+	TemplateWorkloadDesktop         = "server"
```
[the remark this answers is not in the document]
mmartinv: done

pkg/controller/provider/web/openstack/workload.go
yaacov: ok, in current version we only have attachedVolumes

https://github.com/kubev2v/forklift-console-plugin/pull/343/files#diff-9e59aeb6e8ebb0803d37926497b9ef77244ffdd8e597c56b2f2695830a7e17feR46

if we also get the full volumes data when getting the vms that will be great

yaacov: to clarify what the UI need this data for:
when showing the user the posible storage maps,
the UI look at the vm, and show the storage maps that have matching storage type for the volume types of the vm attached volumes.

ovirt and vmware give the storage type for attached volumes in the vm api call, for that use case.

for openstack to get this behavior, we need:
get the vm,
get all volumes, find the volume with matching id.
get the volume type of that volume, and add it to the vm information

then we can match the storage mappings that work with this vm.

PR #240, opened 2023-03-17
controller: vsphere: Use parent path for hosts without cluster

nyoxi:
When there is only a stand-alone host without cluster in the VMware environment VSphere creates a special Cluster resource with the same name in the hierarchy. This leads to duplicated last component in the path to the host resource e.g.: `/DC/my-host.example.com/my-host.example.com`. This confuses libvirt and/or VMware API. To get a proper path and avoid the duplicated last component we need to look at the path of the (artificial) cluster resource instead.


ahadas (approved):

ahadas:
backport: #241

PR #242, opened 2023-03-19
Require KVM to the PVC consumer pod

liranr23:
In order to have the PVC bind with a host with KVM, which is required by the conversion pod. The consumer need to run on a node that has KVM enabled as well.


pkg/controller/plan/kubevirt.go
```diff
@@ -648,6 +648,17 @@ func (r *KubeVirt) createPodToBindPVCs(vm *plan.VMStatus, pvcNames []string) err
 			},
 		})
 	}
+	resourceReq := core.ResourceRequirements{}
```
liranr23, line 651: maybe it should be only for v2v flow? i think it is good anyway

bennyz: makes sense to always use it

ahadas: you mean to require kvm and kubevirt.io/schedulable" = "true" also when importing from ovirt/openstack?
if that's the case then I'm not so sure - if we don't need virtualization and we can't tell if the vm would start on that node later on (or even if the vm would start anytime soon) then why limiting us to such nodes?

and regardless, @liranr23 you're missing the part that sets this on the pod spec, right?

pkg/controller/plan/kubevirt.go
```diff
@@ -661,7 +672,9 @@ func (r *KubeVirt) createPodToBindPVCs(vm *plan.VMStatus, pvcNames []string) err
 			RestartPolicy: core.RestartPolicyNever,
 			Containers: []core.Container{
 				{
-					Name:    "main",
+					Name: "main",
+					// For v2v the consumer pod is only when we are using cold migration on el9.
+					// In that case, it might be useful to cache the image now and save time for the conversion pod.
```
ahadas, line 677: 1. how about separating this to another PR that is correlated with the issue about this?
2. please add the missing part to the first sentence ("used"?)
3.  how about
```suggestion
					// In that case, we could benefit from pulling the image of the conversion pod so it will be present on the node.
```

pkg/controller/plan/kubevirt.go
bennyz: why wouldn't we need virtualization? I thought this would allow us to start the VM where the PVC is bound too, no?

ahadas: yeah, theoretically, but we don't require the same amount of memory / cpu and potentially other resources that the vm may require so if the vm is started on the same node later on it could save us from bounding the PVC(s) on that node.. that's different than when virt-v2v is used because the conversion pod is started right after this PVC-consumer pod is starts during the migration

bennyz: ack

pkg/controller/plan/kubevirt.go
liranr23: done - PR #245

pkg/controller/plan/kubevirt.go
liranr23: you are right about the resources. i made it v2v only.

nyoxi (approved):

nyoxi (changes requested):
It would be nice to factor it out into a function so that this and `guestConversionPod()` share the same code.

pkg/controller/plan/kubevirt.go
```diff
@@ -686,6 +688,23 @@ func (r *KubeVirt) createPodToBindPVCs(vm *plan.VMStatus, pvcNames []string) err
 	return nil
 }
 
+// Sets KVM requirement to the pod and containers.
+func (r *KubeVirt) setKvmOnPodSpec(podSpec core.PodSpec) {
```
nyoxi: You need to pass a pointer if you plan to modify the `podSpec` value later.

pkg/controller/plan/kubevirt.go
```diff
@@ -686,6 +688,20 @@ func (r *KubeVirt) createPodToBindPVCs(vm *plan.VMStatus, pvcNames []string) err
...
 }
 
+// Sets KVM requirement to the pod and container.
+func (r *KubeVirt) setKvmOnPodSpec(podSpec *core.PodSpec) {
+	nodeSelector := make(map[string]string)
+	reqLimit := make(map[core.ResourceName]resource.Quantity)
+	if *r.Plan.Provider.Source.Spec.Type == v1beta1.VSphere && !Settings.VirtV2vDontRequestKVM {
+		nodeSelector["kubevirt.io/schedulable"] = "true"
+		// Ensure that the pod is deployed on a node where /dev/kvm is present.
+		reqLimit["devices.kubevirt.io/kvm"] = resource.MustParse("1")
```
nyoxi, line 698: Why don't you modify the `Limits` directly? That way you would add instead of overwrite the requirements.

nyoxi (changes requested):

pkg/controller/plan/kubevirt.go
```diff
@@ -686,6 +688,23 @@ func (r *KubeVirt) createPodToBindPVCs(vm *plan.VMStatus, pvcNames []string) err
...
+// Sets KVM requirement to the pod and containers.
+func (r *KubeVirt) setKvmOnPodSpec(podSpec core.PodSpec) {
+	resourceReq := core.ResourceRequirements{}
+	nodeSelector := make(map[string]string)
+	if *r.Plan.Provider.Source.Spec.Type == v1beta1.VSphere && !Settings.VirtV2vDontRequestKVM {
+		resourceReq.Limits = core.ResourceList{
+			"devices.kubevirt.io/kvm": resource.MustParse("1"),
+		}
+		// Ensure that the pod is deployed on a node where /dev/kvm is present.
+		nodeSelector["kubevirt.io/schedulable"] = "true"
```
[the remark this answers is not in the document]
liranr23: it doesn't stick that way. maybe i can pull it out and edit it, then do it as podSpec.NodeSelector = data.

pkg/controller/plan/kubevirt.go
liranr23: same answer as above

ahadas: btw, maybe that's the reason for setting both `devices.kubevirt.io/kvm` and `kubevirt.io/schedulable`? why is it a "limit" rather than a "request"?

pkg/controller/plan/kubevirt.go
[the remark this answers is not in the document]
liranr23: if it's empty it's going to be nil, so we must initiate it. otherwise yes, let's add it up.

pkg/controller/plan/kubevirt.go
liranr23: ^same as above

pkg/controller/plan/kubevirt.go
```diff
@@ -686,6 +688,22 @@ func (r *KubeVirt) createPodToBindPVCs(vm *plan.VMStatus, pvcNames []string) err
...
 
+// Sets KVM requirement to the pod and container.
+func (r *KubeVirt) setKvmOnPodSpec(podSpec *core.PodSpec) {
+	if *r.Plan.Provider.Source.Spec.Type == v1beta1.VSphere && !Settings.VirtV2vDontRequestKVM {
+		if podSpec.NodeSelector == nil {
+			podSpec.NodeSelector = make(map[string]string)
+		}
+		podSpec.NodeSelector["kubevirt.io/schedulable"] = "true"
+		container := &podSpec.Containers[0]
+		if container.Resources.Limits == nil {
```
ahadas, line 699: while on it, shouldn't it rather be Requests here?

liranr23: changed to requests.

ahadas (approved):

nyoxi (approved):

PR #243, opened 2023-03-19
Fix default setting for insecureSkipVerify

bkhizgiy:
Setting `insecureSkipVerify` to be false in the populator in case its not set in the secret.

### DIFF
--- a/cmd/ovirt-populator/ovirt-populator.go
+++ b/cmd/ovirt-populator/ovirt-populator.go
@@ -3,6 +3,7 @@ package main
 import (
 	"bufio"
 	"encoding/json"
+	"errors"
 	"flag"
 	"fmt"
 	"net/http"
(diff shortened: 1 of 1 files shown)


cmd/ovirt-populator/ovirt-populator.go
```diff
@@ -165,10 +166,21 @@ func loadEngineConfig(engineURL string) engineConfig {
 		klog.Fatal(err.Error())
 	}
 
-	insecureSkipVerify, err := os.ReadFile("/etc/secret-volume/insecureSkipVerify")
+	var insecureSkipVerify []byte
+	_, err = os.Stat("/etc/secret-volume/insecureSkipVerify")
 	if err != nil {
-		klog.Fatal(err.Error())
```
bkhizgiy, line 170: another suggestion, we can also change all of the logic to switch this line with `insecureSkipVerify = []byte("false")`.
In case we have the flow that it does not exist in secret that's ok.
In the case the value is set to true but the operation failed for some unrelated reason, we will be missing the ca-cert and the operation will fail in the transfer phase, the container will be recreated and the actual value will be set in the insecureSkipVerify.

I'm not sure which approach is better so presenting both options...

bennyz: +1

ahadas: > In the case the value is set to true but the operation failed for some unrelated reason, we will be missing the ca-cert and the operation will fail in the transfer phase, the container will be recreated and the actual value will be set in the insecureSkipVerify.

functionality-wise I think it would work but to ease debugging I think it would be better to be more explicit about the issue, no?

bkhizgiy: @ahadas Yes I was thinking the same, but since the pod is deleted anyway the log is not really saved.

liranr23: well, it deletes anyway currently. if we wish to use the must-gather on the pod, he will have to stay. we do plan on getting logs from it, so in the near future - it won't be deleted.

bennyz: > well, it deletes anyway currently. if we wish to use the must-gather on the pod, he will have to stay. we do plan on getting logs from it, so in the near future - it won't be deleted.

we need to think about this, because this will essentially remove the retry feature we get from the populator recreating the pod...

ahadas: what happens in CDI? they also retry but didn't remove their import pods, no?

bennyz: > what happens in CDI? they also retry but didn't remove their import pods, no?

yes, the retries are limited which we should probably do as well

ahadas (approved):

liranr23 (approved):

PR #244, opened 2023-03-19
Validate dynamic storage class

liranr23:
When using the volume populator, it can't migrate to a static storage class. This patch will prevent from starting such plan, indicating the problem to the user.

### DIFF
--- a/operator/roles/forkliftcontroller/templates/api/deployment-forklift-api.yml.j2
+++ b/operator/roles/forkliftcontroller/templates/api/deployment-forklift-api.yml.j2
@@ -19,6 +19,7 @@ spec:
         app: {{ app_name }}
         service: {{ api_service_name }}
     spec:
+      serviceAccountName: forklift-controller
       containers:
         - name: {{ api_container_name }}
           image: {{ api_image_fqin }}
--- a/operator/roles/forkliftcontroller/templates/api/validatingwebhookconfiguration-forklift-api.yml.j2
+++ b/operator/roles/forkliftcontroller/templates/api/validatingwebhookconfiguration-forklift-api.yml.j2
@@ -39,3 +39,27 @@ webhooks:
     - secrets
     scope: Namespaced
   sideEffects: None
+- admissionReviewVersions:
+  - v1
+  clientConfig:
+    service:
+      name: {{ api_service_name }}
+      namespace: {{ app_namespace }}
+      path: /plan-validate
+      port: 443
+  failurePolicy: Fail
+  matchPolicy: Equivalent
+  name: plans.forklift.konveyor
+  namespaceSelector: {}
+  objectSelector: {}
+  rules:
+  - apiGroups:
+    - forklift.konveyor.io
+    resources:
+    - plans
+    apiVersions:
+    - v1beta1
+    operations:
+    - CREATE
+    - UPDATE
+  sideEffects: None
--- a/pkg/forklift-api/webhooks/validating-webhook/admitters/plan-admitter.go
+++ b/pkg/forklift-api/webhooks/validating-webhook/admitters/plan-admitter.go
@@ -1,0 +1,121 @@
+package admitters
+
+import (
+	"context"
+	v1 "k8s.io/api/storage/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"encoding/json"
+	"fmt"
+	admissionv1 "k8s.io/api/admission/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/rest"
+	"net/http"
+
+	"github.com/konveyor/forklift-controller/pkg/apis"
+	api "github.com/konveyor/forklift-controller/pkg/apis/forklift/v1beta1"
+	"github.com/konveyor/forklift-controller/pkg/forklift-api/webhooks/util"
+)
+
+type PlanAdmitter struct {
+}
+
+func (admitter *PlanAdmitter) Admit(ar *admissionv1.AdmissionReview) *admissionv1.AdmissionResponse {
+	log.Info("Plan admitter was called")
+	raw := ar.Request.Object.Raw
+
+	plan := &api.Plan{}
+	err := json.Unmarshal(raw, plan)
+	if err != nil {
+		return util.ToAdmissionResponseError(err)
+	}
+
+	if plan.Spec.Warm {
+		log.Info("Warm migration supports all storages, passing")
+		return util.ToAdmissionResponseAllow()
+	}
+
+	config, err := rest.InClusterConfig()
+	if err != nil {
+		log.Error(err, "Couldn't get the cluster configuration", err.Error())
+		return util.ToAdmissionResponseError(err)
+	}
+
+	err = api.SchemeBuilder.AddToScheme(scheme.Scheme)
+	if err != nil {
+		log.Error(err, "Couldn't build the scheme", err.Error())
+		return util.ToAdmissionResponseError(err)
+	}
+	err = apis.AddToScheme(scheme.Scheme)
+	if err != nil {
+		log.Error(err, "Couldn't add forklift API to the scheme", err.Error())
+		return util.ToAdmissionResponseError(err)
+	}
+
+	cl, err := client.New(config, client.Options{Scheme: scheme.Scheme})
+	if err != nil {
+		log.Error(err, "Couldn't create a cluster client", err.Error())
+		return util.ToAdmissionResponseError(err)
+	}
+
+	sourceProvider := api.Provider{}
+	err = cl.Get(context.TODO(), client.ObjectKey{Namespace: plan.Spec.Provider.Source.Namespace, Name: plan.Spec.Provider.Source.Name}, &sourceProvider)
+	if err != nil {
+		log.Error(err, "Couldn't get the source provider", err.Error())
+		return util.ToAdmissionResponseError(err)
+	}
+
+	if sourceProvider.Type() == api.VSphere {
+		log.Info("Provider supports all storages, passing")
+		return util.ToAdmissionResponseAllow()
+	}
+
+	destinationProvider := api.Provider{}
+	err = cl.Get(context.TODO(), client.ObjectKey{Namespace: plan.Spec.Provider.Destination.Namespace, Name: plan.Spec.Provider.Destination.Name}, &destinationProvider)
+	if err != nil {
+		log.Error(err, "Couldn't get the destination provider", err.Error())
+		return util.ToAdmissionResponseError(err)
+	}
+
+	if !destinationProvider.IsHost() {
+		log.Info("Migration to a remote provider supports all storages, passing")
+		return util.ToAdmissionResponseAllow()
+	}
+
+	storageClasses := v1.StorageClassList{}
+	err = cl.List(context.TODO(), &storageClasses, &client.ListOptions{})
+	if err != nil {
+		log.Error(err, "Couldn't get the cluster storage classes", err.Error())
+		return util.ToAdmissionResponseError(err)
+	}
+
+	storageMap := api.StorageMap{}
+	err = cl.Get(context.TODO(), client.ObjectKey{Namespace: plan.Spec.Map.Storage.Namespace, Name: plan.Spec.Map.Storage.Name}, &storageMap)
+	if err != nil {
+		log.Error(err, "Couldn't get the storage map", err.Error())
+		return util.ToAdmissionResponseError(err)
+	}
+	storagePairList := storageMap.Spec.Map
+	var badStorageClasses []string
+	for _, storagePair := range storagePairList {
+		scName := storagePair.Destination.StorageClass
+		for _, sc := range storageClasses.Items {
+			if scName == sc.Name && sc.Provisioner == "kubernetes.io/no-provisioner" {
+				badStorageClasses = append(badStorageClasses, sc.Name)
+			}
+		}
+	}
+	if len(badStorageClasses) > 0 {
+		log.Error(fmt.Errorf("storage class(es) '%v' is static, failing", badStorageClasses), "Storage class(es) is static")
+		return &admissionv1.AdmissionResponse{
+			Allowed: false,
+			Result: &metav1.Status{
+				Code:    http.StatusBadRequest,
+				Message: fmt.Sprintf("This plan requires dynamic volume provisioning. Therefore the following destination storage classes cannot be used: %v", badStorageClasses),
+			},
+		}
+	}
+	log.Info("Passed storage validation")
+	return util.ToAdmissionResponseAllow()
+}
--- a/pkg/forklift-api/webhooks/util/util.go
+++ b/pkg/forklift-api/webhooks/util/util.go
@@ -87,3 +87,10 @@ func GeneratePatchPayload(patches ...PatchOperation) ([]byte, error) {
 
 	return payloadBytes, nil
 }
+
+// ToAdmissionResponseAllow returns allowed response
+func ToAdmissionResponseAllow() *admissionv1.AdmissionResponse {
+	return &admissionv1.AdmissionResponse{
+		Allowed: true,
+	}
+}
--- a/pkg/forklift-api/webhooks/validating-webhook.go
+++ b/pkg/forklift-api/webhooks/validating-webhook.go
@@ -10,3 +10,7 @@ import (
 func ServeSecretCreate(resp http.ResponseWriter, req *http.Request) {
 	validating_webhooks.Serve(resp, req, &admitters.SecretAdmitter{})
 }
+
+func ServePlanCreate(resp http.ResponseWriter, req *http.Request) {
+	validating_webhooks.Serve(resp, req, &admitters.PlanAdmitter{})
+}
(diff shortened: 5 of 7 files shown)


pkg/controller/plan/validation.go
```diff
@@ -491,6 +514,9 @@ func (r *Reconciler) validateVM(plan *api.Plan) error {
 	if len(multiplePodNetworkMappings.Items) > 0 {
 		plan.Status.SetCondition(multiplePodNetworkMappings)
 	}
+	if len(staticStorage.Items) > 0 {
+		plan.Status.SetCondition(staticStorage)
```
ahadas, line 518: but that just means that there's at least one pv that was created statically, right?
would checking the 'provisioner' field of the storage class would make sense?

bennyz: I think `provisioner` is a mandatory field so it would be difficult to use it for this

liranr23: it checks per vm, but i didn't verify it yet

liranr23: It does checks it per VM, the plan will be in critical status. On the migration you will be able to see on which VM ID and name you hit the rule.

About the provisioner, we said - no provisioner is ok. but i am not sure it covers all the cases. in such case i think using the current implementation also covers the no provisioner case.

nyoxi (commented):
Could you please describe more in commit message or in comment why the `no-provisioner` is not supported? Explaining that the problem is with the volume populator library.

pkg/controller/plan/adapter/ovirt/validator.go
```diff
@@ -119,3 +120,19 @@ func (r *Validator) MaintenanceMode(_ ref.Ref) (ok bool, err error) {
 	ok = true
 	return
 }
+
+// Validate storage map is dynamic for populator
+func (r *Validator) StorageDynamic(storageClasses v1.StorageClassList) bool {
+	if r.plan.Spec.Warm || !r.plan.Provider.Destination.IsHost() {
+		return true
+	}
```
ahadas, line 128: this part is not so great because it return 'true' also when the storage is not using dynamic provisioning

pkg/controller/plan/adapter/ovirt/validator.go
```diff
@@ -119,3 +120,19 @@ func (r *Validator) MaintenanceMode(_ ref.Ref) (ok bool, err error) {
...
+	for _, storagePair := range r.plan.Referenced.Map.Storage.Spec.Map {
+		scName := storagePair.Destination.StorageClass
+		for _, sc := range storageClasses.Items {
+			if scName == sc.Name && sc.Provisioner == "kubernetes.io/no-provisioner" {
+				return false
+			}
+		}
```
ahadas, line 135: how about extracting this part to reduce the amount of code duplication?

pkg/controller/plan/validation.go
```diff
@@ -491,6 +514,9 @@ func (r *Reconciler) validateVM(plan *api.Plan) error {
...
+	}
```
ahadas, line 519: this means that when a user defines a plan that is going to use a volume-populator and storage map that includes an invalid target namespace, the operation will succeed and only later set a condition on the plan - why not to block this with a validation webhook if we're talking about an invalid input that we can check before persisting the plan?

pkg/forklift-api/webhooks/validating-webhook/admitters/storage-admitter.go
```diff
@@ -0,0 +1,107 @@
...
+	plan := &api.Plan{}
+	err := json.Unmarshal(raw, plan)
+	if err != nil {
+		return &admissionv1.AdmissionResponse{
+			Allowed: false,
+			Result: &metav1.Status{
+				Code:    http.StatusBadRequest,
+				Message: err.Error(),
+			},
+		}
```
bkhizgiy, line 40: You can use `util.ToAdmissionResponseError(err)` instead

pkg/forklift-api/webhooks/validating-webhook/admitters/storage-admitter.go
```diff
@@ -0,0 +1,125 @@
...
+
+	cl, err := client.New(config, client.Options{Scheme: scheme.Scheme})
+	if err != nil {
+		log.Info("Couldn't create a cluster client", "error", err.Error())
+	}
+
+	sourceProvider := api.Provider{}
+	err = cl.Get(context.TODO(), client.ObjectKey{Namespace: plan.Spec.Provider.Source.Namespace, Name: plan.Spec.Provider.Source.Name}, &sourceProvider)
+	if err != nil {
+		log.Info("Couldn't get the source provider", "error", err.Error())
```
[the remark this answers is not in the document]
liranr23: actually, maybe we should allow it when getting to such error. otherwise we will block any creation/edit of a plan if an error occurred in the webhook. i think it's bad to block it in such case (better to have migrations working, as no-provisioner is anyway not commonly used). WDYT?

pkg/forklift-api/webhooks/validating-webhook/admitters/storage-admitter.go
```diff
@@ -0,0 +1,125 @@
...
+func (admitter *StorageAdmitter) Admit(ar *admissionv1.AdmissionReview) *admissionv1.AdmissionResponse {
+	log.Info("storage admitter was called")
+	raw := ar.Request.Object.Raw
+
+	if len(raw) == 0 {
+		// plan is being deleted, passing
+		return &admissionv1.AdmissionResponse{
+			Allowed: true,
+		}
+	}
```
ahadas, line 33: if no validation is required when deleting a plan (which makes sense), we can define the webhook in a way that it's not involved in deletion - see [this](https://github.com/kubev2v/forklift/blob/main/operator/roles/forkliftcontroller/templates/api/validatingwebhookconfiguration-forklift-api.yml.j2#L35-L37)

liranr23: Missed that one! I did saw the new matcher which requires enabling featuregate(alpha) so i went with this workaround. Your suggestion is much better, thanks.

pkg/forklift-api/webhooks/validating-webhook/admitters/storage-admitter.go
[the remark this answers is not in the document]
bkhizgiy: yes it makes sense not to block the plan creation over these errors, so in that case you should return an allowed request but mention in the log and message that we had an error, something like in this block [https://github.com/kubev2v/forklift/blob/cf2ad9146d2c17e5a88668fe73a27563e29fa2ce/pkg/forklift-api/webhooks/validating-webhook/admitters/secret-admitter.go#L72] , but we must have some return statement, otherwise it will cause the API pod to crash (it will try to query info that wasn't retrieved properly) and we don't want it to happen.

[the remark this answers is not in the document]
bkhizgiy: also because you will need to return this in a lot of places of the code you can add a util function for this purpose [here](pkg/forklift-api/webhooks/util/util.go) to make it more readable if you like .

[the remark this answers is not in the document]
bennyz: why would this request fail though? if the plan is going to be successfully executed these requests should work regardless

operator/roles/forkliftcontroller/templates/api/validatingwebhookconfiguration-forklift-api.yml.j2
```diff
@@ -39,3 +39,26 @@ webhooks:
...
+  objectSelector: {}
+  rules:
+  - apiGroups:
+    - forklift.konveyor.io
+    resources:
+    - plans
+    apiVersions:
+    - v1beta1
+    operations:
+    - '*'
```
mnecas, line 63: Do we need to validate the plan even when deleting/connecting?
I would specify the `['CREATE', 'UPDATE']`

pkg/forklift-api/webhooks/validating-webhook/admitters/storage-admitter.go
```diff
@@ -0,0 +1,125 @@
...
+		log.Info("Couldn't build the scheme", "error", err.Error())
+	}
+	err = apis.AddToScheme(scheme.Scheme)
+	if err != nil {
+		log.Info("Couldn't add forklift API to the scheme", "error", err.Error())
+	}
+
+	cl, err := client.New(config, client.Options{Scheme: scheme.Scheme})
+	if err != nil {
+		log.Info("Couldn't create a cluster client", "error", err.Error())
```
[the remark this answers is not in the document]
liranr23: based on below thread, i with with log (like in the block mentioned) and passing it.

operator/roles/forkliftcontroller/templates/api/deployment-forklift-api.yml.j2
```diff
@@ -19,6 +19,7 @@ spec:
         app: {{ app_name }}
         service: {{ api_service_name }}
     spec:
+      serviceAccountName: forklift-controller
```
ahadas, line 22: it would be better to introduce another, more limited, service account for forklift-api but we can do this separately

pkg/forklift-api/webhooks/validating-webhook/admitters/plan-admitter.go
```diff
@@ -0,0 +1,118 @@
...
+
+	if plan.Spec.Warm {
+		log.Info("Warm migration supports all storages, passing")
+		return util.ToAdmissionResponseAllow()
+	}
+
+	config, err := rest.InClusterConfig()
+	if err != nil {
+		log.Info("Couldn't get the cluster configuration", "error", err.Error())
+		return util.ToAdmissionResponseAllow()
```
ahadas, line 42: why allowing the request when we're not able to validate it here and below?

pkg/forklift-api/webhooks/validating-webhook/admitters/plan-admitter.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+
+	sourceProvider := api.Provider{}
+	err = cl.Get(context.TODO(), client.ObjectKey{Namespace: plan.Spec.Provider.Source.Namespace, Name: plan.Spec.Provider.Source.Name}, &sourceProvider)
+	if err != nil {
+		log.Info("Couldn't get the source provider", "error", err.Error())
+		return util.ToAdmissionResponseAllow()
+	}
+
+	// need to query the provider
```
ahadas, line 69: this comment is not placed properly or should be rephrased

pkg/forklift-api/webhooks/validating-webhook/admitters/plan-admitter.go
```diff
@@ -0,0 +1,118 @@
...
+	for _, storagePair := range storagePairList {
+		scName := storagePair.Destination.StorageClass
+		for _, sc := range storageClasses.Items {
+			if scName == sc.Name && sc.Provisioner == "kubernetes.io/no-provisioner" {
+				log.Info(fmt.Sprintf("storage class '%s' is static, failing", scName))
+				return &admissionv1.AdmissionResponse{
+					Allowed: false,
+					Result: &metav1.Status{
+						Code:    http.StatusBadRequest,
+						Message: fmt.Sprintf("Incorrect destination 'StorageClass' value. Static storage class is not supported"),
```
ahadas, line 110: how about adding the name of the storage class? maybe even to include the names of all the problematic storage classes

pkg/forklift-api/webhooks/validating-webhook/admitters/plan-admitter.go
liranr23: i think i wrote about it in another comment. overall the case is not common to have `no-provisioner` SC. I think if it's documented the validation is more as "nice to have". If something went wrong with the pod or whatever, don't you think it's better to it for every SC?

pkg/forklift-api/webhooks/validating-webhook/admitters/plan-admitter.go
liranr23: it's logged in the api pod, but sure we can add it here as well. i think it's better to tell the specific bad used SC other than telling all of them.

pkg/forklift-api/webhooks/validating-webhook/admitters/plan-admitter.go
ahadas: I see it a bit differently - by adding a validation we can ensure that we won't persist invalid configuration. if there's a good argument for persisting it anyway, like we had with providers (the provider may not be accessible when its secret is being defined in MTV), then that's an argument for keep having code that handles such an invalid configuration, but in this case I don't see such an argument - the provider and storage mapping should exist when defining the plan (btw, did you check what happens when defining a plan with a mapping that doesn't exist already?)

liranr23: i checked only non existing mapping (creating within the plan)

pkg/forklift-api/webhooks/validating-webhook/admitters/plan-admitter.go
ahadas: so if I define a storage mapping with two target storage classes that don't support dynamic provisioning, I'll get the error about the first, change/fix it, and then try again to get the error about the second and do the same, right? why not returning a more helpful error here like `This plan requires dynamic volume provisioning. Therefore the following destination storage classes are invalid: ...`

pkg/forklift-api/webhooks/validating-webhook/admitters/plan-admitter.go
ahadas: good, this means that we first create the mappings in this case

pkg/forklift-api/webhooks/validating-webhook/admitters/plan-admitter.go
```diff
@@ -0,0 +1,121 @@
...
+
+	if !destinationProvider.IsHost() {
+		log.Info("Remote migration supports all storages, passing")
+		return util.ToAdmissionResponseAllow()
+	}
+
+	storageClasses := v1.StorageClassList{}
+	err = cl.List(context.TODO(), &storageClasses, &client.ListOptions{})
+	if err != nil {
+		log.Info("Couldn't get the cluster storage classes", "error", err.Error())
```
bennyz, line 89: I think these can be changed to log.Error now, no?

liranr23: i based on https://github.com/kubev2v/forklift/pull/244#discussion_r1178868267

bennyz: but this is rejected now, or am I misunderstanding something?

liranr23: https://github.com/kubev2v/forklift/blob/cf2ad9146d2c17e5a88668fe73a27563e29fa2ce/pkg/forklift-api/webhooks/validating-webhook/admitters/secret-admitter.go#L72 ?

bennyz: yeah, but in the secret admitter the request is not rejected while here it is

liranr23: true :)

bkhizgiy: +1 on changing to `log.Error`.

pkg/forklift-api/webhooks/validating-webhook/admitters/plan-admitter.go
```diff
@@ -0,0 +1,121 @@
...
+
+	destinationProvider := api.Provider{}
+	err = cl.Get(context.TODO(), client.ObjectKey{Namespace: plan.Spec.Provider.Destination.Namespace, Name: plan.Spec.Provider.Destination.Name}, &destinationProvider)
+	if err != nil {
+		log.Error(err, "Couldn't get the destination provider", err.Error())
+		return util.ToAdmissionResponseError(err)
+	}
+
+	if !destinationProvider.IsHost() {
+		log.Info("Remote migration supports all storages, passing")
```
ahadas, line 82:
```suggestion
		log.Info("Migration to a remote provider supports all storages, passing")
```

pkg/forklift-api/webhooks/validating-webhook/admitters/plan-admitter.go
```diff
@@ -0,0 +1,121 @@
...
+	for _, storagePair := range storagePairList {
+		scName := storagePair.Destination.StorageClass
+		for _, sc := range storageClasses.Items {
+			if scName == sc.Name && sc.Provisioner == "kubernetes.io/no-provisioner" {
+				badStorageClasses = append(badStorageClasses, sc.Name)
+			}
+		}
+	}
+	if len(badStorageClasses) > 0 {
+		log.Info(fmt.Sprintf("storage classes '%v' is static, failing", badStorageClasses))
```
ahadas, line 110:
```suggestion
		log.Error(fmt.Sprintf("storage class(es) '%v' is static, failing", badStorageClasses))
```

ahadas (commented):
minor comments about logs, other than that - lgtm

ahadas (approved):

bkhizgiy (approved):
lgtm:)

bennyz (approved):

PR #264, opened 2023-03-26
Switch the KVM resource to limits

### LINKED ISSUE #263: Migration from VMWare fails because of invalid Pod error
Cold migration from VMWare fails at Initialize migration step because of error:
`Pod "mtv-api-tests-23-26-03-21-36-39-82f-plan-vm-19125-pvcinit-fvpbr" is invalid: spec.containers[0].resources.limits: Required value: Limit must be set for non overcommitable resources`

Warm migration from VMWare fails at Convert image step because of error:
`Pod "mtv-api-tests-23-26-03-18-32-07-911-plan-vm-19128-q25m5" is invalid: spec.containers[0].resources.limits: Required value: Limit must be set for non overcommitable resources`

liranr23:
Having the consumer container and the v2v conversion container with the KVM device as required caused:
`spec.containers[0].resources.limits: Required value: Limit must be set for non overcommitable resources`. This patch switch it back to be as Limit of the container as it was before.

Fixes #263

### DIFF
--- a/pkg/controller/plan/kubevirt.go
+++ b/pkg/controller/plan/kubevirt.go
@@ -698,6 +698,10 @@ func (r *KubeVirt) setKvmOnPodSpec(podSpec *core.PodSpec) {
 		}
 		podSpec.NodeSelector["kubevirt.io/schedulable"] = "true"
 		container := &podSpec.Containers[0]
+		if container.Resources.Limits == nil {
+			container.Resources.Limits = make(map[core.ResourceName]resource.Quantity)
+		}
+		container.Resources.Limits["devices.kubevirt.io/kvm"] = resource.MustParse("1")
 		if container.Resources.Requests == nil {
 			container.Resources.Requests = make(map[core.ResourceName]resource.Quantity)
 		}


bennyz:
Can't we use both requirements and limits, similar to: https://kubevirt.io/2018/KVM-Using-Device-Plugins.html
(under Consuming Devices)

liranr23:
> Can't we use both requirements and limits, similar to: https://kubevirt.io/2018/KVM-Using-Device-Plugins.html (under Consuming Devices)

I think we can but what's the benefit?

liranr23:
> > Can't we use both requirements and limits, similar to: https://kubevirt.io/2018/KVM-Using-Device-Plugins.html (under Consuming Devices)
> 
> I think we can but what's the benefit?

Seems there is an agreement to follow the docs. :)

pkg/controller/plan/kubevirt.go
```diff
@@ -698,10 +698,14 @@ func (r *KubeVirt) setKvmOnPodSpec(podSpec *core.PodSpec) {
 		}
 		podSpec.NodeSelector["kubevirt.io/schedulable"] = "true"
 		container := &podSpec.Containers[0]
+		// Ensure that the pod is deployed on a node where /dev/kvm is present.
```
ahadas, line 701: the previous location of this comment was better - it is achieved by setting the 'requests'

ahadas (approved):

ahadas:
@liranr23 please backport to release-v2.4.0

PR #265, opened 2023-03-26
openstack: error handling for snapshots and volumes

bennyz:


pkg/controller/plan/adapter/openstack/builder.go
```diff
@@ -900,6 +900,9 @@ func (r *Builder) BeforeTransferHook(c planbase.Client, vmRef ref.Ref) (ready bo
 		if snapshot.Status != "available" {
 			r.Log.Info("Snapshot not ready yet, recheking...", "snapshot", snap.Name)
 			return false, nil
+		} else if snapshot.Status == "error" {
```
ahadas, line 903: we'll never get here, no? (due to the check of `snapshot.Status != "available"`)

pkg/controller/plan/adapter/openstack/builder.go
```diff
@@ -953,6 +956,9 @@ func (r *Builder) BeforeTransferHook(c planbase.Client, vmRef ref.Ref) (ready bo
 		if volume.Status != "available" && volume.Status != "uploading" {
 			r.Log.Info("Volume not ready yet, recheking...", "volume", vol.Name)
 			return false, nil
+		} else if volume.Status == "error" {
```
ahadas, line 959: same as above

ahadas (approved):

PR #271, opened 2023-03-29
openstack: use openstack time instead of openshift time

bennyz:
Request changes since using the time in the controller may be an issue if there is a time difference between openstack and openshift.

Instead, where possible, this patch attempts to use the time from resource's updated_at field as the value for changes-since.

### DIFF
--- a/pkg/controller/provider/container/openstack/model.go
+++ b/pkg/controller/provider/container/openstack/model.go
@@ -247,7 +247,8 @@ type ImageAdapter struct {
 func (r *ImageAdapter) List(ctx *Context) (itr fb.Iterator, err error) {
 	opts := &ImageListOpts{}
 	imageList := []Image{}
-	now := time.Now()
+	// Set time to epoch start
+	updateTime := time.Unix(0, 0)
 	err = ctx.client.list(&imageList, opts)
 	if err != nil {
 		return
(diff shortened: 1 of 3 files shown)


pkg/controller/provider/container/openstack/model.go
```diff
@@ -259,6 +260,12 @@ func (r *ImageAdapter) List(ctx *Context) (itr fb.Iterator, err error) {
 		}
 		image.ApplyTo(m)
 		list.Append(m)
+		if image.UpdatedAt.After(now) {
+			now = image.UpdatedAt
+		}
+		if image.UpdatedAt.After(now) {
+			now = image.UpdatedAt
+		}
```
ahadas, line 268: duplicate

pkg/controller/provider/container/openstack/model.go
```diff
@@ -247,7 +247,8 @@ type ImageAdapter struct {
 func (r *ImageAdapter) List(ctx *Context) (itr fb.Iterator, err error) {
 	opts := &ImageListOpts{}
 	imageList := []Image{}
-	now := time.Now()
+	// Set time to epoch start
+	now := time.Unix(0, 0)
```
ahadas, line 251: maybe another name? you've proposed "updateTime" :)

ahadas (approved):

ahadas (approved):

Issue #249, opened 2023-03-21
Default insecureSkipVerify is not added to secret when label createdForProviderType value is vsphere or openstack


qiyuann:
When creating a secret with label `createdForProviderType: vsphere` or `createdForProviderType: openstack`, there is no default `insecureSkipVerify` option in the created secret.

bkhizgiy:
Fix for this issue was posted https://github.com/kubev2v/forklift/pull/272

PR #272, opened 2023-03-29
Refactor mutating webhook to support all providers

bkhizgiy:
This change refactors the secret mutator to support all provider types and not only oVirt for the insecureSkipVerify field.

Fixes: #249


et --hard HEAD^
ahadas, line 1: should be dropped

pkg/forklift-api/webhooks/mutating-webhook/mutators/secret-mutator.go
```diff
@@ -38,14 +40,17 @@ func (mutator *OvirtCertMutator) Mutate(ar *admissionv1.AdmissionReview) *admiss
 	}
 
 	var insecure = false
-	if insecureSkipVerify, ok := secret.Data["insecureSkipVerify"]; ok {
-		insecure, err = strconv.ParseBool(string(insecureSkipVerify))
-		if err != nil {
-			log.Error(err, "Failed to parse insecure property from the secret")
-			return util.ToAdmissionResponseError(err)
+	if _, ok := secret.GetLabels()["createdForProviderType"]; ok {
```
ahadas, line 43: this check is redundant - we forward 'create' and 'update' operations to this webhook with objectSelector that checks the `createdForProviderType` exists so we can just replace this with a comment to make it clear that this is the assumption here

pkg/forklift-api/webhooks/mutating-webhook/mutators/secret-mutator.go
```diff
@@ -24,10 +24,12 @@ var log = logging.WithName("mutator")
 
 const NewLine = 0x0a
 
-type OvirtCertMutator struct {
+var secretChanged = false
```
ahadas, line 27: why having a global variable? it can be defined within the `Mutate` function, right?

pkg/forklift-api/webhooks/mutating-webhook/mutators/secret-mutator.go
```diff
@@ -38,14 +40,17 @@ func (mutator *OvirtCertMutator) Mutate(ar *admissionv1.AdmissionReview) *admiss
...
+			}
+		} else {
+			secret.Data["insecureSkipVerify"] = []byte("false")
+			secretChanged = true
 		}
-	} else {
-		secret.Data["insecureSkipVerify"] = []byte("false")
 	}
 
 	if providerType, ok := secret.GetLabels()["createdForProviderType"]; ok && providerType == "ovirt" && !insecure {
```
ahadas, line 56: unrelated to this PR but here we can also assume the `createdForProviderType` label exists

pkg/forklift-api/webhooks/mutating-webhook/mutators/secret-mutator.go
```diff
@@ -86,9 +91,13 @@ func (mutator *OvirtCertMutator) Mutate(ar *admissionv1.AdmissionReview) *admiss
 		if !contains(secret.Data["cacert"], cert) {
 			secret.Data["cacert"] = appendCerts(secret.Data["cacert"], cert)
 			secret.Labels["ca-cert-updated"] = "true"
+			secretChanged = true
 			log.Info("Engine CA certificate was missing, updating the secret")
 		}
+	}
 
+	//In case some of the data in the secret was changed patch the new secret
```
ahadas, line 99:
```suggestion
	// In case the data in the secret changed, patch the secret
```

ahadas (commented):
ah and @bkhizgiy , I've changed the description of the PR (fixed phrasing), please change the commit message as well

pkg/forklift-api/webhooks/mutating-webhook/mutators/secret-mutator.go
bkhizgiy: right, the label still need to be fetch for the ovirt provider check, but the OK check is redundant.

pkg/forklift-api/webhooks/mutating-webhook/mutators/secret-mutator.go
bkhizgiy: I initially tried to separate everything into function to make it more easy to read but it just added more complexity, forgot to return the var back inside.

ahadas (approved):

PR #277, opened 2023-03-30
CDI should copy disks from vSphere to remove clusters

ahadas:
While virt-v2v should copy the data only for cold-migrations to the host cluster.


pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -313,7 +313,12 @@ func (r *Builder) DataVolumes(vmRef ref.Ref, secret *core.Secret, _ *core.Config
 			if disk.Datastore.ID == ds.ID {
 				storageClass := mapped.Destination.StorageClass
 				var dvSource cdi.DataVolumeSource
-				if r.Plan.Spec.Warm {
+				if r.Destination.Provider.IsHost() && !r.Plan.Spec.Warm {
```
nyoxi, line 316: It would be best to use `r.Context.UseEl9VirtV2v()` here.

nyoxi (changes requested):

pkg/controller/plan/adapter/vsphere/builder.go
ahadas: indeed

nyoxi (approved):

PR #280, opened 2023-03-30
Bind PVC with WaitForFirstConsumer storage classes when CDI imports

liranr23:
When executing warm migration from ovirt to a storage class with binding mode `WaitForFirstConsumer` the importer pod won't start until the PVC is bound.

An annotation is added to the DataVolume (passing to the PVC), telling CDI to start immediate without bounding.

### DIFF
--- a/pkg/controller/plan/kubevirt.go
+++ b/pkg/controller/plan/kubevirt.go
@@ -62,6 +62,8 @@ const (
 	AnnOriginalID = "original-ID"
 	// DV deletion on completion
 	AnnDeleteAfterCompletion = "cdi.kubevirt.io/storage.deleteAfterCompletion"
+	// DV immediate bind to WaitForFirstConsumer storage class
+	AnnBindImmediate = "cdi.kubevirt.io/storage.bind.immediate.requested"
 	// Max Length for vm name
 	NameMaxLength = 63
 )
@@ -989,6 +991,9 @@ func (r *KubeVirt) dataVolumes(vm *plan.VMStatus, secret *core.Secret, configMap
 		annotations[AnnDefaultNetwork] = path.Join(
 			r.Plan.Spec.TransferNetwork.Namespace, r.Plan.Spec.TransferNetwork.Name)
 	}
+	if r.Plan.Spec.Warm || !r.Destination.Provider.IsHost() {
+		annotations[AnnBindImmediate] = "true"
+	}
 	// Do not delete the DV when the import completes as we check the DV to get the current
 	// disk transfer status.
 	annotations[AnnDeleteAfterCompletion] = "false"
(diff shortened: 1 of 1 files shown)


pkg/controller/plan/kubevirt.go
```diff
@@ -987,6 +994,9 @@ func (r *KubeVirt) dataVolumes(vm *plan.VMStatus, secret *core.Secret, configMap
 		annotations[AnnDefaultNetwork] = path.Join(
 			r.Plan.Spec.TransferNetwork.Namespace, r.Plan.Spec.TransferNetwork.Name)
 	}
+	if r.Plan.Spec.Warm && r.isOvirt() {
```
liranr23, line 997: possibly just `r.isOvirt()` is enough (only warm plan will enter now since the populator). but it's more self-explanatory

ahadas: so how would you cover cold migration to remote cluster?

ahadas (commented):
@liranr23 please give a better title to this PR `Bind PVC to warm ovirt WaitForFirstConsumer` seems like some code name..

pkg/controller/plan/kubevirt.go
```diff
@@ -797,7 +799,12 @@ func (r *KubeVirt) getStorageProfileModes(storageName string) (accessModes []cor
 
 // Return true when the import is done with OvirtVolumePopulator
 func (r *KubeVirt) useOvirtPopulator(vm *plan.VMStatus) bool {
-	return *r.Plan.Provider.Source.Spec.Type == v1beta1.OVirt && vm.Warm == nil && r.Destination.Provider.IsHost()
+	return r.isOvirt() && vm.Warm == nil && r.Destination.Provider.IsHost()
+}
+
+// Return true when the import is done from oVirt source.
+func (r *KubeVirt) isOvirt() bool {
```
ahadas, line 806: it's a bit awkward to have a method named `isOvirt` on `KubeVirt`.. it can be named `isImportFromOvirt` or something like that but if we just want to make it shorter, how about replacing `*r.Plan.Provider.Source.Spec.Type` with `r.Plan.Provider.Source.Type()` ?

pkg/controller/plan/kubevirt.go
```diff
@@ -987,6 +994,9 @@ func (r *KubeVirt) dataVolumes(vm *plan.VMStatus, secret *core.Secret, configMap
 		annotations[AnnDefaultNetwork] = path.Join(
 			r.Plan.Spec.TransferNetwork.Namespace, r.Plan.Spec.TransferNetwork.Name)
 	}
+	if r.Plan.Spec.Warm && r.isOvirt() {
+		annotations[AnnBindImmediate] = "true"
```
ahadas, line 998: isn't it needed also for warm migrations and cold migrations to remote OCP clusters from vSphere? in that case, maybe we should set this annotation all the time and document that we don't create DVs when using volume populators (maybe to exclude the flow in which el9 image of the conversion pod is used - because it's better to bind the volume on the node that the conversion pod starts on)

pkg/controller/plan/kubevirt.go
liranr23: i just returned it to the way it was.

pkg/controller/plan/kubevirt.go
liranr23: yes, indeed. but i also saw openstack is set with https://github.com/kubev2v/forklift/blob/main/pkg/controller/plan/kubevirt.go#L1837-L1839, shouldn't it be only if the source is openstack? what will happen if a user set warm true or tries remote cluster? it will enter the datavolume creation. @bennyz

bennyz: yeah

ahadas: need to block warm migrations or migrations to remote clusters from openstack until they're implemented

ahadas (approved):

ahadas:
@liranr23 please verify on mtv05 and backport

PR #282, opened 2023-03-30
openstack: use the REST resource for lookup when finalizing

### LINKED ISSUE #281: VMs that are successfully migrated from openstack are shown with Error
I created an openstack Plan with 4 VMs in for newly created NS. all four VMs are migrated and created successfully . I can start/stop them . However the UI screen represents green color for 3 steps migration and Errors at the same time.

ERROR

Discussed in slack with @bennyz - he thinks it is the issue is in the finalization 
If you would like to see the plan in mtv07: 'pa4-four-vms' in 'polina' NS (the target NS with created VMs - polina1)

bennyz:
Use the REST resource type when looking up a resource in the inventory, otherwise lookup will fail 

Fix #281

### DIFF
--- a/pkg/controller/plan/adapter/openstack/client.go
+++ b/pkg/controller/plan/adapter/openstack/client.go
@@ -21,6 +21,7 @@ import (
 	"github.com/konveyor/forklift-controller/pkg/apis/forklift/v1beta1/ref"
 	plancontext "github.com/konveyor/forklift-controller/pkg/controller/plan/context"
 	model "github.com/konveyor/forklift-controller/pkg/controller/provider/model/openstack"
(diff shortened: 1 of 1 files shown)


ahadas (commented):
@bennyz can you please elaborate on the issue? do I get it right that the name of resource.VM is (or may be) different than that of model.VM?

bennyz:
> @bennyz can you please elaborate on the issue? do I get it right that the name of resource.VM is (or may be) different than that of model.VM?
it's not the name, but the type in https://github.com/kubev2v/forklift/blob/4d76d08d4d135180313e9a564a977a9ca3b2d5ff/pkg/controller/provider/web/openstack/client.go#L109

the matching is done on a different type than model.VM, so it goes to the `default`

ahadas (approved):

PR #289, opened 2023-04-03
openstack: initial validations

mmartinv:


validation/policies/io/konveyor/forklift/openstack/cpu_shares.rego
```diff
@@ -0,0 +1,14 @@
+package io.konveyor.forklift.openstack
+
+has_cpushares_enabled {
+	input.flavor.extraSpecs["quota:cpu_shares"]
+}
+
+concerns[flag] {
+	has_cpushares_enabled
```
ahadas, line 8: shouldn't it be a boolean value when we get here? I also see this in host_devices.rego below so maybe there's something about this syntax that I don't realize

validation/policies/io/konveyor/forklift/openstack/vm_image_based.rego
```diff
@@ -0,0 +1,14 @@
...
+image_based_vm {
+	input.imageID != ""
+}
+
+concerns[flag] {
+	image_based_vm
+	flag := {
+		"category": "Critical",
+		"label": "Image based VMs are not supported",
+		"assessment": "The VM is image based, only volume based VMs are supported. Attempting to migrate this VM may fail.",
```
ahadas, line 12: I'm not sure about the last part - if it's critical then we shouldn't say that the migration may succeed, how about
```suggestion
		"assessment": "The VM is image based and migration of VMs that are based on images is not supported. Only migration of VMs that are based on volumes is supported.",
```

mmartinv: What about `The VM is 'Image' based which is not currently supported. Only the migration 'Volume' based VMs is supported. The VM can be migrated but it will not boot.`?

mmartinv: > the first two parts are ok (just missing "of": `The VM is 'Image' based which is not currently supported. Only the migration of 'Volume' based VMs is supported`). 
I will fix that missing `of`.

>about the last part - wouldn't this validation, as a critical one, block the migration?
Indeed, I will remove that last part.

validation/policies/io/konveyor/forklift/openstack/cpu_shares.rego
mmartinv: Ok, I upgraded the opa version and used your latest suggestion.

validation/policies/io/konveyor/forklift/openstack/vm_image_based.rego
mmartinv: Done

ahadas (commented):
I may have missed other places but I commented on three places that should also change to be like cpu_shares.rego

ahadas (approved):
@mmartinv please take a look at the second commit - if the changes are alright, let's get it in

mmartinv:
> 

@ahadas your changes look good to me. Added `vm_os.rego` and `vif_models.rego`, please take a look.

ahadas:
@mmartinv your turn again :)

mmartinv:
> @mmartinv your turn again :)

@ahadas added minor changes, I think we can get this in...

ahadas:
`make test` in /validation passes, which is good enough, getting this one in

PR #293, opened 2023-04-11
openstack add missing constants

mmartinv:
- openstack: Add missing virtual hardware constants
- openstack: Map unsupported "vmxnet3" interface


ahadas (approved):
my only worry was about using virtio as a default since it requires drivers on Windows but I see virtio is the default model in kubevirt so it looks alright

PR #298, opened 2023-04-13
openstack: fix image properties mappings and precedence 

### LINKED ISSUE #300: The vCPU topology configured on image is not applied to the migrated VM
The vCPU topology of VM migrated from OpenStack is not the same as the topology of the source VM if the vCPU topology is configured on image.
Here is the matrix about vCPU topology test:
Screenshot from 2023-04-17 19-27-59

### LINKED ISSUE #303: multi-queue is not mapped after migration from openstack
The source VM in Openstack has hw_vif_multiqueue_enabled = true.
ethtool -l eth0
Channel parameters for eth0:
Combined:	4

scpu
Architecture:        x86_64
CPU op-mode(s):      32-bit, 64-bit
Byte Order:          Little Endian
CPU(s):              4
On-line CPU(s) list: 0-3
Thread(s) per core:  1
Core(s) per socket:  2
Socket(s):           2

after migrating to the openshift cluster the VM yaml is missing multi-queue configuration

mmartinv:
- Fix the retrieval of vCPU topology and multi-queue properties from the Image.
- Prefer the Image properties over the Flavor extra specs.

Fixes #300 , Fixes #303

### DIFF
--- a/pkg/controller/plan/adapter/openstack/builder.go
+++ b/pkg/controller/plan/adapter/openstack/builder.go
@@ -443,51 +443,52 @@ func (r *Builder) mapResources(vm *model.Workload, object *cnv.VirtualMachineSpe
 
 	// Set CPU Sockets/Cores/Threads and Memory requests
 	// TODO support NUMA, CPU pinning
-	var cpuSockets, cpuCores, cpuThreads int
-
-	cpuSockets = vm.Flavor.VCPUs
-	cpuCores = 1
-	cpuThreads = 1
-
-	if flavorCPUSockets, ok := vm.Flavor.ExtraSpecs[FlavorCpuSockets]; ok {
-		if sockets, err := strconv.Atoi(flavorCPUSockets); err == nil {
-			cpuSockets = sockets
-		}
-	} else if imageCPUSockets, ok := vm.Image.Properties[CpuSockets]; ok {
-		if sockets, ok := imageCPUSockets.(int); ok {
-			cpuSockets = sockets
-		}
-	}
-	if flavorCPUCores, ok := vm.Flavor.ExtraSpecs[FlavorCpuCores]; ok {
-		if cores, err := strconv.Atoi(flavorCPUCores); err == nil {
-			cpuCores = cores
-		}
-	} else if imageCPUCores, ok := vm.Image.Properties[CpuCores]; ok {
-		if cores, ok := imageCPUCores.(int); ok {
-			cpuCores = cores
-		}
-	}
-	if flavorCPUThreads, ok := vm.Flavor.ExtraSpecs[FlavorCpuThreads]; ok {
-		if threads, err := strconv.Atoi(flavorCPUThreads); err == nil {
-			cpuThreads = threads
-		}
-	} else if imageCPUThreads, ok := vm.Image.Properties[CpuThreads]; ok {
-		if threads, ok := imageCPUThreads.(int); ok {
-			cpuThreads = threads
-		}
-	}
-
-	resourceRequests := map[core.ResourceName]resource.Quantity{}
-
-	object.Template.Spec.Domain.CPU.Sockets = uint32(cpuSockets)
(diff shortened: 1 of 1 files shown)


pkg/controller/plan/adapter/openstack/builder.go
```diff
@@ -420,42 +420,49 @@ func (r *Builder) mapResources(vm *model.Workload, object *cnv.VirtualMachineSpe
...
 	cpuThreads = 1
 
-	if flavorCPUSockets, ok := vm.Flavor.ExtraSpecs[FlavorCpuSockets]; ok {
-		if sockets, err := strconv.Atoi(flavorCPUSockets); err == nil {
-			cpuSockets = sockets
+	if imageCPUSockets, ok := vm.Image.Properties[CpuSockets]; ok {
+		if socketsStr, ok := imageCPUSockets.(string); ok {
+			if sockets, err := strconv.Atoi(socketsStr); err == nil {
+				cpuSockets = sockets
+			}
```
nyoxi, line 427: Is it really fine to ignore the error here? Should we perhaps at least log it?

ahadas: I guess there is some validation on the openstack side of these values but +1 for logging - it can help to debug misalignment between the original and the imported vms if it ever happens

mmartinv: Done.

pkg/controller/plan/adapter/openstack/builder.go
```diff
@@ -420,42 +420,61 @@ func (r *Builder) mapResources(vm *model.Workload, object *cnv.VirtualMachineSpe
...
 
-	if flavorCPUSockets, ok := vm.Flavor.ExtraSpecs[FlavorCpuSockets]; ok {
-		if sockets, err := strconv.Atoi(flavorCPUSockets); err == nil {
-			cpuSockets = sockets
+	if imageCPUSockets, ok := vm.Image.Properties[CpuSockets]; ok {
+		if socketsStr, ok := imageCPUSockets.(string); ok {
+			if sockets, err := strconv.Atoi(socketsStr); err == nil {
+				cpuSockets = sockets
+			} else {
+				r.Log.Error(err, "unable to parse image property", "property", CpuSockets)
```
ahadas, line 428: should it be something like 
```suggestion
				r.Log.Error(err, "unable to parse image property", "cpuSockets", CpuSockets)
```
here and below?

mmartinv: I don't think so, `CpuSockets` value is `hw_cpu_sockets` which is the image property name, not the value.

nyoxi: Yes, the value is in `vm.Image.Properties[CpuSockets]`. A more likely alternative is to have: `..., "property", CpuSockets, "value", vm.Image.Properties[CpuSockets]`.

ahadas: > Yes, the value is in `vm.Image.Properties[CpuSockets]`. A more likely alternative is to have: `..., "property", CpuSockets, "value", vm.Image.Properties[CpuSockets]`.

yeah, logging the value can be useful

mmartinv: done

ahadas (approved):

nyoxi (approved):

nyoxi:
I have already approved the PR, but codecov is right. We could simplify the code a bit. The parsing of properties shares a common pattern that could be factored out into a small function.

nyoxi:
But we can do that in a separate PR later if we're short on time here.

ahadas:
> But we can do that in a separate PR later if we're short on time here.

there's no time pressure now

bennyz (approved):

ahadas (approved):

PR #305, opened 2023-04-20
openstack: map all the OpenStack VIF models

### LINKED ISSUE #304: An additional mapping for virtual network interface must be added
Currently when configuring an Image on OpenStack side we have a possibility to add some values for virtual network interfaces that come from a VMWare hypervisor and they are not supported. Now the migration of such VM (for example VM with  VirtualE1000) fails after disk transferring while the 'Create VM' phase with the error "admission webhook "[virtualmachine-validator.kubevirt.io](http://virtualmachine-validator.kubevirt.io/)" denied the request: interface spec.template.spec.domain.devices.interfaces[0].name uses model VirtualE1000 that is not supported".

mmartinv:
Add all the missing OpenStack VIF models and map
them to the supported values in OpenShift. Fallback
to `virtio` if the model is not supported by OpenShift

Fixes #304

### DIFF
--- a/pkg/controller/plan/adapter/openstack/builder.go
+++ b/pkg/controller/plan/adapter/openstack/builder.go
@@ -154,13 +154,25 @@ const (
 
 // Vif models
 const (
-	VifModelE1000   = "e1000"
-	VifModelE1000e  = "e1000e"
+	// KVM/Xen/VMWare
+	VifModelE1000 = "e1000"
+	// KVM/VMWare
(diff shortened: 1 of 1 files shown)


ahadas (approved):
we can also add a warning-validation for models that are not supported and going to change during the migration

ahadas:
> we can also add a warning-validation for models that are not supported and going to change during the migration

this requires #289

Issue #306, opened 2023-04-20
Consider on which network volume-populator will transfer disks


liranr23:
Previously we had usage of `transferNetwork` to know on which network we transfer the disks. Now using volume-populator on some flows we don't take it into account.

PR #310, opened 2023-04-25
ovirt,openstack: propagate transferNetwork to volume populators

bennyz:
Propagte the transferNetwork set for the plan to the volume populators via CRDs

Address #306


ahadas (commented):
seems like something useful to contribute back to lib-volume-populator

pkg/lib-volume-populator/populator-machinery/controller.go
```diff
@@ -576,11 +577,23 @@ func (c *controller) syncPvc(ctx context.Context, key, pvcNamespace, pvcName str
 
 		// If the pod doesn't exist yet, create it
 		if pod == nil {
+			transferNetwork, found, err := unstructured.NestedFieldNoCopy(crInstance.Object, "spec", "transferNetwork")
```
liranr23, line 580: why "spec" is needed?

bennyz: this is how it's often used, probably just to ensure the nested field has the right parent. But I am changing it to NestedStringMap to avoid casting

liranr23: actually i tried to do it without it for a PR i have, it didn't work. but when pointing out the tree it does...

bennyz: even better :)

ahadas (approved):

PR #312, opened 2023-04-27
Build our own libguestfs appliance

nyoxi:
The appliance from Kubevirt [1] is missing some packages we need. For legal
reasons it is not possible to include the missing packages there. We need to
build our own fixed appliance.

We imitate multi-stage Dockerfile build in Bazel. Unfortunately
container_run_and_extract() does not work with Podmana [2] and Bazel's
linux-sandbox. Slightly less secure but working processwrapper sandbox
is used as a workaround.

[1] https://github.com/kubevirt/libguestfs-appliance/pull/17
[2] https://github.com/bazelbuild/rules_docker/issues/2251

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -22,3 +22,9 @@ build --action_env=POPULATOR_CONTROLLER_IMAGE=quay.io/kubev2v/populator-controll
 build --action_env=OPENSTACK_POPULATOR_IMAGE=quay.io/kubev2v/openstack-populator:latest
 build --action_env=OVIRT_POPULATOR_IMAGE=quay.io/kubev2v/ovirt-populator:latest
 build --action_env=OPERATOR_IMAGE=quay.io/kubev2v/forklift-operator:latest
+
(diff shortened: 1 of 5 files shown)


ahadas (approved):
nice! there're conflicts due to #317 that need to be resolved
also this can be an alternative to few Dockerfiles/Containerfiles we've added for the populators (at least for oVirt's one that we failed to build with bazel)

ahadas:
> The appliance from Kubevirt [1] is missing some packages we need. For legal reasons it is not possible to include the missing packages there. We need to build our own fixed appliance.
> 
> We imitate multi-stage Dockerfile build in Bazel. Unfortunately container_run_and_extract() does not work with Podmana [2] and Bazel's linux-sandbox. 

@nyoxi can you please fix 'Podmana' in the commit message before merging?

nyoxi:
> @nyoxi can you please fix 'Podmana' in the commit message before merging?

Done

PR #318, opened 2023-05-01
simplify setting ownerRefs

bennyz:
Use MergeFrom which has prior art and is simpler, instead of marshaling a json

### DIFF
--- a/pkg/controller/plan/kubevirt.go
+++ b/pkg/controller/plan/kubevirt.go
@@ -2,7 +2,6 @@ package plan
 
 import (
 	"context"
-	"encoding/json"
 	"encoding/xml"
 	"fmt"
 	"math/rand"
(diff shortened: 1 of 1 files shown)


pkg/controller/plan/kubevirt.go
```diff
@@ -327,17 +326,10 @@ func (r *KubeVirt) EnsureVM(vm *plan.VMStatus) (err error) {
...
-		patchBytes, patchErr := json.Marshal(patch)
-		if patchErr != nil {
-			err = liberr.Wrap(patchErr)
-			return
-		}
-		err = r.Client.Patch(context.TODO(), &pvc, client.RawPatch(types.MergePatchType, patchBytes))
+		pvcCopy := pvc.DeepCopy()
+		pvcCopy.OwnerReferences = ownerRefs
+		patch := client.MergeFrom(pvcCopy)
+		err = r.Client.Patch(context.TODO(), &pvc, patch)
```
ahadas, line 332: I like the idea of using MergeFrom but looking at the tests [here](https://github.com/kubernetes-sigs/controller-runtime/blob/0ef07535542bb44dc642fe5c67dcd4f7a9c06a22/pkg/client/fake/client_test.go), shouldn't it be 
```golang
pvcCopy := pvc.DeepCopy()
pvc.OwnerReferences = ownerRefs // update `pvc` here
patch := client.MergeFrom(pvcCopy)
err = r.Client.Patch(context.TODO(), &pvc, patch)
```

bennyz: Indeed, also in https://github.com/kubev2v/forklift/blob/5c358f5a71e11f9dde4a13e49c06ffc415ea501a/pkg/controller/plan/kubevirt.go#L459

ahadas (approved):

liranr23 (approved):

PR #322, opened 2023-05-03
Use single combination of access mode

### LINKED ISSUE #316: spec.accessModes: Forbidden: may not use ReadWriteOncePod with other access modes
Hi there,

I'm currently trying to migrate some VMs over to openshift, and it is failing on the storage migration part.

Error:
        pipeline:
          - description: Initialize migration.
            error:
              phase: Running
              reasons:
                - >-
                  PersistentVolumeClaim "a95f2952-f020-45eb-9d5c-e62cae09fa8e"
                  is invalid: [spec.accessModes: Unsupported value:
                  "ReadWriteOncePod": supported values: "ReadOnlyMany",
                  "ReadWriteMany", "ReadWriteOnce", spec.accessModes: Forbidden:
                  may not use ReadWriteOncePod with other access modes]
                  
Is there an option to manually set the access mode? I'm using the built-in LVM storage operator and it doesn't look like it has RWOP capability.

Thanks

liranr23:
Using 'ReadWriteOncePod' with others access modes causing a conflict. Basing on CDI usage to determine the access mode based on storage profile, we should fetch only one combination, preferring 'Block' and 'ReadWriteMany'.

Fixes #316


pkg/controller/plan/kubevirt.go
```diff
@@ -795,7 +795,14 @@ func (r *KubeVirt) getStorageProfileModes(storageName string) (accessModes []cor
...
 				}
 				if *claimPropertyVolumeMode == *volumeMode {
-					accessModes = append(accessModes, claimProperty.AccessModes...)
+					for _, am := range claimProperty.AccessModes {
+						if am == core.ReadWriteMany {
+							accessModes = append(accessModes, am)
+							return
+						}
+					}
+					accessModes = append(accessModes, claimProperty.AccessModes[0])
```
ahadas, line 804: how about imitating the logic in CDI  - no need to take all accessModes but to take the ones of the first claim property

ahadas: I referred to [this](https://github.com/kubevirt/containerized-data-importer/blob/b98001422c95ebf54cb41d6bc81faab3f70bd6b1/pkg/controller/datavolume/util.go#L162-L167)

nyoxi: Yes, definitely. The current code is unsafe and should change the array length.

nyoxi (commented):
If we know the sets are sorted as we need and block mode is first then this is ok. Could you however please update a commit message with this information possibly with a link to the relevant documentation.

nyoxi:
So it seems only the default CDI values are sorted and this seems like an arbitrary choice and not something we can rely on. If we really want to prefer Block mode we need to do the sorting ourselves. Otherwise please remove the mention from commit message.

Either way this seems to fix for the associated issue.

pkg/controller/plan/kubevirt.go
```diff
@@ -766,42 +767,22 @@ func (r *KubeVirt) areOvirtPVCsReady(vm ref.Ref, step *plan.Step) (ready bool, e
...
-// Return storage profile access mode based on storage class
-func (r *KubeVirt) getStorageProfileModes(storageName string) (accessModes []core.PersistentVolumeAccessMode, volumeMode *core.PersistentVolumeMode, err error) {
-	storageProfileList := &cdi.StorageProfileList{}
-	err = r.Client.List(context.TODO(), storageProfileList)
+func (r *KubeVirt) getDefaultVolumeAndAccessMode(storageName string) ([]core.PersistentVolumeAccessMode, *core.PersistentVolumeMode, error) {
+	storageProfile := &cdi.StorageProfile{}
+	err := r.Client.Get(context.TODO(), types.NamespacedName{Name: storageName}, storageProfile)
 	if err != nil {
-		return
+		return nil, nil, errors.Wrap(err, "cannot get StorageProfile")
```
bennyz, line 774: why not `liberr.Wrap`?

bennyz:
> So it seems only the default CDI values are sorted and this seems like an arbitrary choice and not something we can rely on. If we really want to prefer Block mode we need to do the sorting ourselves. Otherwise please remove the mention from commit message.
> 
> Either way this seems to fix for the associated issue.

I think if the user overrides the default we should use what the user sets. There is the case where the user might set multiple access modes, but I think setting multiple access modes is only relevant for static PVs which vol-pop doesn't support

bennyz (approved):

ahadas:
> > So it seems only the default CDI values are sorted and this seems like an arbitrary choice and not something we can rely on. If we really want to prefer Block mode we need to do the sorting ourselves. Otherwise please remove the mention from commit message.

we should have used CDI to allocate the PVs but we didn't manage to do it together with volume populators so we ended up duplicating its logic, which is fine, but also trying to improve it - and we see that we face issues that didn't exist in their code. so I think we should try to align with CDI and if we think we can improve it, let's propose it for CDI

> > Either way this seems to fix for the associated issue.
> 
> I think if the user overrides the default we should use what the user sets. There is the case where the user might set multiple access modes, but I think setting multiple access modes is only relevant for static PVs which vol-pop doesn't support

if the user overrides the default in the StorageProfile's 'Spec'? we have no other option but to use since they would be the only values that appear in the StorageProfile's 'Status' section, no?

pkg/controller/plan/kubevirt.go
```diff
@@ -766,42 +767,22 @@ func (r *KubeVirt) areOvirtPVCsReady(vm ref.Ref, step *plan.Step) (ready bool, e
 	return
 }
 
-var filesystemMode = core.PersistentVolumeFilesystem
-
-// Return storage profile access mode based on storage class
-func (r *KubeVirt) getStorageProfileModes(storageName string) (accessModes []core.PersistentVolumeAccessMode, volumeMode *core.PersistentVolumeMode, err error) {
-	storageProfileList := &cdi.StorageProfileList{}
-	err = r.Client.List(context.TODO(), storageProfileList)
+func (r *KubeVirt) getDefaultVolumeAndAccessMode(storageName string) ([]core.PersistentVolumeAccessMode, *core.PersistentVolumeMode, error) {
```
ahadas, line 770: could you please add a comment that refers to CDI's getDefaultVolumeAndAccessMode there?

pkg/controller/plan/kubevirt.go
```diff
@@ -766,42 +767,22 @@ func (r *KubeVirt) areOvirtPVCsReady(vm ref.Ref, step *plan.Step) (ready bool, e
...
-					accessModes = append(accessModes, claimProperty.AccessModes...)
-				}
-			}
-			break
-		}
+
+	if len(storageProfile.Status.ClaimPropertySets) > 0 &&
+		len(storageProfile.Status.ClaimPropertySets[0].AccessModes) > 0 {
+		accessModes := storageProfile.Status.ClaimPropertySets[0].AccessModes
+		volumeMode := storageProfile.Status.ClaimPropertySets[0].VolumeMode
```
ahadas, line 780: AFAIR, we failed in this place before when volume mode was missing (and thus, defaulted to Filesystem) but just to be sure that we don't bring up this issue again (and if we do, we'll need to change the place we fail on), can you please check what happens when `len(storageProfile.Status.ClaimPropertySets[0].AccessModes) > 0` and volumeMode is not specified?

bennyz:
> > I think if the user overrides the default we should use what the user sets. There is the case where the user might set multiple access modes, but I think setting multiple access modes is only relevant for static PVs which vol-pop doesn't support
> 
> if the user overrides the default in the StorageProfile's 'Spec'? we have no other option but to use since they would be the only values that appear in the StorageProfile's 'Status' section, no?

Yes, I meant if for example the user sets something like this for NFS
```
  claimPropertySets:
    - accessModes:
        - ReadWriteOnce
        - ReadWriteMany
      volumeMode: Filesystem
```
we could in theory still prefer ReadWriteMany

ahadas:
> ```
>   claimPropertySets:
>     - accessModes:
>         - ReadWriteOnce
>         - ReadWriteMany
>       volumeMode: Filesystem
> ```
> 
> we could in theory still prefer ReadWriteMany

nah, sure we can, but if VMs that are created by KubeVirt and use that storage class would be set with ReadWriteOnce why should VMs that are created by MTV be different?

bennyz:
> > ```
> >   claimPropertySets:
> >     - accessModes:
> >         - ReadWriteOnce
> >         - ReadWriteMany
> >       volumeMode: Filesystem
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > we could in theory still prefer ReadWriteMany
> 
> nah, sure we can, but if VMs that are created by KubeVirt and use that storage class would be set with ReadWriteOnce why should VMs that are created by MTV be different?

yeah, that's what I meant by respecting the user's setting :)

pkg/controller/plan/kubevirt.go
liranr23: yup, good catch, good that we didn't introduce a problem we already hit.

ahadas (approved):

PR #326, opened 2023-05-03
Fix centos9 image SHA

liranr23:
Bazel requires the exact SHA of the arch it uses instead of the parent SHA.


ahadas (approved):

liranr23:
@nyoxi given the comment i wrote in the previous PR and this one (having to have the arch SHA) - might be worth to change into tag name instead of digest?

Issue #333, opened 2023-05-05
Remote migration fails at Create VM step


qiyuann:
Remote migration fails at Create VM step with error:
`persistentvolumeclaims "remote-rhv-cold-immediate-9200aaa8-b9a9-4ac7-b89f-4jg9rv" not found`

But the VM is already created on the target cluster, and the PVC exists.

PR #334, opened 2023-05-05
Revert "Use patch instead of update to set PVC owner references (backport)"

mnecas:
Reverts kubev2v/forklift#328

This is related to the issue: https://github.com/kubev2v/forklift/issues/333
It does not fix it, it just removes the commit form 2.4.1 so we can have a stable release.


qiyuann:
The issue https://github.com/kubev2v/forklift/issues/333 is solved after https://github.com/kubev2v/forklift/pull/328 is reverted

PR #342, opened 2023-05-09
CI collect systemd journal log for packstack container

eslutsky:
part of the effort to eliminate CI job instability of OpenStack,we want to include  the packstack container journal in the CI artifacts.


eslutsky:
the journal  log file in the artifacts is empty , will investigate later.

ahadas (approved):

PR #343, opened 2023-05-09
ovirt: perform snapshot removal sequentially

bennyz:
Because of ovirt-engine's internal locks, starting a snapshot removal operation will lock the VM, and will prevent other snapshot removals on the same VM, leaving all snapshots around except for the first one.

This patch waits for each snapshot removal to complete before starting the next one. And if one fails it would retry 3 times, because there is a possible race with image transfer, as CDI does not currently wait correctly for the transfer to finish.

### DIFF
--- a/pkg/controller/plan/adapter/ovirt/client.go
+++ b/pkg/controller/plan/adapter/ovirt/client.go
@@ -4,6 +4,8 @@ import (
 	"crypto/md5"
 	"encoding/hex"
 	"fmt"
+	"sync"
+	"time"
 
 	planapi "github.com/konveyor/forklift-controller/pkg/apis/forklift/v1beta1/plan"
 	"github.com/konveyor/forklift-controller/pkg/apis/forklift/v1beta1/ref"
@@ -12,6 +14,7 @@ import (
 	model "github.com/konveyor/forklift-controller/pkg/controller/provider/web/ovirt"
 	liberr "github.com/konveyor/forklift-controller/pkg/lib/error"
 	ovirtsdk "github.com/ovirt/go-ovirt"
+	"k8s.io/apimachinery/pkg/util/wait"
 	cdi "kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1"
 )
 
@@ -61,21 +64,7 @@ func (r *Client) CreateSnapshot(vmRef ref.Ref) (snapshot string, err error) {
 
 // Remove all warm migration snapshots.
 func (r *Client) RemoveSnapshots(vmRef ref.Ref, precopies []planapi.Precopy) (err error) {
-	if len(precopies) == 0 {
-		return
-	}
-	_, vmService, err := r.getVM(vmRef)
-	if err != nil {
-		return
-	}
-	snapsService := vmService.SnapshotsService()
-	for i := range precopies {
-		snapService := snapsService.SnapshotService(precopies[i].Snapshot)
-		_, err = snapService.Remove().Async(true).Query("correlation_id", r.Migration.Name).Send()
-		if err != nil {
-			err = liberr.Wrap(err)
-		}
-	}
+	// Snapshot removal is done in Finalize to avoid race conditions
 	return
 }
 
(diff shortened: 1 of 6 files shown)


pkg/controller/plan/adapter/ovirt/client.go
```diff
@@ -409,6 +418,107 @@ func (r *Client) cacert() []byte {
...
+				for _, disk := range snapshotDisks.Slice() {
+					if transferred, err := r.isDiskBeingTransferred(disk.MustId()); transferred || err != nil {
+						r.Log.Info("Disk is being transferred, retrying..")
+						return false, nil
+					} else {
+						// Try to remove the snapshot again
+						_, err = snapService.Remove().Query("correlation_id", correlationID).Send()
+						if err != nil {
+							err = liberr.Wrap(err)
+							return true, err
```
ahadas, line 482: returning true here in case of an error?

pkg/controller/plan/adapter/ovirt/client.go
```diff
@@ -409,6 +418,107 @@ func (r *Client) cacert() []byte {
...
+				err = liberr.Wrap(err)
+				return
+			}
+
+		}
+
+		for {
+			jobs, err := r.getJobs(correlationID)
+			if err != nil {
+				err = liberr.Wrap(err)
```
ahadas, line 498: but we don't do anything with it - did you intend to log it?

pkg/controller/plan/adapter/ovirt/client.go
```diff
@@ -409,6 +418,107 @@ func (r *Client) cacert() []byte {
 	return nil
 }
 
-func (c *Client) Finalize(vms []*planapi.VMStatus, planName string) {
+func (r Client) Finalize(vms []*planapi.VMStatus, planName string) {
```
ahadas, line 421: it also applies to openstack I guess, but the fact that we don't mark the status of the removal means that we may have leftovers when the controller restarts in the middle, right?

pkg/controller/plan/adapter/ovirt/client.go
```diff
@@ -409,6 +418,107 @@ func (r *Client) cacert() []byte {
...
+		_, err := snapService.Remove().Query("correlation_id", correlationID).Send()
+		r.Log.Error(err, "Remove request for snapshot failed", "snapshotID", snapshotID)
+		if err != nil {
+			// Assume we could not remove because of an ongoing transfer
+			backoff := wait.Backoff{
+				Duration: time.Second * 3,
+				Factor:   0,
+				Jitter:   0,
+				Steps:    3,
+			}
```
ahadas, line 458: the description of the PR says this would apply to the first snapshot but this way we do it for all snapshots, no?

pkg/controller/plan/adapter/ovirt/client.go
```diff
@@ -409,6 +418,107 @@ func (r *Client) cacert() []byte {
...
+	for _, vm := range vms {
+		_, vmService, err := r.getVM(vm.Ref)
+		if err != nil {
+			return
+		}
+
+		go r.removePrecopies(vm.Warm.Precopies, vmService, &wg)
+	}
+
+	wg.Wait()
```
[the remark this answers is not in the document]
bennyz: it was mainly added just to make sure the function doesn't exit until all removals are done, but yeah we can also use it to log :)

pkg/controller/plan/adapter/ovirt/client.go
bennyz: if we couldn't get the snapshot it makes sense to stop waiting no? or I misunderstood?

pkg/controller/plan/adapter/ovirt/client.go
bennyz: yeah, this is kind of a best effort operation, we can however add a field like `finailizeCompleted: false` or something similar and have the controller rerun this, but I'm not sure about the risk/reward ratio here 😅

pkg/controller/plan/adapter/ovirt/client.go
bennyz: yes, updated

pkg/controller/plan/adapter/ovirt/client.go
ahadas: yeah, me neither - after all, the goal is to move the workload to the target cluster, the workloads on the source should not be used after the migration.. just for backup

pkg/settings/migration.go
```diff
@@ -31,6 +34,10 @@ type Migration struct {
 	ImporterRetry int
 	// Warm migration precopy interval in minutes
 	PrecopyInterval int
+	// Snapshot removal timeout in minutes
+	SnapshotRemovalTimeout int
+	// Snapshot status check rate in seconds
+	SnapshotStatusCheckRate int
```
ahadas, line 40: not used yet?

pkg/controller/plan/adapter/ovirt/client.go
```diff
@@ -372,6 +361,26 @@ func (r *Client) getDiskSnapshot(diskID, targetSnapshotID string) (diskSnapshotI
...
+		phase, _ := transfer.Phase()
+		if phase != ovirtsdk.IMAGETRANSFERPHASE_FINISHED_FAILURE && phase != ovirtsdk.IMAGETRANSFERPHASE_FINISHED_SUCCESS {
+			transferDiskID, _ := transfer.MustImage().Id()
+			if transferDiskID == diskID {
+				return true, nil
+			}
+		}
+	}
+
+	return false, nil
```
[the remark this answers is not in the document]
ahadas: the logic works different than I thought, after speaking with @bennyz - that's fine

PR #344, opened 2023-05-09
CI openstack tests with https keystone

eslutsky:


.github/workflows/pull-request.yml
```diff
@@ -38,7 +38,7 @@ jobs:
 
       - name: Build and setup everything with bazel
         id: forkliftci
-        uses: kubev2v/forkliftci@v1.0
+        uses: kubev2v/forkliftci@main
```
liranr23, line 41: since we use versioning now, shouldn't it be v<something> and not main?

eslutsky: yes, but i want to to test the PR against the latest CI code (since  the tests depends on it)

tests/suit/openstack_test.go
```diff
@@ -19,6 +19,63 @@ const (
...
+				"domainName":         "default",
+				"projectName":        "admin",
+				"regionName":         "RegionOne",
+				"insecureSkipVerify": "true",
+				"cacert":             "",
+			}, nil, namespace, "os-test-secret"))
+		Expect(err).ToNot(HaveOccurred())
+
+		By("Create osp provider")
+		pr := utils.NewProvider(openstackProviderName, forkliftv1.OpenStack, namespace, map[string]string{},
```
liranr23, line 71: try a different provider name? maybe the secured should have suffix `secured` and here `insecured`.

eslutsky: okay, i've made the change

.github/workflows/pull-request.yml
ahadas: btw, I tried to make it configurable by using a variable that will appear in both cases but [github doesn't support it](https://github.com/orgs/community/discussions/25246)

tests/suit/openstack_test.go
```diff
@@ -19,6 +19,63 @@ const (
...
+		By("Create Secret from Definition")
+		s, err := utils.CreateSecretFromDefinition(f.K8sClient, utils.NewSecretDefinition(nil,
+			map[string]string{
+				"username":           "admin",
+				"password":           "12e2f14739194a6c",
+				"domainName":         "default",
+				"projectName":        "admin",
+				"regionName":         "RegionOne",
+				"insecureSkipVerify": "true",
+				"cacert":             "",
```
ahadas, line 66: I try to see if I got it right - the difference between this and the provider that is set below is in this line, did you add this test to check adding a provider separately (and not as part of testing it with NetworkMap) or because of this line?

eslutsky: this only tests provider creation, with out running the migration, i can simplify this further and only test the openstack.client  interface  (without creating the provider CRs ie) so the  connection attempt will be made from the test  code (outside from the cluster) and not from the inventory pod.

eslutsky: the name `[test] should create provider with NetworkMap` below is misleading and should be changed.

eslutsky: done

eslutsky:
blocked until https://github.com/kubev2v/forklift/pull/348 is merged

ahadas (approved):

.github/workflows/pull-request.yml
liranr23: should it change now?

ahadas:
dropped the approval as it looks like this adds a memory overhead that the hosted runners cannot cope with, as indicated by @eslutsky

ahadas (approved):

PR #345, opened 2023-05-10
VMware snapshots removal during warm migration

### LINKED ISSUE #301: VMWare snapshots are not deleted after a successful warm migration
The snapshots created on VMWare during warm migration are not deleted after the migration is successfully completed.

nyoxi:
After moving VM creation to the end of the flow checkpoint update became
broken. As a result old snapshots were not removed during incremental
warm migration and checkpoints were not updated on DateVolume CR.

Fixes #301

### DIFF
--- a/pkg/controller/plan/adapter/vsphere/client.go
+++ b/pkg/controller/plan/adapter/vsphere/client.go
@@ -41,6 +41,7 @@ type Client struct {
 
 // Create a VM snapshot and return its ID.
 func (r *Client) CreateSnapshot(vmRef ref.Ref) (id string, err error) {
+	r.Log.V(1).Info("Creating snapshot", "vmRef", vmRef)
 	vm, err := r.getVM(vmRef)
 	if err != nil {
 		return
@@ -56,6 +57,7 @@ func (r *Client) CreateSnapshot(vmRef ref.Ref) (id string, err error) {
 		return
 	}
 	id = res.Result.(types.ManagedObjectReference).Value
+	r.Log.Info("Created snapshot", "vmRef", vmRef, "id", id)
 
 	return
 }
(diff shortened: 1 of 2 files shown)


pkg/controller/plan/adapter/vsphere/client.go
```diff
@@ -41,6 +48,7 @@ type Client struct {
 
 // Create a VM snapshot and return its ID.
 func (r *Client) CreateSnapshot(vmRef ref.Ref) (id string, err error) {
+	log.V(1).Info("Creating snapshot", "vmRef", vmRef)
```
liranr23, line 51: why not using r.Log..?

pkg/controller/plan/migration.go
```diff
@@ -1522,16 +1527,13 @@ func (r *Migration) setDataVolumeCheckpoints(vm *plan.VMStatus) (err error) {
 			return
 		}
 	}
-	var vmCr VirtualMachine
-	found := false
-	if vmCr, found = r.vmMap[vm.ID]; !found {
```
liranr23, line 1527: well, my try to save querying DVs didn't go as expected :<

pkg/controller/plan/adapter/vsphere/client.go
ahadas: yeah, for consistency with other adapters, better to use the logger within the `Context`

pkg/controller/plan/migration.go
```diff
@@ -416,7 +416,12 @@ func (r *Migration) CleanUp(vm *plan.VMStatus) (err error) {
 		return
 	}
 	if vm.Warm != nil {
-		_ = r.provider.RemoveSnapshots(vm.Ref, vm.Warm.Precopies)
+		if errLocal := r.provider.RemoveSnapshots(vm.Ref, vm.Warm.Precopies); errLocal != nil {
+			r.Log.Info(
```
ahadas, line 420: yeah, I wouldn't fail the cleanup on this but not sure about the 'info' level here - I'd expect this to be logged as a warning if we had such a level, so maybe with 'error' level?

pkg/controller/plan/migration.go
ahadas, line 1527: so now that `r.vmMap` is (rightly) not used we can drop the first few lines above, right?

pkg/controller/plan/adapter/vsphere/client.go
nyoxi: I keep forgetting about that one. Done.

pkg/controller/plan/migration.go
nyoxi: right

pkg/controller/plan/migration.go
nyoxi: Error might be little too much, but OK.

liranr23 (approved):

ahadas (approved):

PR #346, opened 2023-05-10
virt-v2v: cold: do not use glibc-2.34-66

nyoxi:
For some reason latest glibc in CentOS Stream 9 causes QEMU in TCG emulation to fail to boot and leads to kernel panic. Revert to previous glibc version when building the libguestfs appliance.

There is another issue somewhere causing that when the appliance build triggers kernel panic the build will not terminate and QEMU is stuck waiting forever. This makes the investigation of a failure complicated, but it's something we cannot fix on our side.

### DIFF
--- a/virt-v2v/cold/BUILD.bazel
+++ b/virt-v2v/cold/BUILD.bazel
@@ -30,6 +30,10 @@ container_run_and_extract(
     commands = [
         "set -x",
         "dnf -y update",
+        # Latest glibc in CentOS Stream 9 causes QEMU in TCG emulation to fail
+        # to boot and leads to kernel panic. If it is installed revert to older
+        # version.
+        "( [ \\\"\\$(rpm -q glibc)\\\" == \\\"glibc-2.34-66.el9.x86_64\\\" ] && dnf -y downgrade glibc-2.34-65.el9.x86_64 || true )",
         "dnf -y install libguestfs libguestfs-appliance libguestfs-xfs libguestfs-winsupport qemu-img supermin",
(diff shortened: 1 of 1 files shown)


liranr23 (approved):
worth commenting in the build file about this?

virt-v2v/cold/BUILD.bazel
```diff
@@ -30,6 +30,7 @@ container_run_and_extract(
     commands = [
         "set -x",
         "dnf -y update",
+        "( [ \\\"\\$(rpm -q glibc)\\\" == \\\"glibc-2.34-66.el9.x86_64\\\" ] && dnf -y downgrade glibc-2.34-65.el9.x86_64 || true )",
```
ahadas, line 33: please add a comment about it above

nyoxi: Done.

ahadas (approved):
@nyoxi this change looks good but did you file a bug on this? the packages on centos stream 9 are updated so quickly that I guess we'll lose that version 2.34.66 soon..

nyoxi:
> @nyoxi this change looks good but did you file a bug on this? the packages on centos stream 9 are updated so quickly that I'm afraid we'll lose that version 2.34.66 soon..

Not yet, but I am still considering opening a bug on QEMU. Although I am quite unsure if it's worth any effort exactly for the reason you mentioned. Chances are next glibc build will not suffer from this.

ahadas:
> Chances are next glibc build will not suffer from this.

yeah, if the next build will be alright then that's ok, but what do we base it on?

PR #348, opened 2023-05-11
openstack: fix insecure connections

mmartinv:
- openstack: rename function
- openstack: support unversioned authentication types
- openstack: configure TLS for https only
- openstack: allow to configure the endpoint interface
- openstack: make sure the TLS configuration is applied.
- openstack: make less connections to openstack

### DIFF
--- a/pkg/controller/provider/container/openstack/client.go
+++ b/pkg/controller/provider/container/openstack/client.go
@@ -53,6 +53,7 @@ const (
 	DefaultDomain               = "defaultDomain"
 	InsecureSkipVerify          = "insecureSkipVerify"
 	CACert                      = "cacert"
+	EndpointAvailability        = "availability"
 )
 
 var supportedAuthTypes = map[string]clientconfig.AuthType{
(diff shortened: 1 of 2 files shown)


eslutsky:
looks like the CI failure is not related to the change:
in k8s-packstack-cinder-volume.log log artifact:
```
2023-05-11 12:24:39.112 1235 ERROR oslo.messaging._drivers.impl_rabbit [-] [b9342e14-c450-4335-9426-a331b116cf5d] AMQP server on 127.0.0.1:5672 is unreachable: Too many heartbeats missed. Trying again in 1 seconds.: amqp.exceptions.ConnectionForced: Too many heartbeats missed
2023-05-11 12:24:44.668 1235 ERROR oslo.messaging._drivers.impl_rabbit [-] [19e60d71-e68c-4fe7-8d78-8c366adc8f43] AMQP server on 127.0.0.1:5672 is unreachable: Too many heartbeats missed. Trying again in 1 seconds.: amqp.exceptions.ConnectionForced: Too many heartbeats missed
2023-05-11 12:24:47.416 1235 ERROR oslo.messaging._drivers.impl_rabbit [-] [b9342e14-c450-4335-9426-a331b116cf5d] AMQP server on 127.0.0.1:5672 is unreachable: <RecoverableConnectionError: unknown error>. Trying again in 1 seconds.: amqp.exceptions.RecoverableConnectionError: <RecoverableConnectionError: unknown error>
```
@ahadas looks like journal didnt catch the root cause , we will have to include rabbitmq logs in the artifacts .

ahadas:
> @ahadas looks like journal didnt catch the root cause , we will have to include rabbitmq logs in the artifacts .

sure, we can do that in a separate PR

pkg/controller/provider/container/openstack/client.go
```diff
@@ -98,93 +100,135 @@ func (r *Client) Connect() (err error) {
...
-		roots := x509.NewCertPool()
-		ok := roots.AppendCertsFromPEM(cacert)
-		if !ok {
-			r.Log.Info("the CA certificate is malformed or was not provided, falling back to system CA cert pool")
-			roots, err = x509.SystemCertPool()
-			if err != nil {
-				err = liberr.New("failed to configure the system's cert pool")
-				return
+	var identityUrl *url.URL
+	identityUrl, err = url.Parse(r.URL)
```
ahadas, line 124: wouldn't it make sense to do that before calling `openstack.NewClient`?

ahadas (approved):
ok, only a minor comment inside so let's get it in and address it separately

pkg/controller/provider/container/openstack/client.go
ahadas: posted #349

PR #350, opened 2023-05-16
Keep populator pod and limit the restarts

liranr23:
This patch will keep the populator pod when it completes the execution. In case of error, it limits the number of restarts of the pod.


pkg/lib-volume-populator/populator-machinery/controller.go
```diff
@@ -761,17 +765,17 @@ func (c *controller) syncPvc(ctx context.Context, key, pvcNamespace, pvcName str
 	// *** At this point the volume population is done and we're just cleaning up ***
 	c.recorder.Eventf(pvc, corev1.EventTypeNormal, reasonPodFinished, "Populator finished")
 
-	// If the pod still exists, delete it
-	if pod != nil {
-		err = c.kubeClient.CoreV1().Pods(populatorNamespace).Delete(ctx, pod.Name, metav1.DeleteOptions{})
+	// If PVC' still exists, delete it
```
ahadas, line 768:
```suggestion
	// If PVC still exists, delete it
```

pkg/lib-volume-populator/populator-machinery/controller.go
```diff
@@ -761,17 +765,17 @@ func (c *controller) syncPvc(ctx context.Context, key, pvcNamespace, pvcName str
...
+		err = c.kubeClient.CoreV1().PersistentVolumeClaims(populatorNamespace).Delete(ctx, pvcPrime.Name, metav1.DeleteOptions{})
 		if err != nil {
 			return err
 		}
 	}
 
 	// If PVC' still exists, delete it
 	if pvcPrime != nil {
-		err = c.kubeClient.CoreV1().PersistentVolumeClaims(populatorNamespace).Delete(ctx, pvcPrime.Name, metav1.DeleteOptions{})
+		err = c.ensureFinalizer(ctx, pvcPrime, pvcKuberenetesFinalizer, false)
```
ahadas, line 778: how about inline this code in the block above?

liranr23: right right, my bad.

ahadas (approved):

pkg/lib-volume-populator/populator-machinery/controller.go
```diff
@@ -761,17 +765,13 @@ func (c *controller) syncPvc(ctx context.Context, key, pvcNamespace, pvcName str
...
+		err = c.kubeClient.CoreV1().PersistentVolumeClaims(populatorNamespace).Delete(ctx, pvcPrime.Name, metav1.DeleteOptions{})
 		if err != nil {
 			return err
 		}
-	}
-
-	// If PVC' still exists, delete it
-	if pvcPrime != nil {
-		err = c.kubeClient.CoreV1().PersistentVolumeClaims(populatorNamespace).Delete(ctx, pvcPrime.Name, metav1.DeleteOptions{})
+		err = c.ensureFinalizer(ctx, pvcPrime, pvcKuberenetesFinalizer, false)
```
bennyz, line 774: still not excited about this 😅 
unless we can prove this is safe

liranr23: few lines above the controller checks for rebind of the PV, making the prime PVC a dummy. it's in the state the populator pod finished, the original code is trying to remove the pvc prime but now since we have it attached to the populator pod and we keep it for logs, we need to remove the finalizer. i think it's safe enough.

ahadas: it would have been alright in my perspective to keep the prime PVC as well until we clean up after the migration, but this looks ok to me as, like Liran wrote above, the prime PVC is not supposed to be used at this point - @bennyz what are you worried about?

liranr23: you are the storage expert, i think you know better than i do :)
however, if it can happen - it shouldn't or be very rare, as we are basing also on k8s code here. we can add check for the bounding of the prime pvc, but clearly we have this check before to cover this.

liranr23: ok, then it suppose to be phase `Lost`. which is what we check few lines above. if it's not `Lost` it will reconcile (without deletion of the pvc prime and removal of the finalizer).

liranr23:
Closing it in favor of https://github.com/kubev2v/forklift/pull/223, as it will cause some problem unless we have the whole thing. About the PVC prime discussion we have an agreement to have it as it is now in the PRs, although leaving the PVC prime in termination state is also OK.

PR #351, opened 2023-05-16
Add test for ovirt insecure connection

bkhizgiy:
Adding new VM in fakeovirt for the insecure test in this commit https://github.com/kubev2v/fakeovirt/pull/10

### DIFF
--- a/tests/suit/framework/ovirt.go
+++ b/tests/suit/framework/ovirt.go
@@ -10,7 +10,7 @@ import (
 
 const DefaultStorageClass = "nfs-csi"
 
-func (r *OvirtClient) SetupClient() (err error) {
+func (r *OvirtClient) SetupClient(insecure bool) (err error) {
 	r.CustomEnv = false
 	if env := os.Getenv("OVIRT_CUSTOM_ENV"); env == "true" {
 		r.CustomEnv = true
@@ -21,25 +21,36 @@ func (r *OvirtClient) SetupClient() (err error) {
 		r.storageClass = sc
 	}
 
-	mandEnvVars := []string{"OVIRT_USERNAME", "OVIRT_PASSWORD", "OVIRT_URL", "OVIRT_CACERT", "OVIRT_VM_ID"}
-	for _, field := range mandEnvVars {
+	envVars := []string{"OVIRT_USERNAME", "OVIRT_PASSWORD", "OVIRT_URL"}
+
+	if !insecure {
+		envVars = append(envVars, "OVIRT_CACERT", "OVIRT_VM_ID")
+	} else {
+		envVars = append(envVars, "OVIRT_INSECURE_VM_ID")
+	}
+	for _, field := range envVars {
 		if env := os.Getenv(field); env == "" {
 			return fmt.Errorf("%s is not set", field)
 		}
 	}
 
-	cacertFile := os.Getenv("OVIRT_CACERT")
-	fileinput, err := os.ReadFile(cacertFile)
-	if err != nil {
-		return fmt.Errorf("could not read %s", cacertFile)
+	if !insecure {
+		cacertFile := os.Getenv("OVIRT_CACERT")
+		fileinput, err := os.ReadFile(cacertFile)
+		if err != nil {
+			return fmt.Errorf("could not read %s", cacertFile)
+		}
+		r.Cacert = fileinput
+		r.Insecure = false
+		r.vmData.testVMId = os.Getenv("OVIRT_VM_ID")
+	} else {
+		r.Insecure = true
+		r.vmData.testVMId = os.Getenv("OVIRT_INSECURE_VM_ID")
 	}
 
 	r.Username = os.Getenv("OVIRT_USERNAME")
 	r.Password = os.Getenv("OVIRT_PASSWORD")
 	r.OvirtURL = os.Getenv("OVIRT_URL")
-	r.vmData.testVMId = os.Getenv("OVIRT_VM_ID")
-	r.Cacert = fileinput
-
 	return
 }
 
@@ -49,11 +60,9 @@ func (r *OvirtClient) LoadSourceDetails() (vm *OvirtVM, err error) {
 	// default test values
 	sdomains := []string{"95ef6fee-5773-46a2-9340-a636958a96b8"}
 	nics := []string{"6b6b7239-5ea1-4f08-a76e-be150ab8eb89"}
-
 	// if a real custom ovirt environment is used.
 	if r.CustomEnv {
 		defer r.Close()
-		// connect to ovirt instance
 		err = r.Connect()
 
 		// get storage domain from the test VM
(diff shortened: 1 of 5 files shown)


tests/suit/ovirt_insecure_test.go
ahadas, line 1: how about using a table in which the insecure variable is passed as a parameter rather than duplicating the code in ovirt_test.go? see [an example in kubevirt](https://github.com/kubevirt/kubevirt/blob/main/tests/vm_test.go#L208-L242)

bkhizgiy:
@ahadas I've tried to convert the test to DescribeTable as you suggested in one of the comments above to avoid redundant code duplication; this option is supported only from ginkgo v2.1.0 (we currently using v.1.16.4).
I was trying to update the version to the right one, but it caused a lot of issues, some parts in the tests were no longer supported, and I had to modify a lot of things, even after all of that, it still had some compilation errors that I didn't manage to solve. I do agree that it would be a better solution, but I don't think all these changes are justified for a single test addition. We can have a separate task for updating the version (if it's needed), but for this one, I returned the tests into two separate files with the same flow and different variables.

+I've added a new VM for the insecure flow in fakeovirt (for some reason I can't add reviewers), its need to be merged first for the CI to pass.
https://github.com/kubev2v/fakeovirt/pull/10

ahadas:
@bkhizgiy ack

bkhizgiy:
The tests are passing now (was tested locally), switch it to sequence run and cleaning the fakeovirt pod between the tests solved the issue.
once the new VM will be merged to fakeovirt the CI will pass.

tests/suit/ovirt_test.go
```diff
@@ -13,77 +14,159 @@ import (
...
+			Expect(err).ToNot(HaveOccurred())
+			Expect(vm).ToNot(BeNil())
+		})
+
+		It("[oVirt MTV] should create insecure provider", func() {
+			namespace := f.Namespace.Name
+			By("Load Source VM Details from Ovirt")
+			vmData, err := f.Clients.OvirtClient.LoadSourceDetails()
+			Expect(err).ToNot(HaveOccurred())
+			By("Reset fakeovirt")
```
liranr23, line 98: i'm not sure how many tests(vms) we are going to add here, but maybe it worth to pull it out to `AfterEach`?

bkhizgiy: good suggestion, although its required only for tests that using image transfer, beside the regular and insecure flow I don't know if we will have something soon, we can change it later if will be needed.

tests/suit/tests_suite_test.go
```diff
@@ -13,10 +13,14 @@ import (
 )
 
 const (
-	networkMapName        = "network-map-test"
-	test_migration_name   = "migration-test"
-	test_plan_name        = "plan-test"
-	test_storage_map_name = "test-storage-map-v"
+	networkMapName                 = "network-map-test"
```
liranr23, line 16: pretty strange that only `networkMapName` is in camelcase convention while the rest is snake case. but it's not in yours change.

liranr23 (approved):

tests/suit/tests_suite_test.go
bkhizgiy: agreed:) I wanted to change that but I think it was used in some other places so I left it as is to not include irreverent changes in this PR.

tests/suit/framework/ovirt.go
```diff
@@ -77,13 +86,21 @@ func (r *OvirtClient) LoadSourceDetails() (vm *OvirtVM, err error) {
 
 // Connect - Connect to the oVirt API.
 func (r *OvirtClient) Connect() (err error) {
```
bennyz, line 88: can we do something like this?
```go
func (r *OvirtClient) Connect() (err error) {
	builder := ovirtsdk.NewConnectionBuilder().
		URL(r.OvirtURL).
		Username(r.Username)

	if r.Insecure {
		builder = builder.Insecure(true)
	} else {
		builder = builder.CACert(r.Cacert)
	}

	r.connection, err = builder.Build()
	if err != nil {
		return err
	}
	return
}
```

bennyz: this was meant to be on the right side 😅

tests/suit/framework/ovirt.go
```diff
@@ -77,13 +86,21 @@ func (r *OvirtClient) LoadSourceDetails() (vm *OvirtVM, err error) {
...
-	r.connection, err = ovirtsdk.NewConnectionBuilder().
-		URL(r.OvirtURL).
-		Username(r.Username).
-		Password(r.Password).
-		CACert(r.Cacert).
-		Insecure(false).
-		Build()
+	if !r.Insecure {
```
bennyz, line 89: can we do something like this?
```go
func (r *OvirtClient) Connect() (err error) {
	builder := ovirtsdk.NewConnectionBuilder().
		URL(r.OvirtURL).
		Username(r.Username)

	if r.Insecure {
		builder = builder.Insecure(true)
	} else {
		builder = builder.CACert(r.Cacert)
	}

	r.connection, err = builder.Build()
	if err != nil {
		return err
	}
	return
}
```

tests/suit/framework/ovirt.go
```diff
@@ -21,25 +21,35 @@ func (r *OvirtClient) SetupClient() (err error) {
 		r.storageClass = sc
 	}
 
-	mandEnvVars := []string{"OVIRT_USERNAME", "OVIRT_PASSWORD", "OVIRT_URL", "OVIRT_CACERT", "OVIRT_VM_ID"}
+	var mandEnvVars []string
```
ahadas, line 24: `mand` as mandatory?  why is it important to have this prefix?

tests/suit/framework/ovirt.go
```diff
@@ -21,25 +21,35 @@ func (r *OvirtClient) SetupClient() (err error) {
...
+	if !insecure {
+		mandEnvVars = []string{"OVIRT_USERNAME", "OVIRT_PASSWORD", "OVIRT_URL", "OVIRT_CACERT", "OVIRT_VM_ID"}
+	} else {
+		mandEnvVars = []string{"OVIRT_USERNAME", "OVIRT_PASSWORD", "OVIRT_URL", "OVIRT_INSECURE_VM_ID"}
+	}
```
ahadas, line 29: how about appending only the relevant elements here?

tests/suit/framework/ovirt.go
```diff
@@ -49,11 +59,10 @@ func (r *OvirtClient) LoadSourceDetails() (vm *OvirtVM, err error) {
 	// default test values
 	sdomains := []string{"95ef6fee-5773-46a2-9340-a636958a96b8"}
 	nics := []string{"6b6b7239-5ea1-4f08-a76e-be150ab8eb89"}
-
 	// if a real custom ovirt environment is used.
 	if r.CustomEnv {
 		defer r.Close()
-		// connect to ovirt instance
+		// connect/insecure connect to ovirt instance
```
ahadas, line 65: I'd change this to
```suggestion
		// connect (possibly insecurely) to ovirt instance
```
but is it really important to mention this??

tests/suit/framework/ovirt.go
ahadas: +1

tests/suit/framework/ovirt.go
bkhizgiy: nope, removing:)

ahadas (approved):

PR #352, opened 2023-05-16
Preserve the VM UUID when migrating from RHV

ahadas:
The UUID can be important for licesning of software that runs within the guest so better to preserve it. Note that this ID is different than the VM of the VirtualMachine entity in k8s.


Makefile
```diff
@@ -71,7 +71,7 @@ e2e-sanity: e2e-sanity-ovirt e2e-sanity-vsphere
 
 e2e-sanity-ovirt:
 	# ovirt suite
-	go test ./tests/suit -v -ginkgo.focus ".*oVirt.*"
+	KUBEVIRT_CLIENT_GO_SCHEME_REGISTRATION_VERSION=v1 go test ./tests/suit -v -ginkgo.focus ".*oVirt.*"
```
liranr23, line 74: why we need it now?

ahadas: because we query a KubeVirt entity (the imported VM), see [this](https://github.com/kubevirt/kubevirt/pull/4749)

liranr23 (approved):

bennyz (approved):

PR #355, opened 2023-05-16
controller: vsphere: make VDDK image optional

nyoxi:
Do not require VDDK image to be configured in the VMware provider. Instead of validating on the provider level we now validate on the plan. For migration that do not use the EL9 virt-v2v image (cold to remote cluster and warm) the VDDK image remains mandatory.

### DIFF
--- a/pkg/controller/plan/kubevirt.go
+++ b/pkg/controller/plan/kubevirt.go
@@ -1252,6 +1252,27 @@ func (r *KubeVirt) guestConversionPod(vm *plan.VMStatus, vmVolumes []cnv.Volume,
 	} else {
 		virtV2vImage = Settings.Migration.VirtV2vImageWarm
 	}
+	// VDDK image
+	var initContainers []core.Container
+	if vddkImage, found := r.Source.Provider.Spec.Settings["vddkInitImage"]; found {
+		initContainers = append(initContainers, core.Container{
+					Name:            "vddk-side-car",
+					Image:           vddkImage,
+					ImagePullPolicy: core.PullIfNotPresent,
+					VolumeMounts: []core.VolumeMount{
+						{
+							Name:      "vddk-vol-mount",
+							MountPath: "/opt",
+						},
+					},
+					SecurityContext: &core.SecurityContext{
+						AllowPrivilegeEscalation: &allowPrivilageEscalation,
+						Capabilities: &core.Capabilities{
+							Drop: []core.Capability{"ALL"},
+						},
+					},
+				})
+	}
 	// pod environment
 	environment, err := r.Builder.PodEnvironment(vm.Ref, r.Source.Secret)
 	if err != nil {
@@ -1271,25 +1292,7 @@ func (r *KubeVirt) guestConversionPod(vm *plan.VMStatus, vmVolumes []cnv.Volume,
 				RunAsNonRoot: &nonRoot,
 			},
 			RestartPolicy: core.RestartPolicyNever,
-			InitContainers: []core.Container{
-				{
-					Name:            "vddk-side-car",
-					Image:           r.Source.Provider.Spec.Settings["vddkInitImage"],
-					ImagePullPolicy: core.PullIfNotPresent,
-					VolumeMounts: []core.VolumeMount{
-						{
-							Name:      "vddk-vol-mount",
-							MountPath: "/opt",
-						},
-					},
-					SecurityContext: &core.SecurityContext{
-						AllowPrivilegeEscalation: &allowPrivilageEscalation,
-						Capabilities: &core.Capabilities{
-							Drop: []core.Capability{"ALL"},
-						},
-					},
-				},
-			},
+			InitContainers: initContainers,
 			Containers: []core.Container{
 				{
 					Name: "virt-v2v",
--- a/pkg/controller/plan/kubevirt.go
+++ b/pkg/controller/plan/kubevirt.go
@@ -520,7 +520,12 @@ func (r *KubeVirt) EnsureDataVolumes(vm *plan.VMStatus, dataVolumes []cdi.DataVo
 		// DataVolume and PVC names are the same
 		pvcNames = append(pvcNames, dv.Name)
 	}
-	if r.UseEl9VirtV2v() {
+	el9, el9Err := r.Context.Plan.VSphereUsesEl9VirtV2v()
+	if el9Err != nil {
+		err = el9Err
+		return
+	}
+	if el9 {
 		err = r.createPodToBindPVCs(vm, pvcNames)
 		if err != nil {
 			return err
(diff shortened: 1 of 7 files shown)


pkg/controller/plan/validation.go
```diff
@@ -626,3 +632,39 @@ func (r *Reconciler) validateHooks(plan *api.Plan) (err error) {
...
+	if source == nil {
+		return nil
+	}
+	destination := plan.Referenced.Provider.Destination
+	if destination == nil {
+		return nil
+	}
+
+	// NOTE: Same set of conditions is used in migration context in
+	// UseEl9VirtV2v(). Thes conditions need to remain synchronized.
```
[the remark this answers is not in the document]
liranr23: I also wonder if we could pull the `UseEl9VirtV2v()` to somewhere we can use in all places..

liranr23:
Are there any pros/cons of using VDDK or not? does the EL9 migration always not using VDDK even when it's provided?

liranr23:
I think I might precede arik and ask about the possibility to do it in a webhook

pkg/controller/plan/kubevirt.go
```diff
@@ -1237,8 +1237,28 @@ func (r *KubeVirt) guestConversionPod(vm *plan.VMStatus, vmVolumes []cnv.Volume,
...
+							},
+						},
+						SecurityContext: &core.SecurityContext{
+							AllowPrivilegeEscalation: &allowPrivilageEscalation,
+							Capabilities: &core.Capabilities{
+								Drop: []core.Capability{"ALL"},
+							},
+						},
+					})
+		}
```
ahadas, line 1261: need to add the vddk side-container in the else-block below, right?

pkg/controller/plan/validation.go
```diff
@@ -134,6 +135,11 @@ func (r *Reconciler) validate(plan *api.Plan) error {
 	if err != nil {
 		return err
 	}
+	// VDDK image
+	err = r.validateVddkImage(plan)
```
ahadas, line 139: as @liranr23 commented earlier, I would prefer to validate it using a webhook since we only check the validity of the configuration of the plan (in light of that of the provider) here

nyoxi:
> Are there any pros/cons of using VDDK or not?

VDDK is faster but getting the library and creating the image may be seen as complex and unnecessary step for small scale migrations.
 
> 
does the EL9 migration always not using VDDK even when it's provided?

No, if the VDDK image is provided in the pod VDDK is used.

nyoxi:
> I think I might precede arik and ask about the possibility to do it in a webhook(validation part)

This may be more complex. If we want to do that in admission webhook we need to validate not only plans but also providers. If you modify existing provider and remove VDDK image some of your plans may become invalid. That entails going through all the plans and verifying them on provider change. Since webhooks should work quickly this may not be the right thing to do. Also, however it seems we're not doing that right now, it is possible to run multiple webhooks in parallel (through load balancer). This could potentially lead to races when plan and associated provider are changed simultaneously. To sum up, I am not convinced that using a webhook here is a good way forward.

pkg/controller/plan/kubevirt.go
nyoxi: Actually this is wrong. It does not belong into this scope.

pkg/controller/plan/validation.go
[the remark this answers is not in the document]
nyoxi: I think we can. It seemed to me we use different types in different contexts but after careful look we don't.

ahadas:
> > I think I might precede arik and ask about the possibility to do it in a webhook(validation part)
> 
> This may be more complex. If we want to do that in admission webhook we need to validate not only plans but also providers. If you modify existing provider and remove VDDK image some of your plans may become invalid. That entails going through all the plans and verifying them on provider change. Since webhooks should work quickly this may not be the right thing to do.

right, but we can prevent changing a provider this way when there're non-archived plans that point to it or to use a mutating webhook when a request to add/update a plan is received:
1. look for the provider that the plan points to as a source provider (mutating webhook)
2. copy the data that is relevant for the migration from the provider to the plan, like the vddk image (mutating webhook)
3. validate the plan based on the data that it stores (validating webhook)
and then the plan no longer depends on the configuration of the provider but on a snapshot of it at the time it was created - I think it's a fair assumption, it's like users didn't expect existing vms to change in RHV when changing the template they were based on

> Also, however it seems we're not doing that right now, it is possible to run multiple webhooks in parallel (through load balancer). This could potentially lead to races when plan and associated provider are changed simultaneously. To sum up, I am not convinced that using a webhook here is a good way forward.

such races also exist in the current code I believe, let's say
1. we define a provider with vddk-image=X
2. we define a plan with that provider as a source
3. the plan was validated and reaches a 'ready' state
4. the provider is changed - vddk-image is cleared
5. the plan is executed before it is re-validated
no?

nyoxi:
Very well then, I'll add validating webhook for plan and provider.

> > > I think I might precede arik and ask about the possibility to do it in a webhook(validation part)
> > 
> > 
> > This may be more complex. If we want to do that in admission webhook we need to validate not only plans but also providers. If you modify existing provider and remove VDDK image some of your plans may become invalid. That entails going through all the plans and verifying them on provider change. Since webhooks should work quickly this may not be the right thing to do.
> 
> right, but we can prevent changing a provider this way when there're non-archived plans that point to it

I prefer this option.

> or to use a mutating webhook when a request to add/update a plan is received:
> 
>     1. look for the provider that the plan points to as a source provider (mutating webhook)
> 
>     2. copy the data that is relevant for the migration from the provider to the plan, like the vddk image (mutating webhook)
> 
>     3. validate the plan based on the data that it stores (validating webhook)
> 
> 
> and then the plan no longer depends on the configuration of the provider but on a snapshot of it at the time it was created - I think it's a fair assumption, it's like users didn't expect existing vms to change in RHV when changing the template they were based on

To me this seems non-intuitive. I prefer to see the provider as the source of truth. I can imagine a case where you not just remove the image from a provider but also from the registry (or perhaps you lose access to the registry). Your plans would start failing only later during the migration.

ahadas:
> Very well then, I'll add validating webhook for plan and provider.

it doesn't have to be part of this PR

> > and then the plan no longer depends on the configuration of the provider but on a snapshot of it at the time it was created - I think it's a fair assumption, it's like users didn't expect existing vms to change in RHV when changing the template they were based on
> 
> To me this seems non-intuitive. I prefer to see the provider as the source of truth. I can imagine a case where you not just remove the image from a provider but also from the registry (or perhaps you lose access to the registry). Your plans would start failing only later during the migration.

yeah this can happen but how likely is it?
btw, that also applies to the 'transfer network': if someone sets a provider with a transfer-network and create a plan that migrates to this provider and no transfer-network is set on the plan, then the provider's transfer-network is set on the plan (if exists) - if at this point someone changes the transfer-network on the provider, it won't propagate to such a plan that already exists

pkg/controller/plan/validation.go
```diff
@@ -626,3 +632,36 @@ func (r *Reconciler) validateHooks(plan *api.Plan) (err error) {
...
+		Message:  "VDDK image is necessary for this type of migration",
+	}
+
+	source := plan.Referenced.Provider.Source
+	if source == nil {
+		return nil
+	}
+	destination := plan.Referenced.Provider.Destination
+	if destination == nil {
+		return nil
```
ahadas, line 651: I see that we do it this way also in `validateVM` - not sure I understand the reason for this but if we get here and have a plan that is not set with a source/destination provider then I guess they were removed?

nyoxi: I guess either removed or not yet created (e.g. from command line). From what I saw we do the check when validating plan and host, but not for network and storage maps.

ahadas (approved):

nyoxi:
> @nyoxi can you please check if the failures with ovirt and openstack are related?

Uhm, it is. Will post a fix shortly.

ahadas (approved):

PR #358, opened 2023-05-16
oVirt provider: Use POST request instead of GET for auth

mnecas:
The oVirt provider uses the GET request instead of the POST to authenticate the provider. The issue is that if the auth fails the url is logged with all credentials to the pod log. Using the POST logs only the url.


bennyz (approved):

ahadas (approved):
it's surprising that it worked with Get.. changing it to Post makes sense regardless of error handling

PR #360, opened 2023-05-19
openstack: block warm migration plans

mmartinv:
Modify the validating webhook to block the creation
of warm migration plans for the OpenStack provider.


ahadas (commented):
the way the plan admitter  was implemented assumed the checks inside would be limited to validate storage classes and therefore had those early returns of util.ToAdmissionResponseAllow() - it would be better to return a positive response only at the end, when all validations above pass, but that's a comment that is unrelated to this change.
this change looks good, but how about adding a test that verifies we return the expected error in case of warm migration to openstack? it would help us when refactoring this part

liranr23:
> the way the plan admitter was implemented assumed the checks inside would be limited to validate storage classes and therefore had those early returns of util.ToAdmissionResponseAllow() - it would be better to return a positive response only at the end, when all validations above pass, but that's a comment that is unrelated to this change. this change looks good, but how about adding a test that verifies we return the expected error in case of warm migration to openstack? it would help us when refactoring this part

not only that, if we intend to add more and more validations here, it make sense to split it into methods per validation.

mmartinv:
>how about adding a test that verifies we return the expected error in case of warm migration to openstack?

@ahadas do you mean adding an e2e test?

> not only that, if we intend to add more and more validations here, it make sense to split it into methods per validation.

@liranr23 We would have two approaches here, use the same validator with different methods or use different validators with different endpoints. First approach seems simpler the second makes the validations independent and offers more performance I would say as the validators are invoked in parallel.

bennyz (approved):

ahadas (approved):

PR #362, opened 2023-05-19
openstack: do not shadow returned errors

mmartinv:
Do not define new `err` local vars shadowing
the named result parameter.


ahadas (commented):
@mmartinv I see that `go vet` doesn't complain about shadowing here, it should be fine - in this case [Redeclaration does not introduce a new variable; it just assigns a new value to the original](https://go.dev/ref/spec#Short_variable_declarations). did you get an error or an unexpected behavior because of this?

PR #363, opened 2023-05-23
fix makefile install and deploy targets

mmartinv:
- Fix install and deploy targets
- Improve Makefile readability
- Rename deprecated field name

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -251,9 +251,29 @@ build-openstack-populator-image: check_container_runtmime
 push-openstack-populator-image: build-openstack-populator-image
 	$(CONTAINER_CMD) push $(OPENSTACK_POPULATOR_IMAGE)
 
-build-all-images: build-api-image build-controller-image build-validation-image build-operator-image build-virt-v2v-image build-virt-v2v-warm-image build-operator-bundle-image build-operator-index-image build-populator-controller-image build-ovirt-populator-image build-openstack-populator-image
+build-all-images: build-api-image \
+                  build-controller-image \
+                  build-validation-image \
+                  build-operator-image \
+                  build-virt-v2v-image \
+                  build-virt-v2v-warm-image \
+                  build-operator-bundle-image \
+                  build-operator-index-image \
+                  build-populator-controller-image \
+                  build-ovirt-populator-image \
+                  build-openstack-populator-image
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -96,12 +96,12 @@ run: generate fmt vet
...
 # Install CRDs into a cluster
 install: manifests
-	kubectl apply -f config/crds
+	kubectl apply -f operator/config/crd/bases
 
 # Deploy controller in the configured Kubernetes cluster in ~/.kube/config
 deploy: manifests
-	kubectl apply -f config/crds
-	kustomize build config/default | kubectl apply -f -
+	kubectl apply -f operator/config/crd/bases
```
ahadas, line 103: I don't get something here - isn't it enough to have the command below?

ahadas (commented):
I fixed deply to deploy in the PR's description - please change the commit message of the third commit as well

Makefile
mmartinv: Yeah, according to `operator/config/default/kustomization.yml`  which by the way has the `bases` deprecated field. 
Sorry I didn't look deeper I just adjusted the actual directories, I think both targets could be simplified with the following, right?:
```
install: manifests
	kustomize build operator/config/crd  | kubectl apply -f -

# Deploy controller in the configured Kubernetes cluster in ~/.kube/config
deploy: manifests
	kustomize build operator/config/default | kubectl apply -f -
```

ahadas: yeah

Makefile
```diff
@@ -95,13 +100,12 @@ run: generate fmt vet
...
-	kubectl apply -f config/crds
+install: manifests kubectl
+	$(KUBECTL) apply -k operator/config/crds
 
 # Deploy controller in the configured Kubernetes cluster in ~/.kube/config
-deploy: manifests
-	kubectl apply -f config/crds
-	kustomize build config/default | kubectl apply -f -
+deploy: manifests kubectl
+	$(KUBECTL) apply -k operator/config/default
```
ahadas, line 108: should be `kustomize build operator/config/default | $(KUBECTL) apply -f `, no?

ahadas (approved):
the `deploy` target was broken already so let's get it in but @mmartinv please take a look at the comment inside

Makefile
mmartinv: That's equivalent to `$(KUBECTL) apply -k operator/config/default`

From `kubectl apply -h`:
```
    -k, --kustomize='':
	Process a kustomization directory. This flag can't be used together with -f or -R.
```

PR #365, opened 2023-05-24
Use provider's transfer network when defining a plan

ahadas:
Add a mutating webhook for a plan that takes into consideration the default transfer network that is set on the target provider in a way that if no transfer network is specified on the plan and the transfer network that is set for the provider exists in the namespace that the plan imports to, the plan's transfer network is set to that of the provider. This logic is applied from the UI client when using the UI wizard for defining a plan so this code is relevant only when using the CLI or when modifying the plan's YAML directly from the UI.


pkg/forklift-api/webhooks/mutating-webhook/mutators/plan-mutator.go
```diff
@@ -0,0 +1,144 @@
...
+			return util.ToAdmissionResponseError(err)
+		}
+
+		if network := targetProvider.Annotations["forklift.konveyor.io/defaultTransferNetwork"]; network != "" {
+			key := client.ObjectKey{
+				Namespace: plan.Spec.TargetNamespace,
+				Name:      network,
+			}
+
+			var rcl client.Client // remote client, i.e., client to a possibly remote cluster
```
liranr23, line 78: can't we use the same client for both local and remote?

pkg/forklift-api/webhooks/mutating-webhook/mutators/plan-mutator.go
```diff
@@ -0,0 +1,144 @@
...
+				}
+				planChanged = true
+			} else if !k8serr.IsNotFound(err) { // TODO: else if !NotFound ...
+				log.Error(err, "Failed to get the network-attachment-definition", err.Error())
+				return util.ToAdmissionResponseError(err)
+			}
+		}
+	}
+
+	if planChanged {
```
liranr23, line 117: worth logging the change?

ahadas: yeah, it's logged in line 104 above

pkg/forklift-api/webhooks/mutating-webhook/mutators/plan-mutator.go
ahadas: I've posted an optimization that uses the existing client to local cluster in case the local cluster is the target but it seems you meant something else here - what did you mean to use the same client for both? the URL is different, the credentials may be different..

ahadas:
seems that there's a race in one of the tests, probably due to missing check of the network-attachment-definition..

pkg/forklift-api/webhooks/mutating-webhook/mutators/plan-mutator.go
liranr23: i meant why using `cl` instead of initiating the `tcl` from the beginning.
```
var tcl client.Client // target client, i.e., client to a possibly remote cluster
			if targetProvider.IsHost() {
				tcl = cl
			} else {
				ref := targetProvider.Spec.Secret
```
doing it above and dropping the `cl`.

ahadas: ah that's because not all entities are available on the target namespace, the mutator needs the provider that exists only on the local cluster

operator/roles/forkliftcontroller/templates/api/mutatingwebhookconfiguration-forklift-api.yml.j2
```diff
@@ -40,4 +40,28 @@ webhooks:
...
+  rules:
+  - apiGroups:
+    - forklift.konveyor.io
+    resources:
+    - plans
+    apiVersions:
+    - v1beta1
+    operations:
+    - CREATE
+    - UPDATE
```
liranr23, line 65: how about having it only with create?

ahadas: @liranr23 the logic in this mutator should also apply to 'update', could you elaborate on why not to handle the case of 'update' here?

liranr23: true it make sense also on update, but that would be on the user fault? i am thinking about future handling of new plans with annotation of labels.

ahadas: I don't think we can tell users that drop a transfer network of a plan that they do something invalid
we can still have code in this mutator that applies only to 'create' by checking `ar.Request.Operation == admissionv1.Create`

liranr23 (approved):

PR #368, opened 2023-05-29
Vddk admission webhook

nyoxi:
Necessary refactoring is performed, extended Plan validation and added Provider validation.

### DIFF
--- a/pkg/forklift-api/webhooks/validating-webhook/admitters/plan-admitter.go
+++ b/pkg/forklift-api/webhooks/validating-webhook/admitters/plan-admitter.go
@@ -38,31 +38,31 @@ func (admitter *PlanAdmitter) Admit(ar *admissionv1.AdmissionReview) *admissionv
 
 	config, err := rest.InClusterConfig()
 	if err != nil {
-		log.Error(err, "Couldn't get the cluster configuration", err.Error())
+		log.Error(err, "Couldn't get the cluster configuration")
 		return util.ToAdmissionResponseError(err)
 	}
 
 	err = api.SchemeBuilder.AddToScheme(scheme.Scheme)
 	if err != nil {
-		log.Error(err, "Couldn't build the scheme", err.Error())
+		log.Error(err, "Couldn't build the scheme")
 		return util.ToAdmissionResponseError(err)
 	}
 	err = apis.AddToScheme(scheme.Scheme)
 	if err != nil {
-		log.Error(err, "Couldn't add forklift API to the scheme", err.Error())
+		log.Error(err, "Couldn't add forklift API to the scheme")
 		return util.ToAdmissionResponseError(err)
 	}
 
 	cl, err := client.New(config, client.Options{Scheme: scheme.Scheme})
 	if err != nil {
-		log.Error(err, "Couldn't create a cluster client", err.Error())
+		log.Error(err, "Couldn't create a cluster client")
 		return util.ToAdmissionResponseError(err)
 	}
 
 	sourceProvider := api.Provider{}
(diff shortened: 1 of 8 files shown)


pkg/forklift-api/webhooks/validating-webhook/admitters/plan-admitter.go
```diff
@@ -116,6 +83,105 @@ func (admitter *PlanAdmitter) Admit(ar *admissionv1.AdmissionReview) *admissionv
...
+	err = admitter.client.Get(
+		context.TODO(),
+		client.ObjectKey{
+			Namespace: admitter.plan.Spec.Provider.Source.Namespace,
+			Name: admitter.plan.Spec.Provider.Source.Name,
+		},
+		&admitter.sourceProvider)
+	if err != nil {
+		log.Error(err, "Couldn't get the source provider")
+		return util.ToAdmissionResponseError(err)
```
ahadas, line 158: I reviewed the provider's admitter before this one - the comment on allowing the request when not all the data is available also applies here (it was incorrect in the original code, before your changes)

pkg/forklift-api/webhooks/validating-webhook/admitters/plan-admitter.go
```diff
@@ -116,6 +83,105 @@ func (admitter *PlanAdmitter) Admit(ar *admissionv1.AdmissionReview) *admissionv
...
+	err = admitter.client.Get(
+		context.TODO(),
+		client.ObjectKey{
+			Namespace: admitter.plan.Spec.Provider.Destination.Namespace,
+			Name: admitter.plan.Spec.Provider.Destination.Name,
+		},
+		&admitter.destinationProvider)
+	if err != nil {
+		log.Error(err, "Couldn't get the destination provider")
+		return util.ToAdmissionResponseError(err)
```
ahadas, line 170: and here

pkg/forklift-api/webhooks/validating-webhook/admitters/plan-admitter.go
```diff
@@ -7,93 +7,60 @@ import (
...
-func (admitter *PlanAdmitter) Admit(ar *admissionv1.AdmissionReview) *admissionv1.AdmissionResponse {
-	log.Info("Plan admitter was called")
-	raw := ar.Request.Object.Raw
-
-	plan := &api.Plan{}
-	err := json.Unmarshal(raw, plan)
-	if err != nil {
-		return util.ToAdmissionResponseError(err)
-	}
+func (admitter *PlanAdmitter) validateStorage() *admissionv1.AdmissionResponse {
```
ahadas, line 30: the refactoring is nice!
nitpicking - how about returning an `error` instead of `*admissionv1.AdmissionResponse` and in the callers to do
```golang
if err != nil {
	return util.ToAdmissionResponseError(err)
}
```
as we never use the allows-response

liranr23:
what's the point of the provider validator? it seems to have a conflict only combined with a plan. if that the case, wouldn't plan validator be enough? until then we don't have any problem?

ahadas:
> what's the point of the provider validator? it seems to have a conflict only combined with a plan. if that the case, wouldn't plan validator be enough? until then we don't have any problem?

imagine that a plan for migration from a given vsphere provider is created and afterwards you edit the provider

liranr23:
> > what's the point of the provider validator? it seems to have a conflict only combined with a plan. if that the case, wouldn't plan validator be enough? until then we don't have any problem?
> 
> imagine that a plan for migration from a given vsphere provider is created and afterwards you edit the provider

oh right, it won't cause an update for the plan. thanks.

liranr23 (approved):

bkhizgiy (approved):
looking good, no comments from my side.

ahadas (approved):

PR #369, opened 2023-05-30
openstack fail early if the root CA is malformed

eslutsky:
Dont fall back to system-CA cert pool.

### DIFF
--- a/pkg/controller/provider/container/openstack/client.go
+++ b/pkg/controller/provider/container/openstack/client.go
@@ -126,14 +126,19 @@ func (r *Client) Authenticate() (err error) {
 		} else {
 			cacert := []byte(r.getStringFromSecret(CACert))
 			roots := x509.NewCertPool()
-			ok := roots.AppendCertsFromPEM(cacert)
-			if !ok {
-				r.Log.Info("the CA certificate is malformed or was not provided, falling back to system CA cert pool")
+
+			if len(cacert) == 0 {
+				r.Log.Info("CA certificate was not provided, falling back to system CA cert pool")
 				roots, err = x509.SystemCertPool()
 				if err != nil {
 					err = liberr.New("failed to configure the system's CA cert pool")
 					return
 				}
+			}
+
+			ok := roots.AppendCertsFromPEM(cacert)
+			if !ok {
+				err = liberr.New("CA certificate is malformed, failed to configure the CA cert pool")
 			}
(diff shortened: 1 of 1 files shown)


pkg/controller/provider/container/openstack/client.go
```diff
@@ -126,15 +126,20 @@ func (r *Client) Authenticate() (err error) {
 		} else {
 			cacert := []byte(r.getStringFromSecret(CACert))
 			roots := x509.NewCertPool()
-			ok := roots.AppendCertsFromPEM(cacert)
-			if !ok {
-				r.Log.Info("the CA certificate is malformed or was not provided, falling back to system CA cert pool")
+
+			if len(cacert) == 0 {
```
mmartinv, line 130: From https://pkg.go.dev/crypto/tls#Config:
>  // RootCAs defines the set of root certificate authorities
>	// that clients use when verifying server certificates.
>	// If RootCAs is nil, TLS uses the host's root CA set.
>	RootCAs *[x509](https://pkg.go.dev/crypto/x509).[CertPool](https://pkg.go.dev/crypto/x509#CertPool)

So maybe we can simplify this block.

pkg/controller/provider/container/openstack/client.go
```diff
@@ -126,15 +126,20 @@ func (r *Client) Authenticate() (err error) {
...
+				r.Log.Info("CA certificate was not provided, falling back to system CA cert pool")
```
ahadas, line 131: then it's not a fall back, how about
```suggestion
				r.Log.Info("CA certificate was not provided, system CA cert pool is used")
```

pkg/controller/provider/container/openstack/client.go
```diff
@@ -126,15 +126,20 @@ func (r *Client) Authenticate() (err error) {
...
 				roots, err = x509.SystemCertPool()
 				if err != nil {
 					err = liberr.New("failed to configure the system's CA cert pool")
 					return
 				}
 			}
+
+			ok := roots.AppendCertsFromPEM(cacert)
```
ahadas, line 139: in an else-block?

eslutsky: moved to else block

pkg/controller/provider/container/openstack/client.go
eslutsky: done

mmartinv (approved):
lgtm

PR #373, opened 2023-05-30
Use ESX transport for EL9 image when Host resource exists

nyoxi:
When there is a Host resource for the ESXi host on which the converted VM is located we use the `esx://` transport when running virt-v2v. The Host resource is normally created when migration network is configured in UI for the provider. There is not switch or option whether to use `vpx://` or `esx://` and the choice is based solely on the existence of the Host resource. This mimics the behavior that exists with CDI.

This also fixes a bug with EL9 image. When user configured migration network, e.g. to be used for warm migrations with CDI, this prevented any cold migrations because we passed wrong credentials in conversion pod secret and no thumbprint.

### DIFF
--- a/pkg/controller/plan/adapter/vsphere/builder.go
+++ b/pkg/controller/plan/adapter/vsphere/builder.go
@@ -187,6 +187,11 @@ func (r *Builder) PodEnvironment(vmRef ref.Ref, sourceSecret *core.Secret) (env 
 		return
 	}
 
+	sslVerify := ""
+	if container.GetInsecureSkipVerifyFlag(sourceSecret) {
+		sslVerify = "no_verify=1"
+	}
+
 	host, err := r.host(vm.Host)
 	if err != nil {
 		err = liberr.Wrap(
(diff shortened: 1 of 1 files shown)


ahadas (approved):

pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -196,37 +201,52 @@ func (r *Builder) PodEnvironment(vmRef ref.Ref, sourceSecret *core.Secret) (env
...
-			path = parent.Path
+		libvirtURL = liburl.URL{
+			Scheme:   "esx",
+			Host:     hostDef.Spec.IpAddress,
+			User:     liburl.User(string(hostSecret.Data["user"])),
+			Path:     "",
+			RawQuery: sslVerify,
+		}
+	} else {
+		// Connecth through VCenter
```
ahadas, line 221:
```suggestion
		// Connect through VCenter
```

pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -247,19 +267,27 @@ func (r *Builder) Secret(vmRef ref.Ref, in, object *core.Secret) (err error) {
...
 			return
 		}
 		in = hostSecret
+		// Get host thumbprint
+		h, nErr := r.host(hostID)
+		if nErr != nil {
+			err = nErr
+			return
+		}
+		thumbprint = h.Thumbprint
```
ahadas, line 284: yeah, I think it's better the way you did it compared to set it on the secret here (in.Data["thumbprint"] = h.Thumbprint)

ahadas (commented):
minor typo, other than that lgtm

pkg/controller/plan/adapter/vsphere/builder.go
nyoxi: It's better not to change the source provider secret. It should always contain the VCenter thumbprint.

pkg/controller/plan/adapter/vsphere/builder.go
nyoxi: Done

pkg/controller/plan/adapter/vsphere/builder.go
ahadas: yeah, I meant without persisting the changes into etcd, just so we'll operate on the secret below - but I do like the way you wrote it more

ahadas (approved):

PR #376, opened 2023-06-01
Drop unused Thumbprint field from HostSpec

ahadas:
This change theoretically breaks backward compatibility but it's really unlikely that someone automated the setting of the default network of ESX hosts, and as this field was not used there's no change in behavior. Scripts that referred to this field, for some reason, would need to change.


nyoxi:
Did you test how/where exactly is controller going to fail if somebody actually specified the thumbprint in Host CR?

I suppose we don't want to break the spec and remove the property from CRD, but perhaps we could at least change the description to "This property is deprecated" or similar.

While I am not completely opposed to this change, I am wondering what good does this actually do. If we want to fail (visibly) if somebody set the thumbprint there than maybe a validation adding a condition to the CR would be a better option.

ahadas:
@nyoxi that's a good point - I do need to verify it works as I expect it to work with existing host-configs that have this field.
I do intend to drop it from the CRD but I didn't generate the CRD for the operator yet - posted that as kind of a draft

ahadas:
> @nyoxi that's a good point - I do need to verify it works as I expect it to work with existing host-configs that have this field.

verified - that works fine
the argument for removing it is to simplify the code - currently, it looks like we store it also on that CR while we always take it from the inventory.

@liranr23 please note this change of a CRD

Issue #366, opened 2023-05-28
Set new plans with the 'populatorLabels' annotation


ahadas:
talked offline, the missing bit here is for new plans - probably to use the mutator to annotate it when a new plan is created. for existing plans it's ok, but it would be better to figure if the plan is new and didn't run to annotate it immediately.

_Originally posted by @liranr23 in https://github.com/kubev2v/forklift/pull/223#discussion_r1205293752_

PR #377, opened 2023-06-01
Webhook: set the annotation regarding populator labels on new plans

liranr23:
When creating a new plan, the controller doesn't need to set the populator
labels because it's a new plan (not from previous versions). The webhook
will set the annotation to prevent the logic of setting the labels.

#366

### DIFF
--- a/pkg/forklift-api/webhooks/mutating-webhook/mutators/plan-mutator.go
+++ b/pkg/forklift-api/webhooks/mutating-webhook/mutators/plan-mutator.go
@@ -18,111 +18,35 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+const (
+	AnnPopulatorLabels = "populatorLabels"
+)
+
 type PlanMutator struct {
+	ar   *admissionv1.AdmissionReview
+	plan api.Plan
 }
 
 func (mutator *PlanMutator) Mutate(ar *admissionv1.AdmissionReview) *admissionv1.AdmissionResponse {
 	log.Info("plan mutator was called")
+	mutator.ar = ar
 	raw := ar.Request.Object.Raw
-	plan := &api.Plan{}
-	err := json.Unmarshal(raw, plan)
+	err := json.Unmarshal(raw, &mutator.plan)
 	if err != nil {
 		log.Error(err, "mutating webhook error, failed to unmarshel plan")
 		return util.ToAdmissionResponseError(err)
 	}
 
-	var planChanged bool
-
-	if plan.Spec.TransferNetwork == nil {
-		config, err := rest.InClusterConfig()
-		if err != nil {
-			log.Error(err, "Couldn't get the cluster configuration", err.Error())
-			return util.ToAdmissionResponseError(err)
-		}
-
-		err = api.SchemeBuilder.AddToScheme(scheme.Scheme)
-		if err != nil {
-			log.Error(err, "Couldn't build the scheme", err.Error())
-			return util.ToAdmissionResponseError(err)
-		}
-		err = apis.AddToScheme(scheme.Scheme)
-		if err != nil {
-			log.Error(err, "Couldn't add forklift API to the scheme", err.Error())
-			return util.ToAdmissionResponseError(err)
-		}
-		err = net.AddToScheme(scheme.Scheme)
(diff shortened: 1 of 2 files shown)


pkg/forklift-api/webhooks/mutating-webhook/mutators/plan-mutator.go
```diff
@@ -18,60 +19,96 @@ import (
...
+			Allowed: true,
+			Result: &metav1.Status{
+				Message: "Certificate retrieval is not required, passing ",
+				Code:    http.StatusOK,
+			},
+		}
+	}
+}
+
+func (mutator *PlanMutator) transferNetwork() (bool, error) {
```
ahadas, line 70: how about a more expressive name like setTransferNetworkFromTargetProviderIfNotSet

pkg/forklift-api/webhooks/mutating-webhook/mutators/plan-mutator.go
```diff
@@ -90,55 +127,72 @@ func (mutator *PlanMutator) Mutate(ar *admissionv1.AdmissionReview) *admissionv1
...
-		patchBytes, err := util.GeneratePatchPayload(util.PatchOperation{
-			Op:    "replace",
-			Path:  "/spec",
-			Value: plan.Spec,
-		})
-
-		if err != nil {
-			log.Error(err, "mutating webhook error, failed to generete paylod for patch request")
-			return util.ToAdmissionResponseError(err)
+func (mutator *PlanMutator) populatorLabelsAnnotation() bool {
```
ahadas, line 156: also here, maybe addPopulatorLabelsAnnotation

pkg/forklift-api/webhooks/mutating-webhook/mutators/plan-mutator.go
```diff
@@ -90,55 +127,72 @@ func (mutator *PlanMutator) Mutate(ar *admissionv1.AdmissionReview) *admissionv1
...
+			Value: mutator.plan.Spec,
+		})
+	} else if metadataChanged {
+		return util.GeneratePatchPayload(util.PatchOperation{
+			Op:    "replace",
+			Path:  "/metadata",
+			Value: mutator.plan.ObjectMeta,
+		})
 	}
+	return nil, fmt.Errorf("no change, although should be one")
```
ahadas, line 197: this function is over-complicated - there are three cases above instead of two and it returns an error that cannot happen. how about returning an array of PatchOperation instead and then the caller can check if it's empty or not - if it's empty then return 'allow' without patching and otherwise patch the plan? I mean something similar to the code we have in secret-mutator

tests/suit/forklift_test.go
```diff
@@ -84,6 +88,7 @@ var _ = FDescribe("Forklift", func() {
 			By("Verify created plan")
 			Expect(plan).ToNot(BeNil())
 			Expect(plan.Spec.TransferNetwork).To(BeNil())
+			Expect(plan.Annotations[AnnPopulatorLabels]).To(Equal("True"))
```
ahadas, line 91: as it's unrelated to the transfer network, how about adding another case for this?

tests/suit/forklift_test.go
```diff
@@ -122,6 +128,7 @@ var _ = FDescribe("Forklift", func() {
 			Expect(plan).ToNot(BeNil())
 			Expect(plan.Spec.TransferNetwork).ToNot(BeNil())
 			Expect(plan.Spec.TransferNetwork.Name).To(Equal(planNetwork))
+			Expect(plan.Annotations[AnnPopulatorLabels]).To(Equal("True"))
 		})
 	})
```
ahadas, line 133: shall we also test 'update'?

pkg/forklift-api/webhooks/mutating-webhook/mutators/plan-mutator.go
liranr23: the caller already checks is spec/metadata changed if not it will allow. as the previous logic was. but regarding making it easier with array - yes make sense.

ahadas (approved):
lgtm but let's wait a bit before we get it in - with that, it would be more difficult to see that plans without that annotation are handled properly

PR #388, opened 2023-06-11
Validate host secrets

ahadas:
The validation webhook for secrets now checks also secrets that are defined for hosts (at the moment ESX hosts that are configured with transfer network) - and blocks the creation / updates of the secrets when connectivity test fails.


operator/roles/forkliftcontroller/templates/api/validatingwebhookconfiguration-forklift-api.yml.j2
```diff
@@ -25,7 +25,7 @@ webhooks:
   namespaceSelector: {}
   objectSelector:
     matchExpressions:
-    - key: createdForProviderType
+    - key: createdForResourceType
```
liranr23, line 28: there is still usage of `createdForProviderType` in the admitter. did resource replace provider? should we have both? only one is enough in order to detect we should valid the entity?

ahadas: @liranr23 there are three labels we use: createdForResourceType, createdForResource and createdForProviderType. we cannot replace createdForProviderType for two reasons:
1. we already documented in 2.4 that users can specify createdForProviderType when working with CLI (without the others)
2. imagine we label a secret with createdForResourceType=providers + createdForResource=<provider name> - it's not enough since we still need to know what type of provider it is (and we can't rely on having it in the database - as a matter of face, we rely on validating the secret before the 'provider' entity is posted)

so back to your questions  - we need to keep all of them and update the documentation accordingly

liranr23: right, so if `createdForResourceType` exist, it guarantees that  `createdForProviderType` exists?

ahadas: nope, createdForProviderType exists only in secrets for providers

liranr23: ok. what i'm trying to ask is: if `createdForResourceType` cover both and we won't miss something?

ahadas: @liranr23 it doesn't cover both, we need all createdForResourceType, createdForResource and createdForProviderType to validate secrets for providers

liranr23 (approved):

pkg/forklift-api/webhooks/validating-webhook/admitters/secret-admitter.go
```diff
@@ -58,6 +79,30 @@ func (admitter *SecretAdmitter) Admit(ar *admissionv1.AdmissionReview) *admissio
 		return webhookutils.ToAdmissionResponseAllow()
 	}
 
+	if resourceType, ok := admitter.secret.GetLabels()["createdForResourceType"]; ok {
+		switch resourceType {
+		case "hosts":
+			if hostName, ok := admitter.secret.GetLabels()["createdForResource"]; ok {
+				if tested, err := admitter.validateHostSecret(hostName); err != nil {
+					if tested {
+						return &admissionv1.AdmissionResponse{
```
bkhizgiy: Just a suggestion, how about editing the `ToAdmissionResponseError` function to have the option to send custom code, and if empty use default `http.StatusBadRequest`.
Or alternatively adding a new function with the code as a parameter.

bkhizgiy (approved):
Other then the small refactoring suggestion looking good to me.

pkg/forklift-api/webhooks/validating-webhook/admitters/secret-admitter.go
ahadas: yeah, that can be done separately

PR #392, opened 2023-06-12
create one trusted ca issuer using cert manager

yaacov:
Create one trusted ca issuer using cert manager


ahadas (approved):

yaacov:
thanks for the review ( and fixes :+1:  ) , I will make update the ui ci once an image is available

PR #393, opened 2023-06-12
ocp: start OCP source provider support

bennyz:
- [ ] Complete inventory? (might not have to)
- [ ] Mapping
	- [x] Storage
	- [ ] Network
- [x] Create target VM
- [x] Figure out namespaces for VM export related pods

~- [ ] Track progress~  No progress for gzip

- [ ] Labels and annotations
- [x] Cleanup - remove VM export
- [x] Fix inventory VM name issue where the namespace makes its way into the name
- [ ] Set transfer network for DVs (may be tricky because it is done on the target env rather than the source)
- [ ] Map volume sources properly
- [x] Implement relevant Client methods, namely PowerOff and PowerOn
- [ ] MAC conflicts


operator/config/rbac/forklift-controller_role.yaml
```diff
@@ -74,6 +74,7 @@ rules:
   resources:
   - virtualmachines
   - virtualmachines/finalizers
+  - virtualmachineexports
```
bennyz, line 77: not needed

pkg/controller/plan/adapter/ocp/builder.go
```diff
@@ -0,0 +1,469 @@
...
+	}
+
+	sourceVm, err := r.generateVmFromDefinition(vmExport)
+	if err != nil {
+		return liberr.Wrap(err)
+	}
+
+	targetVMspec := sourceVm.Spec
+
+	// TODO: move logic to mapDisks
```
ahadas, line 297: +1

pkg/controller/plan/adapter/ocp/builder.go
```diff
@@ -0,0 +1,469 @@
...
+					if _, ok := pvcMap[key]; ok {
+						diskMap[key] = disk
+						continue
+					}
+				}
+			}
+		}
+	}
+
+	// Clear original disks and volumes, will be required for other mapped devices later
```
ahadas, line 332: so maybe we should change generateVmFromDefinition to return a target vm with these empty collections?

ahadas (commented):
reviewed the first 20 files

pkg/apis/forklift/v1beta1/ref/ref.go
```diff
@@ -13,13 +13,16 @@ type Ref struct {
...
 	//   A qualified name.
 	Name string `json:"name,omitempty"`
+	// The VM Namespace
+	// Only relevant for an openshift source.
+	Namespace string `json:"namespace,omitempty"`
 	// Type used to qualify the name.
 	Type string `json:"type,omitempty"`
 }
 
 // Determine if the ref either the ID or Name is set.
```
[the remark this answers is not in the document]
bennyz: yes, because if we use Name + Namespace this will be true, but we don't want it to be
I'll change accordingly

pkg/controller/plan/adapter/ocp/builder.go
```diff
@@ -0,0 +1,469 @@
...
+	if err != nil {
+		r.Log.Error(err, "Failed to get VM export Secret")
+		return liberr.Wrap(err)
+	}
+
+	// Export pod is ready
+	// Create config maps with CA on the destination
+	// Read secret token
+	tokenSecret := &core.Secret{}
+	err = r.Client.Get(context.Background(), client.ObjectKey{Namespace: vmRef.Namespace, Name: *vmExport.Status.TokenSecretRef}, tokenSecret)
```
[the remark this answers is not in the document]
bennyz: ack

pkg/controller/plan/adapter/ocp/builder.go
```diff
@@ -0,0 +1,469 @@
...
+		err = r.Destination.Client.Create(context.TODO(), dataVolume, &client.CreateOptions{})
+		if err != nil {
+			if !k8serr.IsAlreadyExists(err) {
+				r.Log.Error(err, "Failed to create DataVolume")
+				return nil, liberr.Wrap(err)
+			}
+		}
+
+		dv := &cdi.DataVolume{}
+		err = r.Destination.Client.Get(context.TODO(), client.ObjectKey{Namespace: dataVolume.Namespace, Name: dataVolume.Name}, dv)
```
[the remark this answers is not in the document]
bennyz: I need to recheck, probably leftover

pkg/controller/plan/kubevirt.go
```diff
@@ -1062,7 +1062,7 @@ func (r *KubeVirt) virtualMachine(vm *plan.VMStatus) (object *cnv.VirtualMachine
 	//convention it will be automatically changed.
 	var originalName string
 
-	if errs := k8svalidation.IsDNS1123Label(vm.Name); len(errs) > 0 {
+	if errs := k8svalidation.IsDNS1123Label(vm.Name); !r.Plan.IsSourceProviderOCP() && len(errs) > 0 {
```
ahadas, line 1065: how can it be that we get an error from `IsDNS1123Label` when `IsSourceProviderOCP()==true`?

pkg/controller/plan/migration.go
```diff
@@ -649,6 +651,23 @@ func (r *Migration) execute(vm *plan.VMStatus) (err error) {
 		}
 
 		var pvcNames []string
+		var dataVolumes []cdi.DataVolume
```
ahadas, line 654: why moving it here? I'd also keep pvcNames closer to where it's used below

pkg/controller/plan/migration.go
```diff
@@ -1439,10 +1456,29 @@ func (r *Migration) updateCopyProgress(vm *plan.VMStatus, step *plan.Step) (err
...
+								"Could not get PVC for DataVolume.",
+								"vm",
+								vm.String(),
+								"dv",
+								path.Join(dv.Namespace, dv.Name))
+							continue
+						}
+						importer, found, kErr = r.kubevirt.GetImporterPod(*pvc)
+					} else {
+						importer, found, kErr = r.kubevirt.GetImporterPod(*dv.PVC)
```
ahadas, line 1480: I think it would be better to be more explicit here by modifying the changes above in a way that each block would set the pvc differently and then we proceed to a single call to GetImporterPod

pkg/controller/validation/provider.go
```diff
@@ -179,18 +173,8 @@ func (r *ProviderPair) validateDestination(pair provider.Pair) (result libcnd.Co
 		}
 		result.SetCondition(newCnd)
 	}
-	// A non-openshift destination is not supported.
```
ahadas, line 182: non-openshift destination should still not be supported

pkg/controller/plan/migration.go
bennyz: I think because in some revision I used it in the block below, but looks like I don't do that anymore 😅

ahadas:
@bennyz can we move it to `Ready for review`?

pkg/controller/plan/migration.go
```diff
@@ -648,7 +650,24 @@ func (r *Migration) execute(vm *plan.VMStatus) (err error) {
...
+				break
+			}
+			step.MarkCompleted()
+			step.Phase = Completed
+			vm.Phase = r.next(vm.Phase)
+			return
+		}
+
 		var pvcNames []string
+		var dataVolumes []cdi.DataVolume
```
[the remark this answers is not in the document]
bennyz: I addressed it, just moved it to the wrong place 😅

pkg/controller/plan/adapter/ocp/validator.go
```diff
@@ -0,0 +1,103 @@
...
+
+		}
+	}
+
+	return true, nil
+}
+
+// Validate that a VM's networks have been mapped.
+func (r *Validator) NetworksMapped(vmRef ref.Ref) (ok bool, err error) {
+	return true, nil
```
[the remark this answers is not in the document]
bennyz: should be ok now

ahadas (approved):

PR #398, opened 2023-06-19
Makefile deploy/undeploy targets

mmartinv:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ CONTAINER_CMD := $(shell command -v docker)
 endif
 
 REGISTRY ?= quay.io
-REGISTRY_ACCOUNT ?= kubev2v
+REGISTRY_ORG ?= kubev2v
 REGISTRY_TAG ?= devel
 
 VERSION ?= 2.5.0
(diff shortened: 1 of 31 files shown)


Makefile
```diff
@@ -125,11 +190,19 @@ fmt:
...
+# Run go mod vendor against code
+vendor:
+	go mod vendor
+
 # Generate code
 generate: controller-gen
 	$(CONTROLLER_GEN) object:headerFile="./hack/boilerplate.go.txt" paths="./pkg/apis/..."
 
-build-controller-image: check_container_runtmime
+build-controller-image: check_container_runtmime fmt tidy vendor vet
```
ahadas, line 205: I wouldn't expect `tidy` and `vendor` to be called in this context as the dependencies shouldn't change that much.. what's the reason for adding them here?
and about `fmt` and `vet` - there's no reason to execute them per-image, right? how about adding them to build-all-images or push-all-images?

hack/deploy/operator/resources/namespace.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: project.openshift.io/v1
```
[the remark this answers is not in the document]
mmartinv: It looks like I forgot to remove that file. Actually, the namespace is created before applying those manifests by using the regular kubectl command.

hack/deploy/rolebinding/operator/resources/forklift-default.yaml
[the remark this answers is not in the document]
mmartinv: It looks like the operator uses the default service account to pull the validation image.

Makefile
mmartinv: > I wouldn't expect `tidy` and `vendor` to be called in this context as the dependencies shouldn't change that much.. what's the reason for adding them here? 

Yeah, I agree, I added them without thinking much about it. Indeed, I think the missing `bazel-generate` target was the actual reason for `tidy` and `vendor` IIRC.

> and about `fmt` and `vet` - there's no reason to execute them per-image, right? how about adding them to build-all-images or push-all-images?

I also thought to add them to the `build-all-images` but none of us use that target that much I would say. The usual workflow is more like: push all images the first time and then push the controller image (or any other image we are working on), then remove the controller pod and test the changes.

The main reason for adding them per image build was to make sure we do not submit code without correct formatting. Currently if you run `go fmt` in the project some of the files will be modified and this is not the first (nor the last) time it occurs. The same happened in the past with the vendor dirs or the bazel build files which were not regenerated/updated when needed. So this was an attempt to make sure everything is up to date before posting the PRs.

hack/deploy/operator/resources/namespace.yaml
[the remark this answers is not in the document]
mmartinv: Indeed the `kustomization.yaml` file does not contain `namespace.yaml` in its resources list so it's not applied.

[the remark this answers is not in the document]
ahadas: ack so let's drop it

Makefile
mmartinv: Added the `generate-verify` target and it looks like it's working, the build failed 🤣

Makefile
```diff
@@ -109,9 +120,84 @@ run: generate fmt vet
...
+	fi; \
+	echo -n "- Waiting for the catalog pod to exist"; \
+	until $(KUBECTL) -n $(CATALOG_NAMESPACE) get pod -l olm.catalogSource=$(CATALOG_NAME) &>/dev/null; do \
+		sleep 1; echo -n "."; \
+	done; \
+	echo; \
+	if [ $$OKD -eq 0 ]; then \
+	  echo "- Waiting for the catalog pod to be in 'Pending' state"; \
+	  $(KUBECTL) -n $(CATALOG_NAMESPACE) wait --timeout=60s  --for=jsonpath=.status.phase=Pending pod -l olm.catalogSource=$(CATALOG_NAME); \
+	  echo "- Removing the catalog pod"; \
```
ahadas, line 162: why do we remove the catalog pod here?

Makefile
```diff
@@ -109,9 +120,84 @@ run: generate fmt vet
...
+	@$(KUBECTL) get catalogsource -n $(CATALOG_NAMESPACE) -o name $(CATALOG_NAME) 2>/dev/null | xargs -r $(KUBECTL) -n $(CATALOG_NAMESPACE) delete;
+	@echo "- Removing the Operator"
+	@$(KUBECTL) get operator $(OPERATOR_NAME).$(NAMESPACE) -o name 2>/dev/null | xargs -r $(KUBECTL) delete;
+	@echo "- Removing the Webhooks"
+	@$(KUBECTL) get mutatingwebhookconfiguration forklift-api -o name 2>/dev/null | xargs -r $(KUBECTL) delete;
+	@$(KUBECTL) get validatingwebhookconfiguration forklift-api -o name 2>/dev/null | xargs -r $(KUBECTL) delete;
+	@echo "- Removing the ConsolePlugin"
+	@$(KUBECTL) get consoleplugin forklift-console-plugin -o name 2>/dev/null | xargs -r $(KUBECTL) delete;
+	@echo "- Removing the OAuthClient"
+	@$(KUBECTL) get oauthclient forklift-ui -o name 2>/dev/null | xargs -r $(KUBECTL) delete;
```
ahadas, line 192: this can probably be dropped now that we dropped forklift-ui (the standalone UI)

hack/deploy/crc.sh
```diff
@@ -0,0 +1,25 @@
...
+fi
+
+eval $(crc oc-env)
+
+USERNAME=kubeadmin
+PASSWORD=$(crc console --credentials --output json | jq -r .clusterConfig.adminCredentials.password)
+oc login -u ${USERNAME} -p ${PASSWORD}  https://api.crc.testing:6443
+
+export REGISTRY=$(oc get route -n openshift-image-registry default-route -o 'jsonpath={.spec.host}')
+export REGISTRY_ORG=konveyor-forklift
```
ahadas, line 18: why not `kubev2v`?

hack/deploy/kind.sh
```diff
@@ -0,0 +1,18 @@
...
+
+[ "$(type -P kind)" ] || ( echo "kind is not in PATH" ;  exit 2 )
+
+CONTAINER_CMD=$(command -v podman || command -v docker)
+if [ -z "${CONTAINER_CMD}" ]; then
+  echo "Container runtime not detected"
+  exit 1
+fi
+
+export REGISTRY_ORG=konveyor-forklift
```
ahadas, line 11: same - why not `kubev2v`?

hack/deploy/minikube.sh
```diff
@@ -0,0 +1,19 @@
...
+
+CONTAINER_CMD=$(command -v podman || command -v docker)
+if [ -z "${CONTAINER_CMD}" ]; then
+  echo "Container runtime not detected"
+  exit 1
+fi
+
+export REGISTRY_PORT="5000"
+export REGISTRY="$(minikube ip):${REGISTRY_PORT}"
+export REGISTRY_ORG="konveyor-forklift"
```
ahadas, line 13: same here - why not `kubev2v`?

ahadas (commented):
looks good overall, few more minor comments inside and we're ready to go with this

Makefile
mmartinv: Yeah, I can remove it or leave it so older versions can be fully cleaned up, your choice... 😄

Makefile
```diff
@@ -312,3 +328,152 @@ $(DEFAULT_CONTROLLER_GEN):
...
```
ahadas, line 472: @mmartinv let's drop it - forklift-ui was already removed

hack/installation/env.sh
```diff
@@ -0,0 +1,178 @@
...
+# export OPERATOR_NAME="forklift-operator"
+# export CHANNELS="development"
+# export DEFAULT_CHANNEL="development"
+#
+### Catalog configuration
+# export CATALOG_NAMESPACE="konveyor-forklift"
+# export CATALOG_NAME="forklift-catalog"
+# export CATALOG_DISPLAY_NAME="Konveyor Forklift"
+# export CATALOG_PUBLISHER="Community"
+# export CATALOG_NAMESPACE=konveyor-forklift
```
ahadas, line 26: the last line can be dropped (duplicate)

Makefile
```diff
@@ -3,21 +3,41 @@ GOPATH ?= `go env GOPATH`
...
+CONTAINER_CMD := $(shell type -P docker)
+endif
+CONTAINER_RUNTIME=$(shell basename $(CONTAINER_CMD))
+else
+CONTAINER_CMD := $(shell type -P $(CONTAINER_RUNTIME))
 endif
 
 REGISTRY ?= quay.io
-REGISTRY_ACCOUNT ?= kubev2v
+# TODO remove REGISTRY_ORG check once the changes are merged in forkliftci
```
ahadas, line 20: were the changes posted or merged to forkliftci?

mmartinv: Posted them today: kubev2v/forkliftci#106

ahadas (approved):

PR #399, opened 2023-06-20
openstack: Ignore OpenStack floating IP addresses

mmartinv:
Although the OpenStack Floating IPs do not represent any physical
network interface within the VM, they are associated to specific
VM's IP/Port and therefore they both use the same MAC addresses.

As this is causing conflicts in network mappings we will
ignore them.

### DIFF
--- a/validation/policies/io/konveyor/forklift/openstack/floating_ips.rego
+++ b/validation/policies/io/konveyor/forklift/openstack/floating_ips.rego
@@ -1,0 +1,17 @@
+package io.konveyor.forklift.openstack
+
+addresses := input.addresses
+
+floating_ips[i] {
+  some i
+	addresses[i][_]["OS-EXT-IPS:type"] == "floating"
+}
+
+concerns[flag] {
+	count(floating_ips) != 0
+	flag := {
+		"category": "Warning",
+		"label": "Floating IPs detected",
+		"assessment": "The VM has floating IPs assigned. This functionality is not currently supported by OpenShift Virtualization. The VM can be migrated but the Floating IP configuration will be missing in the target environment.",
+	}
+}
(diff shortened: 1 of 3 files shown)


validation/policies/io/konveyor/forklift/openstack/floating_ips.rego
```diff
@@ -0,0 +1,17 @@
...
+  some i
+	addresses[i][_]["OS-EXT-IPS:type"] == "floating"
+}
+
+concerns[flag] {
+	count(floating_ips) != 0
+	flag := {
+		"category": "Warning",
+		"label": "Floating IPs detected",
+		"assessment": "The VM has floating IPs assigned. The Floating IPs will be ignored by the migration process",
```
ahadas, line 15: not sure about the wording here - I think users might interpret this as if the floating IPs just remain as-is during the migration, how about something like the assessment we have for cpu-shares:
```suggestion
		"assessment": "The VM has floating IPs assigned. This functionality is not currently supported by OpenShift Virtualization.",
```
and when looking at [the validation for cpu-share](https://github.com/kubev2v/forklift/blob/main/validation/policies/io/konveyor/forklift/ovirt/cpu_shares.rego), I see its category is `Information` so we'll be inconsistent if we set this to `Warning` even though I think warning could better fit - what do you think?

mmartinv: I will rephrase the assessment to be the same as the CPU shares. Regarding the category, maybe we should modify the CPU shares validation to be `Warning` instead. I will do a quick review of the rest of the rules also, just in case...

mmartinv: What about?:
>"assessment": "The VM has floating IPs assigned. This functionality is not currently supported by OpenShift Virtualization. The VM can be migrated but the Floating IP configuration will be missing in the target environment."

I reviewed the rest of the validation rules and I will submit a different PR with some changes like the above for the warning category. I would also change the shared disk validation category from 'Warning' to 'Critical'. I think we should block those migrations as the resulting migration will not be functional at all. What do you think?

ahadas: The suggested assessment looks good to me as well as changing other places to 'warning' but I wouldn't change the shared disk validation - the shared disk is not the bootable one so it's similar to a CD-ROM, where losing it means that not all the data is available on the destination, however that doesn't block the migration

ahadas (approved):

PR #400, opened 2023-06-21
openstack: update validation rules

mmartinv:

### DIFF
--- a/validation/policies/io/konveyor/forklift/openstack/host_devices.rego
+++ b/validation/policies/io/konveyor/forklift/openstack/host_devices.rego
@@ -12,6 +12,6 @@ concerns[flag] {
 	flag := {
 		"category": "Warning",
 		"label": "VM has mapped host devices",
-		"assessment": "The VM is configured with hardware devices mapped from the host. This functionality is not currently supported by OpenShift Virtualization. The VM can be migrated but it will not have any host devices attached to it in the target environment.",
+		"assessment": "The VM is configured with hardware devices mapped from the host. This functionality is not currently supported by OpenShift Virtualization. The VM can be migrated but it will not have any host device attached to it in the target environment.",
 	}
 }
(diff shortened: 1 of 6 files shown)


validation/policies/io/konveyor/forklift/openstack/host_devices.rego
```diff
@@ -12,6 +12,6 @@ concerns[flag] {
 	flag := {
 		"category": "Warning",
 		"label": "VM has mapped host devices",
-		"assessment": "The VM is configured with hardware devices mapped from the host. This functionality is not currently supported by OpenShift Virtualization.",
+		"assessment": "The VM is configured with hardware devices mapped from the host. This functionality is not currently supported by OpenShift Virtualization. The VM can be migrated it will not have any host devices attached to it in the target environment.",
```
ahadas, line 15:
```suggestion
		"assessment": "The VM is configured with hardware devices mapped from the host. This functionality is not currently supported by OpenShift Virtualization. The VM can be migrated but it will not have any host device attached to it in the target environment.",
```

ahadas (commented):
a minor comment about wording inside, and also for consistency we need to change bios_boot_menu.rego and cpu_shares.rego for ovirt as well

validation/policies/io/konveyor/forklift/openstack/host_devices.rego
mmartinv: Done!

mmartinv:
@ahadas maybe we should also review the rest of the rules with `Information` category, change them to `Warning` and add more context explaining what will happen if the VM is migrated.

ahadas:
> @ahadas maybe we should also review the rest of the rules with `Information` category, change them to `Warning` and add more context explaining what will happen if the VM is migrated.

yeah but that can be done in a separate PR(s)

PR #401, opened 2023-06-22
Fix log for incompatible VM names

ahadas:


pkg/controller/plan/kubevirt.go
```diff
@@ -1069,8 +1069,8 @@ func (r *KubeVirt) virtualMachine(vm *plan.VMStatus) (object *cnv.VirtualMachine
 			r.Log.Error(err, "Failed to update the VM name to meet DNS1123 protocol requirements.")
 			return
 		}
-		r.Log.Info("VM name ", originalName, " was incompatible with DNS1123 RFC, changing to ",
-			vm.Name)
+		r.Log.Info("VM name is incompatible with DNS1123 RFC, renaming",
+			"originalName", originalName, "newName", vm.Name)
```
bkhizgiy, line 1073: is there any reason `originalName` is sent twice?

ahadas: yeah, first time to describe the value (that's the key) and then the value (the name itself)
Log.Info expects arguments: msg, key-1, value-1, key-2, value-2, ...

bkhizgiy (approved):

PR #407, opened 2023-06-25
ocp: upgrade deps in preparation for VirtuaMachineExport

bennyz:


ahadas (approved):
may the odds be ever in our favor :)

PR #410, opened 2023-06-27
openstack: Fix openstack client in Plan adapter

mmartinv:
- Make sure TLS client configuration is applied
- Support new authentication types (#309).

### DIFF
--- a/pkg/controller/plan/adapter/openstack/client.go
+++ b/pkg/controller/plan/adapter/openstack/client.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"net"
 	"net/http"
+	"net/url"
 	"strconv"
 	"time"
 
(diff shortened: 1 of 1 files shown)


pkg/controller/plan/adapter/openstack/client.go
```diff
@@ -26,6 +27,37 @@ import (
...
+	UserDomainID                = "userDomainID"
+	ProjectDomainName           = "projectDomainName"
+	ProjectDomainID             = "projectDomainID"
+	DomainName                  = "domainName"
+	DomainID                    = "domainID"
+	DefaultDomain               = "defaultDomain"
+	InsecureSkipVerify          = "insecureSkipVerify"
+	CACert                      = "cacert"
+	EndpointAvailability        = "availability"
+)
```
ahadas, line 53: would the refactoring/cleanup that you plan to introduce include to reuse the consts that are defined in pkg/controller/provider/container/openstack/client.go also here?

ahadas (commented):
lgtm, just a minor question about the path forward (doesn't have to be implemented as part of this PR)

pkg/controller/plan/adapter/openstack/client.go
mmartinv: Yeah, I will move the common parts to `pkg/lib/client/openstack`, so all the common code will be there

PR #415, opened 2023-06-29
openstack: remove glance metadata from volume

bennyz:


pkg/controller/plan/adapter/openstack/builder.go
```diff
@@ -1056,6 +1057,21 @@ func (r *Builder) PreTransferActions(c planbase.Client, vmRef ref.Ref) (ready bo
 			continue
 		}
 
+		// Workaround for https://bugs.launchpad.net/cinder/+bug/1945500
```
ahadas, line 1060: if we keep this workaround, would it make sense to set those properties back?

ahadas (approved):
lgtm but this time the CI of openstack is really important :)

pkg/controller/plan/adapter/openstack/builder.go
bennyz: I think not, because the fix in cinder is to filter out these properties for the volume[1], but I will confirm in the chat


[1] https://review.opendev.org/c/openstack/cinder/+/885255/2/cinder/image/image_utils.py#1319

PR #419, opened 2023-07-03
Add OVA provider and inventory

bkhizgiy:
Adding new OVA type provider, once the provider is added a new OVA-server deployment will be applied, the controller will query the new server pod and build the provider inventory based on the info that will be provided by it.


Makefile
```diff
@@ -279,7 +288,8 @@ push-all-images:  push-api-image \
                   push-operator-index-image \
                   push-populator-controller-image \
                   push-ovirt-populator-image \
-                  push-openstack-populator-image
+                  push-openstack-populator-image\
+				  push-ova-parses-image
 
 .PHONY: check_container_runtmime
```
nyoxi, line 294: Not related to the patch but I see we have a typo in the word "runtime" here.

cmd/ova-parser/ova-parser.go
```diff
@@ -0,0 +1,336 @@
...
+	}
+
+	for _, ovaFile := range ovaFiles {
+		fmt.Println("Processing OVA file:", ovaFile)
+
+		ovfPath, tmpDir, err := extractOVFFromOVA(ovaFile)
+		if err != nil {
+			fmt.Println("Error extracting OVF from OVA:", err)
+			continue
+		}
```
nyoxi, line 157: I suggest using `defer os.RemoveAll(tmpDir)` here instead calling it separately below. Either way you may want to log the error from `os.RemoveAll`.

cmd/ova-parser/ova-parser.go
```diff
@@ -0,0 +1,336 @@
...
+	}
+
+	return ovaFiles, nil
+}
+
+func extractOVFFromOVA(ovaFile string) (string, string, error) {
+	tmpDir, err := ioutil.TempDir("", "ova_extraction")
+	print(tmpDir)
+	if err != nil {
+		os.RemoveAll(tmpDir)
```
nyoxi, line 202: You can also use `defer` here to clean only on error. This is little bit trickier but doable. You may refer to this article [1] if you're not familiar enough with it.

[1] https://blog.learngoprogramming.com/5-gotchas-of-defer-in-go-golang-part-iii-36a1ab3d6ef1

cmd/ova-parser/ova-parser.go
```diff
@@ -0,0 +1,336 @@
...
+		return "", "", err
+	}
+
+	cmd := exec.Command("tar", "-xf", ovaFile, "-C", tmpDir, "*.ovf")
```
nyoxi, line 206: You need to use `--wildcards` here if you want globing.

cmd/ova-parser/ova-parser.go
```diff
@@ -0,0 +1,336 @@
...
+	ovfFiles, err := filepath.Glob(filepath.Join(tmpDir, "*.ovf"))
+	if err != nil {
+		os.RemoveAll(tmpDir)
+		return "", "", err
+	}
+
+	if len(ovfFiles) == 0 {
+		os.RemoveAll(tmpDir)
+		return "", "", fmt.Errorf("no OVF file found in the OVA")
+	}
```
nyoxi, line 222: Since you only use first OVF you may want to log a warning if there are more files in the archive.

hack/ova-parser/Containerfile
```diff
@@ -0,0 +1,10 @@
+FROM registry.access.redhat.com/ubi8/go-toolset:1.18.9-8 as builder
```
nyoxi, line 1: It would be cleaner if we created a separate directory for the Container files naming them e.g. `ova-parser.Containerfile` and leave the `hack` directory only for support scripts.

pkg/controller/provider/controller.go
```diff
@@ -59,6 +65,17 @@ var log = logging.WithName(Name)
 // Application settings.
 var Settings = &settings.Settings
 
+const (
+	ovaContainerName = "ova"
+	ovaPodPrefix     = "ova"
+	imageVar         = "OVA_PARSER_IMAGE"
```
nyoxi, line 71: Please use more specific name instead of the general `imageVar`, e.g. `ovaImageVar`. But I understand this is out of place and still WIP.

nyoxi (changes requested):
Preliminary comments.

cmd/ova-parser/ova-parser.go
```diff
@@ -0,0 +1,450 @@
...
+	decoder := xml.NewDecoder(xmlFile)
+
+	err = decoder.Decode(&envelope)
+	if err != nil {
+		return &envelope, err
+	}
+
+	fmt.Println("Virtual Systems:")
+	for _, virtualSystem := range envelope.VirtualSystem {
+		fmt.Println("Virtual System ID:", virtualSystem.ID)
```
ahadas, line 312: is that for debugging? if so, why not to use a logger?

hack/ova-parser/Containerfile
ahadas, line 1: we used Container files for populators that were hard to build with bazel, I guess that was the reference for this, right? it should be easy to build this with bazel instead

pkg/controller/provider/controller.go
```diff
@@ -184,6 +201,55 @@ func (r Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (r
 		}
 	}
 
+	//if the provider added is ova we need to lunch a new pod
```
ahadas, line 204: we should post a deployment here rather than a pod and ideally set the provider as its owner

pkg/controller/provider/model/ova/doc.go
```diff
@@ -0,0 +1,14 @@
+package ova
+
+import (
+	"github.com/konveyor/forklift-controller/pkg/controller/provider/model/ocp"
+)
+
+// Build all models.
+func All() []interface{} {
+	return []interface{}{
+		&ocp.Provider{},
```
ahadas, line 10: ?

ahadas (commented):
looks like a good starting point

cmd/ova-parser/ova-parser.go
nyoxi: Using a library is probably preferable if it can be used easily. But you can also tell `tar` to dump the file to stdout with `-O/--to-stdout`. But you also need to call `tar` multiple times -- first to list the files and then get the output. (Just saying.)

ahadas: > But you also need to call `tar` multiple times -- first to list the files and then get the output. (Just saying.)

according to the OVF specification, the OVF descriptor should always be the first entry ("The ordering restriction ensures that it is possible to extract the OVF descriptor from a compressed OVF package without scanning the entire archive. ")

nyoxi: In theory yes. In reality there are plenty of hand crafted OVAs not following the specs. Some of which are not even TAR files (are e.g. ZIP files). And it's fine to discard those as not supported. But you don't want such files to cause you troubles.

Specifically in this context, if you just run `tar xOf somethig.ova --wildcards '*.ovf'` and the archive contains multiple OVF files you will end up with their content concatenated together which will confuse the XML parser and produce bogus errors. That's what I meant.

pkg/controller/provider/model/ova/doc.go
bkhizgiy: set this way for all the providers.

pkg/controller/provider/web/ova/tree.go
```diff
@@ -0,0 +1,187 @@
...
+}
+
+// Get not supported.
+func (h TreeHandler) Get(ctx *gin.Context) {
+	ctx.Status(http.StatusMethodNotAllowed)
+}
+
+// Tree.
+func (h TreeHandler) Tree(ctx *gin.Context) {
+	// status := h.Prepare(ctx)
```
bkhizgiy, line 72: need to be modified for the UI.

ahadas (commented):
partial review

pkg/controller/provider/model/ova/model.go
```diff
@@ -0,0 +1,129 @@
...
+
+import (
+	"github.com/konveyor/forklift-controller/pkg/controller/provider/model/base"
+	libmodel "github.com/konveyor/forklift-controller/pkg/lib/inventory/model"
+)
+
+// Variants.
+const (
+	// Cluster.
+	ComputeResource = "ComputeResource"
```
ahadas, line 11: do we use it anywhere?

pkg/controller/provider/model/ova/model.go
```diff
@@ -0,0 +1,129 @@
...
+type Concern = base.Concern
+type Ref = base.Ref
+
+// Model.
+type Model interface {
+	base.Model
+	GetName() string
+}
+
+// Base VMWare model.
```
ahadas, line 34: OVA?

pkg/controller/provider/validation.go
```diff
@@ -131,6 +131,9 @@ func (r *Reconciler) validateURL(provider *api.Provider) error {
 	if provider.IsHost() {
 		return nil
 	}
+	if provider.Type() == api.Ova {
+		return nil
```
ahadas, line 135: don't we want to validate the URL of the nfs share here?

pkg/controller/provider/controller.go
```diff
@@ -334,6 +354,155 @@ func (r *Reconciler) getSecret(provider *api.Provider) (*core.Secret, error) {
...
+		var replicas int32 = 1
+
+		//OVA server deployment
+		deployment := &appsv1.Deployment{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:        deploymentName,
+				Namespace:   provider.Namespace,
+				Annotations: annotations,
+				Labels:      labels,
+			},
```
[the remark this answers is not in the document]
bkhizgiy: done.

pkg/controller/provider/controller.go
```diff
@@ -334,6 +354,155 @@ func (r *Reconciler) getSecret(provider *api.Provider) (*core.Secret, error) {
...
+		}
+
+		secret, err1 := r.getSecret(provider)
+		if err1 != nil {
+			return
+		}
+
+		podURL := fmt.Sprintf("http://%s:8080", serviceName)
+		secret.Data["pod-url"] = []byte(podURL)
+		err = r.Update(ctx, secret)
```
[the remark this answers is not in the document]
bkhizgiy: since the only place using this value is in the Ova client, I added a new field called serviceURL in it and when it bein initialize (and test connection) I'm saving this value for all future use, using it as part of the status is a bit more complicated since its being override.

pkg/controller/map/storage/handler/ova/handler.go
```diff
@@ -0,0 +1,116 @@
...
+				"map",
+				path.Join(
+					mp.Namespace,
+					mp.Name))
+			r.Enqueue(event.GenericEvent{
+				Object: mp,
+			})
+		}
+	}
+}
```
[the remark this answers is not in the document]
bkhizgiy: no, It was added to avoid errors in the conroller, will be implemented in the second part.

pkg/controller/provider/container/ova/client.go
```diff
@@ -0,0 +1,118 @@
...
+		Transport: &http.Transport{
+			DialContext: (&net.Dialer{
+				Timeout:   15 * time.Second,
+				KeepAlive: 15 * time.Second,
+			}).DialContext,
+			MaxIdleConns: 10,
+		},
+	}
+
+	podUrl := fmt.Sprintf("http://ova-service-%s:8080", provider.Name)
```
ahadas, line 51:
```suggestion
	serviceUrl := fmt.Sprintf("http://ova-service-%s:8080", provider.Name)
```

pkg/controller/provider/controller.go
```diff
@@ -184,6 +200,24 @@ func (r Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (r
 		}
 	}
 
+	if provider.Type() == api.Ova {
+
+		deploymentName := fmt.Sprintf("%s-deployment-%s", ovaServerPrefix, provider.UID)
```
ahadas, line 205:
```suggestion
		deploymentName := fmt.Sprintf("%s-deployment-%s", ovaServerPrefix, provider.Name)
```

pkg/controller/provider/controller.go
```diff
@@ -334,6 +369,140 @@ func (r *Reconciler) getSecret(provider *api.Provider) (*core.Secret, error) {
...
+		},
+	}
+
+	err = r.Create(ctx, service)
+	if err != nil {
+		return
+	}
+}
+
+func makeOvaProviderPodSpec(url string, id string) corev1.PodSpec {
```
ahadas, line 452:
```suggestion
func makeOvaProviderPodSpec(url string, providerId string) corev1.PodSpec {
```
shall we use the provider name here as well?

pkg/controller/provider/controller.go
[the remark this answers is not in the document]
ahadas: ok, we'll look at it later then

pkg/controller/provider/controller.go
```diff
@@ -334,6 +369,140 @@ func (r *Reconciler) getSecret(provider *api.Provider) (*core.Secret, error) {
...
+	if len(splitted) != 2 {
+		fmt.Println("The string does not contain a ':'")
+	}
+	nfsServer := splitted[0]
+	nfsPath := splitted[1]
+
+	imageName, ok := os.LookupEnv(ovaImageVar)
+	if !ok {
+		fmt.Print("Couldn't find Image ", imageName)
+		imageName = defaultOVAServerImage
```
ahadas, line 465: I think we shouldn't set the `defaultOVAServerImage` variable - the image should always be specified

pkg/controller/provider/validation.go
```diff
@@ -131,6 +132,20 @@ func (r *Reconciler) validateURL(provider *api.Provider) error {
 	if provider.IsHost() {
 		return nil
 	}
+	if provider.Type() == api.Ova && provider.Spec.URL != "" {
```
ahadas, line 135: it's better to move this after the check of `if provider.Spec.URL == ""` below

ahadas (commented):
complete review

pkg/controller/provider/controller.go
bkhizgiy: I saw the same was set for virt-v2v, in 99% it should be taken from the env variable, but in the case it will have some issue do we want to fail the provider addition over this?

pkg/controller/provider/controller.go
bkhizgiy: I've changed it to provider.Name so will be consistent with the rest of the naming.

pkg/controller/provider/controller.go
ahadas: yeah, better to fail than to take an image from quay.io when we shouldn't do that, where did you see this for virt-v2v?

operator/roles/forkliftcontroller/tasks/main.yml
```diff
@@ -79,6 +79,11 @@
       definition: "{{ lookup('template', 'controller/route-inventory.yml.j2') }}"
     when: not k8s_cluster|bool
 
+  - name: "Setup forklift-controller security context constraints"
+    k8s:
+      state: present
+      definition: "{{ lookup('template', 'controller/controller-scc.yml.j2') }}"
```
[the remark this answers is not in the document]
bennyz: yeah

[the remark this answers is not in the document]
ahadas: @bennyz  I wondered why you posted it as a separate PR but I see you've closed it :)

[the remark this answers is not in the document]
bennyz: just wanted to see if it passes CI (or more accurately, to see what would be the k8s error, but there wasn't any)

operator/roles/forkliftcontroller/defaults/main.yml
```diff
@@ -115,3 +115,5 @@ must_gather_image_fqin: "{{ lookup( 'env', 'MUST_GATHER_IMAGE') or lookup( 'env'
 virt_v2v_image_fqin: "{{ lookup( 'env', 'VIRT_V2V_IMAGE') or lookup( 'env', 'RELATED_IMAGE_VIRT_V2V') }}"
 virt_v2v_warm_image_fqin: "{{ lookup( 'env', 'VIRT_V2V_WARM_IMAGE') or lookup( 'env', 'RELATED_IMAGE_VIRT_V2V_WARM') }}"
 virt_v2v_dont_request_kvm: "{{ lookup( 'env', 'VIRT_V2V_DONT_REQUEST_KVM') }}"
+
+ova_provider_server_fqin: "{{ lookup( 'env', 'OVA_PROVIDER_SERVER_IMAGE') or lookup( 'env', 'OVA_PROVIDER_SERVER_IMAGE') }}"
```
ahadas, line 119: something went wrong during rebase here

operator/roles/forkliftcontroller/tasks/main.yml
```diff
@@ -79,6 +79,11 @@
       definition: "{{ lookup('template', 'controller/route-inventory.yml.j2') }}"
     when: not k8s_cluster|bool
 
+  - name: "Setup OVA security context constraints"
```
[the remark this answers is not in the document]
ahadas: also here

operator/roles/forkliftcontroller/templates/controller/ova-scc.yml.j2
```diff
@@ -0,0 +1,12 @@
+---
+kind: SecurityContextConstraints
+apiVersion: security.openshift.io/v1
+metadata:
+  name: my-custom-scc
```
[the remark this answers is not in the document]
ahadas: and here :)

pkg/controller/provider/container/ova/model.go
```diff
@@ -0,0 +1,230 @@
...
+		updater := func(tx *libmodel.Tx) (err error) {
+			m := &model.VM{}
+			// 	Base: model.Base{ID: vm.ID},
+			// }
+			if err = tx.Get(m); err != nil {
+				if errors.Is(err, libmodel.NotFound) {
+					vm.ApplyTo(m)
+					err = tx.Insert(m)
+				}
+			} else if !(vm.OvaPath == m.OvaPath) {
```
ahadas, line 166:
```suggestion
			} else if vm.OvaPath != m.OvaPath {
```

pkg/controller/provider/web/ova/workload.go
```diff
@@ -0,0 +1,91 @@
...
+	Handler
+}
+
+// Add routes to the `gin` router.
+func (h *WorkloadHandler) AddRoutes(e *gin.Engine) {
+	e.GET(WorkloadRoot, h.Get)
+}
+
+// List resources in a REST collection.
+func (h WorkloadHandler) List(ctx *gin.Context) {
```
ahadas, line 31: Sonarcloud: `Add a nested comment explaining why this function is empty or complete the implementation.`

ahadas (commented):
adding comments from sonarcloud

ahadas (approved):

PR #420, opened 2023-07-04
bump all dependencies

bennyz:
Includes human readable timestamp


pkg/lib/logging/logger.go
```diff
@@ -45,25 +46,81 @@ type Logger struct {
...
+	Trace(err error, kvpair ...interface{})
+}
+
+type levelLoggerImpl struct {
+	real  logr.Logger
+	level int
+}
+
+func (l *levelLoggerImpl) Info(msg string, kv ...interface{}) {
+	if Settings.allowed(l.level) {
```
ahadas, line 65: Info should always be logged, no?

pkg/lib/logging/logger.go
```diff
@@ -45,25 +46,81 @@ type Logger struct {
...
-	return l
+func (l *levelLoggerImpl) WithName(name string) LevelLogger {
+	return &levelLoggerImpl{
+		real:  l.real.WithName(name),
+		level: l.level,
+	}
+}
+
+func (l *levelLoggerImpl) Trace(err error, kvpair ...interface{}) {
+	l.real.Error(err, None, kvpair...)
```
ahadas, line 100: Trace messages are logged as Error? I'd expect the check of Settings.allowed to be here rather than on Info

ahadas:
@bennyz if you make changes in this PR, please push new commits - it's hard to review the changes with so many modifications...

pkg/lib/logging/logger.go
bennyz: I took the existing the implementation, theoretically someone could want to only log errors

pkg/lib/logging/logger.go
bennyz: perhaps, but this is the current implementation, and not only that, it's used extensively to log errors 😅

pkg/lib/logging/logger.go
ahadas: yeah, ok, if Trace is used for errors then maybe someone envisioned using Info for debug logs...

pkg/lib/logging/logger.go
ahadas: ok, if that's the current implementation.. where can I see that?

ahadas: I see, the comment for Trace is `// Logs an error without a description.` - so that's explain it.. anyway, if it keeps the existing behavior then it's fine

PR #432, opened 2023-07-13
Fix vmMap race

### LINKED ISSUE #281: VMs that are successfully migrated from openstack are shown with Error

bennyz:
r.vmMap may be stale if a previous VM has created it and the current VM didn't exist yet

Fixes #281

### DIFF
--- a/pkg/controller/plan/migration.go
+++ b/pkg/controller/plan/migration.go
@@ -1309,9 +1309,17 @@ func (r *Migration) setRunning(vm *plan.VMStatus, running bool) (err error) {
 	var vmCr VirtualMachine
 	found := false
 	if vmCr, found = r.vmMap[vm.ID]; !found {
-		msg := "VirtualMachine CR not found."
-		vm.AddError(msg)
-		return
+		// Recreate the map and check again, the map may be stale
+		r.vmMap, err = r.kubevirt.VirtualMachineMap()
+		if err != nil {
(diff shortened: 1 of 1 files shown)


ahadas (commented):
it looks good but why do we actually need to cache it?

bennyz:
> it looks good but why do we actually need to cache it?

that is a good question, I didn't want to rattle the cage too much :)

ahadas:
> > it looks good but why do we actually need to cache it?
> 
> that is a good question, I didn't want to rattle the cage too much :)

yeah but I just think that there's no end to that really - what if the vm exists in the cache but in an incorrect 'running' state?

ahadas (approved):
ok, so not getting the vm is significant because we return an error but if we get the vm in down state while it's up then the implication is that we trigger a call to update the vm - that's insignificant
it would be better though to post the VM with the desired 'Running' state instead now that we post the VM at the end of the migration process

PR #433, opened 2023-07-13
Push ova-provider-server to quay.io

ahadas:


liranr23:
where is the image? or you manually set it in kubev2v?

PR #434, opened 2023-07-13
openstack: fix project sync for new authentication methods

mmartinv:
The project list will retrieve user projects only as
users usually lack permissions for the 'list_projects'
operation in nova.

Currently, we only allow migrations from the project
associated with the session/token/applicationcredential
so it makes no sense to sync other projects

The current implementation takes the project name from
the secret. This property maybe missing if token or
applicationcredentials authentication is used so we
need to figure out the project depending on the
authentication type and which fields are required
in each case.


vendor/modules.txt
```diff
@@ -99,6 +99,7 @@ github.com/gophercloud/gophercloud/openstack/compute/v2/flavors
 github.com/gophercloud/gophercloud/openstack/compute/v2/servers
 github.com/gophercloud/gophercloud/openstack/identity/v2/tenants
 github.com/gophercloud/gophercloud/openstack/identity/v2/tokens
+github.com/gophercloud/gophercloud/openstack/identity/v3/applicationcredentials
```
ahadas, line 102: how come that this has changed without changing go.mod?

mmartinv: It was already in go.mod after the new authentication types PR.

ahadas: Ack

ahadas (approved):

PR #448, opened 2023-07-18
fix NPE when libmodel.Error is received

bennyz:
It seems that in
  `pkg/controller/watch/handler/handler.go`

```go
  func (r *Handler) Error(w *libweb.Watch, err error) {
	  log.Info(
		  "event: error.",
		  "id",
		  r.id,
		  "error",
		  err.Error())
	  if !r.ended {
		  _ = w.Repair()
	  }
  }
```

err will always be nil if we get there from:

```go
  case libmodel.Error:
    r.handler.Error(&Watch{reader: r}, nil)
```

### DIFF
--- a/pkg/lib/inventory/web/client.go
+++ b/pkg/lib/inventory/web/client.go
@@ -3,6 +3,7 @@ package web
 import (
 	"bytes"
 	"encoding/json"
+	"errors"
 	"io/ioutil"
 	"net/http"
 	liburl "net/url"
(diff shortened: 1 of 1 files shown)


ahadas (approved):

ahadas (commented):
@bennyz so how would the log be, something like "event: error. id=... error=Error event received"?
maybe we should rather say that the error is unknown or libmodel.Error?

ahadas (approved):
need to resolve conflicts

PR #459, opened 2023-07-24
Adding OVA migration flow

bkhizgiy:
Adding OVA flow for migration


pkg/controller/plan/adapter/ova/builder.go
```diff
@@ -0,0 +1,594 @@
...
+				Name:  "V2V_provider",
+				Value: "ovf",
+			})
+	}
+	return
+}
+
+// Build the DataVolume credential secret.
+func (r *Builder) Secret(vmRef ref.Ref, in, object *core.Secret) (err error) {
+	// TODO only if we want to save some data for DV.
```
ahadas, line 173: what does it mean? if we need an empty volume to store information on, why do we need a secret?

pkg/controller/plan/adapter/ova/builder.go
```diff
@@ -0,0 +1,594 @@
...
+	if err != nil {
+		err = liberr.Wrap(
+			err,
+			"VM lookup failed.",
+			"vm",
+			vmRef.String())
+		return
+	}
+
+	if r.Plan.Spec.Warm {
```
ahadas, line 259: this is too late to discover this - we should validate it before

pkg/controller/plan/adapter/ova/builder.go
```diff
@@ -0,0 +1,594 @@
...
+			}
+		}
+		if len(needed) == 0 {
+			continue
+		}
+		for _, nic := range needed {
+			networkName := fmt.Sprintf("net-%v", numNetworks)
+			numNetworks++
+			if len(nic.MAC) == 0 {
+				nic.MAC = generateRandomMACAddress()
```
ahadas, line 324: do we have to generate one? (doesn't CNV generate MAC address when it's not specified?)

pkg/controller/plan/adapter/ova/builder.go
```diff
@@ -0,0 +1,594 @@
...
+//
+//	Example:
+//	Input: 	[datastore13] my-vm/disk-name-000015.vmdk
+//	Output: [datastore13] my-vm/disk-name.vmdk
+func trimBackingFileName(fileName string) string {
+	return backingFilePattern.ReplaceAllString(fileName, ".vmdk")
+}
+
+func getDiskSourcePath(filePath string, diskName string) string {
+
```
ahadas, line 533: please remove these empty lines here and below

pkg/controller/plan/adapter/ova/builder.go
```diff
@@ -0,0 +1,594 @@
...
+		env = append(env,
+			core.EnvVar{
+				Name:  "V2V_diskPath",
+				Value: vm.OvaPath,
+			},
+			core.EnvVar{
+				Name:  "V2V_provider",
+				Value: "ova",
+			})
+	} else {
```
ahadas, line 154: I think it should still be with `-i ova` and pointing to the folder, did you check this? (CC @nyoxi )

liranr23: it's just provided to the virt-v2v pod. based on that it uses later the `-i ova` in the command.

virt-v2v/cold/entrypoint
```diff
@@ -53,26 +72,44 @@ for disk in /dev/block[0-9]* ; do
...
+      "${args[@]}"\
+      |& /usr/local/bin/virt-v2v-monitor
+fi
+
+if [ "$V2V_provider" == "ovf" ] ; then
+  exec virt-v2v -v -x \
+      -i disk "$V2V_diskPath" \
+      "${args[@]}" -of raw\
+      |& /usr/local/bin/virt-v2v-monitor
+fi
```
liranr23, line 115: drop

pkg/controller/plan/adapter/ova/builder.go
liranr23: we don't. this comment should be removed.

pkg/controller/provider/container/ova/collector.go
```diff
@@ -304,25 +304,25 @@ func (r *Collector) endWatch() {
...
+	// 	}
+	// 	err = r.apply(updates)
+	// 	if err != nil {
+	// 		return
+	// 	}
+	// }
+	// r.log.Info(
+	// 	"Refresh finished.",
+	// 	"duration",
+	// 	time.Since(mark))
```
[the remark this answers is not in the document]
liranr23: the problem i had with it (unless it changed) - that it always updates without any real reason. if we check that something changes - we should have it.

cmd/ova-provider-server/ova-provider-server.go
```diff
@@ -346,20 +368,32 @@ func convertToVmStruct(envelope []Envelope, ovaPath []string) ([]VM, error) {
 			newVM := VM{
 				OvaPath: ovaPath[i],
 				Name:    virtualSystem.Name,
+				UUID:    vmIDMap.GetUUID(ovaPath[i]),
```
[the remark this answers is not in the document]
bkhizgiy: the inventory store this under ID filed, this is only for the server

pkg/controller/plan/kubevirt.go
```diff
@@ -1452,6 +1453,25 @@ func (r *KubeVirt) podVolumeMounts(vmVolumes []cnv.Volume, configMap *core.Confi
...
+	//path from disk
+	volumes = append(volumes, core.Volume{
+		Name: "nfs",
+		VolumeSource: core.VolumeSource{
+			NFS: &core.NFSVolumeSource{
+				Server: nfsServer,
+				Path:   nfsPath,
+			},
+		},
+	})
```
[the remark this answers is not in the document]
bkhizgiy: yes, this code extracts the nfs server and path for the volume mount

pkg/controller/provider/container/ova/collector.go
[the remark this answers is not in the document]
bkhizgiy: currently we do not implement watch properly, but we do need this code (or else no changes will be saved at all) we can reduce the refresh rate for now

pkg/controller/plan/adapter/ova/builder.go
bkhizgiy: @liranr23 doesn't it already verified as part of the migration steps in kubevirt.go? do we need to add a separate validation for OVA?

liranr23: the validator can do the validation if we didn't add it there already (provider == ova and warm).

liranr23: looks like we are doing it in the plan validation (`pkg/controller/plan/adapter/ova/validator.go`):
```// Validate whether warm migration is supported from this provider type.
func (r *Validator) WarmMigration() (ok bool) {
	ok = false
	return
}
```
so we don't need it in the builder.

cmd/ova-provider-server/ova-provider-server.go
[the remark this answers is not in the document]
ahadas: > the inventory store this under ID filed, this is only for the server

ah ok

cmd/ova-provider-server/ova-provider-server.go
```diff
@@ -453,3 +500,37 @@ func getDiskPath(path string) string {
...
+		uuid[6] = (uuid[6] & 0x0f) | 0x40
+		uuid[8] = (uuid[8] & 0x3f) | 0x80
+		id = fmt.Sprintf("%x-%x-%x-%x-%x", uuid[0:4], uuid[4:6], uuid[6:8], uuid[8:10], uuid[10:])
+
+		um.m[filename] = id
+	}
+	return id
+}
+
+func isValidUUID(id string) bool {
```
ahadas, line 533: did you look into using github.com/google/uuid for validating and generating UIDs?

pkg/apis/forklift/v1beta1/plan.go
```diff
@@ -101,8 +101,8 @@ func (p *Plan) VSphereUsesEl9VirtV2v() (bool, error) {
 		return false, liberr.New("Cannot analyze plan, destination provider is missing.")
 	}
 
-	if source.Type() != VSphere {
-		return false, nil
+	if source.Type() == VSphere || source.Type() == Ova {
```
ahadas, line 104: the first part is problematic - if `source.Type() == VSphere` then we need to check `!p.Spec.Warm && destination.IsHost()`

virt-v2v/cold/entrypoint
```diff
@@ -2,15 +2,34 @@
 set -o pipefail
 shopt -s nullglob
 
-if [ -z "$V2V_libvirtURL" ] || \
-    [ -z "$V2V_secretKey" ] || \
-    [ -z "$V2V_vmName" ] ; then
+if [ -z "$V2V_provider"] ; then
```
ahadas, line 5: need to change it also here and below to V2V_source

pkg/controller/plan/kubevirt.go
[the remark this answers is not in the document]
ahadas: so we should check that the source is OVA then so we won't execute this when using other sources, no?

pkg/controller/provider/container/ova/collector.go
```diff
@@ -21,7 +21,7 @@ const (
 	// Retry interval.
 	RetryInterval = 5 * time.Second
 	// Refresh interval.
-	RefreshInterval = 10 * time.Second
+	RefreshInterval = 30 * time.Second
```
ahadas, line 24: so are we going to update the inventory every 30 seconds?

pkg/controller/plan/adapter/ova/builder.go
```diff
@@ -0,0 +1,547 @@
...
+		core.EnvVar{
+			Name:  "V2V_vmName",
+			Value: vm.Name,
+		},
+		core.EnvVar{
+			Name:  "V2V_diskPath",
+			Value: getDiskSourcePath(vm.OvaPath),
+		},
+		core.EnvVar{
+			Name:  "V2V_provider",
```
ahadas, line 144: need to change it here to V2V_source as well

virt-v2v/cold/entrypoint
```diff
@@ -53,26 +72,39 @@ for disk in /dev/block[0-9]* ; do
...
+fi
+
+if [ "$V2V_provider" == "ova" ] ; then
+  exec virt-v2v -v -x \
+      -i ova "$V2V_diskPath" \
+      "${args[@]}"\
+      |& /usr/local/bin/virt-v2v-monitor
+fi
+
+qemu-img info /mnt/disks/disk0/disk.img
```
ahadas, line 110: added for debugging? if so, we should probably execute this only in debug mode

pkg/apis/forklift/v1beta1/plan.go
ahadas: @bkhizgiy I posted https://github.com/kubev2v/forklift/pull/463 to ease extending this function, please take a look

bkhizgiy: Just saw, looks better in as a switch case

pkg/controller/provider/container/ova/collector.go
bkhizgiy: yes, until the watch is fully supported, do you think we need to set a smaller interval?

pkg/controller/plan/kubevirt.go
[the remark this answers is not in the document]
liranr23: we enter here only when using conversion pod (v2v). but yes, i think we need to split the cases here as vsphere won't have any nfs path or whatsoever, while ova won't need/have the vddk and the domain xml.

cmd/ova-provider-server/ova-provider-server.go
bkhizgiy: yes, I've initially used it, but it didn't work in the pod, probably need to add download in the container file i"ll look into that again

pkg/controller/provider/container/ova/collector.go
ahadas: yeah, the collection of OVAs is expected to be relatively stable - I'd change the frequency to execute it every 5 minutes even

cmd/ova-provider-server/ova-provider-server.go
bkhizgiy: I just remembered that we were planning to use the vm hash as an ID (for consistency if the server will crush), changing in the code

cmd/ova-provider-server/ova-provider-server.go
```diff
@@ -51,12 +64,12 @@ type DiskSection struct {
 
 type Disk struct {
 	XMLName                 xml.Name `xml:"Disk"`
-	Capacity                string   `xml:"capacity,attr"`
+	Capacity                int64    `xml:"capacity,attr"`
```
ahadas, line 67: shouldn't it be uint32?

cmd/ova-provider-server/ova-provider-server.go
```diff
@@ -378,14 +420,18 @@ func convertToVmStruct(envelope []Envelope, ovaPath []string) ([]VM, error) {
...
+
 			}
 
-			for _, network := range vmXml.NetworkSection.Networks {
+			for j, network := range vmXml.NetworkSection.Networks {
 				newVM.Networks = append(newVM.Networks, VmNetwork{
 					Name:        network.Name,
 					Description: network.Description,
 				})
+				newVM.Networks[j].ID = networkIDMap.GetUUID(newVM.Networks[j].Name, network.Name)
```
ahadas, line 434: this is a bit confusing - we pass network.Name twice then, right? it would be more clear with something like:
```golang
                                newVM.Networks = append(newVM.Networks, VmNetwork{
					Name:        network.Name,
					Description: network.Description,
                                        ID:          networkIDMap.GetUUID(network.Name, network.Name)
				})
```

cmd/ova-provider-server/ova-provider-server.go
ahadas: ok, that applies to generating the UUID but we can still validate UIDs with uuid.Parse

pkg/controller/plan/kubevirt.go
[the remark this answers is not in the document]
ahadas: > we enter here only when using conversion pod (v2v). but yes, i think we need to split the cases here as vsphere won't have any nfs path or whatsoever, while ova won't need/have the vddk and the domain xml.

also ESX won't need it in the future

virt-v2v/cold/entrypoint
liranr23: and if so it should be within 2 loops (for block and file), like few lines above:
`for disk in /mnt/disks/disk[0-9]* ; do` ... `for disk in /dev/block[0-9]* ; do`
i think it might be nice to have even without being on debug.

bkhizgiy: yes It's for debuging but I also think it's nice to have, will be easier to get the status in the end

cmd/ova-provider-server/ova-provider-server.go
bkhizgiy: yes for the network specific its redundant since we using only the name and not the entire struct.

virt-v2v/cold/entrypoint
ahadas: once we see it works, the output of this command will no longer be useful as the output from virt-v2v indicates whether the conversion completed successfully or not - the output of `qemu-img info` would then only indicate whether the data was written to the file we intended to write to.. it doesn't worth keeping it in production

cmd/ova-provider-server/ova-provider-server.go
```diff
@@ -378,14 +421,18 @@ func convertToVmStruct(envelope []Envelope, ovaPath []string) ([]VM, error) {
...
+
 			}
 
-			for _, network := range vmXml.NetworkSection.Networks {
+			for j, network := range vmXml.NetworkSection.Networks {
 				newVM.Networks = append(newVM.Networks, VmNetwork{
 					Name:        network.Name,
 					Description: network.Description,
 				})
+				newVM.Networks[j].ID = networkIDMap.GetUUID(network.Name, network.Name)
```
ahadas, line 435: ok, that's better but then there's really no need for `j` - we can set the ID above like the Name and the Description

cmd/ova-provider-server/ova-provider-server.go
```diff
@@ -413,6 +469,7 @@ func convertToNetworkStruct(envelope []Envelope) ([]VmNetwork, error) {
 				Name:        network.Name,
 				Description: network.Description,
 			}
+			newNetwork.ID = networkIDMap.GetUUID(newNetwork.Name, network.Name)
```
ahadas, line 472: and also here

pkg/controller/provider/container/ova/collector.go
ahadas: @bkhizgiy I think you missed this thread

bkhizgiy: I did:) changing

ahadas (commented):
@bkhizgiy the changes look good now but please take a look at [the code-smell issues that are reported for this PR](https://sonarcloud.io/project/issues?id=kubev2v_forklift&pullRequest=459&resolved=false&types=CODE_SMELL), specifically the "Define a constant instead of duplicating this literal" ones - the more things (that we're not going to backport) we solve now, the better as future backports would be easier

ahadas (approved):
dropping unused DefaultWindows and DefaultLinux - good!

PR #460, opened 2023-07-25
Add compatibility ID field for OCP base resource type

yaacov:
Add compatibility ID field for OCP base resource type

Why this is needed:
Users of forklift API (for example the UI) expect an "id" json field and not a "uid" field.

### DIFF
--- a/pkg/controller/provider/web/ocp/resource.go
+++ b/pkg/controller/provider/web/ocp/resource.go
@@ -16,8 +16,11 @@ type Resource struct {
 	Name string `json:"name"`
 	// self link.
 	SelfLink string `json:"selfLink"`
-
+	// self path.
 	Path string `json:"path,omitempty"`
+
+	// forklift ID, for compatability with providers using ID instead of UID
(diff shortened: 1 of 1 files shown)


pkg/controller/provider/web/ocp/resource.go
```diff
@@ -16,8 +16,11 @@ type Resource struct {
 	Name string `json:"name"`
 	// self link.
 	SelfLink string `json:"selfLink"`
-
+	// self path.
 	Path string `json:"path,omitempty"`
+
+	// forklift ID, for compatability with providers using ID instead of UID
+	ID string `json:"id"`
```
bennyz, line 23:
```suggestion
	ID string `json:"id,omitempty"`
```

yaacov: fixed

bennyz (approved):

bennyz (approved):

ahadas:
@yaacov @bennyz so how about deprecating the UID field?

PR #467, opened 2023-07-30
openstack: use bazel to build image

bennyz:


ahadas (approved):
looks good, and drop `hack/openstack-populator/Containerfile`?

ahadas (approved):

PR #471, opened 2023-07-30
bazelify ovirt populator

bennyz:


cmd/ovirt-populator/deps.sh
ahadas, line 1: I realize it's a WIP but as an early feedback - this script should move to the `hack` folder

cmd/ovirt-populator/BUILD.bazel
```diff
@@ -28,13 +19,171 @@ go_library(
...
+
+container_run_and_commit_layer(
+    name = "ovirt-imageio-layer-run",
+    commands = [
+        "update-ca-trust",
+        "pip install setuptools wheel --user",
+        "pip install ovirt-engine-sdk-python ovirt-imageio --user",
+        "cp /root/.local/bin/ovirt-img /usr/local/bin/ovirt-img",
+        "cp -r /root/.local/lib/python3.9/site-packages/ /usr/lib/python3.9/",
+        "chown 107:107 /usr/lib/python3.9/site-packages/ovirt_imageio",
```
bennyz, line 52: this is not the way to go, just checking if CI works

WORKSPACE
ahadas, line 1: in order to avoid frequent changes to this file we've introduce other ones (like [this](https://github.com/kubev2v/forklift/blob/main/virt-v2v/cold/WORKSPACE)) for virt-v2v - I think we better do the same also in this case

ahadas (commented):
using dockerfile_image made it look like a bazel target and these changes take us closer to the standard way we create the images using bazel - that's good. it also means we'll change the base image of ovirt's volume-populator from centos steam 9 to ubi 9 which is fine. so overall it looks good to me

ahadas (approved):

PR #480, opened 2023-08-01
Fix LUN warm migration

liranr23:
When migrating a LUN disk from ovirt source in warm migration we need to mark the CopyDisk as completed.

### DIFF
--- a/pkg/apis/forklift/v1beta1/plan/migration.go
+++ b/pkg/apis/forklift/v1beta1/plan/migration.go
@@ -104,9 +104,6 @@ func (r *Step) ReflectTasks() {
 	tasksStarted := 0
 	tasksCompleted := 0
 	completed := int64(0)
-	if len(r.Tasks) == 0 {
-		return
-	}
 	for _, task := range r.Tasks {
 		if task.MarkedStarted() {
 			tasksStarted++


bennyz (approved):

pkg/apis/forklift/v1beta1/plan/migration.go
```diff
@@ -105,6 +105,7 @@ func (r *Step) ReflectTasks() {
 	tasksCompleted := 0
 	completed := int64(0)
 	if len(r.Tasks) == 0 {
+		r.MarkCompleted()
```
ahadas, line 108: can it be that we'll have a step before the tasks are created?
if seems suspicious that we have this if-statement here - it could have been dropped altogether if the intention was to mark the step as completed, no?

liranr23: yes it does. i'll check about having step before tasks. but in my opinion the idea was to not allow a case there are no tasks (no disks).

liranr23: The creation of tasks happens only when entering:
`AllocateDisks, CopyDisks, CopyDisksVirtV2V and Finalize`. In terms of when we enter `ReflectTasks()`, it's when updating the conversion/copy progress. therefore i don't see much risk with this change.

ahadas: I see - so tasks like initialization and hooks won't get it, ok
so how about dropping this if-statement then?

liranr23: done

ahadas (approved):
and we got lower cognitive complexity!

liranr23:
> and we got lower cognitive complexity!

i thought of it as well :laughing:

PR #483, opened 2023-08-02
ovirt client: Add retry for creating of snapshot 

mnecas:
Issue:
If there is some operation on VMs snapshots (creation, deletion). The migration fails because of the locked disks, and the migration ends.

Fix:
Add a retry in the form of a different exception, `ProviderNotReadyError`, which does not end the migration but reconciles. 
There was another issue during this fix. Every try created another failed job with the same correlation ID which failed the migration. The fix for this was to use events and check the proper event codes.

### DIFF
--- a/pkg/controller/plan/adapter/ovirt/client.go
+++ b/pkg/controller/plan/adapter/ovirt/client.go
@@ -482,8 +482,15 @@ func (r Client) removePrecopies(precopies []planapi.Precopy, vmService *ovirtsdk
 			default:
 				time.Sleep(time.Duration(settings.Settings.Migration.SnapshotStatusCheckRate) * time.Second)
 			}
+
+			_, err := snapService.Get().Send()
+			if err != nil {
+				r.Log.Info("The snapshot was removed", "snapshotID", snapshotID)
+				break
+			}
+
 			// Try to remove the snapshot
-			_, err := snapService.Remove().Query("correlation_id", correlationID).Send()
+			_, err = snapService.Remove().Query("correlation_id", correlationID).Send()
 			if err != nil {
 				if strings.Contains(err.Error(), "Cannot remove Snapshot") {
 					err = web.ConflictError{
@@ -494,26 +501,6 @@ func (r Client) removePrecopies(precopies []planapi.Precopy, vmService *ovirtsdk
 				} else {
 					err = liberr.Wrap(err)
 					r.Log.Error(err, "Request to remove snapshot failed", "snapshotID", snapshotID)
-
-					snapshotDisksResponse, err := snapService.DisksService().List().Send()
-					if err != nil {
-						r.Log.Error(err, "Request for snapshotDisksResponse failed")
-						break
-					}
-					snapshotDisks, found := snapshotDisksResponse.Disks()
-					if !found {
-						r.Log.Error(err, "Disks not found")
-						continue
-					}
-					for _, disk := range snapshotDisks.Slice() {
-						if transferred, err := r.isDiskBeingTransferred(disk.MustId()); transferred {
-							r.Log.Info("Disk is being transferred, retrying...", "disk", disk)
-							continue
-						} else if err != nil {
-							err = liberr.Wrap(err)
-							r.Log.Error(err, "Error checking if disk is being transferred")
-						}
-					}
 				}
 			} else {
 				// Check the events of the snapshot
--- a/pkg/controller/plan/adapter/ovirt/client.go
+++ b/pkg/controller/plan/adapter/ovirt/client.go
@@ -17,7 +17,6 @@ import (
 	liberr "github.com/konveyor/forklift-controller/pkg/lib/error"
 	"github.com/konveyor/forklift-controller/pkg/settings"
 	ovirtsdk "github.com/ovirt/go-ovirt"
-	"k8s.io/apimachinery/pkg/util/wait"
 	cdi "kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1"
 )
 
@@ -475,69 +474,7 @@ func (r Client) removePrecopies(precopies []planapi.Precopy, vmService *ovirtsdk
 		snapshotID := precopies[i].Snapshot
 		snapService := snapsService.SnapshotService(snapshotID)
 		correlationID := fmt.Sprintf("%s_finalize", snapshotID[0:8])
-		_, err := snapService.Remove().Query("correlation_id", correlationID).Send()
-		if err != nil {
-			// Assume we could not remove because of an ongoing transfer
-			r.Log.Error(err, "Remove request for snapshot failed", "snapshotID", snapshotID)
-			backoff := wait.Backoff{
-				Duration: time.Second * 3,
-				Factor:   0,
-				Jitter:   0,
-				Steps:    3,
-			}
-
-			err = wait.ExponentialBackoff(backoff, func() (bool, error) {
-				snapshotDisksResponse, err := snapService.DisksService().List().Send()
-				if err != nil {
-					r.Log.Error(err, "snapshotDiksService Request failed")
-					err = liberr.Wrap(err, "Disks not found")
-					return false, err
-				}
-				snapshotDisks, found := snapshotDisksResponse.Disks()
-				if !found {
-					r.Log.Error(err, "disks not found")
-					err = liberr.Wrap(err, "Disks not found")
-					return false, err
-				}
-				for _, disk := range snapshotDisks.Slice() {
-					if transferred, err := r.isDiskBeingTransferred(disk.MustId()); transferred {
-						r.Log.Info("Disk is being transferred, retrying...")
-						return false, nil
-					} else if err != nil {
-						err = liberr.Wrap(err)
-						r.Log.Error(err, "Error checking if disk is being transferred")
-						return false, nil
-					} else {
-						// Try to remove the snapshot again
-						_, err = snapService.Remove().Query("correlation_id", correlationID).Send()
-						if err != nil {
-							err = liberr.Wrap(err)
-							r.Log.Error(err, "Error removing snapshot")
-							return true, err
-						}
-					}
-				}
-				return false, nil
-			})
-			if err != nil {
-				err = liberr.Wrap(err)
-				r.Log.Error(err, "Error waiting for snapshot removal")
-				return
-			}
-
-		}
-
 		for {
-			events, err := r.getEvents(correlationID)
-			if err != nil {
-				err = liberr.Wrap(err)
-				r.Log.Error(err, "Error waiting for snapshot removal")
-				break
-			}
-			if r.isSnapshotRemovalFinished(events) {
-				break
-			}
-
 			select {
 			case <-time.After(time.Duration(settings.Settings.Migration.SnapshotRemovalTimeout) * time.Minute):
 				r.Log.Info("Timeout waiting for snapshot removal")
@@ -545,21 +482,69 @@ func (r Client) removePrecopies(precopies []planapi.Precopy, vmService *ovirtsdk
 			default:
 				time.Sleep(time.Duration(settings.Settings.Migration.SnapshotStatusCheckRate) * time.Second)
 			}
+			// Try to remove the snapshot
+			_, err := snapService.Remove().Query("correlation_id", correlationID).Send()
+			if err != nil {
+				if strings.Contains(err.Error(), "Cannot remove Snapshot") {
+					err = web.ConflictError{
+						Provider: r.Source.Provider,
+						Err:      err,
+					}
+					r.Log.Error(err, "ConflictError failed to remove snapshot", "snapshotID", snapshotID)
+				} else {
+					err = liberr.Wrap(err)
+					r.Log.Error(err, "Request to remove snapshot failed", "snapshotID", snapshotID)
+
+					snapshotDisksResponse, err := snapService.DisksService().List().Send()
+					if err != nil {
+						r.Log.Error(err, "Request for snapshotDisksResponse failed")
+						break
+					}
+					snapshotDisks, found := snapshotDisksResponse.Disks()
+					if !found {
+						r.Log.Error(err, "Disks not found")
+						continue
+					}
+					for _, disk := range snapshotDisks.Slice() {
+						if transferred, err := r.isDiskBeingTransferred(disk.MustId()); transferred {
+							r.Log.Info("Disk is being transferred, retrying...", "disk", disk)
+							continue
+						} else if err != nil {
+							err = liberr.Wrap(err)
+							r.Log.Error(err, "Error checking if disk is being transferred")
+						}
+					}
+				}
+			} else {
+				// Check the events of the snapshot
+				finished, err := r.isSnapshotRemovalFinished(correlationID)
+				if err != nil {
+					err = liberr.Wrap(err)
+					r.Log.Error(err, "Error gathering events about the snapshot removal")
+				}
+				if finished {
+					break
+				}
+			}
 		}
 	}
 }
 
-func (r Client) isSnapshotRemovalFinished(events []*ovirtsdk.Event) bool {
+func (r *Client) isSnapshotRemovalFinished(correlationID string) (finished bool, err error) {
+	events, err := r.getEvents(correlationID)
+	if err != nil {
+		return
+	}
 	for _, event := range events {
 		switch event.MustCode() {
 		case REMOVE_SNAPSHOT_FINISHED_FAILURE:
 			r.Log.Info("Snapshot removal failed!")
-			return true
+			return true, nil
 		case REMOVE_SNAPSHOT_FINISHED_SUCCESS:
-			return true
+			return true, nil
 		}
 	}
-	return false
+	return
 }
 
 func (r *Client) DetachDisks(vmRef ref.Ref) (err error) {
--- a/pkg/controller/plan/adapter/ovirt/client.go
+++ b/pkg/controller/plan/adapter/ovirt/client.go
@@ -34,10 +34,8 @@ const (
 
 // Snapshot event codes
 const (
-	CREATE_SNAPSHOT                  int64 = 45
 	SNAPSHOT_FINISHED_SUCCESS        int64 = 68
 	SNAPSHOT_FINISHED_FAILURE        int64 = 69
-	REMOVE_SNAPSHOT                  int64 = 342
 	REMOVE_SNAPSHOT_FINISHED_SUCCESS int64 = 356
 	REMOVE_SNAPSHOT_FINISHED_FAILURE int64 = 357
 )
@@ -104,14 +102,10 @@ func (r *Client) CheckSnapshotReady(vmRef ref.Ref, snapshot string) (ready bool,
 		err = liberr.New("No event found for correlation ID", "correlationID", correlationID)
 		return
 	}
-	ready = true
 	for _, event := range events {
 		switch event.MustCode() {
 		case SNAPSHOT_FINISHED_FAILURE:
 			err = liberr.New("Snapshot creation failed!", "correlationID", correlationID)
-			ready = false
-		case CREATE_SNAPSHOT:
-			ready = false
 		case SNAPSHOT_FINISHED_SUCCESS:
 			ready = true
 		}
@@ -540,7 +534,7 @@ func (r Client) removePrecopies(precopies []planapi.Precopy, vmService *ovirtsdk
 				r.Log.Error(err, "Error waiting for snapshot removal")
 				break
 			}
-			if eventFinishedSuccesfully(events) {
+			if r.isSnapshotRemovalFinished(events) {
 				break
 			}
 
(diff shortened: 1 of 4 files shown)


pkg/controller/plan/adapter/ovirt/client.go
```diff
@@ -49,6 +50,19 @@ func (r *Client) CreateSnapshot(vmRef ref.Ref) (snapshot string, err error) {
 		return
 	}
 	snapsService := vmService.SnapshotsService()
+
+	locked, err := r.isAnySnapshotLocked(snapsService)
```
ahadas, line 54: yeah, that's a possible approach - but wouldn't it make more sense to initiate the operation and try to address the case that the operation failed because of a lock more broadly so we'll also over the case that the vm is, e.g., being migrated?

pkg/controller/plan/adapter/ovirt/client.go
```diff
@@ -49,6 +50,19 @@ func (r *Client) CreateSnapshot(vmRef ref.Ref) (snapshot string, err error) {
...
+	if err != nil {
+		err = liberr.Wrap(err)
+		return
+	}
+	if locked {
+		err = web.ProviderNotReadyError{
+			Provider: r.Source.Provider,
+		}
+		return
+	}
```
ahadas, line 64: where is the 3-times limit that is mentioned in the description of the PR?

mnecas: not 3-times but every 3s, and that's the default for the reconcile 
https://github.com/kubev2v/forklift/blob/42043cca14acbf58268eb78020722b795c853da4/pkg/controller/plan/migration.go#L33

pkg/controller/plan/adapter/ovirt/client.go
mnecas: I'll check it out but afaik even when the VM is being migrated it does not create the snapshot at the same plus it looks like the oVirt engine allows just one operation on the VM snapshots at once.

pkg/controller/plan/adapter/ovirt/client.go
ahadas: ah, three seconds - ok, that makes more sense :)

pkg/controller/plan/adapter/ovirt/client.go
ahadas: that's right
the thing is that originally oVirt used locks at the database level, like that status of snapshots, and in-memory locks where used for a very limited time - they were acquired when a command started, validated and then they were released when the database-level locks were set. for quite a few years oVirt uses in-memory lock differently - for the whole duration of the command execution. we  still kept the database-level locks (for various reasons - presentation in the UI, tools that query the database directly, etc) but in theory in most-flows they are pretty much redundant

pkg/controller/plan/adapter/ovirt/client.go
```diff
@@ -56,7 +68,13 @@ func (r *Client) CreateSnapshot(vmRef ref.Ref) (snapshot string, err error) {
 			MustBuild(),
 	).Query("correlation_id", correlationID).Send()
 	if err != nil {
-		err = liberr.Wrap(err)
+		if strings.Contains(err.Error(), "Cannot create Snapshot") {
+			err = web.ProviderNotReadyError{
```
ahadas, line 72: ProviderNotReadyError is a bit confusing because we have the ready state - and the provider could be ready in this scenario. I'd rather call it "ConflictError" - something like we had in oVirt that differentiated invalid input from "bad-timing"

pkg/controller/plan/adapter/ovirt/client.go
```diff
@@ -76,22 +94,22 @@ func (r *Client) CheckSnapshotReady(vmRef ref.Ref, snapshot string) (ready bool,
 		err = liberr.Wrap(err)
 		return
 	}
-	jobs, err := r.getJobs(correlationID)
+	events, err := r.getEvents(correlationID)
```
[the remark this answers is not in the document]
mnecas: The events from previous attempts are not created only jobs.

pkg/controller/plan/adapter/ovirt/client.go
mnecas: Ack, I wanted to use the ProviderNotReadyError because it's used a lot in the plan/migration.go I just need to add another error type here
https://github.com/kubev2v/forklift/blob/b8471caa9374fee5a1eddebbd39491781e257e84/pkg/controller/plan/migration.go#L808-L816

pkg/controller/plan/adapter/ovirt/client.go
```diff
@@ -534,15 +552,14 @@ func (r Client) removePrecopies(precopies []planapi.Precopy, vmService *ovirtsdk
...
 
-func allJobsFinished(jobs []*ovirtsdk.Job) bool {
-	for _, job := range jobs {
-		status, _ := job.Status()
-		if status != ovirtsdk.JOBSTATUS_FINISHED && status != ovirtsdk.JOBSTATUS_FAILED {
-			return false
+func eventFinishedSuccesfully(events []*ovirtsdk.Event) bool {
+	for _, event := range events {
+		code := event.MustCode()
+		if code == SNAPSHOT_FINISHED_SUCCESS {
```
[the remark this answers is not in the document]
mnecas: ahh yeah my mistake

pkg/controller/plan/adapter/ovirt/client.go
```diff
@@ -76,22 +94,22 @@ func (r *Client) CheckSnapshotReady(vmRef ref.Ref, snapshot string) (ready bool,
...
-	for _, job := range jobs {
-		switch job.MustStatus() {
-		case ovirtsdk.JOBSTATUS_FAILED, ovirtsdk.JOBSTATUS_ABORTED:
+	for _, event := range events {
+		switch event.MustCode() {
+		case SNAPSHOT_FINISHED_FAILURE:
 			err = liberr.New("Snapshot creation failed!", "correlationID", correlationID)
 			ready = false
-		case ovirtsdk.JOBSTATUS_STARTED, ovirtsdk.JOBSTATUS_UNKNOWN:
+		case CREATE_SNAPSHOT:
```
ahadas, line 112: is this logic correct - with jobs, a job has a single status - meaning that a job with status `JOBSTATUS_STARTED` means there's an ongoing operation that didn't succeed, but here we can get an event of `CREATE_SNAPSHOT` followed by `SNAPSHOT_FINISHED_SUCCESS`, which means the operation succeeded and yet we'll set ready=false, no?

mnecas: Ahh good point! Added switch case for `SNAPSHOT_FINISHED_SUCCESS`. So if the `CREATE_SNAPSHOT` sets the ready to false and there are more events it will rewrite the ready to true

ahadas: are the events ordered?

mnecas: yes, from the go-ovirt client
```
// The following events occur:
// * id="1" - The API logs in the admin user account.
// * id="2" - The API logs out of the admin user account.
// The order of the returned list of events is always garanteed. If the `sortby` clause is included in the
// `search` parameter, then the events will be ordered according to that clause.
...
```

pkg/controller/plan/adapter/ovirt/client.go
```diff
@@ -56,7 +68,13 @@ func (r *Client) CreateSnapshot(vmRef ref.Ref) (snapshot string, err error) {
 			MustBuild(),
 	).Query("correlation_id", correlationID).Send()
 	if err != nil {
-		err = liberr.Wrap(err)
+		if strings.Contains(err.Error(), "Cannot create Snapshot") {
```
bennyz, line 71: posted https://github.com/oVirt/ovirt-engine-sdk-go/pull/240 
it would be useful to detect why we couldn't create the snapshot

mnecas: Added the error to the `ConflictError` so we could propagate it and print the error.

ahadas: and we can then drop the check of `strings.Contains(err.Error(), "Cannot create Snapshot")`, right?

ahadas (commented):
@mnecas please check the changes in the second commit
if the changes look good, I think we can get this PR in and @bennyz will follow up with another PR to leverage the change he posted to ovirt-go

pkg/controller/plan/adapter/ovirt/client.go
```diff
@@ -534,15 +549,17 @@ func (r Client) removePrecopies(precopies []planapi.Precopy, vmService *ovirtsdk
...
-	for _, job := range jobs {
-		status, _ := job.Status()
-		if status != ovirtsdk.JOBSTATUS_FINISHED && status != ovirtsdk.JOBSTATUS_FAILED {
-			return false
+func (r Client) isSnapshotRemovalFinished(events []*ovirtsdk.Event) bool {
+	for _, event := range events {
+		switch event.MustCode() {
+		case REMOVE_SNAPSHOT_FINISHED_FAILURE:
+			r.Log.Info("Snapshot removal failed!")
+			return true
```
ahadas, line 557: @bennyz pointed out that this logic is incorrect since the event might be correlated with a previous attempt.. so on a second thought, it would probably be better to switch back to checking jobs and say that:
1. if there's an ongoing job, wait for it to finish
2. otherwise, the operation succeeds if there's a job that finished successfully

pkg/controller/plan/adapter/ovirt/client.go
```diff
@@ -555,10 +549,13 @@ func (r Client) removePrecopies(precopies []planapi.Precopy, vmService *ovirtsdk
 	}
 }
 
-func eventFinishedSuccesfully(events []*ovirtsdk.Event) bool {
+func (r Client) isSnapshotRemovalFinished(events []*ovirtsdk.Event) bool {
```
liranr23, line 552:
```suggestion
func (r *Client) isSnapshotRemovalFinished(events []*ovirtsdk.Event) bool {
```

pkg/controller/plan/adapter/ovirt/client.go
```diff
@@ -555,10 +549,13 @@ func (r Client) removePrecopies(precopies []planapi.Precopy, vmService *ovirtsdk
...
 	for _, event := range events {
-		code := event.MustCode()
-		if code == REMOVE_SNAPSHOT_FINISHED_SUCCESS {
+		switch event.MustCode() {
```
liranr23, line 554: why not getting the code out and have error check instead of `MustCode()`?

mnecas: what do you mean by "getting code out"?

liranr23: code, err := event.Code()
if err != nil {
...
}
switch code {
...

pkg/controller/plan/adapter/ovirt/client.go
```diff
@@ -460,89 +474,76 @@ func (r Client) removePrecopies(precopies []planapi.Precopy, vmService *ovirtsdk
...
+						break
+					}
+					snapshotDisks, found := snapshotDisksResponse.Disks()
+					if !found {
+						r.Log.Error(err, "Disks not found")
+						continue
+					}
+					for _, disk := range snapshotDisks.Slice() {
+						if transferred, err := r.isDiskBeingTransferred(disk.MustId()); transferred {
+							r.Log.Info("Disk is being transferred, retrying...", "disk", disk)
```
ahadas, line 510: I think we won't get here anymore (need to check) but if we do then we need to return ConflictError to keep trying

pkg/controller/plan/adapter/ovirt/client.go
```diff
@@ -460,89 +474,76 @@ func (r Client) removePrecopies(precopies []planapi.Precopy, vmService *ovirtsdk
...
+							r.Log.Error(err, "Error checking if disk is being transferred")
+						}
+					}
+				}
+			} else {
+				// Check the events of the snapshot
+				finished, err := r.isSnapshotRemovalFinished(correlationID)
+				if err != nil {
+					err = liberr.Wrap(err)
+					r.Log.Error(err, "Error waiting for snapshot removal")
```
ahadas, line 522: this assumes that later the problem would be resolved but since we're in a loop here (and so even if the user cancels the migration we won't stop), maybe we should rather stop in this case

pkg/controller/plan/adapter/ovirt/client.go
```diff
@@ -460,89 +474,76 @@ func (r Client) removePrecopies(precopies []planapi.Precopy, vmService *ovirtsdk
...
-		if status != ovirtsdk.JOBSTATUS_FINISHED && status != ovirtsdk.JOBSTATUS_FAILED {
-			return false
+func (r *Client) isSnapshotRemovalFinished(correlationID string) (finished bool, err error) {
+	events, err := r.getEvents(correlationID)
+	if err != nil {
+		return
+	}
+	for _, event := range events {
+		switch event.MustCode() {
+		case REMOVE_SNAPSHOT_FINISHED_FAILURE:
```
ahadas, line 539: and here again, we cannot see REMOVE_SNAPSHOT_FINISHED_FAILURE that was generated for a previous attempt since if a previous attempt failed, we would have stopped trying to remove that snapshot, right?

pkg/controller/plan/adapter/ovirt/client.go
ahadas: @bennyz another possibility is to create a temporary version of go-ovirt for 2.5 and later do it "properly" when we get https://github.com/kubev2v/forklift/pull/420 in, wdyt?

pkg/controller/plan/adapter/ovirt/client.go
```diff
@@ -460,89 +474,76 @@ func (r Client) removePrecopies(precopies []planapi.Precopy, vmService *ovirtsdk
...
-			}
-
 			select {
 			case <-time.After(time.Duration(settings.Settings.Migration.SnapshotRemovalTimeout) * time.Minute):
 				r.Log.Info("Timeout waiting for snapshot removal")
 				return
 			default:
 				time.Sleep(time.Duration(settings.Settings.Migration.SnapshotStatusCheckRate) * time.Second)
 			}
+			// Try to remove the snapshot
```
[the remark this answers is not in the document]
mnecas: yeah plus when we want to restart the removal we can just use the `continue`

pkg/controller/plan/adapter/ovirt/client.go
```diff
@@ -76,23 +92,21 @@ func (r *Client) CheckSnapshotReady(vmRef ref.Ref, snapshot string) (ready bool,
...
-		err = liberr.New("No jobs found for correlation ID", "correlationID", correlationID)
+	if len(events) < 1 {
+		err = liberr.New("No event found for correlation ID", "correlationID", correlationID)
 		return
 	}
-	ready = true
-	for _, job := range jobs {
-		switch job.MustStatus() {
-		case ovirtsdk.JOBSTATUS_FAILED, ovirtsdk.JOBSTATUS_ABORTED:
+	for _, event := range events {
```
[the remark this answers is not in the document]
mnecas: Yes exactly! Nice description

pkg/controller/plan/adapter/ovirt/client.go
mnecas: yes here is a bit different control logic compared to the creation.
Here is everything controlled by the `continue` and `break` so we don't need the `ConflictError`. I'll add the `continue` to jump to the next retry.

pkg/controller/plan/adapter/ovirt/client.go
mnecas: Ahh, the error log does not have the best description. It should be something like `Error occurred while gathering events about the snapshot removal`. But you are right that we should rather stop in this case

pkg/controller/plan/adapter/ovirt/client.go
mnecas: The `event.Code()` does not return an error but `(int64, bool)`, which is the code itself, and the second parameter checks, if it has a code or code is 0. In case we want to check the code is 0 and not the default golang value
from the go-ovirt
```go
func (p *Event) Code() (int64, bool) {
	if p.code != nil {
		return *p.code, true
	}
	var zero int64
	return zero, false
}
```

pkg/controller/plan/adapter/ovirt/client.go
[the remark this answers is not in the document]
ahadas: restart because of a conflict-error, right? it's ok to wait a bit so if there's an ongoing operation on the vm, we'll give it a chance to end

pkg/controller/plan/adapter/ovirt/client.go
ahadas: @mnecas right but I'd expect remove-snapshot to return a validation error when the disk is being transferred and then `strings.Contains(err.Error(), "Cannot remove Snapshot")` should be true

pkg/controller/plan/adapter/ovirt/client.go
ahadas: I try to recall what I meant here - it was late yesterday when I wrote it
I think I missed the fact that it's within the loop at should also terminate when the timeout is reached..
but about the timeout, we have:
```golang
                        select {
			case <-time.After(time.Duration(settings.Settings.Migration.SnapshotRemovalTimeout) * time.Minute):
				r.Log.Info("Timeout waiting for snapshot removal")
				return
			default:
				time.Sleep(time.Duration(settings.Settings.Migration.SnapshotStatusCheckRate) * time.Second)
			}
```
wouldn't it always get to the `default` part? I mean, `case <-time.After(time.Duration(settings.Settings.Migration.SnapshotRemovalTimeout) * time.Minute):` is supposed to check whether the time since the call to remove-snapshot exceeded `time.Duration(settings.Settings.Migration.SnapshotRemovalTimeout) * time.Minute)` no?
@mnecas @bennyz ^

bennyz: indeed, I missed the noting the start time

pkg/controller/plan/adapter/ovirt/client.go
mnecas: yeah that's true, I never got to the "disk is being transferred..." part always failed in the ConflictError. I'll remove that part

pkg/controller/plan/adapter/ovirt/client.go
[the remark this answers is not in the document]
mnecas: The issue with waiting right after triggering is that the error message is not up to date, and don't see the advantage in it unless we would retry also after the wait.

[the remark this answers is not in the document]
mnecas: I have added a check right before the removal request if the snapshot even exists. Think that adds necessary benefits

pkg/controller/plan/adapter/ovirt/client.go
mnecas: Updated the timeout logic please check it out

pkg/controller/plan/adapter/ovirt/client.go
ahadas: cool, yeah, I also checked it in ovirt-engine - transfer-image locks the vm with a shared lock and remove-snapshot tries to acquire exclusive lock on the vm - so that makes sense

pkg/controller/plan/adapter/ovirt/client.go
ahadas: we'll do it separately

pkg/controller/plan/adapter/ovirt/client.go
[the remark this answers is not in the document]
ahadas: but we still call `time.Sleep(time.Duration(settings.Settings.Migration.SnapshotStatusCheckRate) * time.Second)` before initiating the call to remove-snapshot - why not to move the block in 477-484 to be after the call to remove-snapshot and handing the result?

ahadas (commented):
looks good! minor issues are still discussed

pkg/controller/plan/adapter/ovirt/client.go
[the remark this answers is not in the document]
mnecas: ah yeah, good point with the current code setup don't see any problems with it, updated

pkg/controller/plan/adapter/ovirt/client.go
```diff
@@ -460,89 +481,69 @@ func (r Client) removePrecopies(precopies []planapi.Precopy, vmService *ovirtsdk
...
+func (r *Client) isSnapshotRemovalFinished(correlationID string) (finished bool, err error) {
+	events, err := r.getEvents(correlationID)
+	if err != nil {
+		return
+	}
+	for _, event := range events {
+		code, exists := event.Code()
+		if !exists {
+			err = liberr.New("The event does not have a code.", "event", event)
+			return
```
ahadas, line 536: same as above - return or continue?

pkg/controller/plan/adapter/ovirt/client.go
```diff
@@ -76,23 +92,28 @@ func (r *Client) CheckSnapshotReady(vmRef ref.Ref, snapshot string) (ready bool,
...
 	}
-	ready = true
-	for _, job := range jobs {
-		switch job.MustStatus() {
-		case ovirtsdk.JOBSTATUS_FAILED, ovirtsdk.JOBSTATUS_ABORTED:
+	for _, event := range events {
+		code, exists := event.Code()
+		if !exists {
+			err = liberr.New("The event does not have a code.", "event", event)
+			return
```
ahadas, line 108: it shouldn't really happen that we get an event without a code from ovirt but if we have this check, and we ever get here then we better continue in this case

pkg/controller/plan/adapter/ovirt/client.go
```diff
@@ -76,23 +92,28 @@ func (r *Client) CheckSnapshotReady(vmRef ref.Ref, snapshot string) (ready bool,
...
 		return
 	}
-	ready = true
-	for _, job := range jobs {
-		switch job.MustStatus() {
-		case ovirtsdk.JOBSTATUS_FAILED, ovirtsdk.JOBSTATUS_ABORTED:
+	for _, event := range events {
+		code, exists := event.Code()
+		if !exists {
+			err = liberr.New("The event does not have a code.", "event", event)
```
ahadas, line 107: sorry about not writing about that before but I just noticed that `err` is set in this case - and an invalid event shouldn't lead to a failure, let me post a change for this

bennyz (approved):

PR #489, opened 2023-08-02
plan: check error returned by GetPopulatorTaskName

mmartinv:


ahadas (approved):

liranr23:
Misssing Sign off in the commit. But I can't see the benefit of the error check if within the actual function we don't assign the error.

mmartinv:
> Misssing Sign off in the commit. But I can't see the benefit of the error check if within the actual function we don't assign the error.

I guess you are referring to the oVirt implementation of `GetPopulatorTaskName`, aren't you? In OpenStack an actual error can be returned.

ahadas:
> But I can't see the benefit of the error check if within the actual function we don't assign the error.

yeah, @liranr23 please explain what you meant here

liranr23:
> > But I can't see the benefit of the error check if within the actual function we don't assign the error.
> 
> yeah, @liranr23 please explain what you meant here


Exactly what Miguel wrote:

> > Misssing Sign off in the commit. But I can't see the benefit of the error check if within the actual function we don't assign the error.
> 
> I guess you are referring to the oVirt implementation of `GetPopulatorTaskName`, aren't you? In OpenStack an actual error can be returned.

In that case I think each function should return an error, if there is one. We failed on oVirt therefore I looked there.

ahadas:
> In that case I think each function should return an error, if there is one.

right, that's a good practice in general :)
so if there's no objection to this change, let's get it in then

mmartinv:
> In that case I think each function should return an error, if there is one. We failed on oVirt therefore I looked there.

For OpenStack is there, for oVirt  ther are two possibilities, check if `DataSource` exists in the pvc spec and then return `pvc.Spec.DataSource.Name` if it does and an error if it doesn't or what I was suggesting before, return always `pvc.Name` with no error and then we don't need to check for lun annotation in `updatePopulatorCopyProgress`. In the second case the findTask will return not found (because there is not task associated to lun and will continue with the next pvc)

liranr23:
> > In that case I think each function should return an error, if there is one. We failed on oVirt therefore I looked there.
> 
> For OpenStack is there, for oVirt there are two possibilities:
> 
> 1. check if `DataSource` exists in the PVC spec and then return `pvc.Spec.DataSource.Name` if it does and an error if it doesn't. In this case the error will never be returned because we are filtering the `lun` PVCs in  `updatePopulatorCopyProgress`. If we remove the lun check in `updatePopulatorCopyProgress` we would need to check that the error returned by GetPopulatorTaskName is actually something like `"LunDiskDoesNotHaveAssociatedTask"` or any other kind of error (it can be different for OpenStack) which complicates the logic.
> 2. Return always `pvc.Name` with no error and then we don't need to check for lun annotation in `updatePopulatorCopyProgress`. In this case the `FindTask` will return not found (because there is not task associated to lun ) and will continue with the next pvc.

generally i would go with option 2, to align with others. but since we do label and so on i'm not sure. DataSource is a better check in my opinion as it indicates the PVC is from migration using populator.

PR #490, opened 2023-08-02
OCP: add support for secrets and configmaps

bennyz:


pkg/controller/plan/adapter/ocp/builder.go
```diff
@@ -266,31 +266,53 @@ func (r *Builder) mapDisks(sourceVm *cnv.VirtualMachine, targetVmSpec *cnv.Virtu
...
+					continue
 				}
 			}
 		}
 	}
 
+	configMaps := make(map[string]*core.ConfigMap)
+	secrets := make(map[string]*core.Secret)
+	for i := range sourceVm.Spec.Template.Spec.Volumes {
+		vol := &sourceVm.Spec.Template.Spec.Volumes[i]
```
ahadas, line 295: how about getting the volume from the `range`?

pkg/controller/plan/adapter/ocp/builder.go
```diff
@@ -266,33 +268,47 @@ func (r *Builder) mapDisks(sourceVm *cnv.VirtualMachine, targetVmSpec *cnv.Virtu
...
+
+func createDiskMap(sourceVm *cnv.VirtualMachine, pvcMap map[string]*core.PersistentVolumeClaim, vmRef ref.Ref) map[string]*cnv.Disk {
 	diskMap := make(map[string]*cnv.Disk)
-	for i := range sourceVm.Spec.Template.Spec.Domain.Devices.Disks {
-		disk := &sourceVm.Spec.Template.Spec.Domain.Devices.Disks[i]
-		// Find matching volume
-		// TODO: make this better
+
+	for _, disk := range sourceVm.Spec.Template.Spec.Domain.Devices.Disks {
+		currentDisk := disk
```
ahadas, line 287: isn't `currentDisk` redundant?

pkg/controller/plan/adapter/ocp/builder.go
```diff
@@ -266,33 +268,47 @@ func (r *Builder) mapDisks(sourceVm *cnv.VirtualMachine, targetVmSpec *cnv.Virtu
...
+			case vol.PersistentVolumeClaim != nil:
+				key = pvcSourceName(vmRef.Namespace, vol.PersistentVolumeClaim.ClaimName)
+			case vol.DataVolume != nil:
+				key = pvcSourceName(vmRef.Namespace, vol.DataVolume.Name)
+			case vol.ConfigMap != nil:
+				key = vol.ConfigMap.Name
+			case vol.Secret != nil:
+				key = vol.Secret.SecretName
+			}
+			if vol.Name == disk.Name {
```
ahadas, line 300: that's a bit inefficient to go through the switch-case block above and only then check this, no? how about moving it to the beginning of the for-loop?

pkg/controller/plan/adapter/ocp/builder.go
```diff
@@ -266,33 +268,47 @@ func (r *Builder) mapDisks(sourceVm *cnv.VirtualMachine, targetVmSpec *cnv.Virtu
...
 			}
 		}
 	}
 
-	// Clear original disks and volumes, will be required for other mapped devices later
+	return diskMap
+}
+
+func (r *Builder) mapPVCsToTarget(targetVmSpec *cnv.VirtualMachineSpec, persistentVolumeClaims []core.PersistentVolumeClaim, diskMap map[string]*cnv.Disk) {
+	// Clear original disks and volumes; will be required for other mapped devices later
```
ahadas, line 311: duplicate code with 275-276?

pkg/controller/plan/adapter/ocp/builder.go
```diff
@@ -324,6 +340,144 @@ func (r *Builder) mapDisks(sourceVm *cnv.VirtualMachine, targetVmSpec *cnv.Virtu
...
+			VolumeSource: cnv.VolumeSource{
+				ConfigMap: &cnv.ConfigMapVolumeSource{
+					LocalObjectReference: core.LocalObjectReference{
+						Name: targetConfigMap.Name,
+					},
+				},
+			},
+		}
+
+		if disk, ok := diskMap[sourceConfigMap.Name]; ok {
```
ahadas, line 415: in which scenario would `ok` be false?

pkg/controller/plan/adapter/ocp/builder.go
bennyz: no because of the reference, otherwise we'll only get the last value, the other option is to use the index on the array

pkg/controller/plan/adapter/ocp/builder.go
bennyz: hopefully no scenario, but you never know, I wanted to add an else here

pkg/controller/plan/adapter/ocp/builder.go
ahadas: doesn't it (the reference) change on each iteration?

pkg/controller/plan/adapter/ocp/builder.go
ahadas: or a log..

pkg/controller/plan/adapter/ocp/builder.go
bennyz: the reference of disk or currentDisk?
currentDisk is a new variable on each iteration so we append a different memory location every time

pkg/controller/plan/adapter/ocp/builder.go
bennyz: yeah that's what I did in the else :)

pkg/controller/plan/adapter/ocp/builder.go
ahadas: I meant the `disk` variable but I see now - the values from the range are placed at the same address that `disk` is initially allocated on so you need to duplicate it, ack

pkg/controller/plan/adapter/ocp/builder.go
```diff
@@ -266,36 +268,48 @@ func (r *Builder) mapDisks(sourceVm *cnv.VirtualMachine, targetVmSpec *cnv.Virtu
...
+				key = pvcSourceName(vmRef.Namespace, vol.PersistentVolumeClaim.ClaimName)
+			case vol.DataVolume != nil:
+				key = pvcSourceName(vmRef.Namespace, vol.DataVolume.Name)
+			case vol.ConfigMap != nil:
+				key = vol.ConfigMap.Name
+			case vol.Secret != nil:
+				key = vol.Secret.SecretName
 			}
+
+			diskMap[key] = &currentDisk
```
ahadas, line 305: and we can break, right?

ahadas (commented):
minor comment, other than that, lgtm

pkg/controller/plan/adapter/ocp/builder.go
bennyz: accidentally removed, restored

pkg/controller/plan/adapter/ocp/builder.go
```diff
@@ -324,6 +339,148 @@ func (r *Builder) mapDisks(sourceVm *cnv.VirtualMachine, targetVmSpec *cnv.Virtu
...
+		targetConfigMap := &core.ConfigMap{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:        sourceConfigMap.Name,
+				Namespace:   r.Plan.Spec.TargetNamespace,
+				Labels:      sourceConfigMap.Labels,
+				Annotations: sourceConfigMap.Annotations,
+			},
+			Data: sourceConfigMap.Data,
+		}
+		err := r.Destination.Client.Create(context.Background(), targetConfigMap)
```
liranr23, line 395: i think all the client create should be done in another place (like kubevirt.go but with OCP maybe we can use destination client for that). can be done in another PR, if there is an agreement..

liranr23 (approved):

ahadas (approved):

pkg/controller/plan/adapter/ocp/builder.go
ahadas: even with other source provider types, the client that is used to post entities on the target cluster should be Destination.Client - when we migrate to the local cluster, it's the same client we use for posting the management entities on the local cluster

I agree it would be nicer to generalize the process in a way that builders return back configmaps and secrets from the source but as it looks specific to OCP, it doesn't seem to be cost-effective.. do you see data that should be stored in secrets or configmaps from other sources? (leave TPM aside as we don't get it from RHV and from vSphere it would probably not be useful)

liranr23: nope i don't see that data is needed from other sources.

ahadas:
@bennyz please backport

PR #491, opened 2023-08-03
openstack: fix cognitive complexity

mmartinv:

### DIFF
--- a/pkg/controller/plan/adapter/openstack/builder.go
+++ b/pkg/controller/plan/adapter/openstack/builder.go
@@ -877,13 +877,11 @@ func (r *Builder) PopulatorVolumes(vmRef ref.Ref, annotations map[string]string,
 		err = liberr.Wrap(err)
 		return
 	}
-
 	images, err := r.getImagesFromVolumes(workload)
 	if err != nil {
 		err = liberr.Wrap(err)
 		return
 	}
-
 	if workload.ImageID != "" {
 		var image model.Image
 		image, err = r.getVMSnapshotImage(workload)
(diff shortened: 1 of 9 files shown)


pkg/controller/plan/adapter/openstack/builder.go
```diff
@@ -894,55 +895,61 @@ func (r *Builder) PopulatorVolumes(vmRef ref.Ref, annotations map[string]string,
...
 		if err != nil {
-			if !k8serr.IsNotFound(err) {
-				err = liberr.Wrap(err)
+			if !k8serr.IsAlreadyExists(err) {
+				err = liberr.Wrap(err, "couldn't build the PVC",
+					"image", image.Name)
 				return
 			}
+			err = nil
+			continue
```
ahadas, line 911: shouldn't we proceed to `pvcNames = append(pvcNames, pvc.Name)` in this case?

pkg/controller/plan/adapter/openstack/builder.go
```diff
@@ -894,55 +895,61 @@ func (r *Builder) PopulatorVolumes(vmRef ref.Ref, annotations map[string]string,
...
-		populatorName, err = r.createVolumePopulatorCR(image, secretName, vmRef.ID)
+		populatorName, err = r.createVolumePopulatorCR(*image, secretName, workload.ID)
 		if err != nil {
 			err = liberr.Wrap(err)
 			return
 		}
+		pvc, err = r.ensurePopulatorPVC(workload, image, annotations, populatorName, secretName)
+		return
+	}
+	err = k8serr.NewAlreadyExists(schema.ParseGroupResource("openstackvolumepopulator.forklift.konveyor.io"), populator.Name)
```
ahadas, line 934: I'd expect 'ensure' method to return err == nil when the entity already exists - this behavior is surprising

mmartinv: Yeah maybe that's better approach, but then we would need to check if the returned pvc is `nil` or not

mmartinv: Please take look now, I think it's better now.

pkg/controller/plan/adapter/openstack/builder.go
mmartinv: outdated, please check the latest push

ahadas (commented):
commented but I'll address those comments

pkg/controller/plan/adapter/openstack/client.go
```diff
@@ -932,3 +726,242 @@ func (r *Client) getImagesFromVolumes(vm *libclient.VM) (images []libclient.Imag
...
+func (r *Client) ensureImagesFromVolumesReady(vm *libclient.VM) (ready bool, err error) {
+	var imagesFromVolumes []libclient.Image
+	imagesFromVolumes, err = r.getImagesFromVolumes(vm)
+	if err != nil {
+		err = liberr.Wrap(err)
+		r.Log.Error(err, "error while trying to get the images from the VM volumes",
+			"vm", vm.Name)
+		return
+	}
+	ready = true
```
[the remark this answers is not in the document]
ahadas: ok, that was an incorrect observation but there was still room for improvement here

ahadas:
@bennyz please review the openstack related changes
@liranr23 please review the changes in EnsurePopulatorVolumes (filtering pending PVCs for the binding pod)

pkg/controller/plan/adapter/openstack/builder.go
```diff
@@ -893,56 +891,64 @@ func (r *Builder) PopulatorVolumes(vmRef ref.Ref, annotations map[string]string,
...
-		if err != nil {
-			if !k8serr.IsNotFound(err) {
-				err = liberr.Wrap(err)
-				return
-			}
-		}
+func (r *Builder) getCorrespondingPvc(image model.Image, workload *model.Workload, vmRef ref.Ref, annotations map[string]string, secretName string) (*core.PersistentVolumeClaim, error) {
+	if populatorName, err := r.ensureVolumePopulator(workload, &image, secretName); err == nil {
+		return r.ensureVolumePopulatorPVC(workload, &image, annotations, populatorName)
+	} else {
```
bennyz, line 912: if used named return value we can dropped the else here

ahadas: but with named return values, I'll need to define populatorName separately... I can drop the 'else' now as well..

bennyz: but err will be out of scope (unless defined at the top), no?

ahadas: ah right right

ahadas: changed, also resolved a conflict with #591 - please double check me

liranr23:
> @liranr23 please review the changes in EnsurePopulatorVolumes (filtering pending PVCs for the binding pod)

looks ok, you can also check the storage class (or profile) and see if it's wait for consumer.

liranr23 (approved):

bennyz (approved):

PR #492, opened 2023-08-03
Change the storage mapping for OVA

bkhizgiy:
This fix changes the storage mapping for OVA provider from disk -> SC into general storage name -> SC.
The mapping should be in the following format:

```
apiVersion: forklift.konveyor.io/v1beta1
kind: StorageMap
metadata:
  name: storage-map
  namespace: konveyor-forklift
spec:
  map:
    - destination:
        storageClass: nfs-csi
      source:
        name: VM storage
  provider:
    destination:
      name: host
      namespace: konveyor-forklift
    source:
      name: provider-ova
      namespace: konveyor-forklift
```

### DIFF
--- a/pkg/controller/plan/adapter/ova/validator.go
+++ b/pkg/controller/plan/adapter/ova/validator.go
@@ -97,25 +97,8 @@ func (r *Validator) PodNetwork(vmRef ref.Ref) (ok bool, err error) {
 
 // Validate that a VM's disk backing storage has been mapped.
 func (r *Validator) StorageMapped(vmRef ref.Ref) (ok bool, err error) {
-	if r.plan.Referenced.Map.Storage == nil {
-		return
-	}
-	vm := &model.VM{}
-	err = r.inventory.Find(vm, vmRef)
-	if err != nil {
-		err = liberr.Wrap(
-			err,
-			ErrVMNotFound,
-			"vm",
-			vmRef.String())
-		return
-	}
-
-	for _, disk := range vm.Disks {
-		if !r.plan.Referenced.Map.Storage.Status.Refs.Find(ref.Ref{ID: disk.ID}) {
-			return
-		}
-	}
+	//For OVA providers, we don't have an actual storage connected,
+	// since we use a dummy storage for mapping the function should always return true.
 	ok = true
 	return
 }
--- a/pkg/controller/provider/container/ova/model.go
+++ b/pkg/controller/provider/container/ova/model.go
@@ -3,8 +3,10 @@ package ova
 import (
 	"context"
 	"errors"
+	"fmt"
 
 	"github.com/go-logr/logr"
+	api "github.com/konveyor/forklift-controller/pkg/apis/forklift/v1beta1"
 	model "github.com/konveyor/forklift-controller/pkg/controller/provider/model/ova"
 	fb "github.com/konveyor/forklift-controller/pkg/lib/filebacked"
 	libmodel "github.com/konveyor/forklift-controller/pkg/lib/inventory/model"
@@ -21,6 +23,7 @@ func init() {
 		&NetworkAdapter{},
 		&DiskAdapter{},
 		&VMAdapter{},
+		&StorageAdapter{},
 	}
 }
 
@@ -56,7 +59,7 @@ func (r *Context) canceled() (done bool) {
 // model and the inventory model.
 type Adapter interface {
 	// List REST collections.
-	List(ctx *Context) (itr fb.Iterator, err error)
+	List(ctx *Context, provider *api.Provider) (itr fb.Iterator, err error)
 	// Get object updates
 	GetUpdates(ctx *Context) (updater []Updater, err error)
 }
@@ -71,7 +74,7 @@ type NetworkAdapter struct {
 }
 
 // List the collection.
-func (r *NetworkAdapter) List(ctx *Context) (itr fb.Iterator, err error) {
+func (r *NetworkAdapter) List(ctx *Context, provider *api.Provider) (itr fb.Iterator, err error) {
 	networkList := []Network{}
 	err = ctx.client.list("networks", &networkList)
 	if err != nil {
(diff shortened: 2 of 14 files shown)


pkg/controller/provider/container/ova/model.go
```diff
@@ -228,3 +229,35 @@ func (r *DiskAdapter) GetUpdates(ctx *Context) (updates []Updater, err error) {
...
+
+func (r *StorageAdapter) GetUpdates(ctx *Context) (updates []Updater, err error) {
+	return
+}
+
+// List the collection.
+func (r *StorageAdapter) List(ctx *Context) (itr fb.Iterator, err error) {
+	storageList := []Storage{}
+	storageList = append(storageList, Storage{
+		Name: "VM storage",
```
bkhizgiy, line 245: Any others suggestion for the name + ID.
They will be the same for all OVA providers and we will have one created for each.

pkg/controller/plan/adapter/ova/validator.go
```diff
@@ -111,10 +111,8 @@ func (r *Validator) StorageMapped(vmRef ref.Ref) (ok bool, err error) {
 		return
 	}
 
-	for _, disk := range vm.Disks {
-		if !r.plan.Referenced.Map.Storage.Status.Refs.Find(ref.Ref{ID: disk.ID}) {
-			return
-		}
+	if len(vm.Disks) == 0 {
+		return
```
ahadas, line 115: why do we need this part?

pkg/controller/provider/container/ova/model.go
ahadas: why do we need the storage to be the same for all OVA providers?
can't we go with something like: "Storage of provider <name of the provider>" and use the UID of the provider for the storage as well?

bkhizgiy: just to simplify it for users, instead of searching the correct value for each provider use something more unified

pkg/controller/plan/adapter/ova/validator.go
bkhizgiy: validate that the vm has disks at all, in case its empty do we want to allow the mapping?

ahadas: if a vm has no disks then it doesn't need mappings - every mapping would be valid
we are generally not interested in diskless vms but if we want to block it, it needs to be a different validation, not of the mapping

pkg/controller/provider/container/ova/model.go
ahadas: users are not supposed to write yaml files with our new beautiful UI :) scripts interact with the system through such files and they're expected to query the available storage types before setting something so I don't see this as a goal to make it easy to manually write a mapping configuration
but we can as well go with something like "All disks in OVAs" or something like that

pkg/controller/plan/adapter/ova/validator.go
bkhizgiy: In that case we don't really need to check anything here, will always be true.

pkg/controller/provider/container/ova/model.go
```diff
@@ -228,3 +231,36 @@ func (r *DiskAdapter) GetUpdates(ctx *Context) (updates []Updater, err error) {
...
+}
+
+func (r *StorageAdapter) GetUpdates(ctx *Context) (updates []Updater, err error) {
+	return
+}
+
+// List the collection.
+func (r *StorageAdapter) List(ctx *Context, provider *api.Provider) (itr fb.Iterator, err error) {
+	storageList := []Storage{}
+	storageName := fmt.Sprintf("Provider %s storage", provider.Name)
```
ahadas, line 246: how about `Fake storage for %s` ?

pkg/controller/provider/container/ova/model.go
```diff
@@ -228,3 +231,36 @@ func (r *DiskAdapter) GetUpdates(ctx *Context) (updates []Updater, err error) {
...
+	for _, object := range storageList {
+		m := &model.Storage{
+			Base: model.Base{
+				ID:   object.ID,
+				Name: object.Name,
+			},
+		}
+		object.ApplyTo(m)
+		list.Append(m)
+	}
```
ahadas, line 261: I understand that the code above is based on similar code we have for other providers but considering that we have a single storage, how about simplifying it?

pkg/controller/provider/container/ova/model.go
bkhizgiy: umm how about `Dummy Storage for %s` or `Mock storage for %s`? personally I think fake sounds a bit off

ahadas: it reminds me of [a post by Martin Fowler where he listed fakes, mocks, dummies, stubs, ...](https://martinfowler.com/articles/mocksArentStubs.html) - I don't think 'mock' is not the right term to use here but 'dummy' sounds good, it aligns with `Dummy objects are passed around but never actually used. Usually they are just used to fill parameter lists.` and it doesn't suggest that it's not production-ready like fakes

pkg/controller/provider/container/ova/model.go
bkhizgiy: after looking into it, all the adapters must implement the same interface and return the iterator from the func 
(which needed for the next func under the collector) since its  exists only for the list object I cant change the implementation even thought we having only one resource here.

ahadas:
@bkhizgiy please also check out [the new code smell that is reported](https://sonarcloud.io/project/issues?resolved=false&types=CODE_SMELL&pullRequest=492&id=kubev2v_forklift&open=AYm8G-_BEpRQaJEkLnuc) - how about adding a comment inside the function that explains why the storage mapping is always valid as we discussed in the thread we had about it?

pkg/controller/provider/container/ova/model.go
bkhizgiy: nice, so dummy it is:)

pkg/controller/provider/container/ova/model.go
ahadas: I'm not suggesting to change the signature but creating a list `storageList` that includes a single value and then iterating over it.. would the following work instead:
```golang
storageName := fmt.Sprintf("Provider %s storage", provider.Name)
list := fb.NewList()
m := &model.Storage{
	Base: model.Base{
		ID:   string(provider.UID),
		Name: storageName,
	},
}
list.Append(m)
itr = list.Iter()
return
```
?

ahadas (approved):

ahadas:
@bkhizgiy please backport

PR #497, opened 2023-08-06
Fix unit size calculation for memory and disk size

bkhizgiy:


pkg/controller/plan/adapter/ova/builder.go
```diff
@@ -513,9 +513,10 @@ func getResourceCapacity(capacity int64, units string) (int64, error) {
 	if units == "" {
 		return 0, nil
 	}
-
+	if units == "byte" {
+		return capacity, nil
+	}
```
ahadas, line 518: this looks alright but looking at the code below, seems there are more cases that aren't covered - can we make it closer to [the code in vdsm](https://github.com/oVirt/vdsm/blob/1938e3cb68bd658c53b042335fc830414fc238bd/lib/vdsm/v2v.py#L1381-L1430)?

ahadas: or similar to [the code in ovirt-engine](https://github.com/oVirt/ovirt-engine/blob/master/backend/manager/modules/utils/src/main/java/org/ovirt/engine/core/utils/ovf/OvfOvaReader.java#L251-L270)

ahadas:
@bkhizgiy can you please also add unit-test(s) for this?

pkg/controller/plan/adapter/ova/ova_capacity_test.go
```diff
@@ -0,0 +1,44 @@
...
+		{"byte units", 20, "byte", 20, false},
+		{"pow 10", 2, "byte * 2^10", 2048, false},
+		{"pow 20", 1, "byte * 2^20", 1048576, false},
+		{"invalid pow", 1, "byte * 2^15", 0, true},
+		{"missing pow", 1, "byte * 2", 0, true},
+		{"malformed units", 100, "byte*2^", 0, true},
+		{"no spaces", 1, "byte*2^20", 1048576, false},
+		{"unsupported units", 10, "kilobytes", 0, true},
+		{"unsupported operand", 1, "byte + 2^20", 0, true},
+		{"unsupported base", 1, "byte * 3^20", 0, true},
```
ahadas, line 30: it's not common, but why is it not supported?

pkg/controller/plan/adapter/ova/ova_capacity_test.go
```diff
@@ -0,0 +1,44 @@
...
+	}
```
ahadas, line 31: it's missing another common configuration in the form of `byte * 2 * 2^10`

pkg/controller/plan/adapter/ova/ova_capacity_test.go
bkhizgiy: from what I saw I thought it has to be representation of kb, gb and etc, but if allowed i"ll change it

ahadas: I looked at https://www.dmtf.org/sites/default/files/standards/documents/DSP0004_2.7.0.pdf and it doesn't look like the base has to be `2` so let's generalize it

pkg/controller/plan/adapter/ova/builder.go
```diff
@@ -510,27 +518,36 @@ func getDiskSourcePath(filePath string) string {
...
+				}
+				pow, err := strconv.Atoi(nums[1])
+				if err != nil {
+					return 0, errors.New(fmt.Sprintf("units '%s' are invalid, pow component is invalid: %s", units, item))
+				}
+				capacity = capacity * int64(math.Pow(float64(base), float64(pow)))
+				continue
+			}
+			return 0, errors.New(fmt.Sprintf("units '%s' are invalid, item is invalid: %s", units, item))
+		}
```
ahadas, line 548: @bkhizgiy I didn't verify these changes, please take a look

pkg/controller/plan/adapter/ova/builder.go
```diff
@@ -517,51 +518,36 @@ func getDiskSourcePath(filePath string) string {
...
-	if err != nil {
-		return 0, err
-	}
-
-	result := capacity
-	for _, num := range numbers[:len(numbers)-2] {
-		val, err := strconv.Atoi(num)
-		if err != nil {
-			return 0, err
+	items := strings.Split(units, "*")
```
bkhizgiy, line 521: looking good, the only thing I"ll add is a check if its not an empty string in the beginning of the func and if so return the capacity.
+ if the tests passed it should be verified.

pkg/controller/plan/adapter/ova/builder.go
bkhizgiy: @ahadas I see that the test failing, so I"ll take a look whats missing.

pkg/controller/plan/adapter/ova/ova_capacity_test.go
```diff
@@ -0,0 +1,45 @@
...
+		{"pow 10", 2, "byte * 2^10", 2048, false},
+		{"pow 20", 1, "byte * 2^20", 1048576, false},
+		{"missing pow", 1, "byte * 2", 0, true},
+		{"malformed units", 100, "byte*2^", 0, true},
+		{"no spaces", 1, "byte*2^20", 1048576, false},
+		{"unsupported units", 10, "kilobytes", 0, true},
+		{"unsupported operand", 1, "byte + 2^20", 0, true},
+		{"diffrent base", 1, "byte * 3^10", 59049, false},
+		{"uncommon format", 1, "byte * 2 * 2^10", 2048, false},
+		{"uncommon format no spaces", 1, "byte*2*2^10", 2048, false},
```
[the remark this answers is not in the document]
ahadas: I've changed it

pkg/controller/plan/adapter/ova/builder.go
ahadas: > looking good, the only thing I"ll add is a check if its not an empty string in the beginning of the func and if so return the capacity.

that's what the first test checks, right? :)
it is supposed to work even without special handling to this case

pkg/controller/plan/adapter/ova/builder.go
ahadas: right, I think I've fixed that test that failed - `byte * 2` is supposed to be a valid input

pkg/controller/plan/adapter/ova/ova_capacity_test.go
```diff
@@ -28,7 +28,7 @@ func TestGetResourceCapacity(t *testing.T) {
 		{"unsupported operand", 1, "byte + 2^20", 0, true},
 		{"diffrent base", 1, "byte * 3^10", 59049, false},
 		{"uncommon format", 1, "byte * 2 * 2^10", 2048, false},
-		{"uncommon format no spaces", 1, "byte*2*2^10", 2048, false},
+		{"uncommon format no spaces", 1, "byte*3*2^10", 2048, false},
```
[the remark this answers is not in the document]
ahadas: right, I did

Issue #494, opened 2023-08-03
"No boot device" after successfully migrating VM using the default host provider


prakashmishra1598:
After completing the migration of a VM using the default host provider created on the target cluster, the migration proceeded to run successfully but after starting up the virtual machine and connecting to it through VNC, the VM wasn't able to boot up with the following error: 
```
Booting from Hard Disk...
Boot failed: not a bootable disk

No bootable device.
```

I also found these lines in the logs which specify that the v2v failed:
```
Failed to determine unit we run in, ignoring: No data available
guestfsd: => inspect_get_distro (0x1e4) took 0.01 secs
libguestfs: trace: v2v: inspect_get_distro = "unknown"
i_root = /dev/sda3
i_type = linux
i_distro = unknown
i_osinfo = unknown
i_arch = x86_64
i_major_version = 5
i_minor_version = 0
i_package_format = unknown
i_package_management = unknown
i_product_name = VMware Photon OS 5.0
i_product_variant = unknown
i_windows_systemroot = 
i_windows_software_hive = 
i_windows_system_hive = 
i_windows_current_control_set = 

virt-v2v: error: inspection could not detect the source guest (or physical machine).

Assuming that you are running virt-v2v/virt-p2v on a source which is supported (and not, for example, a blank dis
k), then this should not happen.

Inspection field ‘i_distro’ was ‘unknown’.
rm -rf -- '/tmp/v2vnbdkit.bcTLgc'
rm -rf -- '/tmp/v2v.tCVBMC'
libguestfs: trace: v2v: close
```

Inspite of this the migration pod runs to completion and the migration is marked as succeeded. Gist of the logs: https://gist.github.com/prakashmishra1598/60e6f814b9fc5014ff3dae020e77303b

ahadas:
yeah, the pod should obviously finish with an error in this situation
about cause of the failure - what type of guest is it?

ahadas:
I can confirm it's a regression on the main branch.
when migrating using 2.4.2 the pod's execution ends with an error:
![2-4-2](https://github.com/kubev2v/forklift/assets/6020068/0feeff0f-abff-4f44-a019-9696a2dc9e76)
when migrating using 'latest' images, the pod's execution ends as 'completed':
![main](https://github.com/kubev2v/forklift/assets/6020068/e59bd412-decb-42f8-9086-75b8a1f227e2)
[...]

PR #498, opened 2023-08-06
Exit when virt-v2v cold hit an error

liranr23:
Since we now support two flows in the virt-v2v image, it doesn't end with the actual `virt-v2v` command. In that case the pod may end with success although, it didn't. Now the script will exit with an error on the first error occurrence.

Fix https://github.com/kubev2v/forklift/issues/494#issuecomment-1664686693

### DIFF
--- a/virt-v2v/cold/entrypoint
+++ b/virt-v2v/cold/entrypoint
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-set -o pipefail
+set -o pipefail -e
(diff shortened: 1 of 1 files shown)


bennyz (approved):

ahadas (approved):
I'd prefer us to build the arguments for virt-v2v differently along the flow and still execute virt-v2v once at the end but that also works

ahadas:
@liranr23 please backport

PR #501, opened 2023-08-06
use --spawn_strategy=processwrapper-sandbox

bennyz:
virt-v2v and ovirt-populator podman build is broken without it

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -30,3 +30,6 @@ build --action_env=OVA_PROVIDER_SERVER_IMAGE=quay.io/kubev2v/forklift-ova-provid
 # sandboxes. Use slightly less secure but working processwrapper sandbox.
 # NOTE: Same configuration is in virt-v2v/cold/.bazelrc.
 build --strategy_regexp="Action appliance/libguestfs-appliance.tar"=processwrapper-sandbox
+
+# For populator images, we need to use processwrapper sandbox as well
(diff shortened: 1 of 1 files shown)


ahadas (commented):
did you check the option of extending the configuration in [.bazelrc](https://github.com/kubev2v/forklift/blob/main/.bazelrc#L32)?

bennyz:
> did you check the option of extending the configuration in [.bazelrc](https://github.com/kubev2v/forklift/blob/main/.bazelrc?rgh-link-date=2023-08-06T18%3A14%3A20Z#L32)?

I did now

.bazelrc
```diff
@@ -30,3 +30,6 @@ build --action_env=OVA_PROVIDER_SERVER_IMAGE=quay.io/kubev2v/forklift-ova-provid
 # sandboxes. Use slightly less secure but working processwrapper sandbox.
 # NOTE: Same configuration is in virt-v2v/cold/.bazelrc.
 build --strategy_regexp="Action appliance/libguestfs-appliance.tar"=processwrapper-sandbox
+
+# For populator images, we need to use processwrapper sandbox as well
+build --strategy_regexp="Action ovirt-populator-image"=processwrapper-sandbox
```
bennyz, line 35: @ahadas do we want to just always use it?

ahadas: I think it's good to stick with the default when it works

ahadas (approved):

PR #503, opened 2023-08-07
Fix virt-v2v bracket

liranr23:
fixing the error in the entrypoint script:
line 5: [: missing `]'

### DIFF
--- a/virt-v2v/cold/entrypoint
+++ b/virt-v2v/cold/entrypoint
@@ -2,7 +2,7 @@
 set -o pipefail -e
 shopt -s nullglob
 
-if [ -z "$V2V_source"] ; then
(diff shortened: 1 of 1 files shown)


ahadas (approved):
yeah, I noticed that error in the log and forgot about it
simple enough - please backport

ahadas:
btw, another side effect of the recent changes is that we set that V2V_source to vCenter also when we convert the image after CDI fetches the image from vSphere which is confusing
cc @bkhizgiy

ahadas:
> btw, another side effect of the recent changes is that we set that V2V_source to vCenter also when we convert the image after CDI fetches the image from vSphere which is confusing cc @bkhizgiy

posted https://github.com/kubev2v/forklift/issues/504

PR #508, opened 2023-08-07
ocp: make export token TTL configurable

bennyz:


bennyz:
> I think this granularity (a global configuration) is good enough are we good with CDI's default setting?

well the docs say it's a reasonable default :)
```
	// ttlDuration limits the lifetime of an export
	// If this field is set, after this duration has passed from counting from CreationTimestamp,
	// the export is eligible to be automatically deleted.
	// If this field is omitted, a reasonable default is applied.
	// +optional
```

pkg/controller/plan/adapter/ocp/client.go
```diff
@@ -143,23 +143,31 @@ func (r *Client) PreTransferActions(vmRef ref.Ref) (ready bool, err error) {
...
 				Kind:       "VirtualMachineExport",
 				APIVersion: "kubevirt.io/v1alpha3",
 			},
-			ObjectMeta: meta.ObjectMeta{
+			ObjectMeta: metav1.ObjectMeta{
 				Name:      vmRef.Name,
 				Namespace: vmRef.Namespace,
 			},
 			Spec: export.VirtualMachineExportSpec{
+				TTLDuration: tokenTTLDuration,
```
ahadas, line 170: is it omittedwhen `tokenTTLDuration` is nil?

bennyz: indeed

ahadas: ack

ahadas (approved):

PR #512, opened 2023-08-08
Fix disk size alignment when adding FS overhead

bkhizgiy:


bennyz:
you can place it in the util package under
pkg/plan/util/

bkhizgiy:
@bennyz thanks for the suggestion, was wandering where it can be placed to reduce code duplication:)

+ The fix still being verified

PR #513, opened 2023-08-08
ocp: use namespace from Ref when present

bennyz:
When looking up VMs in plans we need to use the namespace from the Ref, and not rely on splitting the name

### DIFF
--- a/pkg/controller/provider/web/ocp/client.go
+++ b/pkg/controller/provider/web/ocp/client.go
@@ -158,8 +158,16 @@ func (r *Finder) ByRef(resource interface{}, ref base.Ref) (err error) {
 		}
 		name := ref.Name
 		if name != "" {
-			ns, name := path.Split(name)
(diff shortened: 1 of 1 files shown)


pkg/controller/provider/web/ocp/client.go
```diff
@@ -158,8 +158,13 @@ func (r *Finder) ByRef(resource interface{}, ref base.Ref) (err error) {
 		}
 		name := ref.Name
 		if name != "" {
-			ns, name := path.Split(name)
-			ns = strings.TrimRight(ns, "/")
+			var ns string
+			if ref.Namespace != "" {
```
ahadas, line 162: worth documenting when we expect the namespace to be missing?

bennyz: yes, will add

ahadas (approved):

PR #519, opened 2023-08-09
Fix virt-v2v entrypoint script

liranr23:
Since we added OVA we had two options to execute the virt-v2v command. Now the command will be built and will be executed at the end without conditions regarding the source. This will allow us to know if we had an error and fail the pod in such case.

### DIFF
--- a/virt-v2v/cold/entrypoint
+++ b/virt-v2v/cold/entrypoint
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-set -o pipefail -e
+set -eo pipefail
 shopt -s nullglob
 
 if [ -z "$V2V_source" ] ; then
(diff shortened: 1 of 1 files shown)


virt-v2v/cold/entrypoint
```diff
@@ -87,22 +84,11 @@ if [ "$V2V_source" == "vCenter" ] ; then
           -io "vddk-thumbprint=$V2V_thumbprint"
       )
   fi
+  args+=(-- "$V2V_vmName")
```
ahadas, line 87: don't we need to append above (line 81)?

ahadas: I mean, for consistency...

liranr23: no, because line 81 is only when VDDK is available.

ahadas: > no, because line 81 is only when VDDK is available.

let me rephrase, wouldn't it be better to replace line 81 with:
```golang
args+=(
```

ahadas (approved):

ahadas:
Much better this way!

PR #530, opened 2023-08-10
Check claimName on DV status

bennyz:
It seems like dv.PVC is always nil because we don't create a DV from a PVC, this leads to failing to find the DV whenever the importer pod fails and it will restart forever instead of 3 times

### DIFF
--- a/pkg/controller/plan/kubevirt.go
+++ b/pkg/controller/plan/kubevirt.go
@@ -159,7 +159,7 @@ func (r *KubeVirt) ListVMs() ([]VirtualMachine, error) {
 				}
 				vm.DataVolumes = append(
 					vm.DataVolumes,
-					DataVolume{
+					ExtendedDataVolume{
 						DataVolume: dv,
 						PVC:        pvc,
 					})
@@ -537,7 +537,7 @@ func (r *KubeVirt) isDataVolumeExistsInList(dv *cdi.DataVolume, dataVolumeList *
 }
 
 // Return DataVolumes associated with a VM.
-func (r *KubeVirt) getDVs(vm *plan.VMStatus) (dvs []DataVolume, err error) {
+func (r *KubeVirt) getDVs(vm *plan.VMStatus) (edvs []ExtendedDataVolume, err error) {
 	dvsList := &cdi.DataVolumeList{}
 	err = r.Destination.Client.List(
 		context.TODO(),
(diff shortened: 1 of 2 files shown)


liranr23 (approved):
The change is ok but the commit is strange. The PVC is created by the DV. Isn't dv.PVC should be the created PVC by the DV?

bennyz:
> > The change is ok but the commit is strange. The PVC is created by the DV. Isn't dv.PVC should be the created PVC by the DV?
> 
> I'll have to look at this again, I got fooled by the name of the object, this is not the CDI DataVolume, it's our DataVolume[1] wrapper...
> 
> [1]
> 
> https://github.com/kubev2v/forklift/blob/76efe8ab17b7df4b2391aaa620f1239480f43abe/pkg/controller/plan/kubevirt.go#L1690

ok, so the wrapper is only populated when calling VirtualMachineMap

pkg/controller/plan/kubevirt.go
```diff
@@ -1687,13 +1687,13 @@ func (r *KubeVirt) vmAllButMigrationLabels(vmRef ref.Ref) (labels map[string]str
 }
 
 // Represents a CDI DataVolume and add behavior.
-type DataVolume struct {
+type DataVolumeWrapper struct {
```
liranr23, line 1690: it would be nice to drop it as it only holds the DV and PVC. or use it much more. for now the renaming is less confusing.

liranr23 (approved):

pkg/controller/plan/migration.go
```diff
@@ -1389,32 +1389,27 @@ func (r *Migration) updateCopyProgress(vm *plan.VMStatus, step *plan.Step) (err
...
+					}, pvc)
+					if err != nil {
+						log.Error(
+							err,
+							"Could not get PVC for DataVolume.",
+							"vm",
+							vm.String(),
+							"dv",
+							path.Join(dv.Namespace, dv.Name))
+						continue
```
ahadas, line 1409: I'm ok with having this fix as a temporary & simple solution for backporting, but the core issue here is that we retrieve the DVs using getDVs that return "our struct" but without setting the 'PVC' and so in some places we expect it to be set (when returned from ListVMs) and sometimes it's not.. and in addition, in most of the places we don't even use the two functions of "out struct".. maybe it will be a better idea to drop that struct as @liranr23 wrote above and replace the two methods with utility-methods, so that when "our struct" is returned, we can assume both the DV and the PVC were queried, wdyt?

bennyz: yeah I'm also in favor of dropping it, in this particular case ListVMs doesn't help because the VM doesn't exist yet

bennyz: I'm also in favor of dropping it, and in this particular case ListVMs doesn't help us because there's no VM yet

pkg/controller/plan/kubevirt.go
bennyz: renamed

bennyz:
Created https://github.com/kubev2v/forklift/issues/539

ahadas (approved):

PR #531, opened 2023-08-14
ocp: fix storage mapping validation

bennyz:
DV based disks need to be considered as well

### DIFF
--- a/pkg/controller/plan/adapter/ocp/validator.go
+++ b/pkg/controller/plan/adapter/ocp/validator.go
@@ -84,39 +84,44 @@ func (r *Validator) StorageMapped(vmRef ref.Ref) (ok bool, err error) {
 	}
 
 	for _, vol := range vm.Spec.Template.Spec.Volumes {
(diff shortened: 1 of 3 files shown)


pkg/controller/plan/adapter/ocp/validator.go
```diff
@@ -84,39 +84,44 @@ func (r *Validator) StorageMapped(vmRef ref.Ref) (ok bool, err error) {
...
-					"PVC not found.",
-					"pvc",
-					vol.PersistentVolumeClaim.ClaimName)
-				return
-			}
+			pvcName = vol.PersistentVolumeClaim.ClaimName
+		case vol.DataVolume != nil:
+			pvcName = vol.DataVolume.Name
+		default:
+			continue
```
ahadas, line 94: shall we add a log that indicates we skip the volume here?

bennyz: added

PR #532, opened 2023-08-14
ocp: copy loop var when iterating over interfaces

bennyz:
Otherwise we end up using the variable which we copy over the same interface and keep the type

https://github.com/golang/go/wiki/LoopvarExperiment


ahadas (approved):

ahadas:
> Otherwise we end up using the variable which we copy over the same interface and keep the type

yeah, very confusing...

ahadas:
@bennyz please backport

PR #538, opened 2023-08-15
Fix configuration for disabling EMS check with FIPS

ahadas:


nyoxi:
Could you elaborate on why is this needed? You should leave `/usr` to files from packages and inject your own configuration into `/etc` if possible.

ahadas:
> Could you elaborate on why is this needed? You should leave `/usr` to files from packages and inject your own configuration into `/etc` if possible.

@nyoxi  it seems like the configuration of the crypto-policies in /etc are taken from /usr/share/crypto-policies. we previously modified the ones in /etc and the changes were applied when the crypto policy remained DEFAULT but when it changed to FIPS (that happens when the host is set in FIPS mode), the configuration in /etc/ was overridden

ahadas:
@nyoxi I don't rule it out that there's a better way to alter the configuration for FIPS by changing files in /etc/ but I the documentation I saw doesn't cover this, I saw how to add another customized policy that way - which is not what we're interested in

ahadas:
btw, the configuration looks different in UBI9 compared to Fedora on my machine.
on my machine there's symlinks to txt files in /usr/share:
```bash
[ahadas@arik crypto-policies]$ tree
.
├── back-ends
│   ├── bind.config -> /usr/share/crypto-policies/DEFAULT/bind.txt
│   ├── bind.config.rpmnew -> /usr/share/crypto-policies/DEFAULT/bind.txt
│   ├── gnutls.config -> /usr/share/crypto-policies/DEFAULT/gnutls.txt
│   ├── gnutls.config.rpmnew -> /usr/share/crypto-policies/DEFAULT/gnutls.txt
│   ├── java.config -> /usr/share/crypto-policies/DEFAULT/java.txt
│   ├── java.config.rpmnew -> /usr/share/crypto-policies/DEFAULT/java.txt
│   ├── javasystem.config -> /usr/share/crypto-policies/DEFAULT/javasystem.txt
│   ├── krb5.config -> /usr/share/crypto-policies/DEFAULT/krb5.txt
│   ├── krb5.config.rpmnew -> /usr/share/crypto-policies/DEFAULT/krb5.txt
│   ├── libreswan.config -> /usr/share/crypto-policies/DEFAULT/libreswan.txt
│   ├── libreswan.config.rpmnew -> /usr/share/crypto-policies/DEFAULT/libreswan.txt
│   ├── libssh.config -> /usr/share/crypto-policies/DEFAULT/libssh.txt
│   ├── libssh.config.rpmnew -> /usr/share/crypto-policies/DEFAULT/libssh.txt
│   ├── nss.config
│   ├── nss.config.rpmnew -> /usr/share/crypto-policies/DEFAULT/nss.txt
│   ├── openssh.config -> /usr/share/crypto-policies/DEFAULT/openssh.txt
│   ├── openssh.config.rpmnew -> /usr/share/crypto-policies/DEFAULT/openssh.txt
│   ├── opensshserver.config -> /usr/share/crypto-policies/DEFAULT/opensshserver.txt
│   ├── opensshserver.config.rpmnew -> /usr/share/crypto-policies/DEFAULT/opensshserver.txt
│   ├── opensslcnf.config -> /usr/share/crypto-policies/DEFAULT/opensslcnf.txt
│   ├── opensslcnf.config.rpmnew -> /usr/share/crypto-policies/DEFAULT/opensslcnf.txt
│   ├── openssl.config -> /usr/share/crypto-policies/DEFAULT/openssl.txt
│   ├── openssl.config.rpmnew -> /usr/share/crypto-policies/DEFAULT/openssl.txt
│   ├── rpm-sequoia.config -> /usr/share/crypto-policies/DEFAULT/rpm-sequoia.txt
│   └── sequoia.config -> /usr/share/crypto-policies/DEFAULT/sequoia.txt
├── config
├── local.d
│   └── nss-p11-kit.config
├── policies
│   └── modules
└── state
    ├── current
    └── CURRENT.pol
```
while in UBI there are no symlink, and the files that are being used are not that `txt` files

nyoxi:
So this change to FIPS policy happens automatically and we can't (or don't want) to have the FIPS policy configured by default in the container (i.e. during build). Is that correct? And if it happens automatically we cannot apply a (sub-)policy to override the configuration. Do I understand it correctly?

ahadas:
> So this change to FIPS policy happens automatically and we can't (or don't want) to have the FIPS policy configured by default in the container (i.e. during build). Is that correct?

right, it happens automatically based on the fips mode of the host

> And if it happens automatically we cannot apply a (sub-)policy to override the configuration. Do I understand it correctly?

yeah. maybe we can drop the default configuration of FIPS and introduce a new crypto policy called FIPS with different configuration - but I want to have a quick solution for 2.5, we can look at something like this later

nyoxi (approved):

PR #548, opened 2023-08-15
Fix 'undefined: logging.LevelLogger'

ahadas:
pkg/controller/plan/adapter/ocp/validator.go:24:23: undefined: logging.LevelLogger

### DIFF
--- a/pkg/controller/plan/adapter/ocp/validator.go
+++ b/pkg/controller/plan/adapter/ocp/validator.go
@@ -5,9 +5,9 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/go-logr/logr"
 	api "github.com/konveyor/forklift-controller/pkg/apis/forklift/v1beta1"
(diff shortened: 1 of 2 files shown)


ahadas:
@bennyz do you have a better idea? see [this log](https://github.com/kubev2v/forklift/actions/runs/5870529245/job/15917721025)

bennyz:
I didn't understand why this didn't fail, but then noticed it's the stable branch...

pkg/controller/plan/adapter/ocp/validator.go
```diff
@@ -21,7 +21,7 @@ type Validator struct {
 	plan         *api.Plan
 	inventory    web.Client
 	sourceClient k8sclient.Client
-	log          logging.LevelLogger
```
bennyz, line 24: can't we replace it with `logr.Logger`? then it should be the only change

ahadas: you're right, changed

bennyz (approved):

PR #560, opened 2023-08-17
Change ova storage name to represent source provider

bkhizgiy:


bkhizgiy:
@RichardHoch Following the discussion in the demo.

pkg/controller/provider/container/ova/model.go
```diff
@@ -242,7 +242,7 @@ func (r *StorageAdapter) GetUpdates(ctx *Context) (updates []Updater, err error)
 
 // List the collection.
 func (r *StorageAdapter) List(ctx *Context, provider *api.Provider) (itr fb.Iterator, err error) {
-	storageName := fmt.Sprintf("Dummy storage for %s", provider.Name)
+	storageName := fmt.Sprintf("Dummy storage for source provider %s", provider.Name)
```
bkhizgiy, line 245: I'm not sure if the `provider` is needed, and maybe change to something like `dummy source provider storage <provider_name>`

ahadas: `Dummy storage for source provider %s` looks great

ahadas (approved):

PR #561, opened 2023-08-21
Fix secret admitter to allow OVA provider creation

bkhizgiy:
When a new provider being created in the first run we sets the fields of the secrete and in the second iteration the owner reference of the provider, since the second iteration counts as `update` operation the webhook blocks the provider creation.
This fix checks if the URL filed wasn't changed and if not the operation is permitted, this allow to the creation of the provider to pass successfully.

### DIFF
--- a/operator/config/rbac/api/role.yaml
+++ b/operator/config/rbac/api/role.yaml
@@ -27,4 +27,10 @@ rules:
     verbs:
       - get
       - list
+  - apiGroups:
+      - ""
+    resources:
+      - secrets
+    verbs:
+      - get
 
--- a/pkg/forklift-api/webhooks/validating-webhook/admitters/secret-admitter.go
+++ b/pkg/forklift-api/webhooks/validating-webhook/admitters/secret-admitter.go
@@ -76,16 +76,22 @@ func (admitter *SecretAdmitter) validateProviderSecret() *admissionv1.AdmissionR
 	if createdForProviderType, ok := admitter.secret.GetLabels()["createdForProviderType"]; ok {
 		providerType := api.ProviderType(createdForProviderType)
 
-		// The OVA secret has only name and URL fileds,
-		// since the name can't be changed, we assume that all update requests are for the URL.
+		// once OVA provider is creted we block URL changes.
 		if admitter.ar.Request.Operation == admissionv1.Update && providerType == api.Ova {
-			log.Info("secret admitter for OVA")
-			return &admissionv1.AdmissionResponse{
-				Allowed: false,
-				Result: &metav1.Status{
-					Code:    http.StatusBadRequest,
-					Message: "Updating the URL field of an existing OVA provider is forbidden.",
-				},
+			urlChanged, err := admitter.isOvaUrlChanged()
+			if err != nil {
+				return webhookutils.ToAdmissionResponseError(err)
+			}
+
+			if urlChanged {
+				log.Info("secret admitter for OVA")
+				return &admissionv1.AdmissionResponse{
+					Allowed: false,
+					Result: &metav1.Status{
+						Code:    http.StatusBadRequest,
+						Message: "Updating the URL field of an existing OVA provider is forbidden.",
+					},
+				}
 			}
 		}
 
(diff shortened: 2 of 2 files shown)


pkg/forklift-api/webhooks/validating-webhook/admitters/secret-admitter.go
```diff
@@ -76,16 +76,21 @@ func (admitter *SecretAdmitter) validateProviderSecret() *admissionv1.AdmissionR
 	if createdForProviderType, ok := admitter.secret.GetLabels()["createdForProviderType"]; ok {
 		providerType := api.ProviderType(createdForProviderType)
 
-		// The OVA secret has only name and URL fileds,
```
liranr23, line 79: so now we will be allowed to update only the secret name? what's the benefit? is the secret always provided in the request? if not we can use admitter.secret and drop the role changes + query.

operator/config/rbac/api/role.yaml
```diff
@@ -27,4 +27,11 @@ rules:
...
       - get
       - list
-
+  - apiGroups:
+      - ""
+    resources:
+      - secrets
+    verbs:
+      - get
+      - list
```
ahadas, line 36: why does the operator need this?

ahadas (commented):
@bkhizgiy please elaborate in the commit message and PR description on the issue - I don't understand how come that code that is executed when `admitter.ar.Request.Operation == admissionv1.Update` blocks creation of new OVA providers

ahadas:
> I don't understand how come that code that is executed when `admitter.ar.Request.Operation == admissionv1.Update` blocks creation of new OVA providers

@bkhizgiy is it because we update the owner reference?

pkg/forklift-api/webhooks/validating-webhook/admitters/secret-admitter.go
liranr23: ok, so if we have the new secret (only when user specifies it?) then, once we see it we know it's a change. of course it will be better if we check the previous and say: if it's still the same URL ok, go ahead. i wonder if we need it. user wise it's better as you do it now.

pkg/forklift-api/webhooks/validating-webhook/admitters/secret-admitter.go
```diff
@@ -227,3 +233,43 @@ func (admitter *SecretAdmitter) testConnectionToHost(hostName string) (tested bo
...
+	oldSecret := core.Secret{}
+	err = cl.Get(context.TODO(), client.ObjectKey{Namespace: admitter.secret.Namespace, Name: admitter.secret.Name}, &oldSecret)
+	if err != nil {
+		log.Error(err, "Couldn't get the target provider secrete")
+		return false, err
+	}
+
+	url := oldSecret.Data["url"]
+	newURL := admitter.secret.Data["url"]
+	if string(url) != string(newURL) {
```
ahadas, line 271: converting byte arrays to strings just for the comparison is redundant, how about replacing this and the lines below with:
```golang
return !bytes.Equal(url, newURL), nil
```

ahadas (commented):
please also fix the commit message (secrete -> secret, addmitter -> admitter)

ahadas (approved):

liranr23 (approved):

pkg/forklift-api/webhooks/validating-webhook/admitters/secret-admitter.go
```diff
@@ -227,3 +221,60 @@ func (admitter *SecretAdmitter) testConnectionToHost(hostName string) (tested bo
...
+	cl, err := client.New(config, client.Options{Scheme: scheme.Scheme})
+	if err != nil {
+		log.Error(err, "Couldn't create a cluster client")
+		return false, err
+	}
+
+	oldSecret := core.Secret{}
+	err = cl.Get(context.TODO(), client.ObjectKey{Namespace: admitter.secret.Namespace, Name: admitter.secret.Name}, &oldSecret)
+	if err != nil {
+		log.Error(err, "Couldn't get the target provider secrete")
```
liranr23, line 273:
```suggestion
		log.Error(err, "Couldn't get the target provider secret")
```

PR #566, opened 2023-08-24
Fix secret admitter to allow OVA provider creation

ahadas:
When a new provider being created in the first run we sets the fields of the secrete and in the second iteration the owner reference of the provider, since the second iteration counts as update operation the webhook blocks the provider creation.
This fix checks if the URL field wasn't changed and if not the operation is permitted, this allow to the creation of the provider to pass successfully.

backport of #561

### DIFF
--- a/operator/config/rbac/api/role.yaml
+++ b/operator/config/rbac/api/role.yaml
@@ -27,4 +27,10 @@ rules:
     verbs:
       - get
       - list
+  - apiGroups:
+      - ""
+    resources:
+      - secrets
+    verbs:
+      - get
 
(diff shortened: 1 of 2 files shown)


operator/config/rbac/api/role.yaml
```diff
@@ -27,4 +27,10 @@ rules:
     verbs:
       - get
       - list
+  - apiGroups:
+      - ""
+    resources:
+      - secrets
+    verbs:
+      - get
```
mnecas, line 35: @liranr23 Just adding a notification about this RBAC change, so the CSV needs to be regenerated/updated.

PR #569, opened 2023-08-27
openstack: update VM revision on volume change

bennyz:
If a volume of a VM changes we need to update the VM's revision to make sure it is revalidated

### DIFF
--- a/pkg/controller/provider/container/openstack/model.go
+++ b/pkg/controller/provider/container/openstack/model.go
@@ -10,6 +10,7 @@ import (
 	fb "github.com/konveyor/forklift-controller/pkg/lib/filebacked"
 	libmodel "github.com/konveyor/forklift-controller/pkg/lib/inventory/model"
 	"github.com/konveyor/forklift-controller/pkg/lib/logging"
+	"k8s.io/klog/v2"
 )
 
 // All adapters.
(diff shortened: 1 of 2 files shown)


pkg/controller/provider/container/openstack/model.go
```diff
@@ -732,11 +734,36 @@ func (r *VolumeAdapter) GetUpdates(ctx *Context) (updates []Updater, err error)
...
+						vmID := attachment.ServerID
+						vm := &model.VM{
+							Base: model.Base{ID: vmID},
+						}
+						err = tx.Get(vm)
+						if err != nil {
+							klog.Info("VM not found, skipping", "vmID", vmID)
+							continue
+						}
+						vm.Revision += 1
```
ahadas, line 758: don't we risk missing an update this way? I mean, if the vm is also updated in openstack we'll get that the revision didn't change - maybe we should decrease the revision or set it to zero?

bennyz: I can set vm.RevisionValidated to 0 instead

pkg/controller/provider/container/openstack/model.go
```diff
@@ -741,7 +741,7 @@ func (r *VolumeAdapter) GetUpdates(ctx *Context) (updates []Updater, err error)
 
 				volume.ApplyTo(m)
 				err = tx.Update(m)
-				if err == nil {
+				if err != nil {
```
ahadas, line 744: shouldn't it be `err == nil`?

bennyz: yeah... I changed it to see if it reproduces without the fix and forgot to remove before the second commit

ahadas: ah ok

ahadas: and did reproduce? :)

bennyz: yeah 😅

ahadas (approved):

PR #572, opened 2023-08-28
openstack: block image based VMs

bennyz:
They'll currently fail to boot


validation/policies/io/konveyor/forklift/openstack/image_format.rego
ahadas, line 1: maybe we should comment out the part that might set `has_invalid_image_format` to true as we will want this later on, wdyt?

validation/policies/io/konveyor/forklift/openstack/volume_based_test.rego
```diff
@@ -1,25 +1,25 @@
...
+		"imageID": "b749c132-bb97-4145-b86e-a1751cf75e21",
 		"image": {
 			"id": "b749c132-bb97-4145-b86e-a1751cf75e21",
-			"diskFormat": "qcow2",
 		},
 	}
 	results := concerns with input as mock_vm
-	count(results) == 0
+	count(results) == 1
 }
```
ahadas, line 13: aren't we supposed to have two concerns here when the status is not ACTIVE?

validation/policies/io/konveyor/forklift/openstack/image_format.rego
bennyz: I think that once we implement it there shouldn't be any case where the format is invalid

validation/policies/io/konveyor/forklift/openstack/volume_based_test.rego
bennyz: which status? the test passes

ahadas: the status of the VM, we previously had:
```
mock_vm := {
		"name": "test",
		"status": "ACTIVE",
	}
```

validation/policies/io/konveyor/forklift/openstack/image_format.rego
ahadas: yeah, it depends on how we'll implement  - if we anyway "clone" the image then we won't need it, but if the image is of type raw, why not to download it as-is?

validation/policies/io/konveyor/forklift/openstack/volume_based_test.rego
bennyz: ah, that's because there's no status, if it's there and incorrect it would fail

ahadas: okok, in that case it's fine

ahadas (approved):

PR #574, opened 2023-08-28
ova: fix migration with multiple disks

bkhizgiy:
When starting a migration each disk that needed to be migrated should be assigned with task name to follow with it progress and switch between the phases, in the current implantation all disks are assigned with the same task name (the ova/folder path) this causing the first disk to finish successfully but for all other to be stuck in an endless loop.
This fix change the task name to be unique for each disk and composes from path + disk name as it appearing in the OVF file.

### DIFF
--- a/pkg/controller/plan/adapter/ova/builder.go
+++ b/pkg/controller/plan/adapter/ova/builder.go
@@ -59,12 +59,6 @@ const (
 	Unknown = "unknown"
 )
 
-// Annotations
-const (
-	// CDI import backing file annotation on PVC
-	AnnImportBackingFile = "cdi.kubevirt.io/storage.import.backingFile"
-)
-
 // Error messages
 const (
 	ErrVMLookupFailed = "VM lookup failed."
@@ -209,7 +203,7 @@ func (r *Builder) DataVolumes(vmRef ref.Ref, secret *core.Secret, _ *core.Config
 			if dv.ObjectMeta.Annotations == nil {
 				dv.ObjectMeta.Annotations = make(map[string]string)
 			}
-			dv.ObjectMeta.Annotations[planbase.AnnDiskSource] = getDiskSourcePath(disk.FilePath)
+			dv.ObjectMeta.Annotations[planbase.AnnDiskSource] = getDiskFullPath(disk)
 			dvs = append(dvs, *dv)
 		}
 	}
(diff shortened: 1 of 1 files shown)


pkg/controller/plan/adapter/ova/builder.go
```diff
@@ -387,10 +387,6 @@ func (r *Builder) mapDisks(vm *model.VM, persistentVolumeClaims []core.Persisten
 	var kDisks []cnv.Disk
 
 	disks := vm.Disks
-	// TODO might need sort by the disk id (incremental name)
-	/*sort.Slice(disks, func(i, j int) bool {
-		return disks[i].Key < disks[j].Key
-	})*/
```
ahadas, line 393: can we also drop AnnImportBackingFile that is used below?

pkg/controller/plan/adapter/ova/builder.go
```diff
@@ -500,6 +496,10 @@ func trimBackingFileName(fileName string) string {
 	return backingFilePattern.ReplaceAllString(fileName, ".vmdk")
 }
 
+func getDiskFullPath(disk ova.Disk) string {
+	return disk.FilePath + "/" + disk.Name
```
ahadas, line 500: so this will result in something like `/folder/my.ova/myDisk`? in that case maybe it would be better to change it with something like `/folder/my.ova::myDisk`?

pkg/controller/plan/adapter/ova/builder.go
liranr23: the sort is supposed to be (if needed, because i didn't test it) - in cases where we have multiple disks and we get the wrong disk the the one with operation system which may cause to manual changing the disks order/bootable.

bkhizgiy: Its not needed because its already sorted in the ovf file so its being saved like that in the inventory (tested it).
The only thing i'm checking is if the first disk in the ovf is really the bootable one always.

bkhizgiy: @ahadas cool, so just verifying the AnnImportBackingFile change and uploading the changes.

ahadas (commented):
@bkhizgiy does it mean we return empty string from ResolvePersistentVolumeClaimIdentifier? in that case, we better return empty string explicitly
also let's drop the `// Return a stable identifier for a VDDK DataVolume.` part? vddk is not used...

ahadas (approved):

PR #575, opened 2023-08-29
backport: openstack: block image based VMs

bennyz:
https://github.com/kubev2v/forklift/pull/572


ahadas (commented):
@bennyz don't we need to drop the validation of image-format here as well?

bennyz:
> @bennyz don't we need to drop the validation of image-format here as well?

yes, not sure why the cherry pick dropped it

ahadas (approved):

PR #576, opened 2023-08-29
Fix migrations to restricted namespaces

### LINKED ISSUE #173: Conversion pod fails in restricted namespaces

liranr23:
In order to run on restricted namespaces, the v2v/populator pods need to be set with securityContext.seccompProfile.type that is either RuntimeDefault or Localhost. Thus, we set it in the security context constraints of the service account forklift-controller.
This PR also introduces a new service account for the forklift-populator-controller pod. Currently, it uses the same RBAC as forklift-controller service account. Later on, we will restrict the RBAC to what the populator-controller needs.

Fixes #173

### DIFF
--- a/operator/BUILD.bazel
+++ b/operator/BUILD.bazel
@@ -80,7 +80,7 @@ genrule(
     cmd = """
         cd operator;
         export DATE=$$(date +%Y-%m-%dT%H:%M:%SZ);
-        ../$(location :kustomize_bin) build config/manifests --load_restrictor LoadRestrictionsNone | envsubst | ../$(location @operator-sdk//file) generate bundle -q --overwrite --extra-service-accounts forklift-controller,forklift-api --version $${VERSION} --output-dir ../$(RULEDIR)/bundle --channels=$${CHANNELS} --default-channel=$${DEFAULT_CHANNEL}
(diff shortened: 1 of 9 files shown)


ahadas (commented):
@liranr23 I've changed the title and description of the PR - fixed a typo and improved phrasing, please change the commit message as well.

ahadas (approved):

ahadas:
@liranr23 please backport

PR #577, opened 2023-08-29
Tests: make the namespaces restricted

liranr23:
This change will make the target namespaces of the migrations to be restricted.

### DIFF
--- a/tests/suit/framework/framework.go
+++ b/tests/suit/framework/framework.go
@@ -180,9 +180,8 @@ func (f *Framework) CreateNamespace(prefix string, labels map[string]string) (*v
 	if labels == nil {
 		labels = make(map[string]string)
 	}
-	// TODO: restore the pod-security label once https://github.com/kubev2v/forklift/issues/173 is fixed
 	// pod-security.kubernetes.io/<MODE>: <LEVEL>
-	//labels["pod-security.kubernetes.io/enforce"] = "restricted"
+	labels["pod-security.kubernetes.io/enforce"] = "restricted"
 	ns := &v1.Namespace{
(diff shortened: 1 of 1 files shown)


ahadas:
the openstack tests failed with the expected error:
`[pods \"plan-test-3aa1b8f3-d221-4cbc-b055-16adb8b8a868-pvcinit-v2kqj\" is forbidden: violates PodSecurity \"restricted:latest\": seccompProfile (pod or container \"main\" must set securityContext.seccompProfile.type to \"RuntimeDefault\" or \"Localhost\")]`

eslutsky:
> /test-osp-extended

looks like this  ansible job failed with :
```
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: ValueError: Template with name 'fedora-36-gh_forklift_kubev2v' and version 'None' in data center 'OCP-45' was not found
```
there is a periodic  cleanup job on the CI that removes  the old template, this template can be re-created with this action ,

initially it tried to create this template but failed with missing access_token:
```
https://github.com/kubev2v/forklift/actions/runs/6021140224/job/16333519484#step:6:413
```

@ahadas ,we need to add this token to the repo secrets.. 




we can

ahadas:
@eslutsky thanks for looking at it, we didn't change anything in terms of the secrets as far as I remember - which access token is missing?

ahadas:
ok, updated the token, trying again
/test-osp-extended

eslutsky:
> ok, updated the token, trying again /test-osp-extended

looks like its worked, your  the  first attempt  failed because it had some leftovers from previous runs that  didn't cleanup properly . I've removed the temp-vm and run it again.

PR #579, opened 2023-08-30
Fix migrations to restricted namespaces (backport)

liranr23:
In order to run on restricted namespaces, the v2v/populator pods need to be set with securityContext.seccompProfile.type that is either RuntimeDefault or Localhost. Thus, we set it in the security context constraints of the service account forklift-controller.
This PR also introduces a new service account for the forklift-populator-controller pod. Currently, it uses the same RBAC as forklift-controller service account. Later on, we will restrict the RBAC to what the populator-controller needs.


ahadas:
backport of #576

ahadas (approved):

PR #581, opened 2023-08-31
Tests: use restricted namespaces

liranr23:
Until now the target namespace for the migration in the tests were generated by the plan without labeling them. This change will create the target namespace as restricted.


ahadas (approved):
it would be nice to generate the name of the target namespace as well so we could execute the tests multiple times locally without clean up steps, but that's out of scope of this PR

liranr23:
> it would be nice to generate the name of the target namespace as well so we could execute the tests multiple times locally without clean up steps, but that's out of scope of this PR

we can easily do it using the namespace we create the CRs in (it happens in other tests). but i figured we wish to do it as: CRs in one NS and target NS is another one.

ahadas:
> > it would be nice to generate the name of the target namespace as well so we could execute the tests multiple times locally without clean up steps, but that's out of scope of this PR
> 
> we can easily do it using the namespace we create the CRs in (it happens in other tests). but i figured we wish to do it as: CRs in one NS and target NS is another one.

yeah, it's better to have them separated since it's closer to real-world scenario

PR #585, opened 2023-09-03
Disable EMS check without changing /usr/share

### LINKED ISSUE #556: Change FIPS crypto-policy without changing /usr/share
> So this change to FIPS policy happens automatically and we can't (or don't want) to have the FIPS policy configured by default in the container (i.e. during build). Is that correct?

              right, it happens automatically based on the fips mode of the host

              > And if it happens automatically we cannot apply a (sub-)policy to override the configuration. Do I understand it correctly?

              yeah. maybe we can drop the default configuration of FIPS and introduce a new crypto policy called FIPS with different configuration - but I want to have a quick solution for 2.5, we can look at something like this later

_Originally posted by @ahadas in https://github.com/kubev2v/forklift/issues/538#issuecomment-1678809776_

ahadas:
These changes disable EMS checks when FIPS is enabled by changing configuration files in /etc.

The link /etc/pki/tls/fips_local.cnf that points to a configuration file within the FIPS crypto policy is replaced with a regular file, and Options=RHNoEnforceEMSinFIPS is added to the crypto_policy section in /etc/pki/tls/openssl.cnf

Fix #556

### DIFF
--- a/virt-v2v/cold/BUILD.bazel
+++ b/virt-v2v/cold/BUILD.bazel
@@ -63,9 +63,9 @@ container_image(
 container_run_and_commit_layer(
     name = "no-ems-in-fips",
     commands = [
-        "touch /usr/share/crypto-policies/back-ends/FIPS/openssl_fips.config",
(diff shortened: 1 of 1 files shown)


ahadas:
migrations on FIPS-enabled cluster work fine with these changes

liranr23 (approved):

PR #586, opened 2023-09-06
OVA: Change the scanning for ova and ovf files inside the NFS directory

bkhizgiy:
With the current change the files will be scanned as followed:

+ For OVA files we will search the zero level (the NFS itself) and first level (the share sub folders).
For example giving the nfs share is /nfs these files will be selected:  
     -  /nfs/vm.ova 
     -  /nfs/ova/vm.ova

+ For extracted OVA files (ovf format) we will search the first level (the NFS sub folders) and second level (the sub folders sub folders).
For example giving the nfs share is /nfs these files will be selected:  
     -  /nfs/vm/vm.ovf 
     -  /nfs/dc/vm1/vm1.ovf


cmd/ova-provider-server/ova-provider-server.go
```diff
@@ -299,28 +301,56 @@ func findOVAFiles(directory string) ([]string, []string, error) {
...
 
 	var ovaFiles, ovfFiles []string
+
+	// Check for .ova files in the main directory (zero level)
+	for _, child := range childs {
+		path := filepath.Join(directory, child.Name())
+		if hasSuffixIgnoreCase(child.Name(), OvaExt) {
+			ovaFiles = append(ovaFiles, path)
+		}
+	}
```
liranr23, line 311: why we need this loop while having the same loop below? can't we drop the `if !child.IsDir() {` and deal with this case?

cmd/ova-provider-server/ova-provider-server.go
```diff
@@ -299,28 +301,56 @@ func findOVAFiles(directory string) ([]string, []string, error) {
...
+					}
+				}
 			}
 		}
 	}
 	return ovaFiles, ovfFiles, nil
 }
 
+// Checks if the given file has the desired extension
+func hasSuffixIgnoreCase(fileName, suffix string) bool {
```
ahadas, line 350: how about having two functions isOVA and isOVF that either replace this function or calls it?

cmd/ova-provider-server/ova-provider-server.go
```diff
@@ -299,28 +301,56 @@ func findOVAFiles(directory string) ([]string, []string, error) {
...
-			path := filepath.Join(directory, child.Name(), file.Name())
+			path := filepath.Join(newDir, file.Name())
 			switch {
-			case strings.HasSuffix(strings.ToLower(file.Name()), ".ova"):
+			case hasSuffixIgnoreCase(file.Name(), OvaExt):
 				ovaFiles = append(ovaFiles, path)
-			case strings.HasSuffix(strings.ToLower(file.Name()), ".ovf"):
+			case hasSuffixIgnoreCase(file.Name(), OvfExt):
 				ovfFiles = append(ovfFiles, path)
+			case file.IsDir():
```
ahadas, line 330: what if there's a folder named `arik.ova`?
I'd expect this to be the first check

cmd/ova-provider-server/ova-provider-server.go
```diff
@@ -299,28 +301,56 @@ func findOVAFiles(directory string) ([]string, []string, error) {
...
+				// Check for .ovf files in subdirectories (second level)
+				subDir := filepath.Join(newDir, file.Name())
+				subFiles, err := os.ReadDir(subDir)
+				if err != nil {
+					continue
```
ahadas, line 335: shall we add a log?

cmd/ova-provider-server/ova-provider-server.go
```diff
@@ -299,28 +301,56 @@ func findOVAFiles(directory string) ([]string, []string, error) {
...
+		}
+	}
+
+	// Check for .ova and .ovf files in child directories (first level)
 	for _, child := range childs {
 		if !child.IsDir() {
 			continue
 		}
-		newDir := directory + "/" + child.Name()
+		newDir := filepath.Join(directory, child.Name())
```
ahadas, line 318: yep

cmd/ova-provider-server/ova-provider-server.go
ahadas: +1

cmd/ova-provider-server/ova-provider-server.go
```diff
@@ -299,28 +301,56 @@ func findOVAFiles(directory string) ([]string, []string, error) {
 	}
 
 	var ovaFiles, ovfFiles []string
+
+	// Check for .ova files in the main directory (zero level)
+	for _, child := range childs {
```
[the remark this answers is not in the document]
ahadas: +1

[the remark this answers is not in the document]
bkhizgiy: changed the  func to use `walkDir` so it simplified the logic a lot, now only need to be limit the levels of directories that being scan to max 3 for ovf and 2 for ova.

cmd/ova-provider-server/ova-provider-server.go
bkhizgiy: moved to the beginning.

cmd/ova-provider-server/ova-provider-server.go
bkhizgiy: added print for the error, we generally need to change all the prints to logs in this file, will upload a separate refactor for that.

cmd/ova-provider-server/ova-provider-server.go
```diff
@@ -293,34 +295,52 @@ func scanOVAsOnNFS() (envelopes []Envelope, ovaPaths []string) {
...
+
+		case (depth >= ovaMaxDepth && depth <= ovaMaxDepth+1) && isOvf(info.Name()):
+			ovfFiles = append(ovfFiles, path)
+		}
+
+		return nil
+	})
+
+	if err != nil {
+		fmt.Println("Error scanning OVA and OVF files:  ", err)
```
ahadas, line 325: why using `fmt`? it's better to use the logger

cmd/ova-provider-server/ova-provider-server.go
```diff
@@ -293,34 +295,52 @@ func scanOVAsOnNFS() (envelopes []Envelope, ovaPaths []string) {
 }
 
 func findOVAFiles(directory string) ([]string, []string, error) {
-	childs, err := os.ReadDir(directory)
-	if err != nil {
-		return nil, nil, err
-	}
-
 	var ovaFiles, ovfFiles []string
```
ahadas, line 298: how about naming the out parameters in the signature and dropping this part?

cmd/ova-provider-server/ova-provider-server.go
```diff
@@ -293,34 +295,52 @@ func scanOVAsOnNFS() (envelopes []Envelope, ovaPaths []string) {
...
-	for _, child := range childs {
-		if !child.IsDir() {
-			continue
-		}
-		newDir := directory + "/" + child.Name()
-		files, err := os.ReadDir(newDir)
+	var ovaMaxDepth = 2
+
+	err := filepath.WalkDir(directory, func(path string, info os.DirEntry, err error) error {
```
ahadas, line 301: how about extracting the function and adding unit tests for it?

cmd/ova-provider-server/ova-provider-server.go
```diff
@@ -293,34 +295,52 @@ func scanOVAsOnNFS() (envelopes []Envelope, ovaPaths []string) {
...
 		}
+
+		relativePath, _ := filepath.Rel(directory, path)
+		depth := len(strings.Split(relativePath, string(filepath.Separator)))
+
+		switch {
+		case (depth <= ovaMaxDepth) && isOva(info.Name()):
+			ovaFiles = append(ovaFiles, path)
+
+		case (depth >= ovaMaxDepth && depth <= ovaMaxDepth+1) && isOvf(info.Name()):
```
ahadas, line 317: there's a future scenario in which users upload an ova and then import it. the ova will usually be a tar file in this case, but for completeness - we can also inspect ovf files in the root folder

cmd/ova-provider-server/ova-provider-server.go
ahadas: ah ok, already commented about the log on the latest changes but sure, np

cmd/ova-provider-server/ova-provider-server.go
ahadas: yeah, it's different than what we talked about in the past - but with the current mechanism, it seems we can achieve this relatively easily and that would be even better than the solution we did in oVirt

bkhizgiy: cool, changing to match

cmd/ova-provider-server/ova-provider-server.go
```diff
@@ -292,35 +294,52 @@ func scanOVAsOnNFS() (envelopes []Envelope, ovaPaths []string) {
...
 		}
+
+		return nil
+	})
+
+	if err != nil {
+		fmt.Println("Error scanning OVA and OVF files:  ", err)
+		return nil, nil, err
 	}
 	return ovaFiles, ovfFiles, nil
```
ahadas, line 327: we can replace this with a simple `return` now

cmd/ova-provider-server/nfs_test.go
```diff
@@ -0,0 +1,103 @@
...
+			},
+			expectedOVAs: nil,
+			expectedOVFs: nil,
+			expectError:  false,
+		},
+		{
+			name: "incorrect depth",
+			setup: func(directory string) {
+				_ = os.Mkdir(filepath.Join(directory, "subdir1"), 0755)
+				_ = os.Mkdir(filepath.Join(directory, "subdir1", "subdir2"), 0755)
```
ahadas, line 59: how about using MkdirAll instead (also above and below)?

cmd/ova-provider-server/nfs_test.go
```diff
@@ -0,0 +1,103 @@
...
+				_ = ioutil.WriteFile(filepath.Join(directory, "subdir1", "subdir2", "test3.ova"), []byte{}, 0644)
+			},
+			expectedOVAs: nil,
+			expectedOVFs: nil,
+			expectError:  false,
+		},
+		{
+			name: "folder with extantion",
```
ahadas, line 67:
```suggestion
			name: "folder with extension",
```

cmd/ova-provider-server/nfs_test.go
```diff
@@ -0,0 +1,103 @@
...
+			setup: func(directory string) {
+				_ = os.Mkdir(filepath.Join(directory, "subdir1.ova"), 0755)
+				_ = os.Mkdir(filepath.Join(directory, "subdir2.ovf"), 0755)
+			},
+			expectedOVAs: nil,
+			expectedOVFs: nil,
```
ahadas, line 73: why don't we expect to get the ovf?

ahadas: ah, subdir2.ovf is a directory.. ok

bkhizgiy: yeah its a directory, and good idea, i"ll add this case.

cmd/ova-provider-server/nfs_test.go
ahadas: and while on it
```suggestion
				os.Mkdir(filepath.Join(directory, "subdir1", "subdir2"), 0755)
```

liranr23 (approved):

cmd/ova-provider-server/nfs_test.go
```diff
@@ -0,0 +1,114 @@
...
+		expectedOVAs []string
+		expectedOVFs []string
+		expectError  bool
+	}{
+		{
+			name: "basic structure",
+			setup: func(directory string) {
+				ioutil.WriteFile(filepath.Join(directory, "test.ova"), []byte{}, 0644)
+				ioutil.WriteFile(filepath.Join(directory, "test.ovf"), []byte{}, 0644)
+				os.MkdirAll(filepath.Join(directory, "subdir1"), 0755)
```
ahadas, line 27:
```suggestion
				os.MkdirAll(filepath.Join(directory, "subdir1", "subdir2"), 0755)
```
and I think it will be more structured if we place the creation of the directory structure at the beginning - for your consideration

cmd/ova-provider-server/nfs_test.go
```diff
@@ -0,0 +1,114 @@
...
+				ioutil.WriteFile(filepath.Join(directory, "subdir1", "test1.ova"), []byte{}, 0644)
+				ioutil.WriteFile(filepath.Join(directory, "subdir1", "test1.ovf"), []byte{}, 0644)
+				os.MkdirAll(filepath.Join(directory, "subdir1", "subdir2"), 0755)
```
ahadas, line 30: and then we can drop this line

cmd/ova-provider-server/nfs_test.go
```diff
@@ -0,0 +1,114 @@
...
+
+			ovaFiles, ovfFiles, err := findOVAFiles(testDir)
+			if testCase.expectError {
+				g.Expect(err).To(gomega.HaveOccurred())
+			} else {
+				g.Expect(err).To(gomega.BeNil())
+				g.Expect(ovaFiles).To(gomega.ConsistOf(testCase.expectedOVAs))
+				g.Expect(ovfFiles).To(gomega.ConsistOf(testCase.expectedOVFs))
+			}
+			_ = os.RemoveAll(testDir)
```
ahadas, line 111:
```suggestion
			os.RemoveAll(testDir)
```

cmd/ova-provider-server/nfs_test.go
```diff
@@ -0,0 +1,114 @@
+package main
+
+import (
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/onsi/gomega"
```
ahadas, line 9:
```suggestion
	. "github.com/onsi/gomega"
```
and then we can drop `gomega` below

ahadas (approved):

PR #588, opened 2023-09-07
Clean importer pods on success

liranr23:
When running with a migration that uses CDI, e.g warm migration, it may have multiple importer pods. This patch will clean up all of them for the related VM migrated once the migration completes successfully.


pkg/controller/plan/kubevirt.go
```diff
@@ -235,27 +266,29 @@ func (r *KubeVirt) DeleteDataVolumes(vm *plan.VMStatus) (err error) {
...
 
-// Delete the importer pod for a PersistentVolumeClaim.
-func (r *KubeVirt) DeleteImporterPod(pvc core.PersistentVolumeClaim) (err error) {
-	var pod *core.Pod
+// Delete the importer pods for a PersistentVolumeClaim.
+func (r *KubeVirt) DeleteImporterPods(pvc core.PersistentVolumeClaim) (err error) {
+	var pods []core.Pod
 	var found bool
-	pod, found, err = r.GetImporterPod(pvc)
+	pods, found, err = r.GetImporterPods(pvc)
```
ahadas, line 273: how about
```suggestion
	pods, found, err := r.GetImporterPods(pvc)
```
and drop the variable declarations above?

pkg/controller/plan/kubevirt.go
```diff
@@ -235,27 +266,29 @@ func (r *KubeVirt) DeleteDataVolumes(vm *plan.VMStatus) (err error) {
...
 	}
-	err = r.Destination.Client.Delete(context.TODO(), pod)
-	if err != nil {
-		err = liberr.Wrap(err)
-		return
+	for _, pod := range pods {
+		err = r.Destination.Client.Delete(context.TODO(), &pod)
+		if err != nil {
+			err = liberr.Wrap(err)
+			return
```
ahadas, line 281: I don't see a scenario that we'll fail to remove only one of the pods but wouldn't it be better to try to proceed with removing the other pods?

pkg/controller/plan/migration.go
```diff
@@ -757,6 +757,11 @@ func (r *Migration) execute(vm *plan.VMStatus) (err error) {
 			err = liberr.Wrap(err)
 			return
 		}
+		err = r.deleteImporterPods(vm)
+		if err != nil {
+			err = liberr.Wrap(err)
+			return
```
ahadas, line 763: also here, and when handling error for the calls above, why do we return?
when we return an error the reconcile fails and we'll try again, right? in that case, I'd proceed with try to remove all leftovers (data volumes, init pods, importer pods) and return an error if any of them failed

pkg/controller/plan/kubevirt.go
```diff
@@ -219,6 +219,37 @@ func (r *KubeVirt) GetImporterPod(pvc core.PersistentVolumeClaim) (pod *core.Pod
...
+			Namespace:     r.Plan.Spec.TargetNamespace,
+			LabelSelector: labels.SelectorFromSet(map[string]string{"app": "containerized-data-importer"}),
+		},
+	)
+	if err != nil {
+		err = liberr.Wrap(err)
+		return
+	}
+	for _, pod := range podList.Items {
+		if strings.Contains(pod.Name, pvc.Annotations["vmID"]) {
```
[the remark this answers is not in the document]
ahadas: looking at an existing importer pod and [here](https://github.com/kubevirt/containerized-data-importer/blob/18e6607f06d177abcb6b5efccc5e96bf59fb7bda/pkg/controller/import-controller.go#L1046-L1050), probably not - so if we cannot extend the label selector above, can we use a field selector?

pkg/controller/plan/migration.go
liranr23: i changed it a bit.

pkg/controller/plan/kubevirt.go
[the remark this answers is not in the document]
liranr23: having more thought - this case may be problematic:
2 plans: A and B
in plan A we have a vm called: "liran"
in plan B we have a vm called: "liranr"

we ran plan B and it failed
we ran plan A and it passed
it will remove also the importer pod of plan B.

maybe there is another option to know which importer pods we need. but i am unsure about it.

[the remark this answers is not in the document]
ahadas: > i don't think we can. because it won't be equal to the pod name, it just look if the `vmID` is part of it.

yeah, I meant a field selector that selects by name and uses the ID of the VM with wildcards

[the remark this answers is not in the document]
liranr23: AFAIK, regex isn't supported (unless you use tools like `jq`)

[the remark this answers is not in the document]
liranr23: This options may fit: 
Change the string we look for to `importer-<plan_name>-<vmID>` prefix. that should limit the importer pods to the specific plan and specific vm.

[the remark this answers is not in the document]
ahadas: @liranr23 did you try something like they do with the namespaces [here](https://github.com/kubernetes/kubernetes/issues/107053#issuecomment-1193617828)?

pkg/controller/plan/kubevirt.go
```diff
@@ -235,27 +266,36 @@ func (r *KubeVirt) DeleteDataVolumes(vm *plan.VMStatus) (err error) {
...
 	}
-	r.Log.Info(
-		"Deleted importer pod.",
-		"pod",
-		path.Join(
-			pod.Namespace,
-			pod.Name),
-		"pvc",
-		pvc.Name)
 	return
```
ahadas, line 299: if we return an error if any of the pods couldn't be removed, we'll try again?

pkg/controller/plan/kubevirt.go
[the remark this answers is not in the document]
liranr23: yes, that doesn't work(at least using kubectl/oc). the issue he mentioned there is closed due to inactivity.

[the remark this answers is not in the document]
ahadas: ok, on the bright side - the issue was accepted so we may get a fix later on
so as a temporary solution it's ok, but how about proposing to CDI the propagate labels from the DV to the importer pods to make it easier for us? it's done in a similar way with the virt-launcher pods

pkg/controller/plan/kubevirt.go
```diff
@@ -219,6 +219,37 @@ func (r *KubeVirt) GetImporterPod(pvc core.PersistentVolumeClaim) (pod *core.Pod
...
+			Namespace:     r.Plan.Spec.TargetNamespace,
+			LabelSelector: labels.SelectorFromSet(map[string]string{"app": "containerized-data-importer"}),
+		},
+	)
+	if err != nil {
+		err = liberr.Wrap(err)
+		return
+	}
+	for _, pod := range podList.Items {
+		if strings.Contains(pod.Name, fmt.Sprintf("importer-%s-%s", r.Plan.Name, pvc.Annotations[kVM])) {
```
bennyz, line 244: can we propagate labels to the pod via the DV? if so, it will not help now, but perhaps in the future

liranr23: we discussed it here:
https://github.com/kubev2v/forklift/pull/588#discussion_r1318701734
and i opened issue for CDI for this feature.

pkg/controller/plan/kubevirt.go
ahadas: ok

pkg/controller/plan/kubevirt.go
```diff
@@ -219,6 +219,37 @@ func (r *KubeVirt) GetImporterPod(pvc core.PersistentVolumeClaim) (pod *core.Pod
...
+			pods = append(pods, pod)
+		}
+	}
+
+	found = len(pods) > 0
```
ahadas, line 249: why do we need the `found` variable? seems we can drop it..

pkg/controller/plan/kubevirt.go
```diff
@@ -219,6 +219,37 @@ func (r *KubeVirt) GetImporterPod(pvc core.PersistentVolumeClaim) (pod *core.Pod
 	return
 }
 
+// Get the importer pods for a PersistentVolumeClaim.
+func (r *KubeVirt) GetImporterPods(pvc core.PersistentVolumeClaim) (pods []core.Pod, found bool, err error) {
+	pods = []core.Pod{}
+	if pvc.Annotations[AnnImporterPodName] == "" {
```
ahadas, line 225: how about making this more explicit with:
```suggestion
	if _, ok := pvc.Annotations[AnnImporterPodName]; !ok {
```

pkg/controller/plan/migration.go
```diff
@@ -748,14 +748,15 @@ func (r *Migration) execute(vm *plan.VMStatus) (err error) {
 		// Removing unnecessary DataVolumes
 		err = r.kubevirt.DeleteDataVolumes(vm)
 		if err != nil {
-			step.AddError(err.Error())
```
ahadas, line 751: how about separating out the changes here and below to a different PR? I don't understand why skipping the step.AddError and the changes below also don't make much sense (setting error in line 751 and then proceeding to 753 that could override it)

ahadas:
@liranr23 I also rephrased the description of the PR, please change the commit message as well

pkg/controller/plan/kubevirt.go
```diff
@@ -219,6 +219,36 @@ func (r *KubeVirt) GetImporterPod(pvc core.PersistentVolumeClaim) (pod *core.Pod
 	return
 }
 
+// Get the importer pods for a PersistentVolumeClaim.
+func (r *KubeVirt) GetImporterPods(pvc core.PersistentVolumeClaim) (pods []core.Pod, err error) {
+	pods = []core.Pod{}
```
ahadas, line 224: that's redundant, no?

pkg/controller/plan/kubevirt.go
```diff
@@ -219,6 +219,36 @@ func (r *KubeVirt) GetImporterPod(pvc core.PersistentVolumeClaim) (pod *core.Pod
...
+	if _, ok := pvc.Annotations[AnnImporterPodName]; !ok {
+		return
+	}
+
+	podList := &core.PodList{}
+
```
ahadas, line 230: please drop this empty line

ahadas (commented):
the comments inside are rather minor but this PR includes two changes:
1. remove multiple importer pods by a PVC
2. remove importer pods when posting the vm so that the PV and PVC would be removed when removing the VM

only the first part is documented in the commit message
I'd rather separate this PR to two different PRs for each of them but it's also ok to extend the commit message to cover both..

ahadas:
@liranr23 the changes look good
I've updated the description of the PR based on the commit message and fixes the phrasing - please change the commit message

ahadas (approved):

PR #589, opened 2023-09-12
openstack: do not use region when application credential is used

bennyz:


ahadas:
@bennyz so how do we go forward with this one? shall we change the UI to set the region?

bennyz:
> @bennyz so how do we go forward with this one? shall we change the UI to set the region?

yeah

PR #590, opened 2023-09-12
Clean up PVCs when canceling populator flows

liranr23:
This patch will clean up the PVCs (final and prime) associated with VMs that the migration was cancled for them.


ahadas (commented):
we can move our offline discussion about specific handling for canceled migrations here - why do we need special handling, isn't it a concrete case of a migration that completes (unsuccessfully) before posting a vm?

pkg/controller/plan/kubevirt.go
```diff
@@ -893,6 +893,29 @@ func (r *KubeVirt) SetPopulatorPodOwnership(vm *plan.VMStatus) (err error) {
 	return
 }
 
+// DeletePVCs deletes PVCs associated with the VM, including prime PVCs (populator flows)
+func (r *KubeVirt) DeletePVCs(vm *plan.VMStatus) (err error) {
```
liranr23, line 897: could be nicer to move it into destination client

liranr23:
> we can move our offline discussion about specific handling for canceled migrations here - why do we need special handling, isn't it a concrete case of a migration that completes (unsuccessfully) before posting a vm?

yup. it means the vm doesn't exist, there is no owner to the PVCs.

ahadas:
> yup. it means the vm doesn't exist, there is no owner to the PVCs.

ok, but how is it different from a migration that fails during disk transfer?

> it will break a restart tryout 

what the reason 'restart' would fail in this case and work for a migration that failed during disk transfer?

> and if you just cancel because you don't wish to continue you will have leftovers.

I don't see this as a problem - it's ok to have leftovers as long as the migration still exists

liranr23:
> > yup. it means the vm doesn't exist, there is no owner to the PVCs.
> 
> ok, but how is it different from a migration that fails during disk transfer?
> 
it doesn't.
> > it will break a restart tryout
> 
> what's the reason that 'restart' would fail in this case and work for a migration that failed during disk transfer?
> 
i need to check it. but if we have the PVC and try to recreate them i think we will fail. migration that failed can have the same problem.
> > and if you just cancel because you don't wish to continue you will have leftovers.
> 
> I don't see this as a problem - it's ok to have "leftovers" as long as the migration exists, it's important to remove everything when the plan is removed
you suggest having it on every cleanup flow? e.g based on the PVC has ownership?

ahadas:
> > I don't see this as a problem - it's ok to have "leftovers" as long as the migration exists, it's important to remove everything when the plan is removed
> 
> you suggest having it on every cleanup flow? e.g based on the PVC has ownership?

yeah, assuming that by "cleanup flow" you refer to what we do before restarting a plan or during archive operaton

pkg/controller/plan/kubevirt.go
liranr23: looked into it, it will causing code duplication from the kubevirt into 2 destination clients (ovirt and openstack).

liranr23:
> > > I don't see this as a problem - it's ok to have "leftovers" as long as the migration exists, it's important to remove everything when the plan is removed
> > 
> > 
> > you suggest having it on every cleanup flow? e.g based on the PVC has ownership?
> 
> yeah, assuming that by "cleanup flow" you refer to what we do before restarting a plan or during archive operation

Done.
Overall the main issue is that `RestartLimitExceeded` function doesn't work on the populator pod. The pod is terminated and not restarting, causing the limitation to be useless. If that would work we could know that the plan failed without looping "forever".
Currently the populator controller creates the pod with `RestartPolicyNever`, but maybe we wish to change it into `RestartPolicyOnFailure`?

pkg/controller/plan/migration.go
```diff
@@ -414,13 +414,24 @@ func (r *Migration) Cancel() (err error) {
 	return
 }
 
+func (r *Migration) CleanUpPopulatorPVCs(vm *plan.VMStatus) (err error) {
```
ahadas, line 417: can we reduce the visibility of this function?

pkg/controller/plan/kubevirt.go
```diff
@@ -893,6 +893,35 @@ func (r *KubeVirt) SetPopulatorPodOwnership(vm *plan.VMStatus) (err error) {
...
+	pvcs, err := r.getPVCs(vm.Ref)
+	for _, pvc := range pvcs {
+		primePVC := core.PersistentVolumeClaim{}
+		err = r.Destination.Client.Get(context.TODO(), client.ObjectKey{Namespace: r.Plan.Spec.TargetNamespace, Name: fmt.Sprintf("prime-%s", string(pvc.UID))}, &primePVC)
+		if err != nil {
+			continue
```
ahadas, line 903: don't we need to ignore IsNotFound here as well?

ahadas (commented):
the changes look good, minor comments inside
about restarting the populator pod on failure - yeah, it makes sense to align the populator pods with importer pods, but that's a separate task

pkg/controller/plan/kubevirt.go
ahadas: don't we? :)

ahadas: ah I see you've added a check below but we can get it also here from the call to `Get`, no?

liranr23: we can. but in any case of error we will execute `continue` to the loop. the only things that i missed is setting the error to nil, as if it's the last pvc in the loop we will return an error.

ahadas: the problem with this approach is that when the error is not of type IsNotFound, and there's no error later on, we'll return err = nil, how about doing it as we do in quite a few places elsewhere:
```golang
if err != nil {
	if k8serr.IsNotFound(err) {
		err = nil
	}
	continue
}
```

ahadas: I was wrong here, we shouldn't continue when k8serr.IsNotFound(err) != true, posted a fix - @liranr23 please check it out

liranr23: i think we should continue and not return here. we pass on the PVCs if one doesn't have a prime PVC for some reason - fine. let's check the next PVC. why would we just return and skip the rest?

ahadas: we can also take this approach of trying to remove the other items rather then returning and waiting for the next iteration but then the code would become more complex because we'd need to know that there's been an issue in previous iteration, otherwise let's say we failed to get the first prime pvc for some reason and everything afterwards went smoothly - we will return err == nil which doesn't align with the other places in which you return an error

ahadas: > i think we should continue and not return here. we pass on the PVCs if one doesn't have a prime PVC for some reason - fine. let's check the next PVC. why would we just return and skip the rest?

reading this comment again, I see you wrote `if one doesn't have a prime PVC for some reason - fine.` - I agree, the modified code would still do that. at first look I thought you meant to continue when err != nil and !k8serr.IsNotFound(err) == true

liranr23: i think both approaches are overall ok (hitting an error - failing) but i prefer something more into best effort, like: making logs of each error - if we see any. but resuming the deletion? in case we didn't succeed to get the prime we should also avoid trying deleting it?

ahadas: I'm not sure what you mean by "resuming the deletion" but yeah, the modified code now includes an optimization that skips the call to delete when we cannot get the prime pvc

liranr23: and in case we have the PVC (not prime) and  failing to get/delete the prime PVC, why to skip its removal?

ahadas: I'm still not sure I understand what you mean, before the changes I've made we had:
```golang
		// Best effort deletion
		err = r.DeleteObject(&primePVC, vm, "Deleted prime PVC.", "pvc")
		if err != nil && k8serr.IsNotFound(err) {
			err = nil
		}
		if err != nil {
			return
		}
```
do you mean that the last `return` should have been dropped?

wouldn't we have another iteration if we return an error in this function?

liranr23: yeah, in case we fail to delete the prime it might worth to continue and try to delete the PVC itself.

Overall something like this:
In case we didn't get the prime - skip it deletion - try to delete the PVC
In case we didn't succeed to delete the prime - try to delete the PVC.
In case we didn't succeed to delete the PVC - instead of returning, continue to the next PVC.
Log each error in any cases of the above.

ahadas: ok, I looked at the code - in two places we call this function assuming everything there is a best-effort and in one place we check the error, right?
so for the first two, what you write makes sense
but how would it work for the third case?

liranr23: correct. now when i think about it - yes we can say it's best effort for the first two and not for the third.
i don't like the idea of not doing the best effort as it depend where we cancel/fail in the migration to have or not have the prime pvc.

ahadas: @liranr23 I started to change the code to achieve this kind of 'best-effort' attempt to remove entities (it's not just about ignoring the error but also to try to remove as much as possible), see [this commit](https://github.com/kubev2v/forklift/pull/590/commits/9571e3ded1389a9263a1b1035fbf17bfdbcf0a70) - it looks like too much trouble to add logs everywhere and propagate the failOnErr to all places - I'm about to remove it from this PR but if you think it's important, you can do it separately

ahadas: dropped the aforementioned commit, will post it separately

ahadas: gave it another try, @liranr23 please take a look at the last commit

pkg/controller/plan/migration.go
```diff
@@ -388,12 +392,16 @@ func (r *Migration) Cancel() error {
 
 	for _, vm := range r.Plan.Status.Migration.VMs {
 		if vm.HasCondition(Canceled) {
-			if err := r.cleanup(vm); err != nil {
-				r.Log.Error(err,
-					"Couldn't clean up after canceled VM migration.",
-					"vm",
-					vm.String())
+			dontFailOnError := func(err error) bool {
```
liranr23, line 395: couldn't it cause a spam in logs? if we will fail to clean up multiple times we will see the same message? the other option is your previous commit (boolean), but then we will need to check each function in cleanup to print the error. maybe the multiple message is ok

ahadas: it wouldn't spam the log too much since it's a one-time attempt but yeah, we'll have multiple logs with the same message but with different errors that should hint where we failed, that's a good thing in case several calls fail

ahadas (approved):

PR #592, opened 2023-09-13
OVA: changing nfs mount to server and conversion pod to use shared PVC

bkhizgiy:
Until now when adding a new OVA provider we were mount the NFS volume directly to the server/ conversion pod, this worked for some of the cases but wasn't supported on a restricted or different namespaces from forklift/MTV since NFS is not permitted. This fix switch the use of NFS to used shared PVC for both of the cases instead.

### DIFF
--- a/pkg/controller/plan/kubevirt.go
+++ b/pkg/controller/plan/kubevirt.go
@@ -32,7 +32,6 @@ import (
 	libcnd "github.com/konveyor/forklift-controller/pkg/lib/condition"
 	liberr "github.com/konveyor/forklift-controller/pkg/lib/error"
 	core "k8s.io/api/core/v1"
-	v1 "k8s.io/api/core/v1"
 	k8serr "k8s.io/apimachinery/pkg/api/errors"
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
@@ -1185,7 +1184,10 @@ func (r *KubeVirt) findTemplate(vm *plan.VMStatus) (tmpl *template.Template, err
 }
 
 func (r *KubeVirt) guestConversionPod(vm *plan.VMStatus, vmVolumes []cnv.Volume, configMap *core.ConfigMap, pvcs *[]core.PersistentVolumeClaim, v2vSecret *core.Secret) (pod *core.Pod, err error) {
-	volumes, volumeMounts, volumeDevices := r.podVolumeMounts(vmVolumes, configMap, pvcs)
+	volumes, volumeMounts, volumeDevices, err := r.podVolumeMounts(vmVolumes, configMap, pvcs)
+	if err != nil {
+		return
+	}
 
 	// qemu group
 	fsGroup := qemuGroup
@@ -1291,7 +1293,7 @@ func (r *KubeVirt) guestConversionPod(vm *plan.VMStatus, vmVolumes []cnv.Volume,
 	return
 }
 
-func (r *KubeVirt) podVolumeMounts(vmVolumes []cnv.Volume, configMap *core.ConfigMap, pvcs *[]core.PersistentVolumeClaim) (volumes []core.Volume, mounts []core.VolumeMount, devices []core.VolumeDevice) {
+func (r *KubeVirt) podVolumeMounts(vmVolumes []cnv.Volume, configMap *core.ConfigMap, pvcs *[]core.PersistentVolumeClaim) (volumes []core.Volume, mounts []core.VolumeMount, devices []core.VolumeDevice, err error) {
 	pvcsByName := make(map[string]core.PersistentVolumeClaim)
 	for _, pvc := range *pvcs {
 		pvcsByName[pvc.Name] = pvc
@@ -1337,18 +1339,32 @@ func (r *KubeVirt) podVolumeMounts(vmVolumes []cnv.Volume, configMap *core.Confi
 
 	switch r.Source.Provider.Type() {
 	case api.Ova:
-		pvcName := fmt.Sprintf("ova-server-pvc-%s-%s", r.Source.Provider.Name, r.Plan.Name)
-		// If the provider runs in a different namespace then the destination VM,
-		// we need to create the PV and PVC to access the NFS.
-		if r.Plan.Spec.TargetNamespace != r.Source.Provider.Namespace {
-			r.CreatePvcForNfs(pvcName)
+		pvcName := fmt.Sprintf("ova-nfs-pvc-%s-%s", r.Source.Provider.Name, r.Plan.Name)
+		/*
+			TODO: set ownerReferences the PV and PVC deleted once the plan is done.
+			cross namespaces are not possible to be used for plan/provider ownership.
+			optional: PV+PVC per VM, deleted via cleanup. ownership: importer pod? DV?
+			ownerReference := meta.OwnerReference{
+				APIVersion: "forklift.konveyor.io/v1beta1",
+				Kind:       "Plan",
+				Name:       r.Plan.Name,
+				UID:        r.Plan.UID,
+			}
+		*/
+		err = r.CreatePvForNfs()
+		if err != nil {
+			return
+		}
+		err = r.CreatePvcForNfs(pvcName)
+		if err != nil {
+			return
 		}
 
 		//path from disk
 		volumes = append(volumes, core.Volume{
-			Name: "nfs-pvc",
+			Name: "nfs-pv",
 			VolumeSource: core.VolumeSource{
-				PersistentVolumeClaim: &v1.PersistentVolumeClaimVolumeSource{
+				PersistentVolumeClaim: &core.PersistentVolumeClaimVolumeSource{
 					ClaimName: pvcName,
 				},
 			},
@@ -1363,7 +1379,7 @@ func (r *KubeVirt) podVolumeMounts(vmVolumes []cnv.Volume, configMap *core.Confi
 				MountPath: "/opt",
 			},
 			core.VolumeMount{
-				Name:      "nfs-pvc",
+				Name:      "nfs-pv",
 				MountPath: "/ova",
 			},
 		)
@@ -1832,70 +1848,129 @@ func (r *KubeVirt) EnsurePersistentVolume(vmRef ref.Ref, persistentVolumes []cor
 	return
 }
 
-func (r *KubeVirt) CreatePvcForNfs(pvcName string) (err error) {
-	//TODO: set ownerReferenceso the PV and PVC deleted once the plan is done.
-	// ownerReference := meta.OwnerReference{
-	// 	APIVersion: "forklift.konveyor.io/v1beta1",
-	// 	Kind:       "Plan",
-	// 	Name:       r.Plan.Name,
-	// 	UID:        r.Plan.UID,
-	// }
+func (r *KubeVirt) getOvaPvNfs() (found bool, err error) {
+	pv := &core.PersistentVolume{}
+	err = r.Destination.Client.Get(
+		context.TODO(),
+		types.NamespacedName{
+			Name: fmt.Sprintf("ova-nfs-pv-%s", r.Plan.Name),
+		},
+		pv,
+	)
 
+	if err != nil {
+		if k8serr.IsNotFound(err) {
+			return false, nil
+		}
+		err = liberr.Wrap(err)
+		return
+	}
+	return true, nil
+}
+
+func (r *KubeVirt) getOvaPvcNfs() (found bool, err error) {
+	pvc := &core.PersistentVolumeClaim{}
+	err = r.Destination.Client.Get(
+		context.TODO(),
+		types.NamespacedName{
+			Name:      fmt.Sprintf("ova-nfs-pvc-%s-%s", r.Source.Provider.Name, r.Plan.Name),
+			Namespace: r.Plan.Spec.TargetNamespace,
+		},
+		pvc,
+	)
+
+	if err != nil {
+		if k8serr.IsNotFound(err) {
+			return false, nil
+		}
+		err = liberr.Wrap(err)
+		return
+	}
+	return true, nil
+}
+
+// TODO if we use ownership get it to the method
+func (r *KubeVirt) CreatePvForNfs() (err error) {
 	sourceProvider := r.Source.Provider
-	pvName := fmt.Sprintf("ova-server-pv-%s", r.Plan.Name)
 	splitted := strings.Split(sourceProvider.Spec.URL, ":")
-
 	if len(splitted) != 2 {
-		r.Log.Error(nil, "NFS server path doesn't contains :")
+		r.Log.Error(nil, "NFS server path doesn't contains: ", "url", sourceProvider.Spec.URL)
+		return fmt.Errorf("bad source provider %s URL %s", sourceProvider.Name, sourceProvider.Spec.URL)
 	}
 	nfsServer := splitted[0]
 	nfsPath := splitted[1]
 
-	pv := &v1.PersistentVolume{
+	pvName := fmt.Sprintf("ova-nfs-pv-%s", r.Plan.Name)
+	found, err := r.getOvaPvNfs()
+	if err != nil {
+		return
+	}
+	if found {
+		r.Log.Info("The PV for OVA NFS exists", "PV", pvName)
+		return
+	}
+
+	pv := &core.PersistentVolume{
 		ObjectMeta: meta.ObjectMeta{
 			Name: pvName,
+			// OwnerReferences: []meta.OwnerReference{ownerReference},
 		},
-		Spec: v1.PersistentVolumeSpec{
-			Capacity: v1.ResourceList{
-				v1.ResourceStorage: resource.MustParse("1Gi"),
+		Spec: core.PersistentVolumeSpec{
+			Capacity: core.ResourceList{
+				core.ResourceStorage: resource.MustParse("1Gi"),
 			},
-			AccessModes: []v1.PersistentVolumeAccessMode{
-				v1.ReadOnlyMany,
+			AccessModes: []core.PersistentVolumeAccessMode{
+				core.ReadOnlyMany,
 			},
-			PersistentVolumeSource: v1.PersistentVolumeSource{
-				NFS: &v1.NFSVolumeSource{
+			PersistentVolumeSource: core.PersistentVolumeSource{
+				NFS: &core.NFSVolumeSource{
 					Path:   nfsPath,
 					Server: nfsServer,
 				},
 			},
 		},
 	}
-	err = r.Create(context.TODO(), pv)
+	err = r.Destination.Create(context.TODO(), pv)
 	if err != nil {
 		r.Log.Error(err, "Failed to create OVA plan PV")
 		return
 	}
+	return
+}
+
+// TODO if we use ownership get it to the method
+func (r *KubeVirt) CreatePvcForNfs(pvcName string) (err error) {
+	found, err := r.getOvaPvcNfs()
+	if err != nil {
+		return
+	}
+	if found {
+		r.Log.Info("The PVC for OVA NFS exists", "PVC", pvcName)
+		return
+	}
 
 	sc := ""
-	pvc := &v1.PersistentVolumeClaim{
+	pvName := fmt.Sprintf("ova-nfs-pv-%s", r.Plan.Name)
+	pvc := &core.PersistentVolumeClaim{
 		ObjectMeta: meta.ObjectMeta{
 			Name:      pvcName,
 			Namespace: r.Plan.Spec.TargetNamespace,
+			// OwnerReferences: []meta.OwnerReference{ownerReference},
 		},
-		Spec: v1.PersistentVolumeClaimSpec{
-			Resources: v1.ResourceRequirements{
-				Requests: v1.ResourceList{
-					v1.ResourceStorage: resource.MustParse("1Gi"),
+		Spec: core.PersistentVolumeClaimSpec{
+			Resources: core.ResourceRequirements{
+				Requests: core.ResourceList{
+					core.ResourceStorage: resource.MustParse("1Gi"),
 				},
 			},
-			AccessModes: []v1.PersistentVolumeAccessMode{
-				v1.ReadOnlyMany,
+			AccessModes: []core.PersistentVolumeAccessMode{
+				core.ReadOnlyMany,
 			},
 			VolumeName:       pvName,
 			StorageClassName: &sc,
 		},
 	}
-	err = r.Create(context.TODO(), pvc)
+	err = r.Destination.Create(context.TODO(), pvc)
 	if err != nil {
 		r.Log.Error(err, "Failed to create OVA plan PVC")
 		return
@@ -1913,15 +1988,13 @@ func (r *KubeVirt) CreatePvcForNfs(pvcName string) (err error) {
 		select {
 		case <-timeout:
 			r.Log.Error(err, "Timed out waiting for PVC to be bound")
-			return
+			return fmt.Errorf("timeout passed waiting for the OVA PVC %v", pvc)
 		case <-tick:
 			err = r.Get(context.TODO(), pvcNamespacedName, pvc)
-			fmt.Print("this is pvc: ", pvc)
 			if err != nil {
 				r.Log.Error(err, "Failed to bound OVA plan PVC")
 				return
 			}
-
 			if pvc.Status.Phase == "Bound" {
 				return
 			}
--- a/pkg/controller/plan/kubevirt.go
+++ b/pkg/controller/plan/kubevirt.go
@@ -21,6 +21,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	k8svalidation "k8s.io/apimachinery/pkg/util/validation"
+	"k8s.io/apimachinery/pkg/util/wait"
 	cnv "kubevirt.io/api/core/v1"
 	libvirtxml "libvirt.org/libvirt-go-xml"
 
@@ -1339,22 +1340,11 @@ func (r *KubeVirt) podVolumeMounts(vmVolumes []cnv.Volume, configMap *core.Confi
 
 	switch r.Source.Provider.Type() {
 	case api.Ova:
-		pvcName := fmt.Sprintf("ova-nfs-pvc-%s-%s", r.Source.Provider.Name, r.Plan.Name)
-		/*
-			TODO: set ownerReferences the PV and PVC deleted once the plan is done.
-			cross namespaces are not possible to be used for plan/provider ownership.
-			optional: PV+PVC per VM, deleted via cleanup. ownership: importer pod? DV?
-			ownerReference := meta.OwnerReference{
-				APIVersion: "forklift.konveyor.io/v1beta1",
-				Kind:       "Plan",
-				Name:       r.Plan.Name,
-				UID:        r.Plan.UID,
-			}
-		*/
 		err = r.CreatePvForNfs()
 		if err != nil {
 			return
 		}
+		pvcName := getEntityName("pvc", r.Source.Provider.Name, r.Plan.Name)
 		err = r.CreatePvcForNfs(pvcName)
 		if err != nil {
 			return
@@ -1362,7 +1352,7 @@ func (r *KubeVirt) podVolumeMounts(vmVolumes []cnv.Volume, configMap *core.Confi
 
 		//path from disk
 		volumes = append(volumes, core.Volume{
-			Name: "nfs-pv",
+			Name: "store-pv",
 			VolumeSource: core.VolumeSource{
 				PersistentVolumeClaim: &core.PersistentVolumeClaimVolumeSource{
 					ClaimName: pvcName,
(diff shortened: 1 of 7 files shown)


bkhizgiy:
The PR is ready and was tested on different namespaces, both the provider adding and migrations.
the only thing not working is the owner reference, once is set with the plan info the PVC deletes before the conversion finished, for now I've removed it from the code but it means that the PV and PVC are not removed once the plan is removed.

+ the migration wasn't tested on restricted namespace but according to the documentation it should work as well.

pkg/controller/plan/kubevirt.go
```diff
@@ -32,6 +32,7 @@ import (
 	libcnd "github.com/konveyor/forklift-controller/pkg/lib/condition"
 	liberr "github.com/konveyor/forklift-controller/pkg/lib/error"
 	core "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
liranr23, line 35: same as one line above, please remove and use `core`.

pkg/controller/plan/kubevirt.go
```diff
@@ -1336,23 +1337,19 @@ func (r *KubeVirt) podVolumeMounts(vmVolumes []cnv.Volume, configMap *core.Confi
...
-		server := r.Source.Provider.Spec.URL
-		splitted := strings.Split(server, ":")
-
-		if len(splitted) != 2 {
-			r.Log.Info("The NFS server path format is wrong")
-			return
+		pvcName := fmt.Sprintf("ova-server-pvc-%s-%s", r.Source.Provider.Name, r.Plan.Name)
+		// If the provider runs in a different namespace then the destination VM,
+		// we need to create the PV and PVC to access the NFS.
+		if r.Plan.Spec.TargetNamespace != r.Source.Provider.Namespace {
```
liranr23, line 1343: why should it be only on different namespace? shouldn't we always create the PV and PVC?

pkg/controller/plan/kubevirt.go
```diff
@@ -1835,6 +1832,103 @@ func (r *KubeVirt) EnsurePersistentVolume(vmRef ref.Ref, persistentVolumes []cor
 	return
 }
 
+func (r *KubeVirt) CreatePvcForNfs(pvcName string) (err error) {
+	//TODO: set ownerReferenceso the PV and PVC deleted once the plan is done.
```
liranr23, line 1836: the owner should be the conversion pod? anything else uses it? the migration/plan also might be used.

pkg/controller/plan/kubevirt.go
```diff
@@ -1835,6 +1832,103 @@ func (r *KubeVirt) EnsurePersistentVolume(vmRef ref.Ref, persistentVolumes []cor
...
+	// ownerReference := meta.OwnerReference{
+	// 	APIVersion: "forklift.konveyor.io/v1beta1",
+	// 	Kind:       "Plan",
+	// 	Name:       r.Plan.Name,
+	// 	UID:        r.Plan.UID,
+	// }
+
+	sourceProvider := r.Source.Provider
+	pvName := fmt.Sprintf("ova-server-pv-%s", r.Plan.Name)
```
liranr23, line 1845: i would put this line below, right before the `pv := &...` and let the error run first

pkg/controller/plan/kubevirt.go
```diff
@@ -1835,6 +1832,103 @@ func (r *KubeVirt) EnsurePersistentVolume(vmRef ref.Ref, persistentVolumes []cor
...
+	splitted := strings.Split(sourceProvider.Spec.URL, ":")
+
+	if len(splitted) != 2 {
+		r.Log.Error(nil, "NFS server path doesn't contains :")
```
liranr23, line 1849: should we return here?

pkg/controller/plan/kubevirt.go
```diff
@@ -1835,6 +1832,103 @@ func (r *KubeVirt) EnsurePersistentVolume(vmRef ref.Ref, persistentVolumes []cor
...
+	pvcNamespacedName := types.NamespacedName{
+		Namespace: r.Plan.Spec.TargetNamespace,
+		Name:      pvcName,
+	}
+
+	for {
+		select {
+		case <-timeout:
+			r.Log.Error(err, "Timed out waiting for PVC to be bound")
+			return
```
liranr23, line 1916: will it return an error or we will reconcile and try again? if we will try again - will we have the PV or PVC existing and try to create the same ones?

pkg/controller/plan/kubevirt.go
```diff
@@ -1835,6 +1832,103 @@ func (r *KubeVirt) EnsurePersistentVolume(vmRef ref.Ref, persistentVolumes []cor
 	return
 }
 
+func (r *KubeVirt) CreatePvcForNfs(pvcName string) (err error) {
```
liranr23, line 1835: i would split this method to two: one for PV and the other for PVC

pkg/controller/provider/ova-setup.go
```diff
@@ -0,0 +1,205 @@
...
+	ovaServerPrefix     = "ova-server"
+	ovaImageVar         = "OVA_PROVIDER_SERVER_IMAGE"
+	nfsVolumeNamePrefix = "nfs-volume"
+	mountPath           = "/ova"
+	pvSize              = "1Gi"
+)
+
+func (r *Reconciler) CreateOVAServerDeployment(provider *api.Provider, ctx context.Context) {
+
+	ownerReference := metav1.OwnerReference{
```
liranr23, line 27: continuing the comment about the owner above, so you create 2 PVs and 2 PVCs one for the provider and one for the conversion? it might work to use the same one for both (didn't try that)

pkg/controller/provider/ova-setup.go
```diff
@@ -0,0 +1,205 @@
...
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"strings"
+
+	api "github.com/konveyor/forklift-controller/pkg/apis/forklift/v1beta1"
+	appsv1 "k8s.io/api/apps/v1"
+	v1 "k8s.io/api/core/v1"
```
liranr23, line 11: i would stick with `core`

pkg/controller/provider/ova-setup.go
```diff
@@ -0,0 +1,205 @@
...
+		Kind:       "Provider",
+		Name:       provider.Name,
+		UID:        provider.UID,
+	}
+
+	pvName := fmt.Sprintf("%s-pv-%s", ovaServerPrefix, provider.Name)
+	splitted := strings.Split(provider.Spec.URL, ":")
+
+	if len(splitted) != 2 {
+		r.Log.Error(nil, "NFS server path doesn't contains :")
```
liranr23, line 38: should we return?

pkg/controller/provider/ova-setup.go
```diff
@@ -0,0 +1,205 @@
...
+func (r *Reconciler) CreateOVAServerDeployment(provider *api.Provider, ctx context.Context) {
+
+	ownerReference := metav1.OwnerReference{
+		APIVersion: "forklift.konveyor.io/v1beta1",
+		Kind:       "Provider",
+		Name:       provider.Name,
+		UID:        provider.UID,
+	}
+
+	pvName := fmt.Sprintf("%s-pv-%s", ovaServerPrefix, provider.Name)
```
liranr23, line 34: same as above

pkg/controller/provider/ova-setup.go
```diff
@@ -0,0 +1,205 @@
...
+	pvc := &v1.PersistentVolumeClaim{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:            pvcName,
+			Namespace:       provider.Namespace,
+			OwnerReferences: []metav1.OwnerReference{ownerReference},
+		},
+		Spec: v1.PersistentVolumeClaimSpec{
+			Resources: v1.ResourceRequirements{
+				Requests: v1.ResourceList{
+					v1.ResourceStorage: resource.MustParse("1Gi"),
```
liranr23, line 80: `pvSize`?

pkg/controller/provider/ova-setup.go
```diff
@@ -0,0 +1,205 @@
...
+
+const (
+	ovaServerPrefix     = "ova-server"
+	ovaImageVar         = "OVA_PROVIDER_SERVER_IMAGE"
+	nfsVolumeNamePrefix = "nfs-volume"
+	mountPath           = "/ova"
+	pvSize              = "1Gi"
+)
+
+func (r *Reconciler) CreateOVAServerDeployment(provider *api.Provider, ctx context.Context) {
```
liranr23, line 25: might be worth splitting it to parts?

pkg/controller/provider/ova-setup.go
```diff
@@ -0,0 +1,205 @@
...
+
+		Containers: []v1.Container{
+			{
+				Name:  ovaContainerName,
+				Ports: []v1.ContainerPort{{ContainerPort: 8080, Protocol: v1.ProtocolTCP}},
+				Image: imageName,
+				VolumeMounts: []v1.VolumeMount{
+					{
+						Name:      nfsVolumeName,
+						MountPath: "/ova",
```
liranr23, line 189: `mountPath`?

ahadas (commented):
@liranr23 can you please take over this one and polish it?

pkg/controller/plan/kubevirt.go
liranr23: what's done is - if it's the same namespace we don't need to recreate the PV/PVC. we can reuse the existing ones of the OVA server. @bkhizgiy please keep me honest here

pkg/controller/plan/kubevirt.go
liranr23: still need to figure that one

pkg/controller/plan/kubevirt.go
ahadas: yes, that's clear but we don't allocate space here - so there's no reason to have such an optimization, it just adds another flow that we need to test - there's no need for that

pkg/controller/plan/kubevirt.go
liranr23: @ahadas @bkhizgiy It failed because cross namespaces. If the namespace we target is the one that have the plan - everything is fine. It's not our case.
Conversion pod is also problematic - it raises the question whether we wish the PV+PVC to be per VM migrated or reusable by the whole plan. If we are going on per VM approach we will need to update the owners *after* the conversion pod is created.

bkhizgiy: @liranr23 umm it make more scene to use it for the entire plan I think but in that case the PVC is deleted before the conversion pod created so we need to see if we can somehow overcome this issue, maybe another solution should be check the destination namespace before running a plan and if the PVC+PV are not existing create it for the first run and use it for all future plans, then we can set its reference to the provider (and leave it there) , once we deleting the provider it will remove all these PVCs and PVs from the system, since we are using NFS it dosent really take up space from the storage, it currently set to 1 Gib but we can set it to less.

liranr23: i agree it's better to have the plan as owner. but the main problem is that we need the plan(/provider) in the target namespace which most likely will be different.

bkhizgiy:
Still need to be tested, having some issue with my env.

pkg/controller/plan/kubevirt.go
```diff
@@ -1336,23 +1339,33 @@ func (r *KubeVirt) podVolumeMounts(vmVolumes []cnv.Volume, configMap *core.Confi
...
-		splitted := strings.Split(server, ":")
-
-		if len(splitted) != 2 {
-			r.Log.Info("The NFS server path format is wrong")
+		pvcName := fmt.Sprintf("ova-nfs-pvc-%s-%s", r.Source.Provider.Name, r.Plan.Name)
+		/*
+			TODO: set ownerReferences the PV and PVC deleted once the plan is done.
+			cross namespaces are not possible to be used for plan/provider ownership.
+			optional: PV+PVC per VM, deleted via cleanup. ownership: importer pod? DV?
+			ownerReference := meta.OwnerReference{
```
ahadas, line 1347: @liranr23 and I talked about this part earlier today - I'd check if it's possible to set the provider as the owner of a PV that is shared by the server and conversion pods, and remove PVCs during plan's cleanup

after a second thought, what if we migrate to a remote cluster? in that case, we need to create the PV for the conversion pod..

pkg/controller/plan/kubevirt.go
```diff
@@ -1336,23 +1339,33 @@ func (r *KubeVirt) podVolumeMounts(vmVolumes []cnv.Volume, configMap *core.Confi
...
+				APIVersion: "forklift.konveyor.io/v1beta1",
+				Kind:       "Plan",
+				Name:       r.Plan.Name,
+				UID:        r.Plan.UID,
+			}
+		*/
+		err = r.CreatePvForNfs()
```
ahadas, line 1354: and then we could drop that part

pkg/controller/plan/kubevirt.go
```diff
@@ -1835,6 +1848,156 @@ func (r *KubeVirt) EnsurePersistentVolume(vmRef ref.Ref, persistentVolumes []cor
...
+	}
+	err = r.Destination.Create(context.TODO(), pvc)
+	if err != nil {
+		r.Log.Error(err, "Failed to create OVA plan PVC")
+		return
+	}
+
+	// wait until pvc and pv are bounded.
+	timeout := time.After(5 * time.Minute)
+	tick := time.Tick(5 * time.Second)
```
ahadas, line 1977: golintci has a point

pkg/controller/plan/migration.go
```diff
@@ -336,6 +337,29 @@ func (r *Migration) Archive() {
 		return
 	}
 
+	if r.Plan.Provider.Source.Type() == v1beta1.Ova {
+		//delete pvc and pv
+		pvc, _, errPvc := GetOvaPvcNfs(r.Destination.Client, r.Plan.Name, r.Plan.Namespace, r.Plan.Provider.Source.Name)
+		if errPvc != nil || pvc == nil {
+			return
```
ahadas, line 344: missing log

pkg/controller/plan/migration.go
```diff
@@ -336,6 +337,29 @@ func (r *Migration) Archive() {
...
+		}
+
+		err = r.Destination.Client.Delete(context.TODO(), pvc)
+		if err != nil {
+			return
```
ahadas, line 349: missing log

pkg/controller/plan/migration.go
```diff
@@ -336,6 +337,29 @@ func (r *Migration) Archive() {
...
+		}
+
+		pv, _, errPv := GetOvaPvNfs(r.Destination.Client, r.Plan.Name, r.Plan.Provider.Source.Name)
+		if errPv != nil || pv == nil {
+			return
```
ahadas, line 354: missing log

pkg/controller/plan/migration.go
```diff
@@ -336,6 +337,29 @@ func (r *Migration) Archive() {
...
+		}
+
+		err = r.Destination.Client.Delete(context.TODO(), pv)
+		if err != nil {
+			return
```
ahadas, line 359: missing log

pkg/controller/plan/kubevirt.go
bennyz: yeah, the `wait` can be a good candidate here

pkg/controller/plan/kubevirt.go
bkhizgiy: I've tested this scenario before my PTO (to save some PV creation stages) but once the PV is bound to a PVC its not possible to use it for another one, even if it set as read-write-many.
so I think the best will be to keep the creation + clean up as now or change the PV delete policy so it will be deleted once we deleting the PVC, and additionally add labels of the plan and the providers for the PVC + PV.

liranr23: in that case, i think we should create 1 PV and 1 PVC per plan on destination with labels of the plan. i think it's enough to remove them when archiving.

ahadas: > I've tested this scenario before my PTO (to save some PV creation stages) but once the PV is bound to a PVC its not possible to use it for another one, even if it set as read-write-many.

this seems suspicious - do you still have this environment?

ahadas: > in that case, i think we should create 1 PV and 1 PVC per plan on destination with labels of the plan. i think it's enough to remove them when archiving.

right, but the label should refer to the migration

pkg/controller/plan/kubevirt.go
```diff
@@ -1835,6 +1838,155 @@ func (r *KubeVirt) EnsurePersistentVolume(vmRef ref.Ref, persistentVolumes []cor
...
+		if k8serr.IsNotFound(err) {
+			return nil, false, nil
+		}
+		err = liberr.Wrap(err)
+		return
+	}
+	return
+}
+
+// TODO if we use ownership get it to the method
```
liranr23, line 1882: unnecessary comment

pkg/controller/plan/migration.go
```diff
@@ -336,6 +337,33 @@ func (r *Migration) Archive() {
 		return
 	}
 
+	if r.Plan.Provider.Source.Type() == v1beta1.Ova {
+		//delete pvc and pv
+		pvc, _, errPvc := GetOvaPvcNfs(r.Destination.Client, r.Plan.Name, r.Plan.Spec.TargetNamespace, r.Plan.Provider.Source.Name)
+		if errPvc != nil || pvc == nil {
+			r.Log.Error(err, "Failed to get the plan PVC")
```
liranr23, line 344: if there is no error and `pvc == nil` - is it an error that we didn't get it? it hints that the PVC doesn't exists (deleted already?)

pkg/controller/plan/migration.go
```diff
@@ -336,6 +337,33 @@ func (r *Migration) Archive() {
...
+
+		err = r.Destination.Client.Delete(context.TODO(), pvc)
+		if err != nil {
+			r.Log.Error(err, "Failed to delete the plan PVC")
+			return
+		}
+
+		pv, _, errPv := GetOvaPvNfs(r.Destination.Client, r.Plan.Name, r.Plan.Provider.Source.Name)
+		if errPv != nil || pv == nil {
+			r.Log.Error(err, "Failed to get the plan PV")
```
liranr23, line 356: same as above

pkg/controller/plan/migration.go
bkhizgiy: yeah it's mostly for the second case, that the PVC was already deleted and the error is not Found, in this case the GetOvaPvcNfs() isnt returning an error but we don't need to delete anything, I can change the check to be `!found` instead.

pkg/controller/plan/kubevirt.go
```diff
@@ -1835,6 +1838,155 @@ func (r *KubeVirt) EnsurePersistentVolume(vmRef ref.Ref, persistentVolumes []cor
...
+		return
+	}
+	if found {
+		r.Log.Info("The PVC for OVA NFS exists", "PVC", pvcName)
+		return
+	}
+
+	sc := ""
+	pvName := fmt.Sprintf("ova-nfs-pv-%s-%s", r.Source.Provider.Name, r.Plan.Name)
+	pvc := &core.PersistentVolumeClaim{
```
liranr23, line 1944: labels?

pkg/controller/plan/kubevirt.go
```diff
@@ -1835,6 +1838,155 @@ func (r *KubeVirt) EnsurePersistentVolume(vmRef ref.Ref, persistentVolumes []cor
...
+	if err != nil {
+		r.Log.Error(err, "Failed to get ova PV")
+		return
+	}
+	if found {
+		r.Log.Info("The PV for OVA NFS exists", "PV", pvName)
+		return
+	}
+
+	pv := &core.PersistentVolume{
```
liranr23, line 1904: labels?

pkg/controller/plan/migration.go
liranr23: that would be more logical to me
you can do also (i think it's ugly):
```
if errPvc != nil {
  r.log.Error(....
}
if pvc == nil {
  return
}
```
because we return `nil` if there is an error. but if we look on `found` i think it's nicer.

pkg/controller/plan/kubevirt.go
bkhizgiy: @ahadas its not supported, I did double check and search it up, the shared property referring to the fact that many pods can use the same PVC and PV and not that more then one PVC can use a PV.
Screenshot from 2023-10-02 16-16-39

ahadas: alright, if we no longer try to reuse the PV and don't set owner-references then it's fine

bkhizgiy:
Was tested also for restricted namespace and the entire flow was as expected.

pkg/controller/plan/kubevirt.go
```diff
@@ -1835,6 +1838,161 @@ func (r *KubeVirt) EnsurePersistentVolume(vmRef ref.Ref, persistentVolumes []cor
...
+	sourceProvider := r.Source.Provider
+	splitted := strings.Split(sourceProvider.Spec.URL, ":")
+	if len(splitted) != 2 {
+		r.Log.Error(nil, "NFS server path doesn't contains: ", "url", sourceProvider.Spec.URL)
+		return fmt.Errorf("bad source provider %s URL %s", sourceProvider.Name, sourceProvider.Spec.URL)
+	}
+	nfsServer := splitted[0]
+	nfsPath := splitted[1]
+
+	pvName := fmt.Sprintf("ova-nfs-pv-%s-%s", r.Source.Provider.Name, r.Plan.Name)
```
ahadas, line 1892: please move it closer to where it's used below

pkg/controller/plan/kubevirt.go
```diff
@@ -1835,6 +1838,161 @@ func (r *KubeVirt) EnsurePersistentVolume(vmRef ref.Ref, persistentVolumes []cor
...
+	if err != nil {
+		r.Log.Error(err, "Failed to get ova PV")
+		return
+	}
+	if found {
+		r.Log.Info("The PV for OVA NFS exists", "PV", pvName)
+		return
+	}
+
+	labels := map[string]string{"providerName": r.Plan.Provider.Source.Name, "app": "forklift", "migration": r.Migration.Name, "plan": r.Plan.Name}
```
ahadas, line 1903:
```suggestion
	labels := map[string]string{"provider": r.Plan.Provider.Source.Name, "app": "forklift", "migration": r.Migration.Name, "plan": r.Plan.Name}
```
(unless we rely on `providerName` elsewhere)

pkg/controller/plan/kubevirt.go
```diff
@@ -1835,6 +1838,161 @@ func (r *KubeVirt) EnsurePersistentVolume(vmRef ref.Ref, persistentVolumes []cor
...
+		err = r.Get(context.TODO(), pvcNamespacedName, pvc)
+		if err != nil {
+			r.Log.Error(err, "Failed to get OVA plan PVC")
+			return false, err
+		}
+
+		if pvc.Status.Phase == "Bound" {
+			return true, nil
+		}
+		return false, nil
```
ahadas, line 1986: how about replacing the 3 lined above with:
```golang
return pvc.Status.Phase == "Bound", nil
```

pkg/controller/plan/kubevirt.go
```diff
@@ -1835,6 +1838,161 @@ func (r *KubeVirt) EnsurePersistentVolume(vmRef ref.Ref, persistentVolumes []cor
...
+		r.Log.Error(err, "Failed to create OVA plan PVC")
+		return
+	}
+
+	pvcNamespacedName := types.NamespacedName{
+		Namespace: r.Plan.Spec.TargetNamespace,
+		Name:      pvcName,
+	}
+
+	if err = wait.PollUntilContextTimeout(context.TODO(), 5*time.Second, 15*time.Second, true, func(ctx context.Context) (done bool, err error) {
```
ahadas, line 1976: don't we want to wait a bit longer, let's say, 30 or 45 seconds?

pkg/controller/plan/kubevirt.go
```diff
@@ -1835,6 +1838,161 @@ func (r *KubeVirt) EnsurePersistentVolume(vmRef ref.Ref, persistentVolumes []cor
...
+			return false, err
+		}
+
+		if pvc.Status.Phase == "Bound" {
+			return true, nil
+		}
+		return false, nil
+
+	}); err != nil {
+		r.Log.Error(err, "Failed to bound OVA PVC to PV ")
```
ahadas, line 1989:
```suggestion
		r.Log.Error(err, "Failed to bind OVA PVC to PV ")
```

pkg/controller/plan/migration.go
```diff
@@ -336,6 +337,41 @@ func (r *Migration) Archive() {
...
 	}
 
+	if r.Plan.Provider.Source.Type() == v1beta1.Ova {
+		//delete pvc and pv
+		pvc, _, errPvc := GetOvaPvcNfs(r.Destination.Client, r.Plan.Name, r.Plan.Spec.TargetNamespace, r.Plan.Provider.Source.Name)
+		if errPvc != nil {
+			r.Log.Error(err, "Failed to get the plan PVC")
+			return
+		}
+		// The PVC was alredy deleted
```
ahadas, line 347:
```suggestion
		// The PVC was already deleted
```

pkg/controller/provider/ova-setup.go
```diff
@@ -0,0 +1,228 @@
...
+	}
+
+	pvcName := fmt.Sprintf("%s-pvc-%s", ovaServerPrefix, provider.Name)
+	err = r.createPvcForNfs(provider, ctx, ownerReference, pvName, pvcName)
+	if err != nil {
+		r.Log.Error(err, "Failed to create NFS PVC for the OVA server")
+		return
+	}
+
+	labels := map[string]string{"providerName": provider.Name, "app": "forklift"}
```
ahadas, line 46: ok, I see `providerName` here as well, can we replace it with `provider`?

pkg/controller/provider/ova-setup.go
```diff
@@ -0,0 +1,228 @@
...
+			Namespace:       provider.Namespace,
+			Annotations:     annotations,
+			Labels:          labels,
+			OwnerReferences: []metav1.OwnerReference{ownerReference},
+		},
+		Spec: appsv1.DeploymentSpec{
+			Replicas: &replicas,
+			Selector: &metav1.LabelSelector{
+				MatchLabels: map[string]string{
+					"app": "forklift",
```
ahadas, line 145: shouldn't we extend this matcher - if we have another pod with the `app: forklift` label, we may think that we reached the number of desired replicas, no?

ahadas (commented):
the code is better structured and it was verified with restricted namespaces, which is great. commented mostly on the code changes

pkg/controller/provider/ova-setup.go
```diff
@@ -0,0 +1,228 @@
...
+	if err != nil {
+		r.Log.Error(err, "Failed to create OVA server service")
+		return
+	}
+}
+
+func (r *Reconciler) createPvForNfs(provider *api.Provider, ctx context.Context, ownerReference metav1.OwnerReference, pvName string) (err error) {
+	splitted := strings.Split(provider.Spec.URL, ":")
+	if len(splitted) != 2 {
+		r.Log.Error(nil, "NFS server path doesn't contains: ", "provider", provider, "url", provider.Spec.URL)
```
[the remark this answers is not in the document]
bkhizgiy: yeah, we also have it as a validation when adding the provider, technically its not needed.

pkg/controller/provider/ova-setup.go
```diff
@@ -0,0 +1,225 @@
...
+			Annotations:     annotations,
+			Labels:          labels,
+			OwnerReferences: []metav1.OwnerReference{ownerReference},
+		},
+		Spec: appsv1.DeploymentSpec{
+			Replicas: &replicas,
+			Selector: &metav1.LabelSelector{
+				MatchLabels: map[string]string{
+					"app":      "forklift",
+					"provider": provider.Name,
```
ahadas, line 142: another label like `"subapp": "ova-server"` can be useful

pkg/controller/plan/migration.go
```diff
@@ -336,6 +337,41 @@ func (r *Migration) Archive() {
 		return
 	}
 
+	if r.Plan.Provider.Source.Type() == v1beta1.Ova {
```
ahadas, line 340: please extract this block to a separate function - sonar cloud has a point, that function becomes more complicated then it should be

pkg/controller/plan/kubevirt.go
```diff
@@ -1336,23 +1340,22 @@ func (r *KubeVirt) podVolumeMounts(vmVolumes []cnv.Volume, configMap *core.Confi
 
 	switch r.Source.Provider.Type() {
 	case api.Ova:
-		server := r.Source.Provider.Spec.URL
-		splitted := strings.Split(server, ":")
-
-		if len(splitted) != 2 {
-			r.Log.Info("The NFS server path format is wrong")
+		pvcName := getEntityName("pvc", r.Source.Provider.Name, r.Plan.Name)
```
ahadas, line 1343: please move this line below, closer to where `pvcName` is used

ahadas (approved):

liranr23 (approved):

PR #594, opened 2023-09-18
ocp: consider missing namespace in network name

bennyz:
The namespace may be missing from the network name, in this case we need to use the VM's namespace

### DIFF
--- a/pkg/controller/plan/adapter/ocp/validator.go
+++ b/pkg/controller/plan/adapter/ocp/validator.go
@@ -161,15 +161,7 @@ func (r *Validator) NetworksMapped(vmRef ref.Ref) (ok bool, err error) {
 				return false, err
 			}
 		} else if net.Multus != nil {
-			var namespace, name string
-
-			if !strings.Contains(net.Multus.NetworkName, "/") {
-				namespace = vmRef.Namespace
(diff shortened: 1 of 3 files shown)


ahadas (approved):

pkg/controller/plan/adapter/ocp/validator.go
```diff
@@ -161,8 +161,16 @@ func (r *Validator) NetworksMapped(vmRef ref.Ref) (ok bool, err error) {
...
-			name := strings.Split(net.Multus.NetworkName, "/")[1]
+			var namespace, name string
+
+			if !strings.Contains(net.Multus.NetworkName, "/") {
+				namespace = vmRef.Namespace
+				name = net.Multus.NetworkName
+			} else {
+				splitName := strings.Split(net.Multus.NetworkName, "/")
+				namespace, name = splitName[0], splitName[1]
+			}
```
ahadas, line 172: @bennyz how about extracting this to a separate function to reduce cognitive complexity?

bennyz: ack

ahadas (approved):
great, I was about to ask for a unit test as well and forgot

PR #596, opened 2023-09-20
Single webhook per configuration entity

ahadas:
We've noticed an unexpected behavior in OpenShift:
1. ValidatingWebhookConfiguration or MutatingWebhookConfiguration is posted with a webhook
2. CA bundle is injected to the webhook
3. More webhooks are added to the configuration entity
4. Some of the webhooks (in our case it has always been the last one) does not get a CA bundle injected

This is a workaround for this issue - avoid adding webhooks to an existing configuration entity by separating each webhook to a different configuration entity.


liranr23:
will it be possible to hit the issue once all of the existing configurations are set and we change one of them? what will happen in that case? couldn't we just update the existing configurations?

ahadas:
> will it be possible to hit the issue once all of the existing configurations are set and we change one of them? what will happen in that case? couldn't we just update the existing configurations?

there's no problem with changing them, this issue happened when adding webhooks - so as long as you don't add webhooks to existing configuration entity, it would be fine

bennyz (approved):

PR #603, opened 2023-09-26
Add golangci-lint github action

bennyz:


.github/workflows/golangci-lint.yaml
```diff
@@ -0,0 +1,55 @@
+name: golangci-lint
+on:
+  push:
+    branches:
+      - main
+  pull_request:
```
ahadas, line 6: @bennyz shouldn't it run on pull_request?

bennyz: I think it will run, but yes, it would make sense to remove push here

ahadas (approved):

PR #605, opened 2023-09-27
ocp: fix network lookup

bennyz:
name and namespace were in an incorrect order

### DIFF
--- a/pkg/controller/plan/adapter/ocp/validator.go
+++ b/pkg/controller/plan/adapter/ocp/validator.go
@@ -158,19 +158,21 @@ func (r *Validator) NetworksMapped(vmRef ref.Ref) (ok bool, err error) {
 					"vm",
 					vmRef.String(),
 				)
+				r.log.Error(err, "Pod network not found.")
+
 				return false, err
 			}
 		} else if net.Multus != nil {
-			namespace, name := getNetworkNameAndNamespace(net.Multus.NetworkName, &vmRef)
-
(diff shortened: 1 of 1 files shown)


pkg/controller/plan/adapter/ocp/validator.go
```diff
@@ -158,24 +160,27 @@ func (r *Validator) NetworksMapped(vmRef ref.Ref) (ok bool, err error) {
...
 			_, found := r.plan.Referenced.Map.Network.FindNetworkByNameAndNamespace(namespace, name)
 			if !found {
 				err = liberr.Wrap(
 					err,
 					"Multus network not found.",
-					"vm",
+					"network",
 					fmt.Sprintf("%s/%s", namespace, name),
 				)
+				r.log.Error(err, "Multus network not found.")
```
ahadas, line 177: does it include the name of the network?

ahadas (commented):
one question and please remove the debug logs

pkg/controller/plan/adapter/ocp/validator.go
bennyz: no, but do we want to copy the entire thing twice?

ahadas: ok, let's leave it for now

ahadas (approved):

ahadas:
@bennyz please backport

PR #609, opened 2023-09-28
Increase timeout of validation webhooks to 30 sec

ahadas:
10 seconds are usually enough but in some cases, it takes few seconds to instantiate a client for interacting with the cluster and then this 10 seconds timeout is reached. Therefore, aligning the timeout of the validation webhooks with that of mutating webhooks.

backport of https://github.com/kubev2v/forklift/pull/607


ahadas:
backport of #607

PR #616, opened 2023-10-03
Bump images dependencies 

liranr23:


.github/workflows/pull-request.yml
```diff
@@ -104,7 +104,7 @@ jobs:
         uses: actions/setup-go@v2
         with:
           # NOTE: Keep the version in sync with Go toolchain in WORKSPACE.
-          go-version: '1.19.3'
+          go-version: '1.21.1'
```
ahadas, line 107: go-toolset in the red hat catalog is currently based on 1.19
jumping to 1.21 might be too fast, how about jumping to 1.20?

WORKSPACE
ahadas, line 1: I remember that @bennyz suggested to drop few dependencies that are not used, did you check whether all those dependencies that are updated here are actually used? it's a good opportunity for a cleanup

hack/ova-provider-server/Containerfile
```diff
@@ -1,4 +1,4 @@
-FROM registry.access.redhat.com/ubi9/go-toolset:1.19.10-4 as builder
+FROM registry.access.redhat.com/ubi9/go-toolset:1.19.10-14.1695131433 as builder
```
ahadas, line 1: it's not great that the version of go we use here is not aligned with that of bazel - how about converting this to bazel?

hack/ovirt-populator/Containerfile
ahadas, line 1: we can also drop this one, no?

ahadas (commented):
I'd suggest to separate out the changes of virt-v2v dependencies to a separate PR as it's just about running a script so it can get it quickly

ahadas (approved):
looks good

PR #615, opened 2023-10-03
bump ovirt-imageio and virt-v2v dependencies

bkhizgiy:

PR #617, opened 2023-10-03
Bump v2v rpm deps

liranr23:


ahadas:
@bkhizgiy included this in #615

PR #618, opened 2023-10-04
Backport https://github.com/kubev2v/forklift/pull/592

bkhizgiy:
Until now when adding a new OVA provider we were mount the NFS volume directly to the server/ conversion pod, this worked for some of the cases but wasn't supported on a restricted or different namespaces from forklift/MTV since NFS is not permitted. This fix switch the use of NFS to used shared PVC for both of the cases instead.

### DIFF
--- a/pkg/controller/plan/kubevirt.go
+++ b/pkg/controller/plan/kubevirt.go
@@ -21,6 +21,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	k8svalidation "k8s.io/apimachinery/pkg/util/validation"
+	"k8s.io/apimachinery/pkg/util/wait"
 	cnv "kubevirt.io/api/core/v1"
 	libvirtxml "libvirt.org/libvirt-go-xml"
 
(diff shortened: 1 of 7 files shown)


pkg/controller/plan/kubevirt.go
```diff
@@ -1835,6 +1838,163 @@ func (r *KubeVirt) EnsurePersistentVolume(vmRef ref.Ref, persistentVolumes []cor
...
+	pvcNamespacedName := types.NamespacedName{
+		Namespace: r.Plan.Spec.TargetNamespace,
+		Name:      pvcName,
+	}
+
+	timeout := 45 * time.Second
+	ctx, cancel := context.WithTimeout(context.TODO(), timeout)
+	defer cancel()
+
+	// wait until pvc and pv are bounded.
```
ahadas, line 1976:
```suggestion
	// wait until pvc and pv are bound.
```

pkg/controller/plan/kubevirt.go
```diff
@@ -1835,6 +1838,163 @@ func (r *KubeVirt) EnsurePersistentVolume(vmRef ref.Ref, persistentVolumes []cor
...
+		if err != nil {
+			r.Log.Error(err, "Failed to get OVA plan PVC")
+			return false, err
+		}
+		return pvc.Status.Phase == "Bound", nil
+	}, ctx.Done())
+	if err != nil {
+		r.Log.Error(err, "Failed to bind OVA PVC to PV ")
+		return
+
```
ahadas, line 1988: please drop this empty line

pkg/controller/plan/kubevirt.go
```diff
@@ -1835,6 +1838,163 @@ func (r *KubeVirt) EnsurePersistentVolume(vmRef ref.Ref, persistentVolumes []cor
...
+		Namespace: r.Plan.Spec.TargetNamespace,
+		Name:      pvcName,
+	}
+
+	timeout := 45 * time.Second
+	ctx, cancel := context.WithTimeout(context.TODO(), timeout)
+	defer cancel()
+
+	// wait until pvc and pv are bounded.
+	err = wait.PollImmediateUntil(5*time.Second, func() (bool, error) {
```
ahadas, line 1977:
```suggestion
	err = wait.PollImmediateUntil(5 * time.Second, func() (bool, error) {
```

ahadas (commented):
minor comments but we'll need to verify these changes as well

pkg/controller/plan/kubevirt.go
ahadas: ok then don't worry about it

ahadas (approved):

PR #630, opened 2023-10-18
Add e2e test for ova provider

bkhizgiy:


tests/suit/framework/framework.go
```diff
@@ -85,6 +85,8 @@ type Clients struct {
 	OvirtClient OvirtClient
 	// OpenStackClient provides a pointer to Openstack client.
 	OpenStackClient OpenStackClient
+	// OVAClient
```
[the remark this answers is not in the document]
ahadas: these comments (`XClient providers a pointer to X client`) don't add any value in my opinion, we can just drop them, but if we keep them then `OvaClient`..

tests/suit/framework/ova.go
```diff
@@ -0,0 +1,71 @@
...
+	return &r.vmData, nil
+}
+
+func (r *OvaClient) GetNfsServerForOva(k8sClient *kubernetes.Clientset) (string, error) {
+	storageClass, err := k8sClient.StorageV1().StorageClasses().Get(context.TODO(), r.storageClass, metav1.GetOptions{})
+	if err != nil {
+		return "", err
+	}
+	var nfsShare string
+	for parm, val := range storageClass.Parameters {
```
ahadas, line 31: what prevents us from getting nfsShare=share+server (instead of server+share)?
can you give an example of what we expect to get here?

tests/suit/framework/ova.go
```diff
@@ -0,0 +1,71 @@
...
+		if parm == "server" {
+			nfsShare = nfsShare + val
+		}
+		if parm == "share" {
+			nfsShare = nfsShare + val
+		}
+	}
+	if nfsShare != "" {
```
ahadas, line 39: when would nfsShare be ""?

tests/suit/ova_test.go
```diff
@@ -0,0 +1,81 @@
...
+
+		By("Load Source VM Details from OVA")
+		vmData, err := f.Clients.OvaClient.LoadSourceDetails()
+		Expect(err).ToNot(HaveOccurred())
+
+		By("Get NFS share for OVA provider")
+		nfs, err := f.Clients.OvaClient.GetNfsServerForOva(f.K8sClient)
+		Expect(err).ToNot(HaveOccurred())
+		By("Create Secret from Definition")
+		s, err := utils.CreateSecretFromDefinition(f.K8sClient, utils.NewSecretDefinition(nil, nil,
```
ahadas, line 32:
```suggestion
		secret, err := utils.CreateSecretFromDefinition(f.K8sClient, utils.NewSecretDefinition(nil, nil,
```

tests/suit/utils/storagemap.go
```diff
@@ -29,14 +30,26 @@ func NewStorageMap(namespace string, providerIdentifier forkliftv1.Provider, sto
...
-	for _, sd := range storageIDs {
-		pair := forkliftv1.StoragePair{
-			Source: ref.Ref{ID: sd},
-			Destination: forkliftv1.DestinationStorage{
-				StorageClass: storageClass,
-			},
+	if providerIdentifier.Type() != api.Ova {
+		for _, sd := range storageIDs {
+			pair := forkliftv1.StoragePair{
+				Source: ref.Ref{ID: sd},
```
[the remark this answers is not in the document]
ahadas: yeah, we can decrease the code duplication here

ahadas:
@bkhizgiy I've merged the PR for forkliftci, note that there's a new version of forkliftci, v9.0, that includes the OVA. need to bump the version we use here

tests/suit/framework/ova.go
bkhizgiy: If we ruining locally and the env variables wasn't set when running the CI setup, but we can drop it.

tests/suit/framework/ova.go
bkhizgiy: I'm not sure I've understand the question, you mean why we fetch it saperaly? or why the order set this way?

bkhizgiy:
Still missing two CI changes.
+ need to change the vm.ova file used, the current one don't have proper network config for the test.
+ need to understand how to use virt-v2v real image (and not stub) for ALL endpoint in the CI.

bkhizgiy:
Need to merge PR https://github.com/kubev2v/forklift/pull/643 for the tests to pass.

bkhizgiy:
Missing https://github.com/kubev2v/forkliftci/pull/114 for pulling images.

tests/suit/framework/ova.go
bkhizgiy: yeah we need to return error if its empty, added in the code.

ahadas (approved):

PR #637, opened 2023-10-31
Keep the populator pod on failure

liranr23:
This patch will keep the populator pod existence when failing. It won't restart or recreated. This is in the mind of aligning with v2v pod failure and the thought that once we fail, we will most likely keep failing. This change will allow to get the populator logs in case of failure in a easy way.

### DIFF
--- a/pkg/controller/plan/migration.go
+++ b/pkg/controller/plan/migration.go
@@ -89,7 +89,8 @@ const (
 )
 
 const (
-	TransferCompleted = "Transfer completed."
+	TransferCompleted  = "Transfer completed."
+	PopulatorPodPrefix = "populate-"
 )
 
 var (
@@ -1620,7 +1621,7 @@ func (r *Migration) updatePopulatorCopyProgress(vm *plan.VMStatus, step *plan.St
 			continue
 		}
 
-		err = r.isPopulatorFailed(populatorPods, string(pvc.UID))
+		_, err = r.isPopulatorFailed(populatorPods, string(pvc.UID))
 		if err != nil {
 			return
 		}
@@ -1647,17 +1648,18 @@ func (r *Migration) updatePopulatorCopyProgress(vm *plan.VMStatus, step *plan.St
 	return
 }
 
-func (r *Migration) isPopulatorFailed(populatorPods []core.Pod, givenPvcId string) error {
+func (r *Migration) isPopulatorFailed(populatorPods []core.Pod, givenPvcId string) (bool, error) {
 	for _, pod := range populatorPods {
-		pvcId := strings.Split(pod.Name, "populate-")[1]
+		pvcId := pod.Name[len(PopulatorPodPrefix):]
 		if givenPvcId != pvcId {
 			continue
 		}
 		if pod.Status.Phase == core.PodFailed {
-			return fmt.Errorf("populator pod %s/%s failed for PVC %s. Please check the pod logs.", pod.Namespace, pod.Name, pvcId)
+			return true, fmt.Errorf("populator pod %s/%s failed for PVC %s. Please check the pod logs.", pod.Namespace, pod.Name, pvcId)
 		}
+		break
 	}
-	return nil
+	return false, nil
 }
 
 func (r *Migration) setPopulatorPodsWithLabels(vm *plan.VMStatus, migrationID string) {
@@ -1666,7 +1668,7 @@ func (r *Migration) setPopulatorPodsWithLabels(vm *plan.VMStatus, migrationID st
 		return
 	}
 	for _, pod := range podList.Items {
-		if strings.HasPrefix(pod.Name, "populate-") {
+		if strings.HasPrefix(pod.Name, PopulatorPodPrefix) {
 			// it's populator pod
 			if _, ok := pod.Labels["migration"]; !ok {
 				// un-labeled pod, we need to set it
--- a/pkg/controller/plan/migration.go
+++ b/pkg/controller/plan/migration.go
@@ -1453,7 +1453,7 @@ func (r *Migration) updateCopyProgress(vm *plan.VMStatus, step *plan.Step) (err 
 				if r.Plan.Spec.Warm && len(importer.Status.ContainerStatuses) > 0 {
 					vm.Warm.Failures = int(importer.Status.ContainerStatuses[0].RestartCount)
 				}
-				if RestartLimitExceeded(importer) {
+				if restartLimitExceeded(importer) {
 					task.MarkedCompleted()
 					msg, _ := terminationMessage(importer)
 					task.AddError(msg)
@@ -1598,6 +1598,10 @@ func (r *Migration) updatePopulatorCopyProgress(vm *plan.VMStatus, step *plan.St
 	if err != nil {
 		return
 	}
+	populatorPods, err := r.kubevirt.getPopulatorPods()
+	if err != nil {
+		return
+	}
 
 	for _, pvc := range pvcs {
 		if _, ok := pvc.Annotations["lun"]; ok {
@@ -1614,6 +1618,16 @@ func (r *Migration) updatePopulatorCopyProgress(vm *plan.VMStatus, step *plan.St
 		found := false
 		if task, found = step.FindTask(taskName); !found {
 			continue
+		}
+
+		for _, pod := range populatorPods {
+			pvcId := strings.Split(pod.Name, "populate-")[1]
+			if string(pvc.UID) != pvcId {
+				continue
+			}
+			if pod.Status.Phase == core.PodFailed {
+				return fmt.Errorf("populator pod %s/%s failed for PVC %s. Please check the pod logs.", pod.Namespace, pod.Name, pvcId)
+			}
 		}
 
 		if pvc.Status.Phase == core.ClaimBound {
@@ -1701,7 +1715,7 @@ func terminationMessage(pod *core.Pod) (msg string, ok bool) {
 }
 
 // Return whether the pod has failed and restarted too many times.
-func RestartLimitExceeded(pod *core.Pod) (exceeded bool) {
+func restartLimitExceeded(pod *core.Pod) (exceeded bool) {
 	if len(pod.Status.ContainerStatuses) == 0 {
 		return
 	}
--- a/pkg/controller/plan/migration.go
+++ b/pkg/controller/plan/migration.go
@@ -1599,10 +1599,6 @@ func (r *Migration) updatePopulatorCopyProgress(vm *plan.VMStatus, step *plan.St
 	if err != nil {
 		return
 	}
-	populatorPods, err := r.kubevirt.getPopulatorPods()
-	if err != nil {
-		return
-	}
 
 	for _, pvc := range pvcs {
 		if _, ok := pvc.Annotations["lun"]; ok {
@@ -1621,11 +1617,6 @@ func (r *Migration) updatePopulatorCopyProgress(vm *plan.VMStatus, step *plan.St
 			continue
 		}
 
-		_, err = r.isPopulatorFailed(populatorPods, string(pvc.UID))
-		if err != nil {
-			return
-		}
-
 		if pvc.Status.Phase == core.ClaimBound {
 			task.Phase = Completed
 			task.Reason = TransferCompleted
@@ -1641,25 +1632,39 @@ func (r *Migration) updatePopulatorCopyProgress(vm *plan.VMStatus, step *plan.St
 		}
 
 		percent := float64(transferredBytes/0x100000) / float64(task.Progress.Total)
-		task.Progress.Completed = int64(percent * float64(task.Progress.Total))
+		newProgress := int64(percent * float64(task.Progress.Total))
+		if newProgress == task.Progress.Completed {
+			pvcId := string(pvc.UID)
+			populatorFailed := r.isPopulatorPodFailed(pvcId)
+			if populatorFailed {
+				return fmt.Errorf("populator pod failed for PVC %s. Please check the pod logs", pvcId)
+			}
+		}
+		task.Progress.Completed = newProgress
 	}
 
 	step.ReflectTasks()
 	return
 }
 
-func (r *Migration) isPopulatorFailed(populatorPods []core.Pod, givenPvcId string) (bool, error) {
+// Checks if the populator pod failed when the progress didn't change
+func (r *Migration) isPopulatorPodFailed(givenPvcId string) bool {
+	populatorPods, err := r.kubevirt.getPopulatorPods()
+	if err != nil {
+		r.Log.Error(err, "couldn't get the populator pods")
+		return false
+	}
 	for _, pod := range populatorPods {
 		pvcId := pod.Name[len(PopulatorPodPrefix):]
 		if givenPvcId != pvcId {
 			continue
 		}
 		if pod.Status.Phase == core.PodFailed {
-			return true, fmt.Errorf("populator pod %s/%s failed for PVC %s. Please check the pod logs.", pod.Namespace, pod.Name, pvcId)
+			return true
 		}
 		break
 	}
-	return false, nil
+	return false
 }
 
 func (r *Migration) setPopulatorPodsWithLabels(vm *plan.VMStatus, migrationID string) {
--- a/pkg/controller/plan/migration.go
+++ b/pkg/controller/plan/migration.go
@@ -1620,14 +1620,9 @@ func (r *Migration) updatePopulatorCopyProgress(vm *plan.VMStatus, step *plan.St
 			continue
 		}
 
-		for _, pod := range populatorPods {
-			pvcId := strings.Split(pod.Name, "populate-")[1]
-			if string(pvc.UID) != pvcId {
-				continue
-			}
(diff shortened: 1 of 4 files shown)


pkg/lib-volume-populator/populator-machinery/controller.go
```diff
@@ -698,13 +697,6 @@ func (c *controller) syncPvc(ctx context.Context, key, pvcNamespace, pvcName str
 		if corev1.PodSucceeded != pod.Status.Phase {
 			if corev1.PodFailed == pod.Status.Phase {
 				c.recorder.Eventf(pvc, corev1.EventTypeWarning, reasonPodFailed, "Populator failed: %s", pod.Status.Message)
```
liranr23, line 699: the `pod.Status.Message` will most likely be empty string. but i didn't think of something better (not even the containerstatues.

pkg/controller/plan/migration.go
```diff
@@ -1638,6 +1647,19 @@ func (r *Migration) updatePopulatorCopyProgress(vm *plan.VMStatus, step *plan.St
 	return
 }
 
+func (r *Migration) isPopulatorFailed(populatorPods []core.Pod, givenPvcId string) error {
+	for _, pod := range populatorPods {
+		pvcId := strings.Split(pod.Name, "populate-")[1]
```
ahadas, line 1652: how about extracting `popoulate-` to a constant and use it also in getPopulatorPods?
also `Split` is a bit cumbersome for this purpose, why not to slice the string?

pkg/controller/plan/migration.go
```diff
@@ -1638,6 +1647,19 @@ func (r *Migration) updatePopulatorCopyProgress(vm *plan.VMStatus, step *plan.St
 	return
 }
 
+func (r *Migration) isPopulatorFailed(populatorPods []core.Pod, givenPvcId string) error {
```
ahadas, line 1650: that's a bit awkward to have a function called `isPopulatorFailed` that returns an error - usually an error indicates that an error was raised when executing the function.. how about returning a boolean, or a boolean with an error if we need the error?

pkg/controller/plan/migration.go
```diff
@@ -1638,6 +1647,19 @@ func (r *Migration) updatePopulatorCopyProgress(vm *plan.VMStatus, step *plan.St
...
+	for _, pod := range populatorPods {
+		pvcId := strings.Split(pod.Name, "populate-")[1]
+		if givenPvcId != pvcId {
+			continue
+		}
+		if pod.Status.Phase == core.PodFailed {
+			return fmt.Errorf("populator pod %s/%s failed for PVC %s. Please check the pod logs.", pod.Namespace, pod.Name, pvcId)
+		}
```
ahadas, line 1658: can we break if pod.Status.Phase == core.PodFailed?

pkg/lib-volume-populator/populator-machinery/controller.go
```diff
@@ -698,13 +697,6 @@ func (c *controller) syncPvc(ctx context.Context, key, pvcNamespace, pvcName str
...
-				// Delete failed pods so we can try again
-				if !plan.RestartLimitExceeded(pod) {
-					err = c.kubeClient.CoreV1().Pods(populatorNamespace).Delete(ctx, pod.Name, metav1.DeleteOptions{})
-					if err != nil {
-						return err
-					}
-				}
```
ahadas, line 707: I don't understand this - we won't delete the pods here but still check the amount of restarts in restartLimitExceeded?
the description of the PR is not that clear - what's the problem in restarting the pod and keeping the last one in failed state?

liranr23: i removed this part - the pod won't restart and we don't need to check anything regarding it.
`keeping the last one` ?

ahadas: > `keeping the last one` ?

yeah, I see in the description of the PR that "This change will allow to get the populator logs in case of failure in a easy way" but I don't understand why it's easier. let's say we try to populate a disk and we have three attempts and we fail in all of them - so the first populator pod would fail, we'll delete it, start a second populator pod that would also fail, we'll delete it, start a third populator pod that would also fail. why is it more complicated to get the logs from this pod compared to a single pod that we start and fail?

liranr23: if we delete the pod - it will be recreated, resetting the restartCount in that pod. that will cause us to be in an endless loop (restarted 3 times, recreate, restarted 3 times, recreate...).
it is easier because if the pod restarts or recreated, it depends on the phase you failed and the timing (if the pod just recreated/restarted you won't be able to get them).

ahadas: > it is easier because if the pod restarts or recreated, it depends on the phase you failed and the timing (if the pod just recreated/restarted you won't be able to get them).

right, but I don't see this as an issue since the migration should not be set with failure state at this point
if we manage to set the migration as failed only when the last attempt fails then we shouldn't care about the logs from previous attempt

the problem I see with this approach of changing the restart-policy of the populator pod is that currently CDI restarts them several times, now we drop this functionality, and I supposed that when we integrate our populator pods into CDI we'll also have several attempts..

bennyz: In CDI the annotation is set on the PVC, so there isn't really an issue with losing track of the count

pkg/lib-volume-populator/populator-machinery/controller.go
liranr23: we can tell to look into the pod logs: Please check the logs in <namespace/podName>

pkg/lib-volume-populator/populator-machinery/controller.go
liranr23: discussed offline with benny about it. @ahadas how about 3 re-creations and then to stop it, leaving the pod existing and telling the migration failed?
it means the number of creation need to be saved somewhere.

ahadas: yeah, that makes sense
I'd store it on the CR, assuming we can keep it this way also when the volume populators are integrated in CDI

liranr23: currently i created another commit. the pod will be recreated 3 times as agreed. i saved the annotation on the PVC. we can change it to the CR. i don't have strong opinion about it. as @bennyz said, CDI saves it on the PVC. but it depends how you wish it to be in CDI after integration.

pkg/controller/plan/migration.go
```diff
@@ -1638,13 +1648,27 @@ func (r *Migration) updatePopulatorCopyProgress(vm *plan.VMStatus, step *plan.St
...
 }
 
+func (r *Migration) isPopulatorFailed(populatorPods []core.Pod, givenPvcId string) (bool, error) {
+	for _, pod := range populatorPods {
+		pvcId := pod.Name[len(PopulatorPodPrefix):]
+		if givenPvcId != pvcId {
+			continue
+		}
+		if pod.Status.Phase == core.PodFailed {
+			return true, fmt.Errorf("populator pod %s/%s failed for PVC %s. Please check the pod logs.", pod.Namespace, pod.Name, pvcId)
```
ahadas, line 1658: if we keep this function, it's not that good to return both boolean and error here since an error that is returned from a function that is called isPopulatorFailed suggests that when an error is returned, it means we **couldn't check** whether the populator pod failed..
I'd change this to return just an error when this function return true if needed

pkg/lib-volume-populator/populator-machinery/controller.go
```diff
@@ -697,14 +697,28 @@ func (c *controller) syncPvc(ctx context.Context, key, pvcNamespace, pvcName str
...
-				if !plan.RestartLimitExceeded(pod) {
-					err = c.kubeClient.CoreV1().Pods(populatorNamespace).Delete(ctx, pod.Name, metav1.DeleteOptions{})
+				restarts, ok := pvc.Annotations[AnnPopulatorReCreations]
+				if !ok {
+					pvc.Annotations[AnnPopulatorReCreations] = "1"
+					err = c.restartFailedPopulator(ctx, pvc, populatorNamespace, pod.Name)
 					if err != nil {
 						return err
 					}
+					return nil
```
ahadas, line 707: so can we just have `return err`?

pkg/lib-volume-populator/populator-machinery/controller.go
```diff
@@ -697,14 +697,28 @@ func (c *controller) syncPvc(ctx context.Context, key, pvcNamespace, pvcName str
...
+				}
+				if restartsInteger < 3 {
+					pvc.Annotations[AnnPopulatorReCreations] = strconv.Itoa(restartsInteger + 1)
+					err = c.restartFailedPopulator(ctx, pvc, populatorNamespace, pod.Name)
+					if err != nil {
+						return err
+					}
+					return nil
+				}
+				c.recorder.Eventf(pvc, corev1.EventTypeWarning, reasonPodFailed, "Populator failed: Please check the logs in populator pod, %s/%s", populatorNamespace, pod.Name)
```
ahadas, line 721:
```suggestion
				c.recorder.Eventf(pvc, corev1.EventTypeWarning, reasonPodFailed, "Populator failed: Please check the logs of the populator pod, %s/%s", populatorNamespace, pod.Name)
```
we can also say that few (3) attempts failed

pkg/lib-volume-populator/populator-machinery/controller.go
ahadas: but isn't there a practical reason to store it on the PVC? I see the current code, it handles all populator pods - if we store the counter on the CR, we'll need to differentiate ovirt and openstack, no?

pkg/lib-volume-populator/populator-machinery/controller.go
```diff
@@ -799,6 +813,26 @@ func (c *controller) syncPvc(ctx context.Context, key, pvcNamespace, pvcName str
 	return nil
 }
 
+func (c *controller) restartFailedPopulator(ctx context.Context, pvc *corev1.PersistentVolumeClaim, namespace, podName string) (err error) {
+	err = c.updatePvc(ctx, pvc, namespace)
```
ahadas, line 817: I'd separate this to a separate function - it's not part of restarting the populator pod

pkg/controller/plan/migration.go
```diff
@@ -1616,6 +1621,11 @@ func (r *Migration) updatePopulatorCopyProgress(vm *plan.VMStatus, step *plan.St
 			continue
 		}
 
+		_, err = r.isPopulatorFailed(populatorPods, string(pvc.UID))
```
ahadas, line 1624: why do we need to check it here?

pkg/lib-volume-populator/populator-machinery/controller.go
liranr23: yes, if we will store it on the CR we will need to differentiate, as it's unstructured.

ahadas: ack, so let's keep it on the PVC then

pkg/controller/plan/migration.go
ahadas: but we'd know when the failure happened from the logs of the populator pod, right?
it seems redundant to do this as part of the progress tracking if we anyway track the status of the pod elsewhere

pkg/lib-volume-populator/populator-machinery/controller.go
ahadas: if you say that this methods is responsible for triggering another attempt, including updating the counter, that's fine but then let's make it do it (increment the counter)

pkg/controller/plan/migration.go
liranr23: based on offline discussions - CDI also check the pod status.
i think it's the best option. but i will try to pass it to the end of the function, if the progress is the same, query if the progress is the same. this might reduce unnecessary queries.

pkg/controller/plan/migration.go
```diff
@@ -1631,20 +1632,47 @@ func (r *Migration) updatePopulatorCopyProgress(vm *plan.VMStatus, step *plan.St
...
+		task.Progress.Completed = newProgress
 	}
 
 	step.ReflectTasks()
 	return
 }
 
+// Checks if the populator pod failed when the progress didn't change
+func (r *Migration) checkPopulatorPod(newProgress, oldProgress int64, givenPvcId string) error {
+	if newProgress != oldProgress {
```
ahadas, line 1649: it's a bit weird to pass the progress to this function just to check that they are different here, how about doing that in updatePopulatorCopyProgress?

pkg/controller/plan/migration.go
```diff
@@ -1631,20 +1632,47 @@ func (r *Migration) updatePopulatorCopyProgress(vm *plan.VMStatus, step *plan.St
 		}
 
 		percent := float64(transferredBytes/0x100000) / float64(task.Progress.Total)
-		task.Progress.Completed = int64(percent * float64(task.Progress.Total))
+		newProgress := int64(percent * float64(task.Progress.Total))
+		err = r.checkPopulatorPod(newProgress, task.Progress.Completed, string(pvc.UID))
+		if err != nil {
+			return
```
ahadas, line 1638: what kind of error do we expect here? I see we can get an error when the pod failed  (the one that is created in checkPopulatorPod) or an error that is returned by getPopulatorPods - should we return an error here when we fail to query for pods?

liranr23: depends on us... it does hints something is wrong. but if we will return nil in such case we will reconcile and try again. maybe better to log and reconcile.

pkg/lib-volume-populator/populator-machinery/controller.go
```diff
@@ -799,6 +805,27 @@ func (c *controller) syncPvc(ctx context.Context, key, pvcNamespace, pvcName str
 	return nil
 }
 
+func (c *controller) retryFailedPopulator(ctx context.Context, pvc *corev1.PersistentVolumeClaim, namespace, podName, counter string) (err error) {
+	pvc.Annotations[AnnPopulatorReCreations] = counter
```
ahadas, line 809: how about converting the counter to string here? it makes more sense for counter to be a number when it's passed here

pkg/lib-volume-populator/populator-machinery/controller.go
```diff
@@ -697,14 +697,20 @@ func (c *controller) syncPvc(ctx context.Context, key, pvcNamespace, pvcName str
...
-				// Delete failed pods so we can try again
-				if !plan.RestartLimitExceeded(pod) {
-					err = c.kubeClient.CoreV1().Pods(populatorNamespace).Delete(ctx, pod.Name, metav1.DeleteOptions{})
-					if err != nil {
-						return err
-					}
+				restarts, ok := pvc.Annotations[AnnPopulatorReCreations]
+				if !ok {
+					err = c.retryFailedPopulator(ctx, pvc, populatorNamespace, pod.Name, "1")
+					return err
```
ahadas, line 703: how about inline `err`?

pkg/lib-volume-populator/populator-machinery/controller.go
```diff
@@ -697,14 +697,20 @@ func (c *controller) syncPvc(ctx context.Context, key, pvcNamespace, pvcName str
...
+				}
+				restartsInteger, err := strconv.Atoi(restarts)
+				if err != nil {
+					return err
 				}
+				if restartsInteger < 3 {
+					err = c.retryFailedPopulator(ctx, pvc, populatorNamespace, pod.Name, strconv.Itoa(restartsInteger+1))
+					return err
```
ahadas, line 711: how about inline `err`?

pkg/lib-volume-populator/populator-machinery/controller.go
```diff
@@ -799,6 +805,27 @@ func (c *controller) syncPvc(ctx context.Context, key, pvcNamespace, pvcName str
...
+	}
+	return
+}
+
+func (c *controller) updatePvc(ctx context.Context, pvc *corev1.PersistentVolumeClaim, namespace string) (err error) {
+	_, err = c.kubeClient.CoreV1().PersistentVolumeClaims(namespace).Update(ctx, pvc, metav1.UpdateOptions{})
+	if err != nil {
+		return err
+	}
+	return
```
ahadas, line 826: return err?

pkg/lib-volume-populator/populator-machinery/controller.go
```diff
@@ -799,6 +805,27 @@ func (c *controller) syncPvc(ctx context.Context, key, pvcNamespace, pvcName str
 	return nil
 }
 
+func (c *controller) retryFailedPopulator(ctx context.Context, pvc *corev1.PersistentVolumeClaim, namespace, podName, counter string) (err error) {
```
ahadas, line 808: I'd prefer to change this to:
```suggestion
func (c *controller) retryFailedPopulator(ctx context.Context, pvc *corev1.PersistentVolumeClaim, namespace, podName, counter string) error {
```
since I see no value in naming the returned value here, but that's subjective - for your consideration

ahadas (commented):
looks good, minor comments

ahadas (approved):

PR #643, opened 2023-11-07
Add support for ova provider creation for k8s in restricted namespace

bkhizgiy:
This fix checks on which platform it running on and accordingly setting the appropriate pod security setting.
- For openShift we have this configured in the SCC file.
- For k8s we need to set it as art of the pod security while creation.

### DIFF
--- a/pkg/controller/provider/ova-setup.go
+++ b/pkg/controller/provider/ova-setup.go
@@ -12,6 +12,8 @@ import (
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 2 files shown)


pkg/controller/provider/ova-setup.go
```diff
@@ -188,44 +191,92 @@ func (r *Reconciler) createServerService(provider *api.Provider, ctx context.Con
...
+			FSGroup: &user,
+			SeccompProfile: &core.SeccompProfile{
+				Type: core.SeccompProfileTypeRuntimeDefault,
+			},
+		}
+	}
+	return podSpec
+}
+
+func (r *Reconciler) runningInOpenShift() bool {
```
ahadas, line 257: there should be a "cleaner" way to know the cluster type (by propagating this from the operator to an environment variable of the pod) but please clarify first why we need to differentiate them here

bkhizgiy:
replace this PR with https://github.com/kubev2v/forklift/pull/660, closing this one.

PR #645, opened 2023-11-09
Make the filesystem overhead configurable

liranr23:
This patch will allow the user to configure the filesystem overhead. Currently it defaults to 10%. The overhead is mainly required for ext4 and the root partition.

### DIFF
--- a/operator/roles/forkliftcontroller/templates/controller/deployment-controller.yml.j2
+++ b/operator/roles/forkliftcontroller/templates/controller/deployment-controller.yml.j2
@@ -74,7 +74,7 @@ spec:
 {% endif %}
 {% if controller_cdi_export_token_ttl is number %}
         - name: CDI_EXPORT_TOKEN_TTL
-          value: "{{ CDI_EXPORT_TOKEN_TTL }}"
+          value: "{{ controller_cdi_export_token_ttl }}"
 {% endif %}
 {% if controller_filesystem_overhead is number %}
         - name: FILESYSTEM_OVERHEAD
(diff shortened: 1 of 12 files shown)


operator/roles/forkliftcontroller/templates/controller/deployment-controller.yml.j2
```diff
@@ -76,6 +76,10 @@ spec:
         - name: CDI_EXPORT_TOKEN_TTL
           value: "{{ CDI_EXPORT_TOKEN_TTL }}"
```
liranr23, line 77: @bennyz might be a bug here? shouldn't it be `controller_cdi_export_token_ttl`?

bennyz: yes

bennyz (approved):

pkg/settings/settings.go
```diff
@@ -70,14 +72,20 @@ func (r *ControllerSettings) Load() error {
 // Get positive integer limit from the environment
 // using the specified variable name and default.
 func getEnvLimit(name string, def int) (int, error) {
+	return getEnvLimitMinimal(name, def, 1)
+}
+
+// Get an integer limit from the environment
+// using the specified variable name and default.
+func getEnvLimitMinimal(name string, def, minimum int) (int, error) {
```
ahadas, line 80: `getEnvLimitMinimal` doesn't indicate what the 'Minimal' part stands for clearly
how about renaming `getEnvLimit` to `getPositiveEnvLimit` and introduce another `getNonNegativeEnvLimit`?

operator/roles/forkliftcontroller/templates/controller/deployment-controller.yml.j2
liranr23: done in a new commit

ahadas (approved):

PR #646, opened 2023-11-10
ovirt: Reconcile disk on USER_UPDATE_VM_DISK

nijinashok:
When a disk is extended in RHV, the size is not updated in the inventory causing the VM to be imported with old disk size. This is now solved by reconciling the disk on USER_UPDATE_VM_DISK.

https://issues.redhat.com/browse/MTV-829


ahadas (commented):
thanks @nijinashok !
yeah, USER_UPDATE_VM_DISK is not handled properly - we should really handle it as a disk-property change rather than a vm-change, I think we can drop it from VMAdapter's handling

ahadas:
> yeah, USER_UPDATE_VM_DISK is not handled properly - we should really handle it as a disk-property change rather than a vm-change, I think we can drop it from VMAdapter's handling

after a second look, the VM-specific properties are retrieved by VMAdapter so it should still track USER_UPDATE_VM_DISK events, so no need to change VMAdapter. I pushed a minor change with another commit, that places USER_UPDATE_VM_DISK under `// Disk` in the DiskAdapter

ahadas (approved):

PR #651, opened 2023-11-13
Bump sqlite

liranr23:
The current go-sql dependency contains CVE-2022-35737. Therefore, we need to bump the version of it.


bennyz (approved):

ahadas:
@liranr23 please don't forget to run `hack/post-vendor.sh`

PR #653, opened 2023-11-13
Add negative test for ovf wrong depth

bkhizgiy:


cmd/ova-provider-server/nfs_test.go
```diff
@@ -61,6 +61,18 @@ func TestFindOVAFiles(t *testing.T) {
 			expectedOVFs: nil,
 			expectError:  false,
 		},
+		{
+			name: "incorrect depth ovf",
+			setup: func(directory string) {
+				//nolint:errcheck
```
bkhizgiy, line 67: since its a test and the error checked automatically it makes no sense to add error handle here, excluded from lint.

ahadas (approved):

PR #655, opened 2023-11-16
Draft: test CI

bkhizgiy:


liranr23:
shouldn't you change the forkliftci `ref: v8.0` in github actions to v9?

PR #657, opened 2023-11-20
Add support of direct FC storage from ovirt

liranr23:
This PR will allow migrations of direct FC storage from oVirt.


pkg/controller/plan/adapter/ovirt/validator.go
```diff
@@ -105,20 +108,96 @@ func (r *Validator) StorageMapped(vmRef ref.Ref) (ok bool, err error) {
...
+				}
+			}
+		}
+	}
+	ok = true
+	return
+}
+
+// Checks the version for ovirt direct LUN/FC
+func (r *Validator) checkOvirtVersion() (ok bool, err error) {
```
bennyz, line 153: suggestion for slight simplification:
```go
func (r *Validator) checkOvirtVersion() (bool, error) {
    rl := container.Build(nil, r.plan.Referenced.Provider.Source, r.plan.Referenced.Secret)
    major, minor, build, revision, err := rl.Version()
    if err != nil {
        return false, err
    }

    majorVal, err := strconv.Atoi(major)
    if err != nil {
        return false, err
    }
    minorVal, err := strconv.Atoi(minor)
    if err != nil {
        return false, err
    }
    buildVal, err := strconv.Atoi(build)
    if err != nil {
        return false, err
    }
    revisionVal, err := strconv.Atoi(revision)
    if err != nil {
        return false, err
    }

    currentVersion := majorVal*1000 + minorVal*100 + buildVal*10 + revisionVal

    const minVersion = 4521

    return currentVersion >= minVersion, nil
}

```

pkg/controller/plan/adapter/ovirt/client.go
```diff
@@ -286,6 +286,22 @@ func (r *Client) getVM(vmRef ref.Ref) (ovirtVm *ovirtsdk.Vm, vmService *ovirtsdk
 			vmRef.String())
 		return
 	}
+	//system, err := r.connection.SystemService().Get().Send()
```
liranr23, line 289: need to remove this part

pkg/controller/plan/adapter/base/doc.go
```diff
@@ -106,6 +106,8 @@ type Client interface {
 type Validator interface {
 	// Validate that a VM's disk backing storage has been mapped.
 	StorageMapped(vmRef ref.Ref) (bool, error)
+	// Validate that a VM's direct LUN/FC has the details (oVirt only)
```
ahadas, line 109:
```suggestion
	// Validate that a VM's direct LUN/FC has the required details (oVirt only)
```

pkg/controller/plan/adapter/ovirt/client.go
ahadas: > need to remove this part

yep

pkg/controller/plan/adapter/ocp/validator.go
```diff
@@ -179,3 +179,8 @@ func (r *Validator) NetworksMapped(vmRef ref.Ref) (ok bool, err error) {
 
 	return true, nil
 }
+
+// NO-OP
+func (r *Validator) DirectStorage(vmRef ref.Ref) (bool, error) {
+	return true, nil
```
ahadas, line 185: it's actually interesting to see how CDI handles direct LUNs, maybe it shouldn't remain NO-OP for OCP-to-OCP
@bennyz

pkg/controller/plan/adapter/ovirt/validator.go
```diff
@@ -105,20 +108,79 @@ func (r *Validator) StorageMapped(vmRef ref.Ref) (ok bool, err error) {
 			vmRef.String())
 		return
 	}
+
 	for _, da := range vm.DiskAttachments {
 		if da.Disk.StorageType != "lun" {
 			if !r.plan.Referenced.Map.Storage.Status.Refs.Find(ref.Ref{ID: da.Disk.StorageDomain}) {
 				return
 			}
-		} else if len(da.Disk.Lun.LogicalUnits.LogicalUnit) > 0 && da.Disk.Lun.LogicalUnits.LogicalUnit[0].Address == "" {
```
ahadas, line 113: good, so with this, how about changing line 113 to:
```golang
if da.Disk.StorageType != "lun" && !r.plan.Referenced.Map.Storage.Status.Refs.Find(ref.Ref{ID: da.Disk.StorageDomain}) {
```

pkg/controller/plan/adapter/ovirt/validator.go
```diff
@@ -105,20 +108,79 @@ func (r *Validator) StorageMapped(vmRef ref.Ref) (ok bool, err error) {
...
+			"VM not found in inventory.",
+			"vm",
+			vmRef.String())
+		return
+	}
+
+	for _, da := range vm.DiskAttachments {
+		if da.Disk.StorageType == "lun" {
+			if len(da.Disk.Lun.LogicalUnits.LogicalUnit) > 0 {
+				if ok, err := r.checkOvirtVersion(); err != nil || !ok {
```
ahadas, line 142:
```suggestion
				if ok, err := r.checkOvirtVersion(); !ok {
```

pkg/controller/plan/adapter/ovirt/validator.go
```diff
@@ -105,20 +108,79 @@ func (r *Validator) StorageMapped(vmRef ref.Ref) (ok bool, err error) {
...
+				}
+			}
+		}
+	}
+	ok = true
+	return
+}
+
+// Checks the version for ovirt direct LUN/FC
+func (r *Validator) checkOvirtVersion() (bool, error) {
```
ahadas, line 153:
```suggestion
func (r *Validator) canImportDirectDisksFromProvider() (bool, error) {
```

pkg/controller/plan/adapter/ovirt/validator.go
```diff
@@ -105,20 +108,79 @@ func (r *Validator) StorageMapped(vmRef ref.Ref) (ok bool, err error) {
...
-			return
 		}
 	}
 	ok = true
 	return
 }
 
+// Validates oVirt version in case we use direct LUN/FC storage
+func (r *Validator) DirectStorage(vmRef ref.Ref) (ok bool, err error) {
+	if r.plan.Referenced.Map.Storage == nil {
```
ahadas, line 125: when does it happen that `r.plan.Referenced.Map.Storage == nil`?

pkg/controller/provider/container/ovirt/collector.go
```diff
@@ -133,6 +133,21 @@ func (r *Collector) Test() (status int, err error) {
...
+	if err != nil {
+		return
+	}
+	version := strings.Split(system.Product.Version.FullVersion, ".")
+	major = version[0]
+	minor = version[1]
+	build = version[2]
+	if len(version) > 3 {
+		revision = strings.Split(version[3], "-")[0]
+	}
```
ahadas, line 147: revision should otherwise be set to `"0"`, no?

pkg/controller/plan/validation.go
```diff
@@ -342,6 +344,14 @@ func (r *Reconciler) validateVM(plan *api.Plan) error {
 		Message:  "VM has unmapped storage.",
 		Items:    []string{},
 	}
+	unsupportedStorage := libcnd.Condition{
+		Type:     VMStorageNotSupported,
+		Status:   True,
+		Reason:   NotValid,
+		Category: Critical,
+		Message:  "VM has unsupported storage. Direct LUN/FC in oVirt requires version 4.5.2.1",
```
ahadas, line 352:
```suggestion
		Message:  "VM has unsupported storage. Migration of Direct LUN/FC from oVirt is supported as from version 4.5.2.1",
```

pkg/controller/provider/container/ovirt/resource.go
```diff
@@ -13,10 +13,11 @@ type System struct {
...
 		Version struct {
-			Build    string `json:"build"`
-			Major    string `json:"major"`
-			Minor    string `json:"minor"`
-			Revision string `json:"revision"`
+			Build       string `json:"build"`
+			Major       string `json:"major"`
+			Minor       string `json:"minor"`
+			Revision    string `json:"revision"`
+			FullVersion string `json:"full_version"`
```
ahadas, line 20: do we use `FullVersion` anywhere?

pkg/lib/inventory/container/container.go
```diff
@@ -148,4 +148,6 @@ type Collector interface {
 	Test() (int, error)
 	// Reset
 	Reset()
+	// Get the system verion - only ovirt
```
ahadas, line 151:
```suggestion
	// Get the system verion - currently, ovirt-only
```

ahadas (commented):
nice
as we talked offline - it would have been better to add the version to the data in the inventory so that we can add a validation in the validation-service that is displayed when selecting the VMs, but that's good enough now that we declare that we only support RHV 4.4 SP1

pkg/controller/provider/container/ovirt/resource.go
liranr23: yes, in the version method above. unfortunately revision is usually set to `0` even so the full version differ.

pkg/controller/plan/adapter/ovirt/validator.go
ahadas: ok, that's even more misleading then since we don't validate the storagemap here...
so do we actually need to check it here?

pkg/lib/inventory/container/container.go
```diff
@@ -148,4 +148,6 @@ type Collector interface {
 	Test() (int, error)
 	// Reset
 	Reset()
+	// Get the system verion - currently, ovirt-only
```
ahadas, line 151:
```suggestion
	// Get the system version - currently, ovirt-only
```

pkg/controller/provider/container/ovirt/resource.go
ahadas: ah ok, that's the struct we read from - so do we need the other properties of the Version struct? it looks like we have too much stuff here

pkg/controller/plan/adapter/ocp/validator.go
ahadas: ok, we wrote below "current, oVirt-only" so let's think about it separately

pkg/controller/provider/container/ovirt/resource.go
liranr23: not really. but it doesn't harm. we query ovirt API anyway, it gets these details. i don't see a reason to drop them. although we can.

liranr23: because only full version is the way to determine the exact ovirt version.
e.g:
```
<version>
  <build>3</build>
  <full_version>4.5.3.9-1.el8ev</full_version>
  <major>4</major>
  <minor>5</minor>
  <revision>0</revision>
</version> 
```

i'm ok overall to drop it. since there is no real use of these values. but we query them anyway.

ahadas (approved):

PR #660, opened 2023-11-23
Fix creation of OVA server pod under enforced restricted namespaces

bkhizgiy:
When a new namespace is created in OpenShift, it is automatically labeled with `pod-security.kubernetes.io/audit`, which refers to a secondary, restricted security level. The situation diffrent in Kubernetes and for the default namespace, where such automatic labeling does not occur. When a namespace is marked as restricted, an additional label, `pod-security.kubernetes.io/enforce`, is added. This label represents the highest security level, where violations are not permitted. When deploying an OVA server pod, it is created with standard settings across all namespaces, except those marked only with the enforce option. In these cases, extra configuration is required to eliminate any potential security violations.

### DIFF
--- a/pkg/controller/provider/ova-setup.go
+++ b/pkg/controller/provider/ova-setup.go
@@ -12,14 +12,18 @@ import (
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 const (
-	ovaServer           = "ova-server"
-	ovaImageVar         = "OVA_PROVIDER_SERVER_IMAGE"
-	nfsVolumeNamePrefix = "nfs-volume"
-	mountPath           = "/ova"
-	pvSize              = "1Gi"
+	ovaServer              = "ova-server"
+	ovaImageVar            = "OVA_PROVIDER_SERVER_IMAGE"
+	nfsVolumeNamePrefix    = "nfs-volume"
+	mountPath              = "/ova"
+	pvSize                 = "1Gi"
+	auditRestrictedLabel   = "pod-security.kubernetes.io/audit"
+	enforceRestrictedLabel = "pod-security.kubernetes.io/enforce"
+	qemuGroup              = 107
 )
 
 func (r Reconciler) CreateOVAServerDeployment(provider *api.Provider, ctx context.Context) {
@@ -143,7 +147,7 @@ func (r *Reconciler) createServerDeployment(provider *api.Provider, ctx context.
 				ObjectMeta: metav1.ObjectMeta{
 					Labels: labels,
 				},
-				Spec: r.makeOvaProviderPodSpec(pvcName, provider.Name),
+				Spec: r.makeOvaProviderPodSpec(pvcName, provider.Name, provider.Namespace),
 			},
 		},
 	}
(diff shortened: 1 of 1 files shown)


pkg/controller/provider/ova-setup.go
```diff
@@ -185,47 +189,83 @@ func (r *Reconciler) createServerService(provider *api.Provider, ctx context.Con
...
 	}
+
+	podSpec := core.PodSpec{
+		Containers: []core.Container{container},
+		Volumes:    []core.Volume{volume},
+	}
+	return podSpec
+}
+
+func (r *Reconciler) checkAndUpdateNamespaces(namespaceName string) bool {
```
bennyz, line 251: suggestion:
```go
func (r *Reconciler) checkAndUpdateNamespaces(namespaceName string) bool {
	ns := core.Namespace{}
	err := r.Get(context.TODO(), client.ObjectKey{Name: namespaceName}, &ns)
	if err != nil {
		r.Log.Error(err, "Error getting namespace for restriction check")
		return false
	}

	commonLabel, commonExists := ns.Labels[enforceRestrictedLable]
	ocpLabel, ocpExists := ns.Labels[auditRestrictedLable]

	return commonExists && commonLabel == "restricted" && !(ocpExists && ocpLabel == "restricted")
}
```

pkg/controller/provider/ova-setup.go
```diff
@@ -12,14 +12,18 @@ import (
...
-	ovaImageVar         = "OVA_PROVIDER_SERVER_IMAGE"
-	nfsVolumeNamePrefix = "nfs-volume"
-	mountPath           = "/ova"
-	pvSize              = "1Gi"
+	ovaServer              = "ova-server"
+	ovaImageVar            = "OVA_PROVIDER_SERVER_IMAGE"
+	nfsVolumeNamePrefix    = "nfs-volume"
+	mountPath              = "/ova"
+	pvSize                 = "1Gi"
+	auditRestrictedLable   = "pod-security.kubernetes.io/audit"
```
liranr23, line 24:
```suggestion
	auditRestrictedLabel   = "pod-security.kubernetes.io/audit"
```

pkg/controller/provider/ova-setup.go
```diff
@@ -12,14 +12,18 @@ import (
...
+	enforceRestrictedLable = "pod-security.kubernetes.io/enforce"
```
liranr23, line 25:
```suggestion
	enforceRestrictedLabel = "pod-security.kubernetes.io/enforce"
```

pkg/controller/provider/ova-setup.go
```diff
@@ -185,47 +189,83 @@ func (r *Reconciler) createServerService(provider *api.Provider, ctx context.Con
...
+
+	restricted := r.checkAndUpdateNamespaces(providerNamespace)
+	if restricted {
+		seccompProfile := &core.SeccompProfile{
+			Type: core.SeccompProfileTypeRuntimeDefault,
+		}
+		container.SecurityContext.RunAsUser = &user
+		container.SecurityContext.RunAsNonRoot = &nonRoot
+		container.SecurityContext.SeccompProfile = seccompProfile
+	}
```
liranr23, line 233: i would do it before the container creation, setting it straight into the `commonSecurityContext`. Or, dropping `commonSecurityContext`, setting it in the container and then doing this part as it is.

pkg/controller/provider/ova-setup.go
liranr23, line 251: what does it updates? namespaces?
how about `isEnforcedRestrictionNamespame`?

pkg/controller/provider/ova-setup.go
```diff
@@ -185,47 +189,83 @@ func (r *Reconciler) createServerService(provider *api.Provider, ctx context.Con
...
+		Name: namespaceName,
+	}
+	err := r.Get(context.TODO(), key, &ns)
+	if err != nil {
+		r.Log.Error(err, "Error getting namespace for restriction check")
+		return false
+	}
+
+	//common label for restricted namesapce
+	if commonLabel, exists := ns.Labels[enforceRestrictedLable]; exists && commonLabel == "restricted" {
```
liranr23, line 263: `restricted` is here twice, you can pull it out as a const.

liranr23 (commented):
Did you check other pods we create(virt-v2v, populators)? In my opinion they work in the default namespace. I don't remember we ever checked the restriction level. Can the OVA provider work with the additional restriction here without checking the namespace?

pkg/controller/provider/ova-setup.go
bkhizgiy: yes my mistake I changed the name and forgot to push the change

bkhizgiy:
After testing it with virt-v2v pod its working without the extra configuration, but it seams that the difference relays on the fact the ova server runs as a deployment while the virt-v2v created directly as a pod. for the deployment in the highest security level  there is extra enforcement for the user that can run the pod while for the others its not required and determine automatically.

pkg/controller/provider/ova-setup.go
```diff
@@ -185,47 +189,75 @@ func (r *Reconciler) createServerService(provider *api.Provider, ctx context.Con
...
 	}
+
+	podSpec := core.PodSpec{
+		Containers: []core.Container{container},
+		Volumes:    []core.Volume{volume},
+	}
+	return podSpec
+}
+
+func (r *Reconciler) isEnforcedRestrictionNamespame(namespaceName string) bool {
```
liranr23, line 251:
```suggestion
func (r *Reconciler) isEnforcedRestrictionNamespace(namespaceName string) bool {
```
i had a typo in my comment

bkhizgiy: haha just noticed, uploaded new version:)

liranr23:
> After testing it with virt-v2v pod its working without the extra configuration, but it seams that the difference relays on the fact the ova server runs as a deployment while the virt-v2v created directly as a pod. for the deployment in the highest security level there is extra enforcement for the user that can run the pod while for the others its not required and determine automatically.

so, can we do it as restricted all the time on any namespace?

bkhizgiy:
> > After testing it with virt-v2v pod its working without the extra configuration, but it seams that the difference relays on the fact the ova server runs as a deployment while the virt-v2v created directly as a pod. for the deployment in the highest security level there is extra enforcement for the user that can run the pod while for the others its not required and determine automatically.
> 
> so, can we do it as restricted all the time on any namespace?

No, for the audit security level when specifying the user + not root the pod creation being blocked,  plus I've tried to specify what is the the permitted UID range as part of the controller SCC  but its allowed for use only when setting the security to `enforce` (so same behavior).

liranr23 (approved):

PR #665, opened 2023-11-27
Add a warning and support of TPM

liranr23:
When having a TPM device from the source(oVirt - based on having windows 2022 or windows 11 OS),
create it also for the destination VM.
The data is persist in oVirt. Therefore, if supported by kubevirt the TPM will be persistent TPM,
otherwise the TPM without persistent data.

**A warning will be shown as the data will be lost during the migration**

In order to have the TPM persistent it is required to have kubvirt
v1.0.0 and adding configuration manually as described in:
https://kubevirt.io/user-guide/virtual_machines/persistent_tpm_and_uefi_state/

It is also supported in CNV running on OCP 4.14, and you need to follow:
https://access.redhat.com/documentation/en-us/openshift_container_platform/4.14/html/virtualization/virtual-machines#virt-using-vtpm-devices

https://issues.redhat.com/browse/MTV-378

### DIFF
--- a/pkg/controller/plan/adapter/ovirt/builder.go
+++ b/pkg/controller/plan/adapter/ovirt/builder.go
@@ -287,6 +287,7 @@ func (r *Builder) VirtualMachine(vmRef ref.Ref, object *cnv.VirtualMachineSpec, 
 	r.mapMemory(vm, object)
 	r.mapClock(vm, object)
 	r.mapInput(object)
+	r.mapTpm(vm, object)
 	err = r.mapNetworks(vm, object)
 	if err != nil {
 		return
@@ -502,6 +503,13 @@ func (r *Builder) mapDisks(vm *model.Workload, persistentVolumeClaims []core.Per
 	}
 	object.Template.Spec.Volumes = kVolumes
 	object.Template.Spec.Domain.Devices.Disks = kDisks
+}
+
+func (r *Builder) mapTpm(vm *model.Workload, object *cnv.VirtualMachineSpec) {
+	if vm.OSType == "windows_2022" || vm.OSType == "windows_11" {
+		persistData := true
+		object.Template.Spec.Domain.Devices.TPM = &cnv.TPMDevice{Persistent: &persistData}
+	}
 }
 
 // Build tasks.
--- a/pkg/controller/plan/adapter/vsphere/builder.go
+++ b/pkg/controller/plan/adapter/vsphere/builder.go
@@ -462,6 +462,7 @@ func (r *Builder) VirtualMachine(vmRef ref.Ref, object *cnv.VirtualMachineSpec, 
 	r.mapMemory(vm, object)
 	r.mapClock(host, object)
 	r.mapInput(object)
+	r.mapTpm(vm, object)
 	err = r.mapNetworks(vm, object)
 	if err != nil {
 		return
@@ -637,6 +638,13 @@ func (r *Builder) mapDisks(vm *model.VM, persistentVolumeClaims []core.Persisten
 	}
 	object.Template.Spec.Volumes = kVolumes
 	object.Template.Spec.Domain.Devices.Disks = kDisks
+}
+
+func (r *Builder) mapTpm(vm *model.VM, object *cnv.VirtualMachineSpec) {
+	if vm.TpmEnabled {
+		persistData := true
+		object.Template.Spec.Domain.Devices.TPM = &cnv.TPMDevice{Persistent: &persistData}
+	}
 }
 
 // Build tasks.
--- a/validation/policies/io/konveyor/forklift/ovirt/tpm.rego
+++ b/validation/policies/io/konveyor/forklift/ovirt/tpm.rego
@@ -1,0 +1,16 @@
+package io.konveyor.forklift.ovirt
+
+default has_tpm_os = false
+
+has_tpm_os = true {
+    regex.match(`windows_2022|windows_11`, input.osType)
+}
+
+concerns[flag] {
+    has_tpm_os
+    flag := {
+        "category": "Warning",
+        "label": "VM configured with a TPM device",
+        "assessment": "The VM is detected with an operation system that must have a TPM device. TPM data is not transferred during the migration."
+    }
+}
--- a/validation/policies/io/konveyor/forklift/ovirt/tpm_test.rego
+++ b/validation/policies/io/konveyor/forklift/ovirt/tpm_test.rego
@@ -1,0 +1,25 @@
+package io.konveyor.forklift.ovirt
+ 
+test_without_tpm_enabled {
+    mock_vm := { "name": "test",
+                 "osType": "rhel_9x64"
+                }
+    results = concerns with input as mock_vm
+    count(results) == 0
+}
+
+test_with_tpm_enabled_w11 {
+    mock_vm := { "name": "test",
+                 "osType": "windows_11"
+                }
+    results = concerns with input as mock_vm
+    count(results) == 1
+}
+
+test_with_tpm_enabled_w2k22 {
+    mock_vm := { "name": "test",
+                 "osType": "windows_2022"
+                }
+    results = concerns with input as mock_vm
+    count(results) == 1
+}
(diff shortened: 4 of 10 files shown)


liranr23:
vsphere needs to be verified(mainly the inventory). Also the warning might change to use luks(future?).
I didn't find a real answer if the data is persistent on vsphere, my guts says yes, and based on [their docs](https://docs.vmware.com/en/VMware-vSphere/7.0/com.vmware.vsphere.security.doc/GUID-31F2B3D0-259A-4EFD-B675-F89BC27ACE55.html) i think it does.

pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -639,6 +640,13 @@ func (r *Builder) mapDisks(vm *model.VM, persistentVolumeClaims []core.Persisten
 	object.Template.Spec.Domain.Devices.Disks = kDisks
 }
 
+func (r *Builder) mapDevices(vm *model.VM, object *cnv.VirtualMachineSpec) {
```
ahadas, line 643: same comment about the name of the function

pkg/controller/plan/adapter/ovirt/builder.go
```diff
@@ -485,6 +486,13 @@ func (r *Builder) mapDisks(vm *model.Workload, persistentVolumeClaims []core.Per
 	object.Template.Spec.Domain.Devices.Disks = kDisks
 }
 
+func (r *Builder) mapDevices(vm *model.Workload, object *cnv.VirtualMachineSpec) {
```
ahadas, line 489: this name is too generic since other devices are handled by other function, how about:
```suggestion
func (r *Builder) mapTpm(vm *model.VM, object *cnv.VirtualMachineSpec) {
```

pkg/controller/provider/container/ovirt/model.go
```diff
@@ -837,7 +837,7 @@ func (r *VMAdapter) Event() []int {
 // List the collection.
 func (r *VMAdapter) List(ctx *Context) (itr fb.Iterator, err error) {
 	vmList := VMList{}
-	err = ctx.client.list("vms", &vmList, r.follow())
+	err = ctx.client.list("vms", &vmList, r.follow(), libweb.Param{Key: "all_content", Value: "true"})
```
ahadas, line 840: that's surprising, I guess you did this because you didn't see the tpm_enabled without it, right?
forklift gets snapshots and other things that are not returned by non-extended queries, did you check how we get them?

validation/policies/io/konveyor/forklift/ovirt/tpm.rego
```diff
@@ -0,0 +1,16 @@
...
+has_tpm_enabled = value {
+   value :=  input.tpmEnabled
+}
+
+concerns[flag] {
+    has_tpm_enabled
+    flag := {
+        "category": "Warning",
+        "label": "VM configured with TPM device",
+        "assessment": "The VM is configured with TPM device. TPM data will be loss during the migration."
```
ahadas, line 14:
```suggestion
        "assessment": "The VM is configured with a TPM device. TPM data is not transferred during the migration."
```
("will be loss" may sound like it's removed from the source, so let's be more explicit about what would happen)

validation/policies/io/konveyor/forklift/vmware/tpm_enabled.rego
```diff
@@ -0,0 +1,16 @@
...
+has_tpm_enabled = true {
+    input.tpmEnabled == true
+}
+
+concerns[flag] {
+    has_tpm_enabled
+    flag := {
+        "category": "Warning",
+        "label": "VM configured with TPM device",
+        "assessment": "The VM is configured with TPM device. TPM data will be loss during the migration."
```
ahadas, line 14: same comment as in the validation for oVirt

ahadas:
> Also the warning might change to use luks(future?).

the warning would still hold as we won't be able to transfer the data but yeah, we can then extend the warning by saying that luks can be specified as part of the plan when we do that

> I didn't find a real answer if the data is persistent on vsphere, my guts says yes, and based on [their docs](https://docs.vmware.com/en/VMware-vSphere/7.0/com.vmware.vsphere.security.doc/GUID-31F2B3D0-259A-4EFD-B675-F89BC27ACE55.html) i think it does.

they support windows 11 so it needs to be persistent

pkg/controller/provider/container/ovirt/model.go
liranr23: yeah we use follow params. `?follow=snapshots....`
however here we need all content as this is the only way to get it into the vm. in ovirt it queries the vm devices and then indicate it for us.

ahadas: yes, I'm concerned about the performance penalty by this change
we have validations for watchdog and balloon devices - you say that the tpm device behaves differently?

liranr23: for watchdog we use follow. [here](https://github.com/oVirt/ovirt-engine/blob/master/backend/manager/modules/restapi/jaxrs/src/main/java/org/ovirt/engine/api/restapi/resource/BackendVmResource.java#L640) we can see the list of devices we need the use of querying the vm devices and using all-content.

pkg/controller/plan/adapter/ovirt/builder.go
liranr23: I preferred to leave it as workload, to align with all the methods.

ahadas: sure, I don't remember how the type changed - I meant to rename the function

pkg/controller/provider/container/ovirt/model.go
ahadas: so how about adding the tpm with a follow? we could do something more clever then like querying tpm only for VMs with UEFI

ahadas: I meant in oVirt, we can still make change in RHV

liranr23: let's say we do (it means adding a follow link and a new backendTpmDevice +-), supposedly it will be an addition to the option of all_content, in order to not break back compatibility. but then, you will need a very late version of ovirt to support the follow part. i don't see it a reasonable solution.

ahadas: let's summarize here what we discussed offline:
adding an indication of TPM to oVirt/RHV is possible but complex
we can alternatively do that by the operating system - generate a warning only for Win 11 and 2022 that require TPM (it won't cover all cases, like Linux with TPM, but it should cover most of the cases with much lower effort)

ahadas:
@liranr23 I summarized our offline discussion about determining whether the VM has TPM in oVirt [here](https://github.com/kubev2v/forklift/pull/665#discussion_r1435819138)

validation/policies/io/konveyor/forklift/ovirt/tpm_test.rego
```diff
@@ -0,0 +1,17 @@
...
+    mock_vm := { "name": "test",
+                 "tpmEnabled": false
+                }
+    results = concerns with input as mock_vm
+    count(results) == 0
+}
+
+test_with_tpm_enabled {
+    mock_vm := { "name": "test",
+                 "tpmEnabled": true
```
ahadas, line 13: aren't we supposed to check it with certain operating system types?

liranr23: yup, my bad. i only changed the rule and not the test.

ahadas: @liranr23 and how come that the test passed then?..

ahadas: > @liranr23 and how come that the test passed then?..

what I meant to ask is how come that we didn't see a failure in the CI?

liranr23: our makefile executing `go test ./pkg/... ./cmd/... -coverprofile cover.out`
however for the validation we need something like `opa test`. we don't test our validation test units in the CI. we need to fix that.

ahadas (approved):

PR #670, opened 2023-12-05
OVA: Fix phases display for OVA provider creation

bkhizgiy:
While the OVA provider pod is being created, the pod status currently displays `connectionFailed` until the OVA pod is actually up and running, this gives the wrong impression that something went wrong during the OVA provider creation. This fix changes the phases for the OVA provider. Once the provider is added, the status will immediately show `staging`. If there is a failure during the OVA provider creation process, or if the pod is not running after 10 minutes (consider as a timeout), the status will change to `connectionFailed`.

### DIFF
--- a/pkg/controller/provider/controller.go
+++ b/pkg/controller/provider/controller.go
@@ -22,6 +22,7 @@ import (
 	"os"
 	"path/filepath"
 	"sync"
+	"time"
 
 	api "github.com/konveyor/forklift-controller/pkg/apis/forklift/v1beta1"
 	"github.com/konveyor/forklift-controller/pkg/controller/base"
@@ -52,7 +53,8 @@ import (
 
 const (
 	// Name.
-	Name = "provider"
+	Name       = "provider"
+	OvaTimeout = 10
 )
 
 // Package logger.
@@ -196,10 +198,45 @@ func (r Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (r
 			deployment)
 
 		// If the deployment does not exist
-		if k8serr.IsNotFound(err) {
-			r.CreateOVAServerDeployment(provider, ctx)
-		} else if err != nil {
+		if err != nil {
+			if k8serr.IsNotFound(err) {
+				r.CreateOVAServerDeployment(provider, ctx)
+				provider.Status.Phase = initializing
+				provider.Status.SetCondition(
+					libcnd.Condition{
+						Type:     initializing,
+						Status:   True,
+						Category: Required,
+						Message:  "The OVA server being inizialized.",
+					})
+				err = r.Status().Update(context.TODO(), provider.DeepCopy())
+				if err != nil {
+					return
+				}
+				result.RequeueAfter = 5 * time.Second
+				return
+			}
 			return
+		}
+
+		// The ova server pod is not running yet
+		if deployment.Status.AvailableReplicas == 0 {
+			if time.Since(provider.CreationTimestamp.Time).Minutes() <= OvaTimeout {
+				result.RequeueAfter = 5 * time.Second
+				return
+			} else {
+				// Timeout reached
+				provider.Status.Phase = ServerCreationFailed
+				provider.Status.SetCondition(
+					libcnd.Condition{
+						Type:     ServerCreationFailed,
+						Status:   True,
+						Category: Critical,
+						Message:  "The OVA provider server creation failed.",
+					})
+				err = r.Status().Update(context.TODO(), provider.DeepCopy())
+				return
+			}
 		}
 	}
 
--- a/pkg/controller/provider/controller.go
+++ b/pkg/controller/provider/controller.go
@@ -206,10 +206,10 @@ func (r Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (r
 					r.handleServerCreationFailure(provider, err)
 					return
 				}
-				provider.Status.Phase = Initializing
+				provider.Status.Phase = Staging
 				provider.Status.SetCondition(
 					libcnd.Condition{
-						Type:     Initializing,
+						Type:     Staging,
 						Status:   True,
 						Category: Required,
 						Message:  "The OVA server being inizialized.",
@@ -386,21 +386,6 @@ func (r *Reconciler) getSecret(provider *api.Provider) (*v1.Secret, error) {
 	}
 
 	return secret, nil
-}
-
-func (r *Reconciler) handleServerCreationFailure(provider *api.Provider, err error) {
-	provider.Status.Phase = ServerCreationFailed
-	msg := fmt.Sprint("The OVA provider server creation failed -", err)
-	provider.Status.SetCondition(
-		libcnd.Condition{
-			Type:     ServerCreationFailed,
-			Status:   True,
-			Category: Critical,
-			Message:  msg,
-		})
-	if updateErr := r.Status().Update(context.TODO(), provider.DeepCopy()); updateErr != nil {
-		log.Error(updateErr, "Failed to update provider status")
-	}
 }
 
 // Provider catalog.
--- a/pkg/controller/provider/controller.go
+++ b/pkg/controller/provider/controller.go
@@ -200,11 +200,15 @@ func (r Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (r
 		// If the deployment does not exist
 		if err != nil {
 			if k8serr.IsNotFound(err) {
-				r.CreateOVAServerDeployment(provider, ctx)
-				provider.Status.Phase = initializing
+				err = r.CreateOVAServerDeployment(provider, ctx)
+				if err != nil {
+					r.handleServerCreationFailure(provider, err)
+					return
+				}
+				provider.Status.Phase = Initializing
 				provider.Status.SetCondition(
 					libcnd.Condition{
-						Type:     initializing,
+						Type:     Initializing,
 						Status:   True,
 						Category: Required,
 						Message:  "The OVA server being inizialized.",
@@ -224,17 +228,9 @@ func (r Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (r
 			if time.Since(provider.CreationTimestamp.Time).Minutes() <= OvaTimeout {
 				result.RequeueAfter = 5 * time.Second
 				return
-			} else {
-				// Timeout reached
-				provider.Status.Phase = ServerCreationFailed
-				provider.Status.SetCondition(
-					libcnd.Condition{
-						Type:     ServerCreationFailed,
-						Status:   True,
-						Category: Critical,
-						Message:  "The OVA provider server creation failed.",
-					})
-				err = r.Status().Update(context.TODO(), provider.DeepCopy())
+			} else { // Timeout reached
+				err = fmt.Errorf("the server creation timed out. Please ensure that the NFS export is set correctly")
+				r.handleServerCreationFailure(provider, err)
 				return
 			}
 		}
@@ -389,6 +385,22 @@ func (r *Reconciler) getSecret(provider *api.Provider) (*v1.Secret, error) {
 	}
 
 	return secret, nil
+}
+
+func (r *Reconciler) handleServerCreationFailure(provider *api.Provider, errMsg error) {
+	provider.Status.Phase = ServerCreationFailed
+	msg := fmt.Sprint("The OVA provider server creation failed -", errMsg)
+	provider.Status.SetCondition(
+		libcnd.Condition{
+			Type:     ServerCreationFailed,
+			Status:   True,
+			Category: Critical,
+			Message:  msg,
+		})
+	err := r.Status().Update(context.TODO(), provider.DeepCopy())
+	if err != nil {
+		log.Error(err, "Failed to update provider status")
+	}
 }
 
 // Provider catalog.
--- a/pkg/controller/provider/controller.go
+++ b/pkg/controller/provider/controller.go
@@ -54,7 +54,7 @@ import (
 const (
 	// Name.
 	Name               = "provider"
-	OvaTimeout         = 10
+	OvaTimeout         = 10 * time.Minute
 	OvaReconcilerRetry = 5 * time.Second
 )
 
@@ -212,12 +212,9 @@ func (r Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (r
 						Type:     Staging,
 						Status:   True,
 						Category: Required,
-						Message:  "The OVA server being inizialized.",
+						Message:  "The OVA server is being inizialized.",
 					})
 				err = r.Status().Update(context.TODO(), provider.DeepCopy())
-				if err != nil {
-					return
-				}
 				result.RequeueAfter = OvaReconcilerRetry
 				return
 			}
@@ -226,11 +223,11 @@ func (r Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (r
 
 		// The ova server pod is not running yet
 		if deployment.Status.AvailableReplicas == 0 {
-			if time.Since(provider.CreationTimestamp.Time).Minutes() <= OvaTimeout {
+			if provider.CreationTimestamp.Add(OvaTimeout).After(time.Now()) {
 				result.RequeueAfter = OvaReconcilerRetry
 				return
 			} else { // Timeout reached
-				err = fmt.Errorf("the server creation timed out. Please ensure that the NFS export is set correctly")
+				err = fmt.Errorf("the OVA provider server creation timed out. Please ensure that the NFS export is set correctly")
 				r.handleServerCreationFailure(provider, err)
 				return
 			}
--- a/pkg/controller/provider/controller.go
+++ b/pkg/controller/provider/controller.go
@@ -53,8 +53,9 @@ import (
 
 const (
 	// Name.
-	Name       = "provider"
-	OvaTimeout = 10
+	Name               = "provider"
+	OvaTimeout         = 10
+	OvaReconcilerRetry = 5 * time.Second
 )
 
 // Package logger.
@@ -217,7 +218,7 @@ func (r Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (r
 				if err != nil {
 					return
 				}
-				result.RequeueAfter = 5 * time.Second
+				result.RequeueAfter = OvaReconcilerRetry
 				return
 			}
 			return
@@ -226,7 +227,7 @@ func (r Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (r
 		// The ova server pod is not running yet
 		if deployment.Status.AvailableReplicas == 0 {
 			if time.Since(provider.CreationTimestamp.Time).Minutes() <= OvaTimeout {
-				result.RequeueAfter = 5 * time.Second
+				result.RequeueAfter = OvaReconcilerRetry
 				return
 			} else { // Timeout reached
 				err = fmt.Errorf("the server creation timed out. Please ensure that the NFS export is set correctly")
@@ -387,9 +388,9 @@ func (r *Reconciler) getSecret(provider *api.Provider) (*v1.Secret, error) {
 	return secret, nil
 }
 
-func (r *Reconciler) handleServerCreationFailure(provider *api.Provider, errMsg error) {
+func (r *Reconciler) handleServerCreationFailure(provider *api.Provider, err error) {
 	provider.Status.Phase = ServerCreationFailed
-	msg := fmt.Sprint("The OVA provider server creation failed -", errMsg)
+	msg := fmt.Sprint("The OVA provider server creation failed -", err)
 	provider.Status.SetCondition(
 		libcnd.Condition{
 			Type:     ServerCreationFailed,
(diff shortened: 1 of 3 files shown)


pkg/controller/provider/controller.go
```diff
@@ -196,11 +198,46 @@ func (r Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (r
...
+				r.CreateOVAServerDeployment(provider, ctx)
+				provider.Status.Phase = initializing
+				provider.Status.SetCondition(
+					libcnd.Condition{
+						Type:     initializing,
+						Status:   True,
+						Category: Required,
+						Message:  "The OVA server being inizialized.",
+					})
+				err = r.Status().Update(context.TODO(), provider.DeepCopy())
```
bkhizgiy, line 212: The `.DeepCopy()` func was missing for copying the correct values.

pkg/controller/provider/validation.go
```diff
@@ -58,10 +58,12 @@ const (
...
-	ValidationFailed = "ValidationFailed"
-	ConnectionFailed = "ConnectionFailed"
-	Ready            = "Ready"
-	Staging          = "Staging"
+	ValidationFailed     = "ValidationFailed"
+	ConnectionFailed     = "ConnectionFailed"
+	Ready                = "Ready"
+	Staging              = "Staging"
+	initializing         = "initializing"
+	ServerCreationFailed = "ServerCreationFailed"
```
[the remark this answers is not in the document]
bkhizgiy: yes, I think the change is required, I did a test with existing phase just changed the name and saw the same behavior so seams that the new phases need to be added to the current existing ones.

pkg/controller/provider/controller.go
bkhizgiy: We talked about it in the sync, the status was applied and everything was reset after it, so I just added a comment that instead using simply `err = r.Status().Update(context.TODO(), provider)`  I had to add the DeepCopy() func to make sure e all of the fields are copying properly.

pkg/controller/provider/controller.go
```diff
@@ -196,11 +198,46 @@ func (r Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (r
...
+				}
+				result.RequeueAfter = 5 * time.Second
+				return
+			}
 			return
 		}
+
+		// The ova server pod is not running yet
+		if deployment.Status.AvailableReplicas == 0 {
+			if time.Since(provider.CreationTimestamp.Time).Minutes() <= OvaTimeout {
```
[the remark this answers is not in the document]
bkhizgiy: umm i don't think its matter if the time is measured from the provider creation or the deployment, if the deployment failing for any reason the pod check should cover it (its simply wont run). but really I didn't address the case the deployment isnt creating at all, I"ll add this check as well.

pkg/controller/provider/controller.go
liranr23: ah, yeah.
we hit this issue since we upgraded the controller-runtime once...
https://github.com/kubev2v/forklift/commit/71d5b9bb71600dcfc3917422b6a2a62434abc478

pkg/controller/provider/controller.go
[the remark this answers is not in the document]
bkhizgiy: so `CreateOVAServerDeployment` functoin already handling errors, so I simply returned it to the controller and added the error msg to the condition as well, I think it should cover all cases now.

Screenshot from 2023-12-06 13-44-14

ahadas:
> Currently while the OVA provider pod is being created the pod status despised `connectionFailed`...

@bkhizgiy  I've rephrased the description a bit - did you mean "displayed" here?

pkg/controller/provider/controller.go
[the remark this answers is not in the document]
ahadas: > so `CreateOVAServerDeployment` functoin already handling errors, so I simply returned it to the controller and added the error msg to the condition as well, I think it should cover all cases now.
> 
> Screenshot from 2023-12-06 13-44-14

that this particular issue can't happen now because we are supposed to create the provider only when it doesn't exist, no?

pkg/controller/provider/controller.go
```diff
@@ -196,11 +198,42 @@ func (r Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (r
...
+						Type:     Initializing,
+						Status:   True,
+						Category: Required,
+						Message:  "The OVA server being inizialized.",
+					})
+				err = r.Status().Update(context.TODO(), provider.DeepCopy())
+				if err != nil {
+					return
+				}
+				result.RequeueAfter = 5 * time.Second
```
ahadas, line 220: how about setting this as a const and use it here and below?

pkg/controller/provider/controller.go
```diff
@@ -354,6 +387,22 @@ func (r *Reconciler) getSecret(provider *api.Provider) (*v1.Secret, error) {
 	return secret, nil
 }
 
+func (r *Reconciler) handleServerCreationFailure(provider *api.Provider, errMsg error) {
+	provider.Status.Phase = ServerCreationFailed
```
ahadas, line 391: why not to use `ConnectionFailed`?

pkg/controller/provider/controller.go
```diff
@@ -354,6 +387,22 @@ func (r *Reconciler) getSecret(provider *api.Provider) (*v1.Secret, error) {
 	return secret, nil
 }
 
+func (r *Reconciler) handleServerCreationFailure(provider *api.Provider, errMsg error) {
```
ahadas, line 390:
```suggestion
func (r *Reconciler) handleServerCreationFailure(provider *api.Provider, err error) {
```
errMsg is a bit confusing here - we get an error, not a message

pkg/controller/provider/controller.go
```diff
@@ -354,6 +387,22 @@ func (r *Reconciler) getSecret(provider *api.Provider) (*v1.Secret, error) {
...
+	provider.Status.Phase = ServerCreationFailed
+	msg := fmt.Sprint("The OVA provider server creation failed -", errMsg)
+	provider.Status.SetCondition(
+		libcnd.Condition{
+			Type:     ServerCreationFailed,
+			Status:   True,
+			Category: Critical,
+			Message:  msg,
+		})
+	err := r.Status().Update(context.TODO(), provider.DeepCopy())
```
ahadas, line 400: how about:
```suggestion
	if updateErr := r.Status().Update(context.TODO(), provider.DeepCopy()); updateErr != nil {
```

pkg/controller/provider/controller.go
```diff
@@ -196,11 +198,42 @@ func (r Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (r
...
-			r.CreateOVAServerDeployment(provider, ctx)
-		} else if err != nil {
+		if err != nil {
+			if k8serr.IsNotFound(err) {
+				err = r.CreateOVAServerDeployment(provider, ctx)
+				if err != nil {
+					r.handleServerCreationFailure(provider, err)
+					return
+				}
+				provider.Status.Phase = Initializing
```
ahadas, line 208: can't we reuse the `Staging` phase and check `deployment.Status.AvailableReplicas` as part of the "test connection" phase?

pkg/controller/provider/controller.go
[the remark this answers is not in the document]
bkhizgiy: It can be since the server creation depends on several steps like the PV, PVC creation and then the deployment + service, in that case for any error in the creation we will change the status immediately to failed, without retries. if you referring to the specific PV issue, its shouldn't happen (it currently does because we have a bug which i'm currently working on - [the clean up issue](https://github.com/kubev2v/forklift/issues/672)) just added it as an example to show this case.

pkg/controller/provider/controller.go
bkhizgiy: because I think its a different scenario that `ConnectionFailed`, we are failing to create the server, the connection failed should indicate about issue between the connectivity to the created pod. we don't have this extra steps for any one of the other providers but IMO it make sense to add new phases.

pkg/controller/provider/controller.go
bkhizgiy: same here, staging should be a different phase for the creation of the inventory, here we first only setting up the sever and everything related to it before, so I think it should be a separate phase as well (also more easy for the debug in case any errors happened)

pkg/controller/provider/controller.go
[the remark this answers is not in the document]
ahadas: ok, if that can happen wouldn't it make more sense to say that "already exists" is not an issue? after all, we get the desired state we wish for

pkg/controller/provider/controller.go
ahadas: I agree that the OVA provider has additional steps compared to the other providers but the question is whether we have to reflect them to users? it's a matter of perspective - let's say that the server pod doesn't start, ok? as a user, I get the same result - fail to connect to the collection of VMs to import

the addition of the extra steps doesn't come without a cost - we add phases, we break the general flow with OVA-specific things (some we have to have, like the deployment of the server pod's deployment). if it's not very useful for users, we should minimize those provider-specific things

 for debugging we have other ways to expose data - with logs or by setting messages in conditions in a way that better reflect the underlying error

pkg/controller/provider/controller.go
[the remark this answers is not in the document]
bkhizgiy: I think the error msg is a bit confusing, in this specific case the creation of the server failed since I used a same name for the provider that was existed earlier and when I deleted it the PV wasn't removed (the bug I attached) so the next time I've tried to create provider with the same name it failed because the PV for it cant be created (the name already in use) so we got back this message (its the error that was returned that displayed as part of the condition) so its not indicating that the deployment already exist, simply that it failed so in this case it the desire outcome. and again this wont we happening once the second bug is fixed only used it as an example for this flow.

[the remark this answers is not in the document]
ahadas: @bkhizgiy ok, so in the case you describe, the problem is not with the name of the provider but with the name of the PV that already exists - imagine that someone would have created the PV with that name (without a provider, just a PV with the name we generate according to the provider name), we would have got the same error
so in that case, we should have said "the provider already exists - great, let's try to add the PV" - and repeat this until we are able to
the problem in our code is that, unlike k8s, we don't always generate unique names - in our case, we shouldn't set the name of the PV in the ObjectMeta but the `GenerateName` so the provided string would be the prefix of a unique name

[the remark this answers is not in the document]
ahadas: for identification we can/should set labels - and then it's easier to query stuff

pkg/controller/provider/validation.go
```diff
@@ -362,6 +360,21 @@ func (r *Reconciler) inventoryCreated(provider *api.Provider) error {
 	return nil
 }
 
+func (r *Reconciler) handleServerCreationFailure(provider *api.Provider, err error) {
+	provider.Status.Phase = ConnectionAuthFailed
+	msg := fmt.Sprint("The OVA provider server creation failed -", err)
```
bkhizgiy, line 365: Do we want to leave the error msg at disply or only log it?

pkg/controller/provider/ova-setup.go
```diff
@@ -262,3 +264,15 @@ func (r *Reconciler) isEnforcedRestrictionNamespace(namespaceName string) bool {
 
 	return enforceExists && enforceLabel == "restricted" && !(auditExists && auditLabel == "restricted")
 }
+
+func generateRandomSuffix() (randomSeed string) {
```
bkhizgiy, line 268: + should  add the random suffix for deployment and the service as well?

pkg/controller/provider/controller.go
```diff
@@ -52,7 +53,9 @@ import (
 
 const (
 	// Name.
-	Name = "provider"
+	Name               = "provider"
+	OvaTimeout         = 10
```
ahadas, line 57: it would be better to make this a duration as the const below in order to have units

pkg/controller/provider/controller.go
```diff
@@ -196,11 +199,42 @@ func (r Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (r
...
+					r.handleServerCreationFailure(provider, err)
+					return
+				}
+				provider.Status.Phase = Staging
+				provider.Status.SetCondition(
+					libcnd.Condition{
+						Type:     Staging,
+						Status:   True,
+						Category: Required,
+						Message:  "The OVA server being inizialized.",
```
ahadas, line 215:
```suggestion
						Message:  "The OVA server is being inizialized.",
```

pkg/controller/provider/controller.go
```diff
@@ -196,11 +199,42 @@ func (r Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (r
...
+					})
+				err = r.Status().Update(context.TODO(), provider.DeepCopy())
+				if err != nil {
+					return
```
ahadas, line 219: don't we need to set RequeueAfter here as well?

pkg/controller/provider/controller.go
```diff
@@ -196,11 +199,42 @@ func (r Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (r
...
 			return
 		}
+
+		// The ova server pod is not running yet
+		if deployment.Status.AvailableReplicas == 0 {
+			if time.Since(provider.CreationTimestamp.Time).Minutes() <= OvaTimeout {
+				result.RequeueAfter = OvaReconcilerRetry
+				return
+			} else { // Timeout reached
+				err = fmt.Errorf("the server creation timed out. Please ensure that the NFS export is set correctly")
```
ahadas, line 233:
```suggestion
				err = fmt.Errorf("the OVA provider server creation timed out. Please ensure that the NFS export is set correctly")
```

pkg/controller/provider/ova-setup.go
```diff
@@ -26,21 +28,21 @@ const (
...
+func (r Reconciler) CreateOVAServerDeployment(provider *api.Provider, ctx context.Context) (err error) {
 	ownerReference := metav1.OwnerReference{
 		APIVersion: "forklift.konveyor.io/v1beta1",
 		Kind:       "Provider",
 		Name:       provider.Name,
 		UID:        provider.UID,
 	}
-	pvName := fmt.Sprintf("%s-pv-%s-%s", ovaServer, provider.Name, provider.Namespace)
-	err := r.createPvForNfs(provider, ctx, ownerReference, pvName)
+	pvName := fmt.Sprintf("%s-pv-%s-%s-%s", ovaServer, provider.Name, provider.Namespace, generateRandomSuffix())
```
ahadas, line 38: why generating the name ourselves rather than letting k8s do that (by setting `GenerateName`)?

pkg/controller/provider/ova-setup.go
ahadas: nope

pkg/controller/provider/validation.go
ahadas: if it looks alright (the error message) then I don't see why not to show it

ahadas (commented):
much better, thanks @bkhizgiy, commented on few things inside

pkg/controller/provider/ova-setup.go
bkhizgiy: because we need it later on for the PVC creation, but I can add a query to get it value

ahadas: no need to query, we should get it the PV back with the content from the server

bkhizgiy: yeah, I accidentally returned `generatedName` instead `name` so the random suffix was missing, now working.

pkg/controller/provider/controller.go
ahadas: so we can drop lines 218-221 then ;)

pkg/controller/provider/ova-setup.go
```diff
@@ -26,9 +26,9 @@ const (
 	qemuGroup              = 107
 )
 
-func (r Reconciler) CreateOVAServerDeployment(provider *api.Provider, ctx context.Context) {
-	pvName := fmt.Sprintf("%s-pv-%s-%s", ovaServer, provider.Name, provider.Namespace)
-	err := r.createPvForNfs(provider, ctx, pvName)
+func (r Reconciler) CreateOVAServerDeployment(provider *api.Provider, ctx context.Context) (err error) {
+	pvNameTemplate := fmt.Sprintf("%s-pv-%s-%s", ovaServer, provider.Name, provider.Namespace)
```
ahadas, line 30:
```suggestion
	pvNamePrefix := fmt.Sprintf("%s-pv-%s-%s", ovaServer, provider.Name, provider.Namespace)
```

pkg/controller/provider/ova-setup.go
```diff
@@ -40,8 +40,9 @@ func (r Reconciler) CreateOVAServerDeployment(provider *api.Provider, ctx contex
 		Name:       provider.Name,
 		UID:        provider.UID,
 	}
-	pvcName := fmt.Sprintf("%s-pvc-%s", ovaServer, provider.Name)
-	err = r.createPvcForNfs(provider, ctx, ownerReference, pvName, pvcName)
+
+	pvcNameTemplate := fmt.Sprintf("%s-pvc-%s", ovaServer, provider.Name)
```
ahadas, line 44:
```suggestion
	pvcNamePrefix := fmt.Sprintf("%s-pvc-%s", ovaServer, provider.Name)
```

pkg/controller/provider/ova-setup.go
```diff
@@ -92,15 +94,15 @@ func (r *Reconciler) createPvForNfs(provider *api.Provider, ctx context.Context,
 	if err != nil {
 		return
 	}
-	return
+	return pv.GetObjectMeta().GetName(), nil
```
ahadas, line 97: can be replaced with `pv.Name` but how about returning the PV instead? it makes more sense for a function named `createPvForNfs` in my opinion

pkg/controller/provider/ova-setup.go
```diff
@@ -122,7 +124,7 @@ func (r *Reconciler) createPvcForNfs(provider *api.Provider, ctx context.Context
 	if err != nil {
 		return
 	}
-	return
+	return pvc.GetObjectMeta().GetName(), nil
```
ahadas, line 127: same comment as the one above

pkg/controller/provider/validation.go
ahadas: maybe just for consistency:
```suggestion
	msg := fmt.Sprint("The OVA provider server creation failed: ", err)
```

ahadas (commented):
minor comments, looks good overall

pkg/controller/provider/controller.go
```diff
@@ -196,11 +199,43 @@ func (r Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (r
...
+				}
+				result.RequeueAfter = OvaReconcilerRetry
+				return
+			}
 			return
 		}
+
+		// The ova server pod is not running yet
+		if deployment.Status.AvailableReplicas == 0 {
+			if time.Since(provider.CreationTimestamp.Time) <= OvaTimeout {
```
ahadas, line 230: actually, does it work? I think we need something like `provider.CreationTimestamp.Add(OvaTimeout).Before(time.now())`

bkhizgiy: I've tested it and it worked in the current implementation, but I can change if its not clear enough.

ahadas: ah ok, I see why - Duration is defined as `type Duration int64` so we get int64 on both sides
I prefer the fluent API I wrote above (didn't test that it even compiles though) but for your consideration

pkg/controller/provider/validation.go
```diff
@@ -360,6 +360,21 @@ func (r *Reconciler) inventoryCreated(provider *api.Provider) error {
 	return nil
 }
 
+func (r *Reconciler) handleServerCreationFailure(provider *api.Provider, err error) {
+	provider.Status.Phase = ConnectionAuthFailed
+	msg := fmt.Sprint("The OVA provider server creation failed:", err)
```
ahadas, line 365: need to add space at the end of the first string:
```suggestion
	msg := fmt.Sprint("The OVA provider server creation failed: ", err)
```

pkg/controller/provider/controller.go
bkhizgiy: changed to your suggestion + tested.

ahadas (approved):
@bkhizgiy this PR changed quite a bit, please update its title and description

pkg/controller/provider/validation.go
```diff
@@ -362,10 +362,10 @@ func (r *Reconciler) inventoryCreated(provider *api.Provider) error {
 
 func (r *Reconciler) handleServerCreationFailure(provider *api.Provider, err error) {
 	provider.Status.Phase = ConnectionAuthFailed
-	msg := fmt.Sprint("The OVA provider server creation failed -", err)
+	msg := fmt.Sprint("The OVA provider server creation failed: ", err)
 	provider.Status.SetCondition(
 		libcnd.Condition{
-			Type:     ConnectionAuthFailed,
+			Type:     ConnectionFailed,
```
bkhizgiy, line 368: I've changed the type, it should be `ConnectionFailed` and not `ConnectionAuthFailed`
and included the changes you added here as well so they are not override.

liranr23: i don't get this comment.
do we have auth problems? only connection problems?
if we have only `ConnectionFailed` do we still need `ConnectionAuthFailed` few lines above?

which changes were added to not be overridden?

ahadas: > I've changed the type, it should be `ConnectionFailed` and not `ConnectionAuthFailed` 

right, ack

> and included the changes you added here as well so they are not override.

cool, but you dropped changes in the createPvForNfs and createPvcForNfs functions, intentionally?

bkhizgiy: @ahadas no I didn't drop them, because the return now is same as it was in the function originally it doesn't catch it as a change anymore so its not displayed.

@liranr23 you right the above statues need to be also changed.

PR #676, opened 2023-12-07
Support LUKS keys for cold migrations from vSphere

### LINKED ISSUE #567: Migrate VMs with LUKS-encrypted disks from vSphere
Add support for specifying the `--key` flag for `virt-v2v` as per the [virt-v2v docs](https://www.libguestfs.org/virt-v2v.1.html):
image

Possibly via an additional option in the `StorageMap` CRD?

liranr23:
Add LUKS key to the vm
This patch adds the LUKS keys to the VM specs using a secret ref. It
expected to have list of strings. Each value will use the `all` selector
option based on virt-v2v docs
(https://www.libguestfs.org/virt-v2v.1.html).

The secret should be provided in the destination namespace, the key
should be `luks` and the values should be the arguments provided to
virt-v2v, such as:

```
passphrase1
passphrase2
...
```

Fixes #567


pkg/apis/forklift/v1beta1/plan.go
```diff
@@ -44,6 +44,11 @@ type PlanSpec struct {
 	TransferNetwork *core.ObjectReference `json:"transferNetwork,omitempty"`
 	// Whether this plan should be archived.
 	Archived bool `json:"archived,omitempty"`
+	// LUKs keys for virt-v2v (https://www.libguestfs.org/virt-v2v.1.html)
+	// References a secret containing credentials and
+	// other confidential information.
+	// +optional
+	Secret *core.ObjectReference `json:"secret" ref:"Secret"`
```
liranr23, line 51: looks like i need to drop the pointer

ahadas:
> This patch adds the LUKS keys to the plan specs using a secret ref. It expected to have list of strings. Each value of the string should contain the ID or name of the device, and the selector as pointed out in virt-v2v docs (https://www.libguestfs.org/virt-v2v.1.html).
> 
> **The secret should be provided in the destination namespace, the key should be `luks` and the values should be the arguments provided to virt-v2v, such as:**
> 
> ```
> UUID:key:passphrase
> UUID:clevis
> DEVICE_NAME:key:passphrase
> ```
> 
> * Once https://issues.redhat.com/browse/RHEL-18142 will be implemented,
>   it will be possible to use `all:key:passphrase` and ease on the user.

I see that it's already in progress, doesn't it make more sense to wait for it?

virt-v2v/cold/entrypoint
```diff
@@ -71,6 +71,14 @@ for disk in /dev/block[0-9]* ; do
 	ln -s "$disk" "$DIR/$V2V_vmName-sd$(gen_name "$((num+1))")"
 done
 
+# Adds LUKS keys, if existing.
```
ahadas, line 74:
```suggestion
# Adds LUKS keys, if exist.
```

virt-v2v/cold/entrypoint
```diff
@@ -71,6 +71,14 @@ for disk in /dev/block[0-9]* ; do
...
+if [ -d "/etc/luks" ]; then
+    IFS=$'\n'
+    for line in $(cat /etc/luks/luks); do
+        args+=(--key "$line")
```
ahadas, line 78: as I commented at the top-level, I think we can wait to get the --all option because otherwise, users would have to specify the name or the UIID of the device that may be different for different VMs in the plan (we can find the device UID probably, not sure about OVAs though, but as the --all option is just around the corner...)

pkg/apis/forklift/v1beta1/plan.go
```diff
@@ -44,6 +44,11 @@ type PlanSpec struct {
 	TransferNetwork *core.ObjectReference `json:"transferNetwork,omitempty"`
 	// Whether this plan should be archived.
 	Archived bool `json:"archived,omitempty"`
+	// LUKs keys for virt-v2v (https://www.libguestfs.org/virt-v2v.1.html)
+	// References a secret containing credentials and
+	// other confidential information.
+	// +optional
+	Secret core.ObjectReference `json:"secret" ref:"Secret"`
```
ahadas, line 51: `Secret` is a very generic name :) how about LUKS or EncryptionKeys?

ahadas (commented):
nice, looks good overall

ahadas:
> > **The secret should be provided in the destination namespace, the key should be `luks` and the values should be the arguments provided to virt-v2v, such as:**
> > ```
> > UUID:key:passphrase
> > UUID:clevis
> > DEVICE_NAME:key:passphrase
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * Once https://issues.redhat.com/browse/RHEL-18142 will be implemented,
> >   it will be possible to use `all:key:passphrase` and ease on the user.
> 
> I see that it's already in progress, doesn't it make more sense to wait for it?

BTW, need to check if the support for LUKS exists in el8 but if we rely on the --all then we anyway need to limit this to the flows we use virt-v2v on el9 (hopefully, all cold migration flows in 2.6)

liranr23:
> > **The secret should be provided in the destination namespace, the key should be `luks` and the values should be the arguments provided to virt-v2v, such as:**
> > ```
> > UUID:key:passphrase
> > UUID:clevis
> > DEVICE_NAME:key:passphrase
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * Once https://issues.redhat.com/browse/RHEL-18142 will be implemented,
> >   it will be possible to use `all:key:passphrase` and ease on the user.
> 
> I see that it's already in progress, doesn't it make more sense to wait for it?

yeah we can.
in this PR we give the chance for the user to choose (from all options).
there is a clear performance benefit if the user specify the device. i thought it might be better to give him the option. once we consume the change of virt-v2v, he can set `all:key:passphrase`. currently, i am not sure if we will have fallback to `clevis`, that means, we can either "force" it on the virt-v2v command and the user won't need to specify it at all or, adding `all:clevis` as well. 
Bottom line, it's "all options[and possible better performance]" versus easy usage of the user (specifying list of keys and that's all).

liranr23:
> Member

it's not as written in the 3rd commit

pkg/apis/forklift/v1beta1/plan.go
liranr23: not sure if you meant only for the name we use in our code or also on the CRD. i changed only the name we use in the code.

ahadas: isn't it generated (the CRD)?

liranr23: yes, but i changed only `Secret`. to change the CRD I need:
```LUKS core.ObjectReference `json:"luks" ref:"Secret"```

**Comment by ahadas** [comment]
> > > **The secret should be provided in the destination namespace, the key should be `luks` and the values should be the arguments provided to virt-v2v, such as:**
> > > ```
> > > UUID:key:passphrase
> > > UUID:clevis
> > > DEVICE_NAME:key:passphrase
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > 
> > > * Once https://issues.redhat.com/browse/RHEL-18142 will be implemented,
> > >   it will be possible to use `all:key:passphrase` and ease on the user.
> > 
> > 
> > I see that it's already in progress, doesn't it make more sense to wait for it?
> 
> yeah we can. in this PR we give the chance for the user to choose (from all options). there is a clear performance benefit if the user specify the device. i thought it might be better to give him the option. once we consume the change of virt-v2v, he can set `all:key:passphrase`. currently, i am not sure if we will have fallback to `clevis`, that means, we can either "force" it on the virt-v2v command and the user won't need to specify it at all or, adding `all:clevis` as well. Bottom line, it's "all options[and possible better performance]" versus easy usage of the user (specifying list of keys and that's all).

if we like to consider the option of specifying devices, we need to see how to make it practical, especially in scale - specifying UUIDs/names is problematic, users would need to select disk(s) and a practical way...

**Comment by ahadas** [comment]
> > BTW, need to check if the support for LUKS exists in el8 but if we rely on the all option then we anyway need to limit this to the flows we use virt-v2v on el9 (hopefully, all cold migration flows in 2.6)
> 
> it's not as written in the 3rd commit

ok, so need to add it to [the plan admitter](https://github.com/kubev2v/forklift/blob/main/pkg/forklift-api/webhooks/validating-webhook/admitters/plan-admitter.go)

**Comment by liranr23** [comment]
> > > > **The secret should be provided in the destination namespace, the key should be `luks` and the values should be the arguments provided to virt-v2v, such as:**
> > > > ```
> > > > UUID:key:passphrase
> > > > UUID:clevis
> > > > DEVICE_NAME:key:passphrase
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > 
> > > > * Once https://issues.redhat.com/browse/RHEL-18142 will be implemented,
> > > >   it will be possible to use `all:key:passphrase` and ease on the user.
> > > 
> > > 
> > > I see that it's already in progress, doesn't it make more sense to wait for it?
> > 
> > 
> > yeah we can. in this PR we give the chance for the user to choose (from all options). there is a clear performance benefit if the user specify the device. i thought it might be better to give him the option. once we consume the change of virt-v2v, he can set `all:key:passphrase`. currently, i am not sure if we will have fallback to `clevis`, that means, we can either "force" it on the virt-v2v command and the user won't need to specify it at all or, adding `all:clevis` as well. Bottom line, it's "all options[and possible better performance]" versus easy usage of the user (specifying list of keys and that's all).
> 
> if we like to consider the option of specifying devices, we need to see how to make it practical, especially in scale - specifying UUIDs/names is problematic, users would rather need to select disk(s) and a practical way...

OK, so we can wait for virt-v2v and change the commits + in the virt-v2v to handle it by adding `all:key:` prefix. Then the user will just provide keys in the secret.
We can also add `--key all:clevis` to support it as a fallback (in case virt-v2v won't do it).
This should be the most simple way to a user.

File: pkg/forklift-api/webhooks/validating-webhook/admitters/plan-admitter.go [hunk 5]
```diff
@@ -116,6 +116,34 @@ func (admitter *PlanAdmitter) validateWarmMigrations() error {
 	return nil
 }
 
+func (admitter *PlanAdmitter) validateLuks() error {
```
ahadas, line 119:
```suggestion
func (admitter *PlanAdmitter) validateLUKS() error {
```

pkg/forklift-api/webhooks/validating-webhook/admitters/plan-admitter.go
```diff
@@ -116,6 +116,34 @@ func (admitter *PlanAdmitter) validateWarmMigrations() error {
...
+	providerType := admitter.sourceProvider.Type()
+	if providerType != api.VSphere && providerType != api.Ova {
+		log.Info("Provider type (non-VSphere & non-OVA) does not support LUKS, passing")
```
ahadas, line 122: I'd expect to see a different logic here: if LUKS is specified and the provider type is not vSphere or OVA, then return an error. you probably took validateVDDK as a reference - but here, it's less obvious that LUKS are not being used when migrating from other sources

pkg/forklift-api/webhooks/validating-webhook/admitters/plan-admitter.go
```diff
@@ -116,6 +116,34 @@ func (admitter *PlanAdmitter) validateWarmMigrations() error {
...
+	}
+	if el9 {
+		// LUKS is optional when EL9 virt-v2v image is in use
+		log.Info("LUKS secret is optional when EL9 virt-v2v image is in use, passing")
+		return nil
+	}
+
+	luksRef := &admitter.plan.Spec.LUKS
+	if luksRef != nil {
+		err := liberr.New("LUKS encryption is forbidden for this type of migration")
```
ahadas, line 139:
```suggestion
		err := liberr.New("migration of encrypted disks is not supported for this migration")
```

ahadas:
the `--key all:<selector>` for LUKS [was added in virt-v2v-2.3.7-1.el9](https://issues.redhat.com/browse/RHEL-18142) which is not yet available on centos9-stream

pkg/apis/forklift/v1beta1/plan.go
```diff
@@ -44,6 +44,11 @@ type PlanSpec struct {
 	TransferNetwork *core.ObjectReference `json:"transferNetwork,omitempty"`
 	// Whether this plan should be archived.
 	Archived bool `json:"archived,omitempty"`
+	// LUKs keys for virt-v2v (https://www.libguestfs.org/virt-v2v.1.html)
```
ahadas, line 47:
```suggestion
	// LUKS keys for virt-v2v (https://www.libguestfs.org/virt-v2v.1.html)
```

pkg/controller/plan/kubevirt.go
```diff
@@ -1476,6 +1477,22 @@ func (r *KubeVirt) podVolumeMounts(vmVolumes []cnv.Volume, configMap *core.Confi
 			EmptyDir: &core.EmptyDirVolumeSource{},
 		},
 	})
+	if libref.RefSet(&r.Plan.Spec.LUKS) {
```
ahadas, line 1480: why not adding that to the pod's environment variables?

virt-v2v/cold/entrypoint
ahadas: since a selector of all devices will probably not be available for forklift 2.6, we decided not to wait for it
but I talked with @dankenigsberg, raised a vaild point that users that provide such input may expect us to support key per-devices also when that selector would be available.
@liranr23 can we save the user from specifying the device IDs, as we're supposed to have them in the inventory, and ask only for the encryption keys from users?

pkg/apis/forklift/v1beta1/plan.go
```diff
@@ -44,6 +44,11 @@ type PlanSpec struct {
...
+	// References a secret containing credentials and
+	// other confidential information.
+	// +optional
+	LUKS core.ObjectReference `json:"luks" ref:"Secret"`
```
ahadas, line 51: based on the discussion on #873, we better be consistent here and place this per-VM rather than at the plan level
1. to be aligned with other per-VM configuration - hooks, primary disk
2. it's more flexible, allowing users to specify different keys for different VMs that are migrated to the same namespace without requiring different plans (while still specifying the keys at the UI-level per-plan to ease scale migrations)

liranr23: As I wrote in the related PR, I think LUKs make more sense as a plan, since we have the `all` option and to have the "mass migration" idea of MTV.
Your points are also valid.

ahadas: > since we have the `all` option

the `all` argument just means that the encryption keys would apply to all devices
by setting the encryption keys per VMs, we would still be able to say "for VM A, use encryption keys X, while for VM B, use encryption keys Y" (in other words, we get a better granularity)

> and to have the "mass migration" idea of MTV

it wouldn't conflict with the mindset of mass migration if the UI would provide similar functionality to that of hooks (no one is going to write yaml for mass migration, it would likely be generated by a script and then it doesn't really matter whether to place it in one place or per-VM)

@liranr23 in my opinion, we should think about it differently - the functional benefits of placing it per-VM are clear, do you see any functional benefit of placing it at the plan level?

liranr23: no, i was in mind it's theoretically fits better with the overall idea of the migrations.  
let's go with per-vm, seems it's more beneficial and aligns with other properties.

virt-v2v/cold/entrypoint.go
```diff
@@ -127,6 +128,19 @@ func buildCommand() []string {
 				virtV2vArgs = append(virtV2vArgs, "--mac", macToIp)
 			}
 		}
+		// Adds LUKS keys, if exist.
+		if file, err := os.Open("/etc/luks/luks"); err == nil {
```
bennyz, line 132: maybe I'm misreading, but it looks like we would ignore the error, no? I suppose we want to log it at least if it's not related to the file being missing

pkg/forklift-api/webhooks/validating-webhook/admitters/plan-admitter.go
```diff
@@ -116,6 +116,40 @@ func (admitter *PlanAdmitter) validateWarmMigrations() error {
 	return nil
 }
 
+func (admitter *PlanAdmitter) validateLUKS() error {
+	hasLUKS := false
+	for _, vm := range admitter.plan.Spec.VMs {
+		if vm.LUKS.Name != "" {
```
ahadas, line 122: why not to use `RefSet` here as well?

pkg/forklift-api/webhooks/validating-webhook/admitters/plan-admitter.go
```diff
@@ -116,6 +116,40 @@ func (admitter *PlanAdmitter) validateWarmMigrations() error {
...
+			hasLUKS = true
+		}
+	}
+	if !hasLUKS {
+		return nil
+	}
+
+	providerType := admitter.sourceProvider.Type()
+	if providerType != api.VSphere && providerType != api.Ova {
+		err := liberr.New("migration of encrypted disks is not supported for this migration")
```
ahadas, line 132:
```suggestion
		err := liberr.New("migration of encrypted disks from source provider of type <add provider type> is not supported ")
```

pkg/forklift-api/webhooks/validating-webhook/admitters/plan-admitter.go
```diff
@@ -116,6 +116,40 @@ func (admitter *PlanAdmitter) validateWarmMigrations() error {
...
+	}
+
+	el9, el9Err := admitter.plan.VSphereUsesEl9VirtV2v()
+	if el9Err != nil {
+		log.Error(el9Err, "Could not analyze plan, failing")
+		return el9Err
+	}
+	if el9 {
+		// LUKS is optional when EL9 virt-v2v image is in use
+		log.Info("LUKS secret is optional when EL9 virt-v2v image is in use, passing")
```
ahadas, line 144: why do we need this log?

pkg/forklift-api/webhooks/validating-webhook/admitters/plan-admitter.go
```diff
@@ -116,6 +116,40 @@ func (admitter *PlanAdmitter) validateWarmMigrations() error {
...
+	} else {
+		err := liberr.New("migration of encrypted disks is not supported for this migration")
```
ahadas, line 146: need to change the message here with a more informative one

virt-v2v/cold/entrypoint.go
```diff
@@ -127,6 +128,21 @@ func buildCommand() []string {
...
+			scanner := bufio.NewScanner(file)
+			for scanner.Scan() {
+				virtV2vArgs = append(virtV2vArgs, "--key", fmt.Sprintf("all:key:%s", scanner.Text()))
+			}
+			if err := scanner.Err(); err != nil {
+				fmt.Println("Error scanning the LUKS keys file")
+				panic(fmt.Errorf("error %w", err))
+			}
+		} else {
+			fmt.Println("Error opening the LUKS keys file ", err)
```
ahadas, line 144: but what if the file doesn't exist because LUKS keys were not specified?
I know I'm jumping between these two options, but could be simpler to set the data from the secret as environment variables..

pkg/apis/forklift/v1beta1/plan/vm.go
```diff
@@ -30,6 +30,11 @@ type VM struct {
 	ref.Ref `json:",inline"`
 	// Enable hooks.
 	Hooks []HookRef `json:"hooks,omitempty"`
+	// LUKS keys for virt-v2v (https://www.libguestfs.org/virt-v2v.1.html)
+	// References a secret containing credentials and
```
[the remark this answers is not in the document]
liranr23: right

pkg/forklift-api/webhooks/validating-webhook/admitters/plan-admitter.go
liranr23: please revisit

pkg/apis/forklift/v1beta1/plan/vm.go
```diff
@@ -30,6 +30,10 @@ type VM struct {
 	ref.Ref `json:",inline"`
 	// Enable hooks.
 	Hooks []HookRef `json:"hooks,omitempty"`
+	// LUKS keys for virt-v2v
+	// References a secret containing disks decryption keys
```
ahadas, line 34: multi-line is not rendered nicely in the yaml and we see from the definition below that it references a Secret so how about replacing the two lines above with:
```suggestion
	// Disk decryption LUKS keys
```
?

pkg/forklift-api/webhooks/validating-webhook/admitters/plan-admitter.go
```diff
@@ -116,6 +117,38 @@ func (admitter *PlanAdmitter) validateWarmMigrations() error {
...
+		return err
+	}
+
+	el9, el9Err := admitter.plan.VSphereUsesEl9VirtV2v()
+	if el9Err != nil {
+		log.Error(el9Err, "Could not analyze plan, failing")
+		return el9Err
+	}
+	if !el9 {
+		err := liberr.New("migration of encrypted disks is not supported warm migrations")
```
ahadas, line 145:
```suggestion
		err := liberr.New("migration of encrypted disks is not supported for warm migrations or migrations to remote providers")
```

virt-v2v/cold/entrypoint.go
ahadas: > we show this message in this case. if the vm disk is encrypted - the v2v pod will fail with the relevant error. otherwise it will continue as usual.

ok, I meant when LUKS keys are not specified and the VM doesn't have encrypted disks (which is the common case) - we shouldn't display this error then..

liranr23: ok, in that case either we don't show anything.
i see two options:
- if the disk with encryption, we will fail later. troubleshooting would be checking the command can tell if we had keys and it continues there..
- switching to environment var as you suggested

liranr23: i went with env var at the end

pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -226,6 +227,20 @@ func (r *Builder) PodEnvironment(vmRef ref.Ref, sourceSecret *core.Secret) (env
 			Value: macsToIps,
 		})
 	}
+
+	if libref.RefSet(&vm.LUKS) {
```
liranr23, line 231: `libref.RefSet` checks both name and namespace. when adding envVar from secret we can set only the name.
@ahadas do you think we need to change this to check only the name isn't empty(both here an API)?

ahadas (commented):
to summarize an offline discussion about this - it would be better to switch back to passing the keys with a volume and ask virt-v2v to get it from the file, so that the keys won't be included in logs that are gathered from the cluster

virt-v2v/cold/entrypoint.go
```diff
@@ -127,6 +128,25 @@ func buildCommand() []string {
...
+			if err != nil {
+				fmt.Println("Error reading files in LUKS directory ", err)
+				os.Exit(1)
+			}
+			for _, file := range files {
+				virtV2vArgs = append(virtV2vArgs, "--key", fmt.Sprintf("all:file:%s", file))
+			}
+		}
+		// fallback to clevis, harmless
+		virtV2vArgs = append(virtV2vArgs, "--key", "all:clevis")
```
liranr23, line 149: i put it anyway, but we can say we set it only when a secret exists (even without relevant passphrase). i think this way it is now is good enough - being ignored from virt-v2v if no LUKS needed.

pkg/controller/plan/kubevirt.go
```diff
@@ -1942,6 +1961,12 @@ func (r *KubeVirt) secretDataSetterForCDI(vmRef ref.Ref) func(*core.Secret) erro
 	}
 }
 
+func (r *KubeVirt) isSecretExist(name, namespace string) (err error) {
```
ahadas, line 1964: this function should either return a boolean that indicates whether the secret exists or renamed

pkg/controller/plan/kubevirt.go
```diff
@@ -1760,7 +1759,27 @@ func (r *KubeVirt) podVolumeMounts(vmVolumes []cnv.Volume, configMap *core.Confi
 			EmptyDir: &core.EmptyDirVolumeSource{},
 		},
 	})
-
+	if vm.LUKS.Name != "" {
+		err = r.isSecretExist(vm.LUKS.Name, r.Plan.Spec.TargetNamespace)
```
ahadas, line 1763: this assumes the secret with the keys exists on the target namespace but how is it supposed to get there? there should be a step in the migration that ensures the secret exists on the target namespace..

virt-v2v/cold/entrypoint.go
ahadas: are you sure there is no performance implication for specifying this? it doesn't seem right to specify this without checking whether clevis is defined, unless we're sure about the implications

liranr23: i think if there is no encryption, nothing will happen. if there is, my guess is that it will first try provided keys as they precedent args, if non of them fit, it will try clevis. sure it can harm performance. what i wrote is just theory how it works, we can ask richard or remove this part completely

ahadas: I'd just drop it - it's counter intuitive to specify that when the user doesn't provide any indication for having encrypted disks

virt-v2v/cold/entrypoint.go
```diff
@@ -127,6 +128,25 @@ func buildCommand() []string {
 				virtV2vArgs = append(virtV2vArgs, "--mac", macToIp)
 			}
 		}
+		// Adds LUKS keys, if exist.
+		if _, err := os.Stat(LUKSDIR); os.IsNotExist(err) {
+			// do nothing
```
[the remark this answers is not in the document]
liranr23: that depends on when we wish to have a real error/message.
if the directory isn't there, it might mean the user don't user LUKS. so why should we error/log?

pkg/controller/plan/kubevirt.go
liranr23: it should be created by the user/UI. we don't know the passphrases

virt-v2v/cold/entrypoint.go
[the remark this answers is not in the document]
ahadas: > if the directory isn't there, it might mean the user don't user LUKS. so why should we error/log?

I don't have an answer for this - I didn't think of raising an error or writing a log here

what I wrote above is to handle only the case of `!os.IsNotExist(err)` (in other words - not doing anthing when `os.IsNotExist(err)`)

pkg/controller/plan/kubevirt.go
ahadas: it's a problem to expect the client to interact with the destination provider - we shouldn't assume things are set on the target namespace by other entities

the way I see it is as follow:
1. the user specifies keys
2. Forklift creates a secret along with the plan (on the same namespace)
3. during the migration, Forklift creates a secret on the target namespace (this can happen right after Forklift creates the target namespace)

and then when we get here, we'll find the secret on the target namespace

virt-v2v/cold/entrypoint.go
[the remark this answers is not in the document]
liranr23: i think the better way is as it is now (in terms of error/log). but in case you have strong opinion to change it, i will.

pkg/controller/plan/kubevirt.go
```diff
@@ -1942,14 +1968,26 @@ func (r *KubeVirt) secretDataSetterForCDI(vmRef ref.Ref) func(*core.Secret) erro
 	}
 }
 
+func (r *KubeVirt) secretLUKS(name, namespace string) func(*core.Secret) error {
+	return func(secret *core.Secret) error {
+		sourceSecret := &core.Secret{}
+		err := r.Destination.Client.Get(context.TODO(), client.ObjectKey{Name: name, Namespace: namespace}, sourceSecret)
```
ahadas, line 1974: @liranr23 shouldn't we use the client of the source provider here?

ahadas (commented):
I've made some cleanup, besides the comment inside it looks good

pkg/controller/plan/kubevirt.go
```diff
@@ -523,7 +523,8 @@ func (r *KubeVirt) DeleteVM(vm *plan.VMStatus) (err error) {
 }
 
 func (r *KubeVirt) DataVolumes(vm *plan.VMStatus) (dataVolumes []cdi.DataVolume, err error) {
-	secret, err := r.ensureSecret(vm.Ref, r.secretDataSetterForCDI(vm.Ref), false)
+	labels := r.vmLabels(vm.Ref)
```
liranr23, line 526: @ahadas why do we need it here?

pkg/controller/plan/kubevirt.go
```diff
@@ -540,7 +541,8 @@ func (r *KubeVirt) DataVolumes(vm *plan.VMStatus) (dataVolumes []cdi.DataVolume,
 }
 
 func (r *KubeVirt) PopulatorVolumes(vmRef ref.Ref) (pvcs []*core.PersistentVolumeClaim, err error) {
-	secret, err := r.ensureSecret(vmRef, r.copyDataFromProviderSecret, false)
+	labels := r.vmLabels(vmRef)
```
liranr23, line 544: same

ahadas (approved):

PR #680, opened 2023-12-11
Make the block overhead configurable

liranr23:
This patch will allow the user to configure the block overhead.
Currently it defaults to 0. The overhead is required when migrating to
encrypted Ceph RBD (block). The pod may see less space on the disk and
therefore the migration will fail when it tries to allocate the last
sectors. See more about encrypted Ceph RBD in:
https://docs.ceph.com/en/quincy/rbd/rbd-encryption/

The value needs to be set in the ForkliftController spec under the
value: `controller_block_overhead`. The addition will be the fixed
number provided to the size of the disk.

### DIFF
--- a/pkg/controller/plan/adapter/ovirt/builder.go
+++ b/pkg/controller/plan/adapter/ovirt/builder.go
@@ -793,11 +793,7 @@ func (r *Builder) getDefaultVolumeAndAccessMode(storageClassName string) ([]core
 // Build a PersistentVolumeClaim with DataSourceRef for VolumePopulator
 func (r *Builder) persistentVolumeClaimWithSourceRef(diskAttachment model.XDiskAttachment, storageClassName string, populatorName string,
 	annotations map[string]string) (pvc *core.PersistentVolumeClaim, err error) {
-
-	// We add 10% overhead because of the fsOverhead in CDI, around 5% to ext4 and 5% for root partition.
-	// This value is configurable using `FILESYSTEM_OVERHEAD`
 	diskSize := diskAttachment.Disk.ProvisionedSize
-
 	var accessModes []core.PersistentVolumeAccessMode
 	var volumeMode *core.PersistentVolumeMode
 	accessModes, volumeMode, err = r.getDefaultVolumeAndAccessMode(storageClassName)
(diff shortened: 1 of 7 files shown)


pkg/controller/plan/util/utils.go
```diff
@@ -20,8 +21,14 @@ func roundUp(requestedSpace, multiple int64) int64 {
...
 }
 
-func CalculateSpaceWithOverhead(requestedSpace int64) int64 {
+func CalculateSpaceWithOverhead(requestedSpace int64, volumeMode *core.PersistentVolumeMode) int64 {
+	var overhead int
+	if *volumeMode == core.PersistentVolumeFilesystem {
+		overhead = settings.Settings.FileSystemOverhead
+	} else {
+		overhead = settings.Settings.BlockOverhead
+	}
```
ahadas, line 30: I'm not sure the block overhead should also be in percentage - for things like the encryption header on Ceph, we can use a fixed overhead instead

pkg/controller/plan/adapter/ovirt/builder.go
```diff
@@ -796,6 +796,8 @@ func (r *Builder) persistentVolumeClaimWithSourceRef(diskAttachment model.XDiskA
 
 	// We add 10% overhead because of the fsOverhead in CDI, around 5% to ext4 and 5% for root partition.
 	// This value is configurable using `FILESYSTEM_OVERHEAD`
+	// Encrypted Ceph RBD causes the pod to see less space, it might be required an overhead in this case.
```
ahadas, line 799:
```suggestion
	// Encrypted Ceph RBD makes the pod see less space, this possible overhead needs to be taken into account.
```

and wouldn't it make more sense to move this comment to the call to CalculateSpaceWithOverhead?

pkg/controller/plan/util/utils.go
```diff
@@ -20,8 +21,14 @@ func roundUp(requestedSpace, multiple int64) int64 {
...
+func CalculateSpaceWithOverhead(requestedSpace int64, volumeMode *core.PersistentVolumeMode) int64 {
+	overhead := 0
+	if *volumeMode == core.PersistentVolumeFilesystem {
+		overhead = settings.Settings.FileSystemOverhead
+	} else {
+		requestedSpace += int64(settings.Settings.BlockOverhead)
+	}
 	alignedSize := roundUp(requestedSpace, DefaultAlignBlockSize)
-	spaceWithOverhead := int64(math.Ceil(float64(alignedSize) / (1 - float64(settings.Settings.FileSystemOverhead)/100)))
+	spaceWithOverhead := int64(math.Ceil(float64(alignedSize) / (1 - float64(overhead)/100)))
```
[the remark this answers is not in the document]
liranr23: here, we may miss the alignment. or we don't care? i tried to respect it when adding a fixed value.

pkg/controller/plan/adapter/ovirt/builder.go
liranr23: it is.

pkg/controller/plan/util/utils.go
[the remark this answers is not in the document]
ahadas: > rethinking - the user adds the value, if he wish he can add something that aligns.

yeah, the overhead would typically be value that aligns with the storage device, like 1G

ahadas (approved):

PR #681, opened 2023-12-12
Make the block overhead configurable (backport)

liranr23:
This patch will allow the user to configure the block overhead. Currently it defaults to 0. The overhead is required when migrating to encrypted Ceph RBD (block). The pod may see less space on the disk and therefore the migration will fail when it tries to allocate the last sectors. See more about encrypted Ceph RBD in:
https://docs.ceph.com/en/quincy/rbd/rbd-encryption/

The value needs to be set in the ForkliftController spec under the value: `controller_block_overhead`. The addition will be the fixed number provided to the size of the disk.


ahadas (approved):

ahadas:
@liranr23 let's include the changes in https://github.com/kubev2v/forklift/pull/684 please

ahadas (approved):

PR #682, opened 2023-12-12
Add missing OS from mapping

liranr23:
Adds el9, win11 and win2k22 for the OS mappings when importing from ovirt. This will help selecting the right template for the VM in the destination cluster.

The OS templates names are taken using:
https://gitlab.com/libosinfo/osinfo-db/-/tree/main/data/os

### DIFF
--- a/pkg/controller/plan/adapter/ovirt/builder.go
+++ b/pkg/controller/plan/adapter/ovirt/builder.go
@@ -89,6 +89,7 @@ var osMap = map[string]string{
 	"rhel_7_s390x":         "rhel7.7",
 	"rhel_7x64":            "rhel7.7",
 	"rhel_8x64":            "rhel8.1",
+	"rhel_9x64":            "rhel9.1",
 	"sles_11_ppc64":        "opensuse15.0",
 	"sles_11":              "opensuse15.0",
 	"sles_12_s390x":        "opensuse15.0",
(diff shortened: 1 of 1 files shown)


pkg/controller/plan/adapter/ovirt/builder.go
```diff
@@ -88,7 +88,8 @@ var osMap = map[string]string{
 	"rhel_7_ppc64":         "rhel7.7",
 	"rhel_7_s390x":         "rhel7.7",
 	"rhel_7x64":            "rhel7.7",
-	"rhel_8x64":            "rhel8.1",
+	"rhel_8x64":            "rhel8.9",
+	"rhel_9x64":            "rhel9.3",
```
ahadas, line 92: @liranr23 did you make sure the osinfo version used by kubevirt already supports these versions? they are fairly new

liranr23: to me it looked as they just pointing to the libosinfo

ahadas: yeah but the question is whether all the supported versions of kubevirt and openshift virtualization use a version of libosinfo that already has support for, e.g., rhel 8.9

ahadas: or maybe I don't understand your previous message - you mean that they handle both rhel 8.9 and 8.1 the same way?

liranr23: in 4.13 they have 8.8 and 9.2.. so i change to that...

ahadas (approved):

PR #686, opened 2023-12-21
Keep the custom CPU model

liranr23:
When importing from oVirt, a VM may set with custom CPU model.
This PR will preserve it. In other case, the user may wish to migrate and preserve the cluster CPU model.
Therefore, in the plan spec it will be possible now to set `PerserveClusterCPUModel` to `true` and forklift
will set the VMs with the source cluster CPU model.


pkg/controller/plan/adapter/ovirt/builder.go
```diff
@@ -391,6 +449,26 @@ func (r *Builder) mapCPU(vm *model.Workload, object *cnv.VirtualMachineSpec) {
...
+	cpuFlags := strings.Split(fullCpu, ",")
+	for _, val := range cpuFlags[1:] {
+		var flag string
+		flag, found := strings.CutPrefix(val, "+")
+		if !found {
+			flag, _ = strings.CutPrefix(val, "-")
+			object.Template.Spec.Domain.CPU.Features = append(object.Template.Spec.Domain.CPU.Features, cnv.CPUFeature{Name: flag, Policy: "disable"})
+		} else {
+			object.Template.Spec.Domain.CPU.Features = append(object.Template.Spec.Domain.CPU.Features, cnv.CPUFeature{Name: flag, Policy: "require"})
+		}
```
ahadas, line 470: but we can also have features without '+' or '-', right?

pkg/controller/plan/adapter/ovirt/builder.go
```diff
@@ -391,6 +449,26 @@ func (r *Builder) mapCPU(vm *model.Workload, object *cnv.VirtualMachineSpec) {
...
 		object.Template.Spec.Domain.CPU.DedicatedCPUPlacement = true
 	}
+	var fullCpu string
+	if vm.CustomCpuModel != "" {
+		fullCpu = vm.CustomCpuModel
+	} else if r.Plan.Spec.PerserveClusterCPUModel {
+		fullCpu = clusterCpuMap[vm.Cluster.CPU.Type]
+	} else {
+		return
+	}
```
ahadas, line 459: how about replacing the if-else statements with switch?

ahadas (commented):
it looks good, but how about separating out the changes for the custom CPU from those for the cluster CPU?

pkg/controller/plan/adapter/ovirt/builder.go
liranr23: i didn't see that in ovirt, but it is possible in theory. i saw that the default case is `+`, therefore it defaults here as well.

ahadas: ok, I'd prefer to delegate this decision to libvirt (to determine the default) but if you want to set it here, it's also fine
but are you sure this code default to 'require'? from a brief look, it seems that the logic is: "if the prefix is `+`, set the policy to `required`, otherwise assume the prefix is `-`.."

liranr23: in [ovirt](https://github.com/oVirt/ovirt-engine/blob/master/backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/builder/vminfo/LibvirtVmXmlBuilder.java#L503-L518) seem it is delegate to libvirt. in libvirt: `Since 0.8.5 the policy attribute can be omitted and will default to require.` [libvirt doc](https://libvirt.org/formatdomain.html#cpu-model-and-topology). i think we can drop it and to like ovirt (delegate).

pkg/controller/plan/adapter/ovirt/builder.go
```diff
@@ -391,7 +420,23 @@ func (r *Builder) mapCPU(vm *model.Workload, object *cnv.VirtualMachineSpec) {
 	if vm.CpuPinningPolicy == model.Dedicated {
 		object.Template.Spec.Domain.CPU.DedicatedCPUPlacement = true
 	}
+	if vm.CustomCpuModel != "" {
+		r.setCpuFlags(vm.CustomCpuModel, object)
+	}
+}
 
+func (r *Builder) setCpuFlags(fullCpu string, object *cnv.VirtualMachineSpec) {
+	object.Template.Spec.Domain.CPU.Model = cpuMap[fullCpu]
```
ahadas, line 429: the idea here is to use the same CPU as the VM would have in oVirt, right? so why not [to take the first item as the model ](https://github.com/oVirt/ovirt-engine/blob/master/backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/builder/vminfo/LibvirtVmXmlBuilder.java#L452)?

ahadas (approved):

ahadas:
> When importing from oVirt, a VM may set with custom CPU model. This PR will preserve it. In other case, the user may wish to migrate and preserve the cluster CPU model. Therefore, in the plan spec it will be possible now to set `PerserveClusterCPUModel` to `true` and forklift will set the VMs with the source cluster CPU model.

@liranr23 please update the description of the PR before merging

PR #687, opened 2023-12-24
Address code smell issues

ahadas:


pkg/controller/plan/migration.go
```diff
@@ -1481,52 +1481,54 @@ func (r *Migration) updateConversionProgress(vm *plan.VMStatus, step *plan.Step)
...
+			err = el9Err
+			return
+		}
 	}
+
 	return
 }
 
 func (r *Migration) updateConversionProgressEl9(pod *core.Pod, step *plan.Step) (err error) {
-	if pod.Status.PodIP == "" {
```
liranr23, line 1516: but then you build a bad url which you later do nothing with it(trying to get) and then you will get an error?

pkg/controller/plan/migration.go
```diff
@@ -1481,52 +1481,54 @@ func (r *Migration) updateConversionProgress(vm *plan.VMStatus, step *plan.Step)
...
+		step.MarkCompleted()
+		step.AddError("Guest conversion failed. See pod logs for details.")
+	default:
+		if pod.Status.PodIP == "" {
+			// we get the progress from the pod and we cannot connect to the pod without PodIP
+			break
+		}
+
+		el9, el9Err := r.Context.Plan.VSphereUsesEl9VirtV2v()
+		switch {
```
liranr23, line 1504: i don't see this switch making it any easier. it was simpler in my opinion with a regular `if el9Err != nil`

ahadas: the idea is to say that we switch between different states of el9 and el9Err that are returned from `VSphereUsesEl9VirtV2v`
I think what made it more complicated for you is that first `el9Err == nil && el9` was checked and then `el9Err != nil` while the former can be simplified if it's checked after `el9Err != nil` - what do you think of the last changes?

liranr23: that's much better, thanks.

pkg/controller/plan/migration.go
liranr23: i did not.

liranr23 (approved):

PR #688, opened 2023-12-24
Keep the cluster CPU

liranr23:
If the user may wish to migrate and preserve the cluster CPU model.
Therefore, in the plan spec it will be possible now to set `PreserveClusterCPUModel` to `true` and forklift
will set the VMs with the source cluster CPU model.

### DIFF
--- a/pkg/controller/plan/adapter/ovirt/builder.go
+++ b/pkg/controller/plan/adapter/ovirt/builder.go
@@ -394,7 +394,21 @@ func (r *Builder) mapCPU(vm *model.Workload, object *cnv.VirtualMachineSpec) {
 	}
 	if vm.CustomCpuModel != "" {
 		r.setCpuFlags(vm.CustomCpuModel, object)
+	} else if r.Plan.Spec.PreserveClusterCPUModel {
+		r.setCpuFlags(r.getClusterCpu(vm), object)
 	}
+}
+
+func (r *Builder) getClusterCpu(vm *model.Workload) string {
+	var cpuAndFlags string
+	cpus := strings.Split(vm.ServerCpu.SystemOptionValue[0].Value, ";")
+	for _, values := range cpus {
+		if strings.Contains(values, vm.Cluster.CPU.Type) {
+			cpuAndFlags = strings.Split(values, ":")[3]
+			break
+		}
+	}
+	return cpuAndFlags
 }
 
 func (r *Builder) setCpuFlags(fullCpu string, object *cnv.VirtualMachineSpec) {
--- a/pkg/controller/provider/container/ovirt/resource.go
+++ b/pkg/controller/provider/container/ovirt/resource.go
@@ -77,6 +77,9 @@ type Cluster struct {
 		Enabled string `json:"enabled"`
 	} `json:"ksm"`
 	BiosType string `json:"bios_type"`
+	CPU      struct {
+		Type string `json:"type"`
+	} `json:"cpu"`
 }
 
 // Apply to (update) the model.
@@ -87,6 +90,7 @@ func (r *Cluster) ApplyTo(m *model.Cluster) {
 	m.HaReservation = r.bool(r.HaReservation)
 	m.KsmEnabled = r.bool(r.KSM.Enabled)
 	m.BiosType = r.BiosType
+	m.CPU.Type = r.CPU.Type
 }
 
 // Cluster (list).
--- a/pkg/controller/provider/container/ovirt/resource.go
+++ b/pkg/controller/provider/container/ovirt/resource.go
@@ -80,6 +80,10 @@ type Cluster struct {
 	CPU      struct {
 		Type string `json:"type"`
 	} `json:"cpu"`
+	Version struct {
+		Minor string `json:"minor"`
+		Major string `json:"major"`
+	} `json:"version"`
 }
 
 // Apply to (update) the model.
@@ -91,11 +95,26 @@ func (r *Cluster) ApplyTo(m *model.Cluster) {
 	m.KsmEnabled = r.bool(r.KSM.Enabled)
 	m.BiosType = r.BiosType
 	m.CPU.Type = r.CPU.Type
+	m.Version.Minor = r.Version.Minor
+	m.Version.Major = r.Version.Major
 }
 
 // Cluster (list).
 type ClusterList struct {
 	Items []Cluster `json:"cluster"`
+}
+
+// ServerCpu.
+type ServerCpu struct {
+	Base
+	Values struct {
+		SystemOptionValues []SystemOptionValue `json:"system_option_value"`
+	} `json:"values"`
+}
+
+type SystemOptionValue struct {
+	Value   string `json:"value"`
+	Version string `json:"version"`
 }
 
 // Host.
--- a/pkg/controller/provider/web/ovirt/workload.go
+++ b/pkg/controller/provider/web/ovirt/workload.go
@@ -3,6 +3,7 @@ package ovirt
 import (
 	"errors"
 	"net/http"
+	"strings"
 
 	"github.com/gin-gonic/gin"
 	api "github.com/konveyor/forklift-controller/pkg/apis/forklift/v1beta1"
@@ -207,6 +208,7 @@ type Workload struct {
 	Host       *Host      `json:"host"`
 	Cluster    Cluster    `json:"cluster"`
 	DataCenter DataCenter `json:"dataCenter"`
+	ServerCpu  ServerCpu  `json:"serverCpu"`
 }
 
 // Build self link (URI).
@@ -220,6 +222,7 @@ func (r *Workload) Link(p *api.Provider) {
 	r.XVM.Link(p)
 	r.Cluster.Link(p)
 	r.DataCenter.Link(p)
+	r.ServerCpu.Link(p)
 	if r.Host != nil {
 		r.Host.Link(p)
 	}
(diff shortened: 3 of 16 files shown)


pkg/controller/plan/adapter/ovirt/builder.go
```diff
@@ -391,7 +449,21 @@ func (r *Builder) mapCPU(vm *model.Workload, object *cnv.VirtualMachineSpec) {
 	if vm.CpuPinningPolicy == model.Dedicated {
 		object.Template.Spec.Domain.CPU.DedicatedCPUPlacement = true
 	}
-
+	var fullCpu string
+	if r.Plan.Spec.PreserveClusterCPUModel {
+		fullCpu = clusterCpuMap[vm.Cluster.CPU.Type]
+	}
+	object.Template.Spec.Domain.CPU.Model = cpuMap[fullCpu]
```
liranr23, line 456: most likely need an `if` to check if we need to do all of this

pkg/controller/plan/adapter/ovirt/builder.go
```diff
@@ -120,6 +120,35 @@ var osMap = map[string]string{
...
+	"Secure Intel Cascadelake Server Family": "Cascadelake-Server-noTSX,-mpx",
+	"Intel Icelake Server Family":            "Icelake-Server-noTSX,-mpx",
+	"Secure Intel Icelake Server Family":     "Icelake-Server-noTSX,+arch-capabilities,+rdctl-no,+ibrs-all,+skip-l1dfl-vmentry,+mds-no,+pschange-mc-no,-mpx",
+	"AMD Opteron G4":                         "Opteron_G4",
+	"AMD Opteron G5":                         "Opteron_G5",
+	"AMD EPYC":                               "EPYC",
+	"Secure AMD EPYC":                        "EPYC,+ibpb,+virt-ssbd",
+	"IBM POWER8":                             "POWER8",
+	"IBM POWER9":                             "POWER9",
+}
```
ahadas, line 150: about this mapping - did you take it from the latest cluster version on upstream?

liranr23: it's from cluster level 4.5 and above: https://github.com/oVirt/ovirt-engine/blob/master/packaging/dbscripts/upgrade/pre_upgrade/0000_config.sql#L1261

ahadas: yeah, it can also be different on downstream..
did you try to get the cluster values from the system options endpoint (BackendSystemOptionResource)?

liranr23: ok, using the endpoint `api/options/ServerCPUList?version=4.5` in ovirt we can get the list of specific version (or without the param and get all of them). so it is possible to be based on ovirt and not on hard-coded map.

ahadas: cool so let's do that

pkg/controller/provider/container/ovirt/collector.go
```diff
@@ -344,6 +344,18 @@ func (r *Collector) beginWatch() (err error) {
 	} else {
 		return
 	}
+	// ServerCPU
```
liranr23, line 347: need to drop, we don't need a watch over the list.

pkg/controller/plan/adapter/ovirt/builder.go
```diff
@@ -394,9 +394,27 @@ func (r *Builder) mapCPU(vm *model.Workload, object *cnv.VirtualMachineSpec) {
 	}
 	if vm.CustomCpuModel != "" {
 		r.setCpuFlags(vm.CustomCpuModel, object)
+	} else if r.Plan.Spec.PreserveClusterCPUModel {
+		r.setCpuFlags(r.getClusterCpu(vm), object)
 	}
 }
 
+func (r *Builder) getClusterCpu(vm *model.Workload) string {
+	clusterVersion := vm.Cluster.Version.Major + "." + vm.Cluster.Version.Minor
```
ahadas, line 403: how about:
```suggestion
	clusterVersion := strings.Join([]string{vm.Cluster.Version.Major, vm.Cluster.Version.Minor}, ".")
```

pkg/controller/provider/container/ovirt/resource.go
```diff
@@ -87,13 +94,41 @@ func (r *Cluster) ApplyTo(m *model.Cluster) {
...
 // Cluster (list).
 type ClusterList struct {
 	Items []Cluster `json:"cluster"`
 }
 
+// ServerCpu.
+type ServerCpu struct {
+	Base
+	Values struct {
+		SystemOptionValue []struct {
```
ahadas, line 111:
```suggestion
		SystemOptionValues []struct {
```

pkg/controller/provider/container/ovirt/watch.go
```diff
@@ -615,3 +615,42 @@ func (r *DiskProfileHandler) validate(profile *model.DiskProfile) {
...
+	}
+}
+
+// Report errors.
+func (r *ServerCpuEventHandler) Error(err error) {
+	r.log.Error(liberr.Wrap(err), err.Error())
+}
+
+func (r *ServerCpuEventHandler) validate(serverCpu *model.ServerCpu) {
+	list := []model.Host{}
```
ahadas, line 642: Host?..

pkg/controller/provider/web/ovirt/workload.go
```diff
@@ -262,6 +264,15 @@ func (r *Workload) Expand(db libmodel.DB) (err error) {
 		return err
 	}
 	r.DataCenter.With(dataCenter)
+	// Server CPU
+	serverCpu := &model.ServerCpu{
+		Base: model.Base{ID: "ServerCPUList"},
```
ahadas, line 269: how about making `ServerCPUList` a const?
and can't we leverage the cluster we queried here to get only the relevant CPUs for the cluster rather than those for all the cluster levels?

pkg/controller/plan/adapter/ovirt/builder.go
```diff
@@ -394,9 +394,27 @@ func (r *Builder) mapCPU(vm *model.Workload, object *cnv.VirtualMachineSpec) {
...
+			continue
+		}
+		cpus := strings.Split(serverCpuOption.Value, ";")
+		for _, values := range cpus {
+			if strings.Contains(values, vm.Cluster.CPU.Type) {
+				return strings.Split(values, ":")[3]
+			}
+		}
+	}
+	return ""
```
ahadas, line 415: this can't happen, right? if it can't happen, I'd declare a variable before the for-loop and update it within the for-loop. this way it seems like we expect to return an empty string in some cases..

pkg/controller/plan/adapter/ovirt/builder.go
```diff
@@ -394,9 +394,27 @@ func (r *Builder) mapCPU(vm *model.Workload, object *cnv.VirtualMachineSpec) {
...
 		r.setCpuFlags(vm.CustomCpuModel, object)
+	} else if r.Plan.Spec.PreserveClusterCPUModel {
+		r.setCpuFlags(r.getClusterCpu(vm), object)
 	}
 }
 
+func (r *Builder) getClusterCpu(vm *model.Workload) string {
+	clusterVersion := vm.Cluster.Version.Major + "." + vm.Cluster.Version.Minor
+	for _, serverCpuOption := range vm.ServerCpu.SystemOptionValue {
+		if serverCpuOption.Version != clusterVersion {
```
ahadas, line 405: I commented on this below as well - can't we do it differently so we'll get here only with the CPUs for the compatibility level of the cluster? it seems odd to query all the values and filtering them out here

ahadas (commented):
much better now!

pkg/controller/provider/web/ovirt/workload.go
liranr23: hmm, maybe if we change they way we set it in our DB. currently it's taken "as is" from ovirt.

ahadas: ok, so let's check the possibility of changing that please

pkg/controller/provider/container/ovirt/watch.go
liranr23: we don't watch this endpoint, so let's drop the event handler.

pkg/controller/provider/container/ovirt/resource.go
```diff
@@ -87,13 +94,32 @@ func (r *Cluster) ApplyTo(m *model.Cluster) {
...
 // Cluster (list).
 type ClusterList struct {
 	Items []Cluster `json:"cluster"`
 }
 
+// ServerCpu.
+type ServerCpu struct {
+	Base
+	Values struct {
+		SystemOptionValues []struct {
```
ahadas, line 111: isn't it possible to do something like:
```suggestion
		SystemOptionValues []SystemOptionValue `json:"system_option_value"`
```

ahadas (commented):
the changes look good
one minor comment inside, and how about adding an end-to-end test now that you added a mock to fakeovirt?

liranr23:
> the changes look good one minor comment inside, and how about adding an end-to-end test now that you added a mock to fakeovirt?

a whole e2e test for this? it is possible, but we need to think here on multiple cases (i tested manually): e.g:
1. have the cluster pass as is(`preserveClusterCpuModel: true`).
2. do as option 1, but have the vm set with custom cpu
of course negative tests are optional as well. we will be interested only in 1, right? and do you think it really worth e2e case?

ahadas (approved):

PR #692, opened 2023-12-26
openstack: convert qcow2 snapshots for image based VMs

bennyz:
This PR currently introduces the image-converter image that will perform the conversion. It has to jobs:
1. Measure the required size, it will receive a PVC and will run a job to measure the required size for the conversion.
    forklift-controller will create an additional PVC for the output of the conversion
2. Another job that will start the pod with `convert` command and will perform the conversion itself


ahadas (commented):
didn't look at the details yet but there might be a need for a 'converter' also for export-to-OVA - can't we reuse the virt-v2v image for this though? it should have qemu-img..

pkg/controller/plan/adapter/openstack/builder.go
```diff
@@ -1250,3 +1257,227 @@ func (r *Builder) GetPopulatorTaskName(pvc *core.PersistentVolumeClaim) (taskNam
...
+			return false, nil
+		}
+
+		r.Log.Info("Convert job status", "pvc", pvc.Name, "status", convertJob.Status)
+		for _, condition := range convertJob.Status.Conditions {
+			switch condition.Type {
+			case batchv1.JobComplete:
+				r.Log.Info("Convert job completed", "pvc", pvc.Name)
+				// remove job after completion
+				err = r.Destination.Client.Delete(context.Background(), convertJob, &client.DeleteOptions{})
```
bennyz, line 1306: move to archive

pkg/controller/plan/kubevirt.go
```diff
@@ -297,6 +297,88 @@ func (r *KubeVirt) DeleteImporterPods(pvc core.PersistentVolumeClaim) (err error
...
+			"Deleted job.",
+			"job",
+			path.Join(
+				job.Namespace,
+				job.Name))
+
+		jobNames = append(jobNames, job.Name)
+	}
+
+	// Remove pods created by jobs
```
ahadas, line 341: that's surprising, aren't they (the pods) set with the job as their owner?

Makefile
```diff
@@ -296,6 +297,17 @@ build-ova-provider-server-image: check_container_runtime
 		$(BAZEL_OPTS) \
 		--action_env CONTAINER_CMD=$(CONTAINER_CMD)
 
+build-image-converter-image: check_container_runtime
```
ahadas, line 300: the changes in forklift-controller look good overall but adding a new image for qemu-img that already exists in the virt-v2v image is debatable, did you try to see if we can reuse the el9 image with virt-v2v instead?

cmd/image-converter/README.md
```diff
@@ -0,0 +1,64 @@
+# Image Converter
+
+Convert the format of images. Since KubeVirt requires RAW images, we sometimes have to convert images before attaching them to a VM.
+
+## Converting
+
+1. Create an empty PVC with the measured size
```
ahadas, line 7: this is outdated, right? (since we don't measure the size)

pkg/controller/plan/kubevirt.go
bennyz: I may have forgotten to set PropagationPolicy

pkg/controller/plan/adapter/openstack/builder.go
```diff
@@ -520,13 +523,36 @@ func (r *Builder) mapDisks(vm *model.Workload, persistentVolumeClaims []core.Per
...
+		// image is volume based, check if it's bootable
+		if volumeID, ok := image.Properties[forkliftPropertyOriginalVolumeID]; ok && volumeID != "" {
+			volume := &model.Volume{}
+			err = r.Source.Inventory.Get(volume, volumeID.(string))
+			if err != nil {
+				r.Log.Error(err, "volume not found in inventory", "volumeID", volumeID)
+				return
+			}
+
+			if volume.Bootable == "true" {
```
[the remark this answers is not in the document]
bennyz: I am not entirely against, but that means checking errors 😅

pkg/controller/plan/adapter/openstack/builder.go
```diff
@@ -891,9 +929,13 @@ func (r *Builder) PopulatorVolumes(vmRef ref.Ref, annotations map[string]string,
 		}
 		images = append(images, image)
 	}
+
 	for _, image := range images {
+		if imageID, ok := image.Properties[forkliftPropertyOriginalImageID]; ok && imageID == workload.ImageID {
+			annotations[planbase.AnnRequiresConversion] = "true"
```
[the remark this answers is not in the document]
bennyz: the snapshot will always qcow2 in this case, it wouldn't hurt to check, but it would have to be set elsewhere

Makefile
ahadas: maybe we don't need to binary part if all it does is to invoke qemu-img, we can enhance [the entrypoint script in that image](https://github.com/kubev2v/forklift/blob/main/virt-v2v/cold/entrypoint) instead

cmd/image-converter/README.md
```diff
@@ -0,0 +1,64 @@
...
+        persistentVolumeClaim:
+          claimName: cirros
+      - name: scratch
+        persistentVolumeClaim:
+          claimName: scratch-pvc
+      serviceAccountName: forklift-populator-controller' | kubectl apply -f -
+```
+
+
+NOTE: `virtctl upload-image` converts the image to RAW, so we the examples converts raw -> qcow2
```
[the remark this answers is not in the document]
ahadas: `otherwise it's harder to create a PVC with a qcow2 file` - I didn't get this part, can you elaborate?

pkg/controller/plan/adapter/openstack/builder.go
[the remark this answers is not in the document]
ahadas: ack, in that case, a comment would be enough

cmd/image-converter/README.md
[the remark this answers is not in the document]
bennyz: I wanted the example to run through the steps in isolation, so I added a step that populates a PVC and uploading with `virtctl` seems like that simplest way (but we get a RAW file instead of qcow2)

Makefile
bennyz: I thought about it, but I feel like the progress monitoring part will be simpler if it's wrapped in go 😅

ahadas: :)
we anyway need to sort out progress reporting with the populator controller for virt-v2v (it's really about time to start working on that!)

pkg/controller/plan/adapter/openstack/builder.go
```diff
@@ -1250,3 +1295,241 @@ func (r *Builder) GetPopulatorTaskName(pvc *core.PersistentVolumeClaim) (taskNam
...
+				},
+			},
+		},
+	}
+}
+
+func makeConversionContainer(pvc *core.PersistentVolumeClaim, srcFormat, dstFormat string) core.Container {
+	container := core.Container{
+		Name:  "convert",
+		Image: base.Settings.VirtV2vImageCold,
```
bennyz, line 1438: for the record, I am against this 😅

pkg/controller/plan/kubevirt.go
bennyz: keeping this as PropagationPolicy doesn't seem to do its thign

pkg/controller/plan/migration.go
```diff
@@ -847,6 +854,39 @@ func (r *Migration) execute(vm *plan.VMStatus) (err error) {
...
+		ready, err := r.builder.ConvertPVCs(pvcs)
+		if err != nil {
+			step.AddError(err.Error())
+			err = nil
+			break
+		}
+
+		if !ready {
+			r.Log.Info("Conversion isn't ready yet")
+			return err
```
ahadas, line 883:
```suggestion
			return nil
```
or simply `return`? `return err` suggests that err might not be nil

pkg/controller/plan/adapter/openstack/builder.go
```diff
@@ -520,13 +523,36 @@ func (r *Builder) mapDisks(vm *model.Workload, persistentVolumeClaims []core.Per
...
+		} else if volumeID, ok := image.Properties[forkliftPropertyOriginalVolumeID]; ok && volumeID != "" {
+			// Image is volume based, check if it's bootable
+			volume := &model.Volume{}
+			err = r.Source.Inventory.Get(volume, volumeID.(string))
+			if err != nil {
+				r.Log.Error(err, "Failed to get volume from inventory", "volumeID", volumeID)
+				return
+			}
+
+			if bootable, err := strconv.ParseBool(volume.Bootable); err != nil && bootable {
```
ahadas, line 550:
```suggestion
			if bootable, err := strconv.ParseBool(volume.Bootable); err == nil && bootable {
```

pkg/controller/plan/adapter/openstack/builder.go
```diff
@@ -1250,3 +1295,241 @@ func (r *Builder) GetPopulatorTaskName(pvc *core.PersistentVolumeClaim) (taskNam
...
+		Image: base.Settings.VirtV2vImageCold,
+		SecurityContext: &core.SecurityContext{
+			AllowPrivilegeEscalation: pointer.Bool(false),
+			RunAsNonRoot:             pointer.Bool(true),
+			RunAsUser:                pointer.Int64(107),
+			Capabilities: &core.Capabilities{
+				Drop: []core.Capability{"ALL"},
+			},
+		},
+		Command: []string{"/usr/local/bin/image-converter"},
```
ahadas, line 1447: do we really need image-converter? what I had in mind is that by using the existing image, we can extend its `entrypoint` script to initiate conversion using qemu-img rather than initiating virt-v2v...

pkg/controller/plan/adapter/openstack/builder.go
ahadas: noted!

cmd/image-converter/README.md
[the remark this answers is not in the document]
ahadas: ah ok, I see your point now

pkg/controller/plan/adapter/openstack/builder.go
bennyz: I know, I just didn't want to mess with the bash script and risk breaking ova and vsphere... also it would've been easier if the virt-v2v part would've been wrapped too, and not broken into go (for progress monitoring) and bash

pkg/controller/plan/migration.go
bennyz: tried naked return originally but doesn't work because of the scoping of `err`

pkg/controller/plan/adapter/openstack/builder.go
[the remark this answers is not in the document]
bennyz: added for now, but I am confirming this with the openstack team

pkg/controller/plan/adapter/openstack/builder.go
ahadas: yeah, [we reused the original script we had before moving the virt-v2v image to this repository](https://github.com/kubev2v/forklift-virt-v2v/blob/main/bin/entrypoint) - but we could have triggered virt-v2v with go.
ok, I see no good reason to prefer a script over this code so let's keep it

cmd/image-converter/README.md
```diff
@@ -0,0 +1,64 @@
...
+kind: Job
+metadata:
+  name: convert-pvc
+  namespace: default
+spec:
+  template:
+    spec:
+      containers:
+      - name: convert-pvc
+        image: quay.io/bzlotnik/image-converter:latest
```
ahadas, line 37: not sure we really need this readme but if we keep it, this needs to change

cmd/image-converter/image-converter.go
```diff
@@ -0,0 +1,77 @@
...
+	flag.StringVar(&dstFormat, "dst-format", "", "Format of the target volume")
+
+	flag.Parse()
+
+	klog.Info("srcVolPath: ", srcVolPath, " dstVolPath: ", dstVolPath, " sourceFormat: ", srcFormat, " targetFormat: ", dstFormat)
+	err := convert(srcVolPath, dstVolPath, srcFormat, dstFormat)
+	if err != nil {
+		klog.Fatal(err)
+	}
+}
```
ahadas, line 26: once/if we integrate progress reporting here then I have no objection for triggering virt-v2v this way

pkg/controller/plan/adapter/openstack/builder.go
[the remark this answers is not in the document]
ahadas: you told me you'll add a test of the format here - that would be better

pkg/controller/plan/adapter/openstack/builder.go
```diff
@@ -520,13 +523,37 @@ func (r *Builder) mapDisks(vm *model.Workload, persistentVolumeClaims []core.Per
 		bus = VirtioBus
 	}
 
+	var bootOrder *uint
+	var imagePVC *core.PersistentVolumeClaim
 	for _, pvc := range persistentVolumeClaims {
+		// Handle loopvar https://go.dev/wiki/LoopvarExperiment
```
ahadas, line 529: it would be great to upgrade to go 1.21 when we can to get not only to get rid of this thingie but also to have slice.Contains ;)

pkg/controller/plan/kubevirt.go
```diff
@@ -842,7 +911,7 @@ func (r *KubeVirt) GetPodsWithLabels(podLabels map[string]string) (pods *core.Po
 }
 
 // Deletes an object from destination cluster associated with the VM.
-func (r *KubeVirt) DeleteObject(object client.Object, vm *plan.VMStatus, message, objType string) (err error) {
+func (r *KubeVirt) DeleteObject(object client.Object, vm *plan.VMStatus, message, objType string, options ...client.DeleteOption) (err error) {
```
ahadas, line 914: `options` is unused but ok, it should be at some point

ahadas (approved):

pkg/controller/plan/kubevirt.go
bennyz: ah yes, I tried PropagationPolicy again 😅

PR #693, opened 2024-01-01
openstack: check image is in the inventory before progressing

bennyz:
Image based migrations can get stuck since there is no check to ensure they were created in the inventory before attempting to progress, this will lead to the migration progressing to the next phase, but the PVC will not be created.

This patch adds the relevant property and checks the snapshot image is in the inventory before progressing

### DIFF
--- a/pkg/controller/plan/adapter/openstack/builder.go
+++ b/pkg/controller/plan/adapter/openstack/builder.go
@@ -1101,8 +1101,7 @@ func (r *Builder) persistentVolumeClaimWithSourceRef(image model.Image, storageC
 	}
 
 	var accessModes []core.PersistentVolumeAccessMode
-	var volumeMode *core.PersistentVolumeMode
-	accessModes, volumeMode, err = r.getVolumeAndAccessMode(storageClassName)
+	accessModes, volumeMode, err := r.getVolumeAndAccessMode(storageClassName)
 	if err != nil {
 		err = liberr.Wrap(err)
 		return
@@ -1116,6 +1115,8 @@ func (r *Builder) persistentVolumeClaimWithSourceRef(image model.Image, storageC
 	} else if originalImageId, ok := image.Properties["forklift_original_image_id"]; ok {
 		annotations[AnnImportDiskId] = originalImageId.(string)
 		r.Log.Info("the image comes from a vm snapshot", "imageID", originalImageId)
+	} else {
+		r.Log.Error(nil, "the image has no volume or vm snapshot associated to it", "image", image.Name)
 	}
 
 	pvc = &core.PersistentVolumeClaim{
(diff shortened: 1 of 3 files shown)


pkg/controller/plan/adapter/openstack/client.go
```diff
@@ -500,7 +500,15 @@ func (r *Client) createVmSnapshotImage(vm *libclient.VM) (vmImage *libclient.Ima
 	}
 	// The vm is image based and we need to create the snapsots of the
 	// volumes attached to it.
-	if _, ok := vm.Image["id"]; ok {
+	if imageID, ok := vm.Image["id"]; ok {
+		// Update property for image based
+		imageUpdateOpts := &libclient.ImageUpdateOpts{}
+		imageUpdateOpts.AddImageProperty(forkliftPropertyOriginalVolumeID, imageID.(string))
```
bennyz, line 506: this can break in builder.go, can't reuse the same property key

pkg/controller/plan/adapter/openstack/builder.go
```diff
@@ -1113,9 +1113,9 @@ func (r *Builder) persistentVolumeClaimWithSourceRef(image model.Image, storageC
 	if originalVolumeDiskId, ok := image.Properties["forklift_original_volume_id"]; ok {
 		annotations[AnnImportDiskId] = originalVolumeDiskId.(string)
 		r.Log.Info("the image comes from a volume", "volumeID", originalVolumeDiskId)
-	} else {
-		annotations[AnnImportDiskId] = image.ID
-		r.Log.Info("the image comes from a vm snapshot", "imageID", image.ID)
+	} else if originalImageId, ok := image.Properties["forklift_original_image_id"]; ok {
+		annotations[AnnImportDiskId] = originalImageId.(string)
+		r.Log.Info("the image comes from a vm snapshot", "imageID", originalImageId)
 	}
```
ahadas, line 1119: this code suggests that there might be another case where both properties don't exist even though it wouldn't make sense, right? maybe adding an `else` with a log?

pkg/controller/plan/adapter/openstack/client.go
```diff
@@ -500,7 +500,15 @@ func (r *Client) createVmSnapshotImage(vm *libclient.VM) (vmImage *libclient.Ima
 	}
 	// The vm is image based and we need to create the snapsots of the
```
ahadas, line 501: while on it (2)
```suggestion
	// The vm is image based and we need to create the snapshots of the
```

pkg/controller/plan/adapter/openstack/builder.go
ahadas: while on it, can you please replace 
`accessModes, volumeMode, err = r.getVolumeAndAccessMode(storageClassName)`
above with:
`accessModes, volumeMode, err := r.getVolumeAndAccessMode(storageClassName)`

ahadas (approved):

PR #694, opened 2024-01-02
Disable TPM gathering from vSphere temporarily

ahadas:
The addition of the property "summary.config.tpmPresent" breaks compatibility with vSphere 6.5 ("InvalidProperty" error is returned when building the inventory). This patch reverts this part temporarily until we change this to work also with older versions of vSphere.

Related to #665

### DIFF
--- a/pkg/controller/provider/container/vsphere/collector.go
+++ b/pkg/controller/provider/container/vsphere/collector.go
@@ -715,7 +715,7 @@ func (r *Collector) propertySpec() []types.PropertySpec {
 				fExtraConfig,
 				fGuestName,
 				fGuestID,
-				fTpmPresent,
+				//fTpmPresent,
 				fBalloonedMemory,
 				fVmIpAddress,
 				fStorageUsed,


pkg/controller/provider/container/vsphere/collector.go
```diff
@@ -715,7 +715,7 @@ func (r *Collector) propertySpec() []types.PropertySpec {
 				fExtraConfig,
 				fGuestName,
 				fGuestID,
-				fTpmPresent,
+				//fTpmPresent,
```
bkhizgiy, line 718: This value isnt used later on  in the flow of the creation? I mean wont it cause issues in the VM migration if we wont set it with some kind of default value it in the inventory?

ahadas: it won't be reported by vsphere so from our side it will appear as if the vm doesn't have tpm in the rest of the flow

bkhizgiy (approved):

liranr23 (approved):
that's unfortunate..

PR #695, opened 2024-01-02
Add e2e test for preserve cluster CPU

liranr23:


liranr23:
@ahadas i added a new case here, to make it work i need to change the test to handle another vm in fakeovirt or delete the existing vm. i thought on another option, to just check the CPU within existing test. WDYT?

ahadas:
> @ahadas i added a new case here, to make it work i need to change the test to handle another vm in fakeovirt or delete the existing vm. i thought on another option, to just check the CPU within existing test. WDYT?

why do we need to do that, is it because there's a limit on the number of queries from fakeovirt?

liranr23:
> > @ahadas i added a new case here, to make it work i need to change the test to handle another vm in fakeovirt or delete the existing vm. i thought on another option, to just check the CPU within existing test. WDYT?
> 
> why do we need to do that, is it because there's a limit on the number of queries from fakeovirt?

the restart of fakeovirt - yes.
the other vm - conflict of existing vm (e.g mac conflict)

ahadas:
> > > @ahadas i added a new case here, to make it work i need to change the test to handle another vm in fakeovirt or delete the existing vm. i thought on another option, to just check the CPU within existing test. WDYT?
> > 
> > 
> > why do we need to do that, is it because there's a limit on the number of queries from fakeovirt?
> 
> the restart of fakeovirt - yes. the other vm - conflict of existing vm (e.g mac conflict)

it would be best in my opinion to use the existing vm as a way to verify the CPU is not set when PreserveClusterCPUModel is not set, and to add another vm that will be migrated with PreserveClusterCPUModel set

tests/suit/ovirt_test.go
```diff
@@ -172,5 +172,101 @@ var _ = Describe("[level:component]Migration tests for oVirt providers", func()
...
+			Expect(err).ToNot(HaveOccurred())
+			err = utils.WaitForMigrationSucceededWithTimeout(f.CrClient, provider.Namespace, test_migration_name, 300*time.Second)
+			Expect(err).ToNot(HaveOccurred())
+			By("Verifying imported VM exists")
+			vm, err = utils.GetImportedVm(f.CrClient, targetNS.Name, func(vm cnv.VirtualMachine) bool {
+				return vm.Spec.Template.Spec.Domain.Firmware.UUID == vmId
+			})
+			Expect(err).ToNot(HaveOccurred())
+			Expect(vm).ToNot(BeNil())
+			Expect(vm.Spec.Template.Spec.Domain.CPU.Model).To(Equal("Westmere"))
```
bkhizgiy, line 269: maybe would be nicer to hold this value as a const.

liranr23:
> > > > @ahadas i added a new case here, to make it work i need to change the test to handle another vm in fakeovirt or delete the existing vm. i thought on another option, to just check the CPU within existing test. WDYT?
> > > 
> > > 
> > > why do we need to do that, is it because there's a limit on the number of queries from fakeovirt?
> > 
> > 
> > the restart of fakeovirt - yes. the other vm - conflict of existing vm (e.g mac conflict)
> 
> it would be best in my opinion to use the existing vm as a way to verify the CPU is not set when PreserveClusterCPUModel is not set, and to add another vm that will be migrated with PreserveClusterCPUModel set

OK, in that matter we already have 2 VMs (fors secure and insecure). We can check one and see we don't have the CPU model, and the other to check we do have it.

bkhizgiy:
> > > > > @ahadas i added a new case here, to make it work i need to change the test to handle another vm in fakeovirt or delete the existing vm. i thought on another option, to just check the CPU within existing test. WDYT?
> > > > 
> > > > 
> > > > why do we need to do that, is it because there's a limit on the number of queries from fakeovirt?
> > > 
> > > 
> > > the restart of fakeovirt - yes. the other vm - conflict of existing vm (e.g mac conflict)
> > 
> > 
> > it would be best in my opinion to use the existing vm as a way to verify the CPU is not set when PreserveClusterCPUModel is not set, and to add another vm that will be migrated with PreserveClusterCPUModel set
> 
> OK, in that matter we already have 2 VMs (fors secure and insecure). We can check one and see we don't have the CPU model, and the other to check we do have it.
Yep, we have two VMs I think the combination with the insecure test can be OK for this case, I don't think its worth the work of adding a new VM and dont see any reason of this colliding with the insecure check so it should be fine IMO.

bkhizgiy (approved):

tests/suit/ovirt_test.go
```diff
@@ -171,6 +176,8 @@ var _ = Describe("[level:component]Migration tests for oVirt providers", func()
 			})
 			Expect(err).ToNot(HaveOccurred())
 			Expect(vm).ToNot(BeNil())
+			By("Verifying VM created with the cluster CPU")
```
ahadas, line 179:
```suggestion
			By("Verifying VM is created with the cluster CPU")
```

ahadas (commented):
looks good, minor comments about phrasing

ahadas (approved):

PR #697, opened 2024-01-03
Support adding an ESXi host as a source provider

ahadas:
This PR adds the ability to add a vSphere provider that connects to SDK that is served by an ESXi host, instead of that of  vCenter. This allows users to import directly from environments that comprise of ESXi hosts, without vCenter.

We don't add a new type of provider since most of the code is similar in both cases, the SDK that is exposed from ESXi aligns with the one exposed from vCenter. The only difference is that in case we operate with an ESXi host, we need to set the libvirt URL with `esx://` rather than `vpx://` for virt-v2v.

For vSphere providers that interact with an ESXi hosts, the backend doesn't assume clients specify the credentials for the host as these credentials were already specified for the provider - a mutating webhook copies the credentials from the secret of the Provider to the secret of the Host in this case.

### DIFF
--- a/pkg/forklift-api/webhooks/mutating-webhook/mutators/provider-mutator.go
+++ b/pkg/forklift-api/webhooks/mutating-webhook/mutators/provider-mutator.go
@@ -1,0 +1,68 @@
+package mutators
+
+import (
+	"encoding/json"
+
+	api "github.com/konveyor/forklift-controller/pkg/apis/forklift/v1beta1"
+	"github.com/konveyor/forklift-controller/pkg/forklift-api/webhooks/util"
+	admissionv1 "k8s.io/api/admission/v1beta1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type ProviderMutator struct {
+	ar       *admissionv1.AdmissionReview
+	provider api.Provider
+	Client   client.Client
+}
+
+func (mutator *ProviderMutator) Mutate(ar *admissionv1.AdmissionReview) *admissionv1.AdmissionResponse {
+	log.Info("provider mutator was called")
+	mutator.ar = ar
+	raw := ar.Request.Object.Raw
+	if err := json.Unmarshal(raw, &mutator.provider); err != nil {
+		log.Error(err, "mutating webhook error, failed to unmarshel provider")
+		return util.ToAdmissionResponseError(err)
+	}
+
+	specChanged := mutator.setSdkEndpointIfNeeded()
+
+	if specChanged {
+		patches := mutator.patchPayload()
+		patchBytes, err := util.GeneratePatchPayload(patches...)
+		if err != nil {
+			log.Error(err, "mutating webhook error, failed to generate payload for patch request")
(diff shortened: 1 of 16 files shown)


pkg/forklift-api/webhooks/mutating-webhook/mutators/provider-mutator.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+}
+
+func (mutator *ProviderMutator) setVSphereEndpointIfNeeded() bool {
+	var providerChanged bool
+
+	if mutator.provider.Type() == api.VSphere {
+		if _, ok := mutator.provider.Spec.Settings[api.EndpointType]; !ok {
+			log.Info("endpoint type was not specified for vSphere provider, assuming vCenter")
+			mutator.provider.Spec.Settings[api.EndpointType] = api.ESXI
```
ahadas, line 54: this needs to change, did this for development..

liranr23 (approved):
LGTM,
one question about the mutator/validation - does it make any sense to validate the given URLs of each provider to see it has `/api`? / `/sdk`?

ahadas:
> LGTM, one question about the mutator/validation - does it make any sense to validate the given URLs of each provider to see it has `/api`? / `/sdk`?

we chose not to enforce it - the UI shows a warning in this case and let the user to proceed with posting the provider

ahadas:
> For vSphere providers that interact with an ESXi hosts, the backend doesn't assume clients specify the credentials for the host as these credentials were already specified for the provider - a mutating webhook copies the credentials from the secret of the Provider to the secret of the Host in this case.

@yaacov that's what agreed on offline, are you also ok with naming the new setting `sdkEndpoint` (that can be set to `vcenter` or `esxi`)?

tests/suit/ova_test.go
```diff
@@ -32,6 +32,7 @@ var _ = Describe("[level:component]Migration tests for OVA provider", func() {
 		secret, err := utils.CreateSecretFromDefinition(f.K8sClient, utils.NewSecretDefinition(
 			map[string]string{
 				"createdForProviderType": "ova",
+				"createdForResourceType": "providers",
```
ahadas, line 35: @bkhizgiy that took me quite some time yesterday, to realize why that's needed - apparently OVA providers **must** be processed by the secret mutator since we validate their secrets have the `insecureSkipVerify` property - we don't have to force having it since we should support handling existing providers that don't have that setting, could you please change that?

ahadas: @bkhizgiy I posted https://github.com/kubev2v/forklift/pull/703 for this, please take a look

bkhizgiy: sorry I've missed the previous comment from yesterday, I just comment on the PR, so yes we can entirely drop it, I will also add a fix for the mutator so this property wont be set on the secret (but this part is not a blocker).

yaacov:
> @yaacov that's what we agreed on offline, are you also ok with naming the new setting sdkEndpoint (that can be set to vcenter or esxi)?

Ack, that is good for me too

PR #701, opened 2024-01-04
Enable TPM gathering from vSphere

liranr23:
Since #694, we disabled the TPM gathering to avoid "InvalidProperty" error.
vSphere started supporting TPM API in 6.7, based on that version we will gather the TPM on supported versions.

### DIFF
--- a/pkg/controller/provider/container/vsphere/collector.go
+++ b/pkg/controller/provider/container/vsphere/collector.go
@@ -697,39 +697,50 @@ func (r *Collector) propertySpec() []types.PropertySpec {
 			},
 		},
 		{ // VM
-			Type: VirtualMachine,
-			PathSet: []string{
-				fName,
-				fParent,
-				fUUID,
-				fFirmware,
-				fFtInfo,
-				fCpuAffinity,
-				fCpuHotAddEnabled,
-				fCpuHotRemoveEnabled,
-				fMemoryHotAddEnabled,
-				fNumCpu,
-				fNumCoresPerSocket,
-				fMemorySize,
-				fDevices,
-				fExtraConfig,
-				fGuestName,
-				fGuestID,
-				//fTpmPresent,
-				fBalloonedMemory,
-				fVmIpAddress,
-				fStorageUsed,
-				fDatastore,
-				fNetwork,
-				fRuntimeHost,
-				fPowerState,
-				fConnectionState,
-				fIsTemplate,
-				fSnapshot,
-				fChangeTracking,
-			},
+			Type:    VirtualMachine,
+			PathSet: r.vmPathSet(),
 		},
 	}
+}
+
+func (r *Collector) vmPathSet() []string {
+	pathSet := []string{
+		fName,
+		fParent,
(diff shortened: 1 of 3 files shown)


pkg/controller/provider/container/vsphere/collector.go
```diff
@@ -697,41 +697,53 @@ func (r *Collector) propertySpec() []types.PropertySpec {
...
+		fNetwork,
+		fRuntimeHost,
+		fPowerState,
+		fConnectionState,
+		fIsTemplate,
+		fSnapshot,
+		fChangeTracking,
+	}
+
+	if majorVal >= 6 && minorVal >= 7 {
```
ahadas, line 741: that's not good since it would fail on things like 7.0
how about adding a unit test for this?

pkg/controller/provider/container/vsphere/collector.go
```diff
@@ -697,41 +697,53 @@ func (r *Collector) propertySpec() []types.PropertySpec {
...
+			Type:    VirtualMachine,
+			PathSet: r.vmPathSet(),
 		},
 	}
 }
 
+func (r *Collector) vmPathSet() []string {
+	apiVer := strings.Split(r.client.ServiceContent.About.ApiVersion, ".")
+	majorVal, _ := strconv.Atoi(apiVer[0])
+	minorVal, _ := strconv.Atoi(apiVer[1])
```
ahadas, line 709: please do that closer to where the variables are used

ahadas:
> Since #694, which discovered vSphere validates the API endpoint requested, we disabled the TPM gathering to avoid "InvalidProperty" error. vSphere started supporting TPM API in 6.7, based on that version we will gather the TPM on supported versions.

I'm not sure it's actually an API endpoint, it seems like a properties that are retrieved from a specific API endpoint

ahadas (commented):
looks good now, @liranr23 how about adding a unit test for the new function?

pkg/controller/provider/container/vsphere/collector_test.go
```diff
@@ -0,0 +1,36 @@
...
+	g.Expect(err).To(gomega.BeNil())
+	vimClient := &vim25.Client{
+		Client:         soap.NewClient(url, false),
+		ServiceContent: types.ServiceContent{},
+	}
+	collector.client = &govmomi.Client{
+		SessionManager: session.NewManager(vimClient),
+		Client:         vimClient,
+	}
+	collector.client.ServiceContent.About.ApiVersion = "6.5"
```
bkhizgiy, line 28: maybe worth adding name/description for each test so it will be easier to understand what they are testing and desire outcome.

liranr23: i added comments. i think the other option is to split tests (i don't think changing `t.Name` every line is fine).

ahadas (commented):
there's some room for improvement in the test, but other than that this looks good. let's improve the tests separately

ahadas (approved):

PR #703, opened 2024-01-09
Allow OVA provider secret without 'insecureSkipVerify'

ahadas:


pkg/controller/provider/validation.go
```diff
@@ -258,7 +258,6 @@ func (r *Reconciler) validateSecret(provider *api.Provider) (secret *core.Secret
 	case api.Ova:
 		keyList = []string{
 			"url",
-			"insecureSkipVerify",
```
bkhizgiy, line 261: Its really not needed here, can be drooped. The mutator will add this value any way, so if we want to entirely remove this value from the secret we need to exclude ova from there as well.

bkhizgiy: following the other PR comment we can entirely drop it.

PR #708, opened 2024-01-10
extract image converter

bennyz:
Extract the image conversion from openstack to a generic converter which implements an interface making it testable and possibly reusable


pkg/controller/plan/migration.go
```diff
@@ -860,6 +864,25 @@ func (r *Migration) execute(vm *plan.VMStatus) (err error) {
 			vm.AddError(fmt.Sprintf("Step '%s' not found", r.step(vm)))
 			break
 		}
+
+		if r.converter == nil {
+			labels := map[string]string{
```
bennyz, line 869: these two are somewhat questionable, it'd be better if they are constructed outside per provider

ahadas: +1
there is some similarity between this conversion and the guest conversion in case of warm migration from vsphere (in both cases we transfer the image(s) and then initiate another pod to convert the data we got. but they are also different in the sense that here it's per-PVC while the pod in case of warm migration should get all the volumes..
but it would be nice if we generalize this somehow and extract the warm migration to be specific to the vsphere provider, and this part specific to the OpenStack provider under a 'conversion' phase.

ahadas (commented):
generalizing the volume converter so it can be reusable also for export OVA is an interesting direction but need to think whether we'll want the export OVA flow to be:
1. convert the volumes to qcow2/vmdk
2. initiate another pod that copies their data into an OVA
or to initiate a pod that gets all the volumes and convert the data directly into the OVA (as we do in oVirt using a loopback device). the latter option makes this more similar to the flow of warm migration from vsphere

bennyz:
> generalizing the volume converter so it can be reusable also for export OVA is an interesting direction but need to think whether we'll want the export OVA flow to be:
> 
> 1. convert the volumes to qcow2/vmdk
> 2. initiate another pod that copies their data into an OVA
> 
> or to initiate a pod that gets all the volumes and convert the data directly into the OVA (as we do in oVirt using a loopback device). the latter option makes this more similar to the flow of warm migration from vsphere

I think we can still do that if I change the interface back to:
```go
type PVCConverter interface {
	ConvertPVC(pvcs []*v1.PersistentVolumeClaim, srcFormat, dstFormat string) (ready bool, err error)
}
```

(I now noticed I left it `pvcs`)

But then `srcFormat` is no good, because in theory we can have PVCs with different source formats, in that case we can switch it to function

```go
type GetSourceFormatFn func(pvc *v1.PersistentVolumeClaim) string

type PVCConverter interface {
	ConvertPVC(pvcs *v1.PersistentVolumeClaim, fn GetSourceFormatFn, dstFormat string) (ready bool, err error)
}
```

And then we'll have a separate implementation for an `OVAConverter` which will start one job for all volumes

I believe this is good regardless of reusability, because this approach makes it easier to test this flow isolation with https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/client/fake

ahadas:
> > generalizing the volume converter so it can be reusable also for export OVA is an interesting direction but need to think whether we'll want the export OVA flow to be:
> > 
> > 1. convert the volumes to qcow2/vmdk
> > 2. initiate another pod that copies their data into an OVA
> > 
> > or to initiate a pod that gets all the volumes and convert the data directly into the OVA (as we do in oVirt using a loopback device). the latter option makes this more similar to the flow of warm migration from vsphere
> 
> I think we can still do that if I change the interface back to:
> 
> ```go
> type PVCConverter interface {
> 	ConvertPVC(pvcs []*v1.PersistentVolumeClaim, srcFormat, dstFormat string) (ready bool, err error)
> }
> ```
> 
> (I now noticed I left it `pvcs`)
> 
> But then `srcFormat` is no good, because in theory we can have PVCs with different source formats, in that case we can switch it to function
> 
> ```go
> type GetSourceFormatFn func(pvc *v1.PersistentVolumeClaim) string
> 
> type PVCConverter interface {
> 	ConvertPVC(pvcs *v1.PersistentVolumeClaim, fn GetSourceFormatFn, dstFormat string) (ready bool, err error)
> }
> ```
> 
> And then we'll have a separate implementation for an `OVAConverter` which will start one job for all volumes
> 
> I believe this is good regardless of reusability, because this approach makes it easier to test this flow isolation with https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/client/fake

yeah, that sounds good - and then we won't allocate new volumes for the converted images but the pod that would start for export OVA would create the OVA and convert the disks directly to it

pkg/controller/plan/adapter/converter.go
```diff
@@ -0,0 +1,259 @@
...
+	"k8s.io/apimachinery/pkg/api/resource"
+	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/pointer"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type FilterFn func(pvc *v1.PersistentVolumeClaim) bool
+
+type PVCConverter interface {
+	ConvertPVC(pvcs *v1.PersistentVolumeClaim, srcFormat, dstFormat string) (ready bool, err error)
```
ahadas, line 23: and following up on the other thread, then we won't need dstFormat at that level because each provider would have a different one (OSP - raw, vSphere - none, export OVA - qcow2/vmdk)

ahadas: actually not sure about the vSphere part

forklift_suite_test.go
ahadas, line 1: we talked about this offline - it's weird that we need this at the top level of the repository..

go.mod
```diff
@@ -14,6 +14,7 @@ require (
 	github.com/k8snetworkplumbingwg/network-attachment-definition-client v1.4.0
 	github.com/mattn/go-sqlite3 v1.14.18
 	github.com/onsi/ginkgo v1.16.5
+	github.com/onsi/ginkgo/v2 v2.13.2
```
ahadas, line 17: we also talked about this offline - making this a direct dependency is good but I find the changed vendored files suprising..

pkg/controller/plan/adapter/converter.go
```diff
@@ -0,0 +1,285 @@
...
+			completed++
+			continue
+		}
+
+		scratchPVC, err := c.ensureScratchPVC(pvc)
+		if err != nil {
+			return false, err
+		}
+
+		if scratchPVC == nil {
```
ahadas, line 54: don't we get `scratchPVC == nil` only when `err != nil`?

pkg/controller/plan/adapter/converter.go
```diff
@@ -0,0 +1,285 @@
...
+			c.Log.Info("Scratch PVC status is unknown", "pvc", scratchPVC.Name, "status", scratchPVC.Status.Phase)
+			return false, nil
+		}
+
+		convertJob, err := c.ensureJob(pvc, srcFormat(pvc), dstFormat)
+		if err != nil {
+			return false, err
+		}
+
+		if convertJob == nil {
```
ahadas, line 78: same as above

pkg/controller/plan/kubevirt.go
```diff
@@ -795,7 +795,7 @@ func (r *KubeVirt) getListOptionsNamespaced() (listOptions *client.ListOptions)
 }
 
 // Ensure the guest conversion (virt-v2v) pod exists on the destination.
-func (r *KubeVirt) EnsureGuestConversionPod(vm *plan.VMStatus, vmCr *VirtualMachine, pvcs *[]core.PersistentVolumeClaim) (err error) {
+func (r *KubeVirt) EnsureGuestConversionPod(vm *plan.VMStatus, vmCr *VirtualMachine, pvcs []*core.PersistentVolumeClaim) (err error) {
```
ahadas, line 798: I don't think that it's that significant to pass a slice of PVCs or a reference for that slice since VMs wouldn't have so many disks, it's ok

pkg/controller/plan/migration.go
```diff
@@ -860,6 +863,26 @@ func (r *Migration) execute(vm *plan.VMStatus) (err error) {
 			vm.AddError(fmt.Sprintf("Step '%s' not found", r.step(vm)))
 			break
 		}
+
+		var filterFn func(pvc *core.PersistentVolumeClaim) bool
```
ahadas, line 867: why defining filterFn here?

pkg/controller/plan/migration.go
```diff
@@ -871,7 +894,11 @@ func (r *Migration) execute(vm *plan.VMStatus) (err error) {
 			break
 		}
 
-		ready, err := r.builder.ConvertPVCs(pvcs)
+		srcFormatFn := func(pvc *core.PersistentVolumeClaim) string {
+			return pvc.Annotations[base.AnnSourceFormat]
```
ahadas, line 898: so if we say that for export-OVA (@liranr23 @bkhizgiy ) we won't need this, why having this generalization rather than taking the source format from the annotation within the converter?

bennyz: we can, but it feels weird in terms of API usability, it makes the format hidden that way

ahadas: I wouldn't say it's hidden that way, but yeah - it relies on the annotation that we rely on also this way..

ahadas:
@bennyz ping?

pkg/controller/plan/migration.go
ahadas: ok, a bit of an over-spec here but it's not too complex..

ahadas (approved):

bennyz:
> > proposal
> 
> @bennyz can you please write few words here?

added

PR #711, opened 2024-01-11
openstack: fix conversion for block volumes

### LINKED ISSUE #710: Fix openstack qcow2 snapshot conversion for volumeMode: Block
1. The destination path is not correctly set
2. Copy data between block devices

bennyz:
fix #710

### DIFF
--- a/cmd/image-converter/image-converter.go
+++ b/cmd/image-converter/image-converter.go
@@ -9,29 +9,55 @@ import (
 )
 
 func main() {
-	var srcVolPath, dstVolPath, srcFormat, dstFormat string
+	var srcVolPath, dstVolPath, srcFormat, dstFormat, volumeMode string
 
 	flag.StringVar(&srcVolPath, "src-path", "", "Source volume path")
 	flag.StringVar(&dstVolPath, "dst-path", "", "Target volume path")
 	flag.StringVar(&srcFormat, "src-format", "", "Format of the source volume")
 	flag.StringVar(&dstFormat, "dst-format", "", "Format of the target volume")
+	flag.StringVar(&volumeMode, "volume-mode", "", "Format of the target volume")
 
 	flag.Parse()
 
 	klog.Info("srcVolPath: ", srcVolPath, " dstVolPath: ", dstVolPath, " sourceFormat: ", srcFormat, " targetFormat: ", dstFormat)
-	err := convert(srcVolPath, dstVolPath, srcFormat, dstFormat)
+	err := convert(srcVolPath, dstVolPath, srcFormat, dstFormat, volumeMode)
 	if err != nil {
 		klog.Fatal(err)
 	}
 }
 
-func convert(srcVolPath, dstVolPath, sourceFormat, targetFormat string) error {
+func convert(srcVolPath, dstVolPath, srcFormat, dstFormat, volumeMode string) error {
+	err := qemuimgConvert(srcVolPath, dstVolPath, srcFormat, dstFormat)
+	if err != nil {
+		return err
+	}
+
+	// Copy dst over src
+	switch volumeMode {
+	case "Block":
+		err = qemuimgConvert(dstVolPath, srcVolPath, dstFormat, dstFormat)
+		if err != nil {
+			return err
+		}
+	case "Filesystem":
+		// Use mv for files as it's faster than qemu-img convert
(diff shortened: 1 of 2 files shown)


cmd/image-converter/image-converter.go
```diff
@@ -57,21 +72,5 @@ func convert(srcVolPath, dstVolPath, sourceFormat, targetFormat string) error {
 		return err
 	}
 
-	// Replace /mnt/disk.img with /output/disk.img
-	cmd = exec.Command(
```
liranr23, line 61: why mv or cp can't work while qemu-img convert does?

bennyz: it works only for files, for block volumes you need to copy the content of the device

although now that I think about it may actually make sense to keep the `mv` for files, because it should be much faster

ahadas: > although now that I think about it may actually make sense to keep the `mv` for files, because it should be much faster

are you sure qemu-img doesn't have this optimization? AFAIU vdsm always invokes qemu-img for this and I find it surprising if it could have been optimized

bennyz: I suspect that's because of the other parameters vdsm uses, I'd be very surprised if qemu-img will do what `mv` does

liranr23 (approved):

cmd/image-converter/image-converter.go
```diff
@@ -25,13 +27,39 @@ func main() {
...
+		}
+	} else {
+		// Use mv for files as its faster than qemu-img convert
+		cmd := exec.Command("mv", dstVolPath, srcVolPath)
+		err := cmd.Run()
+		if err != nil {
+			return err
+		}
+
+		return nil
```
ahadas, line 50: if the don't have `return nil` in the block above, why having it here? I'd rather drop it (or add it to the block above and drop it below)

cmd/image-converter/image-converter.go
```diff
@@ -57,21 +85,20 @@ func convert(srcVolPath, dstVolPath, sourceFormat, targetFormat string) error {
...
-	if err != nil {
-		return err
+	stat, ok := fileInfo.Sys().(*syscall.Stat_t)
+	if !ok {
+		klog.Fatal("failed to get raw syscall.Stat_t data")
 	}
 
-	return nil
+	// Check if it's a block device
+	return (stat.Mode & syscall.S_IFMT) == syscall.S_IFBLK
```
ahadas, line 103: what happens on rbd that supports both (when we get a filesystem on block storage)?

bennyz: it depends on the volumeMode on the PVC, which determines how it's attached to the node, so it shouldn't matter

cmd/image-converter/image-converter.go
bennyz: bad copy paste...

cmd/image-converter/image-converter.go
ahadas: ok, then (stat.Mode & syscall.S_IFMT) will return a filesystem in that case and both source and destination volumes are of the same mode so it looks correct, but why not to propagate this from the controller that knows the mode already?

cmd/image-converter/image-converter.go
```diff
@@ -25,13 +27,37 @@ func main() {
...
+	}
+
+	// Copy dst over src
+	if isBlockDevice(srcVolPath) {
+		err = qemuimgConvert(dstVolPath, srcVolPath, dstFormat, dstFormat)
+		if err != nil {
+			return err
+		}
+	} else {
+		// Use mv for files as its faster than qemu-img convert
```
ahadas, line 43:
```suggestion
		// Use mv for files as it's faster than qemu-img convert
```

cmd/image-converter/image-converter.go
bennyz: i think it's weird from the standpoint of the container to request something that can be determined

ahadas: yeah, it's less standalone but on the other hand it's simplify the code - I think it looks better now, thanks for changing that

ahadas (approved):
@bennyz I've added another commit to handle the case of `pvc.Spec.VolumeMode == nil` when setting the volume-mode parameter, please take a look at that change

PR #716, opened 2024-01-22
Support kubevirt instancetypes

liranr23:
This patch add the ability to use the virtualmachineclusterpreferences by kubevirt. These adds some properties to the virtual machine based on the OS within the guest. In case we failed to find the relevant instancetype, we fallback to template.


pkg/controller/plan/adapter/openstack/builder.go
```diff
@@ -781,20 +781,22 @@ func (r *Builder) DataVolumes(vmRef ref.Ref, secret *core.Secret, configMap *cor
...
-// Build tasks.
-func (r *Builder) TemplateLabels(vmRef ref.Ref) (labels map[string]string, err error) {
+func (r *Builder) InstanceTypeName(vmRef ref.Ref) (name string, err error) {
 	vm := &model.Workload{}
 	err = r.Source.Inventory.Find(vm, vmRef)
 	if err != nil {
 		err = liberr.Wrap(
 			err,
-			VM_LOOKUP_FAILED,
+			"VM lookup failed.",
```
liranr23, line 790: need to switch to the const

pkg/controller/plan/kubevirt.go
```diff
@@ -1132,12 +1133,18 @@ func (r *KubeVirt) virtualMachine(vm *plan.VMStatus) (object *cnv.VirtualMachine
 	}
 
 	var ok bool
-	object, ok = r.vmTemplate(vm)
+	object, ok = r.instancetype(vm)
 	if !ok {
-		r.Log.Info("Building VirtualMachine without template.",
+		r.Log.Info("Building VirtualMachine without instancetype.",
```
ahadas, line 1138:
```suggestion
		r.Log.Info("Building VirtualMachine without an instance type.",
```

pkg/controller/plan/adapter/ocp/builder.go
```diff
@@ -230,6 +230,12 @@ func (r *Builder) Tasks(vmRef ref.Ref) (list []*planapi.Task, err error) {
 	return
 }
 
+func (r *Builder) InstanceTypeName(vmRef ref.Ref) (name string, err error) {
+	// The VM is build from configuration, we don't need the instancetype
```
ahadas, line 234:
```suggestion
	// The VM is built from configuration, we don't need the instancetype
```

pkg/controller/plan/adapter/openstack/builder.go
```diff
@@ -821,25 +823,64 @@ func (r *Builder) TemplateLabels(vmRef ref.Ref) (labels map[string]string, err e
...
+					os = fmt.Sprintf("%s2k19", os)
+				} else if strings.Contains(version, "2k22") || strings.Contains(version, "2022") {
+					os = fmt.Sprintf("%s2k22", os)
+				} else if len(version) >= 2 && version[:2] == "11" {
+					os = fmt.Sprintf("%s%s", os, version)
+				}
+			}
+		}
+	} else {
+		if os != UnknownOS && version != "" {
```
ahadas, line 847: what's the difference between those blocks (templates vs. instance types)?
if there's a difference, I'd extract those to separate functions rather than passing a boolean argument for indicating whether we look for a template or an instance type

pkg/controller/plan/adapter/ovirt/builder.go
```diff
@@ -559,6 +590,25 @@ func (r *Builder) Tasks(vmRef ref.Ref) (list []*plan.Task, err error) {
...
+		err = liberr.Wrap(
+			err,
+			"VM lookup failed.",
+			"vm",
+			vmRef.String())
+		return
+	}
+	name, ok := osInstanceTypeMap[vm.OSType]
+	if !ok {
+		err = fmt.Errorf("nothing fits the input OS")
```
ahadas, line 606: please use liberr instead

pkg/controller/plan/adapter/ova/builder.go
```diff
@@ -440,6 +440,12 @@ func (r *Builder) Tasks(vmRef ref.Ref) (list []*plan.Task, err error) {
 	return
 }
 
+func (r *Builder) InstanceTypeName(vmRef ref.Ref) (name string, err error) {
+	// The VM is build from configuration, we don't need the instancetype
```
ahadas, line 444: not sure I get this part - why don't we need the instance type? (and specifically, when we otherwise we set the template)

pkg/controller/plan/kubevirt.go
```diff
@@ -1166,6 +1173,33 @@ func (r *KubeVirt) virtualMachine(vm *plan.VMStatus) (object *cnv.VirtualMachine
...
+	err = r.Destination.Client.Get(
+		context.TODO(),
+		client.ObjectKey{Name: instancetypeName},
+		instanceTypePreference)
+	if err != nil {
+		r.Log.Info("could not find instance type preference for destination VM.",
+			"vm",
+			vm.String(),
+			"error",
+			err)
```
ahadas, line 1193: then it seems more appropriate to log it in error-level

pkg/controller/plan/kubevirt.go
```diff
@@ -1166,6 +1173,33 @@ func (r *KubeVirt) virtualMachine(vm *plan.VMStatus) (object *cnv.VirtualMachine
...
 }
 
+// Attempt to find a suitable instancetype preference.
+func (r *KubeVirt) instancetype(vm *plan.VMStatus) (virtualMachine *cnv.VirtualMachine, ok bool) {
+	instancetypeName, err := r.Builder.InstanceTypeName(vm.Ref)
+	if err != nil {
+		return
+	}
+
+	instanceTypePreference := &instancetype.VirtualMachineClusterPreference{}
```
ahadas, line 1183: I see that there's also a namespaced entity (virtualmachinepreferences.instancetype.kubevirt.io) - I guess we should first try to look for that one, but need to check this with kubevirt guys

pkg/controller/plan/adapter/ocp/builder.go
```diff
@@ -230,6 +230,12 @@ func (r *Builder) Tasks(vmRef ref.Ref) (list []*planapi.Task, err error) {
...
+	err = liberr.New("instancetypes are not used by this provider")
```
ahadas, line 235: did you check what happens in this case of migration of an VM with instance type from one cluster to another, in terms of the instance type specification that is exposed by CDI? it can preserve the original instance type that the vm was based on (which is fine) or "collapse" the properties into the exposed configuration (which is not that great since we lose the instance type that the vm originally had, but on the other hand - if the instance type preferences are different in both clusters than it might be better). need to decide on what's the right direction here

ahadas (commented):
I see that this PR is still in a draft mode, which is fine, and yet - the title should change if we use only the 'preferences' part of the instance type. this PR is more like "replacing templates with instance type preferences" than "support instance types"

@fabiand can you please suggest who can review this one from the kubevirt side?

fabiand:
@0xFelix can you please give this a review?
The goal is that forklift is - roughly (instanceTypes and Preferences are not an exact science) - map VMs to instanceTypes and preferences.

We should avoid creating custom ones.

cc @dominikholler

pkg/controller/plan/adapter/ovirt/builder.go
```diff
@@ -120,6 +120,37 @@ var osMap = map[string]string{
 	"windows_11":           "win11",
 }
 
+// Map of ovirt guest ids to instance type preference names.
+var osInstanceTypeMap = map[string]string{
```
fabiand, line 124: Can we try to make this dynamic?

Please note that instanceTypes and preferences carry a lot of labels which can be used for the mapping.

pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -120,6 +120,25 @@ var osMap = map[string]string{
 	"windows9Server64Guest": "win2k19",
 }
 
+// Map of vmware guest ids to instancetype names.
+var osInstanceTypeMap = map[string]string{
```
fabiand, line 124: Same here, can we make this dynamic?

fabiand (commented):
Awesome to see this work

pkg/controller/plan/adapter/base/doc.go
```diff
@@ -78,6 +78,8 @@ type Builder interface {
 	GetPopulatorTaskName(pvc *core.PersistentVolumeClaim) (taskName string, err error)
 	// Convert the PVC to a given format
 	ConvertPVCs(pvcs []core.PersistentVolumeClaim) (ready bool, err error)
+	// Get the instancetype name
+	InstanceTypeName(vmRef ref.Ref) (name string, err error)
```
0xFelix, line 82: IIUC you got this wrong. Instancetypes are about resource sizing, e.g. how many cores and how much memory your VM gets. Preferences are about runtime preferences, resp. hardware configuration a specific OS expects. Throughout this whole PR you use the term `InstanceType`, when you are actually refering to preferences.

pkg/controller/plan/adapter/ocp/builder.go
0xFelix: Instancetypes and preferences can vary between clusters. Also, a VM always uses a specific ControllerRevision of an instancetype or preference. If you move the VM definition over to another cluster without moving the ControllerRevision of instancetype/preference too, the VM cannot start. 

Options would be:

- Clearing the `revisionName` of the instancetype or preference (Caution: Instancetype or preference on dest cluster might look different or are nonexistent at all)
- Use the Expand endpoint of the KubeVirt API to expand the use instancetype/preference into the VM definition (you lose the abstraction, your VM definition becomes larger)

pkg/controller/plan/adapter/openstack/builder.go
0xFelix: +1

Looks like this can be deduplicated.

pkg/controller/plan/adapter/openstack/builder.go
```diff
@@ -821,25 +823,64 @@ func (r *Builder) TemplateLabels(vmRef ref.Ref) (labels map[string]string, err e
...
-				os = fmt.Sprintf("%s2k19", os)
-			} else if strings.Contains(version, "2k22") || strings.Contains(version, "2022") {
-				os = fmt.Sprintf("%s2k22", os)
-			} else if len(version) >= 2 && version[:2] == "11" {
-				os = fmt.Sprintf("%s%s", os, version)
+	if template {
+		if os != UnknownOS && version != "" {
+			os = fmt.Sprintf("%s.%s", os, version)
+			if distro == CentOS && len(version) >= 1 && (version[:1] == "8" || version[:1] == "9") {
+				os = fmt.Sprintf("%s-stream%s", distro, version)
```
0xFelix, line 830: CentOS doesn't necessarily has to be Stream. It could be an old 7/8 CentOS VM too.

pkg/controller/plan/adapter/openstack/builder.go
```diff
@@ -821,25 +823,64 @@ func (r *Builder) TemplateLabels(vmRef ref.Ref) (labels map[string]string, err e
...
+					os = fmt.Sprintf("%s2k12", os)
+				} else if strings.Contains(version, "2k16") || strings.Contains(version, "2016") {
+					os = fmt.Sprintf("%s2k16", os)
+				} else if strings.Contains(version, "2k19") || strings.Contains(version, "2019") {
+					os = fmt.Sprintf("%s2k19", os)
+				} else if strings.Contains(version, "2k22") || strings.Contains(version, "2022") {
+					os = fmt.Sprintf("%s2k22", os)
+				} else if len(version) >= 2 && version[:2] == "11" {
+					os = fmt.Sprintf("%s%s", os, version)
+				}
```
0xFelix, line 843: A switch/case might look nicer.

pkg/controller/plan/adapter/openstack/builder.go
```diff
@@ -821,25 +823,64 @@ func (r *Builder) TemplateLabels(vmRef ref.Ref) (labels map[string]string, err e
...
+				os = DefaultWindows
+				if strings.Contains(version, "2k12") || strings.Contains(version, "2012") {
+					os = fmt.Sprintf("%s2k12", os)
+				} else if strings.Contains(version, "2k16") || strings.Contains(version, "2016") {
+					os = fmt.Sprintf("%s2k16", os)
+				} else if strings.Contains(version, "2k19") || strings.Contains(version, "2019") {
+					os = fmt.Sprintf("%s2k19", os)
+				} else if strings.Contains(version, "2k22") || strings.Contains(version, "2022") {
+					os = fmt.Sprintf("%s2k22", os)
+				} else if len(version) >= 2 && version[:2] == "11" {
```
0xFelix, line 841: What about Windows 10?

pkg/controller/plan/adapter/ovirt/builder.go
```diff
@@ -120,6 +120,37 @@ var osMap = map[string]string{
...
+	"ubuntu_12_10":       "ubuntu",
+	"ubuntu_13_04":       "ubuntu",
+	"ubuntu_13_10":       "ubuntu",
+	"ubuntu_14_04_ppc64": "ubuntu",
+	"ubuntu_14_04":       "ubuntu",
+	"ubuntu_16_04_s390x": "ubuntu",
+	"windows_10":         "windows.10", // windows also have virtio: e.g windows.10.virtio might be good to imports from ovirt?
+	"windows_10x64":      "windows.10",
+	"windows_2003":       "windows.10",
+	"windows_2003x64":    "windows.10",
```
0xFelix, line 140: Will likely not work. Windows 10 will use EFI starting with the next release of common-instancetypes.

pkg/controller/plan/adapter/ovirt/builder.go
```diff
@@ -120,6 +120,37 @@ var osMap = map[string]string{
 	"windows_11":           "win11",
 }
 
+// Map of ovirt guest ids to instance type preference names.
+var osInstanceTypeMap = map[string]string{
+	"rhel_7_ppc64":       "rhel.7",
+	"rhel_7_s390x":       "rhel.7",
```
0xFelix, line 126: These architectures likely won't work with KubeVirt right now.

0xFelix: Also the preference are mostly suited for x64 guests.

pkg/controller/plan/adapter/ovirt/builder.go
```diff
@@ -120,6 +120,37 @@ var osMap = map[string]string{
...
+	"windows_2012R2x64":  "win2k12",
+	"windows_2012x64":    "win2k12",
+	"windows_2016x64":    "win2k16",
+	"windows_2019x64":    "win2k19",
+	"windows_2022":       "win2k22",
+	"windows_7":          "windows.10",
+	"windows_7x64":       "windows.10",
+	"windows_8":          "windows.10",
+	"windows_8x64":       "windows.10",
+	"windows_xp":         "windows.10",
```
0xFelix, line 150: Same, might not work.

pkg/controller/plan/adapter/ovirt/builder.go
```diff
@@ -559,6 +590,25 @@ func (r *Builder) Tasks(vmRef ref.Ref) (list []*plan.Task, err error) {
 	return
 }
 
+func (r *Builder) InstanceTypeName(vmRef ref.Ref) (name string, err error) {
+	vm := &model.Workload{}
+	err = r.Source.Inventory.Find(vm, vmRef)
+	if err != nil {
```
0xFelix, line 596: Inline into if?

pkg/controller/plan/adapter/openstack/builder.go
```diff
@@ -821,25 +823,64 @@ func (r *Builder) TemplateLabels(vmRef ref.Ref) (labels map[string]string, err e
...
 		}
 	}
+	return
+}
+
+// Build tasks.
+func (r *Builder) TemplateLabels(vmRef ref.Ref) (labels map[string]string, err error) {
+	vm := &model.Workload{}
+	err = r.Source.Inventory.Find(vm, vmRef)
+	if err != nil {
```
0xFelix, line 874: Inline into if?

pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -120,6 +120,25 @@ var osMap = map[string]string{
...
+	"centos7_64Guest":       "centos.7",
+	"centos7Guest":          "centos.7",
+	"centos8_64Guest":       "centos.8stream",
```
0xFelix, line 127: CentOS 8 can be Stream/Non-Stream, preference is called `centos.stream8`.

pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -686,6 +705,25 @@ func (r *Builder) Tasks(vmRef ref.Ref) (list []*plan.Task, err error) {
 	return
 }
 
+func (r *Builder) InstanceTypeName(vmRef ref.Ref) (name string, err error) {
+	vm := &model.Workload{}
+	err = r.Source.Inventory.Find(vm, vmRef)
+	if err != nil {
```
0xFelix, line 711: Same

pkg/controller/plan/kubevirt.go
```diff
@@ -1166,6 +1173,33 @@ func (r *KubeVirt) virtualMachine(vm *plan.VMStatus) (object *cnv.VirtualMachine
 	return
 }
 
+// Attempt to find a suitable instancetype preference.
```
0xFelix, line 1176: You are looking for just a preference.

pkg/controller/plan/kubevirt.go
```diff
@@ -1166,6 +1173,33 @@ func (r *KubeVirt) virtualMachine(vm *plan.VMStatus) (object *cnv.VirtualMachine
...
+	if err != nil {
+		r.Log.Info("could not find instance type preference for destination VM.",
+			"vm",
+			vm.String(),
+			"error",
+			err)
+		return
+	}
+	virtualMachine = r.emptyVm(vm)
+	perMatcher := &cnv.PreferenceMatcher{Name: instanceTypePreference.Name}
```
0xFelix, line 1197: You need to set the `kind` if you want to use namespaced resources. Defaults to cluster-wide.

pkg/controller/plan/kubevirt.go
0xFelix: Yes, depending on the destination namespace, there might be "local" instancetypes/preferences there too.

pkg/controller/plan/kubevirt.go
```diff
@@ -1166,6 +1173,33 @@ func (r *KubeVirt) virtualMachine(vm *plan.VMStatus) (object *cnv.VirtualMachine
...
+	virtualMachine.Spec.Preference = perMatcher
+	ok = true
```
0xFelix, line 1199: That feels non-golang like. If something went wrong, return an `err`?

0xFelix (commented):
Good idea, let's get this straightened out! Did you consider using instancetypes (resource sizing) too?

liranr23:
> Good idea, let's get this straightened out! Did you consider using instancetypes (resource sizing) too?

Yes.
I decided not to go with `instancetypes` because of:
- It's mainly for memory and CPU resources, which we override based on the source VM.
- For "special" type(memory, compute) we can't know what the VM is intended to do, therefore it makes it difficult to choose which profile we need.

pkg/controller/plan/adapter/openstack/builder.go
liranr23: done

pkg/controller/plan/kubevirt.go
liranr23: added

pkg/controller/plan/kubevirt.go
```diff
@@ -1132,12 +1133,20 @@ func (r *KubeVirt) virtualMachine(vm *plan.VMStatus) (object *cnv.VirtualMachine
 	}
 
 	var ok bool
-	object, ok = r.vmTemplate(vm)
-	if !ok {
-		r.Log.Info("Building VirtualMachine without template.",
+	object, err = r.vmPreference(vm)
+	if err != nil {
+		r.Log.Info("Building VirtualMachine without a preference.",
```
ahadas, line 1138: it is a bit ambiguous this way, how about replacing 'a preference' with VirtualMachinePreference?

pkg/controller/plan/kubevirt.go
ahadas: ok, I'll comment on the change

pkg/controller/plan/adapter/ovirt/builder.go
ahadas: > Can we try to make this dynamic?

+1
loading it from some config map would be nice

> but i do think we can query the list of preferences, iterate and check relevant name/annotation..

right, but with proper label(s) we can rather do it in a "k8s way" using a label selector
I also can't find a good label to use for selection though, for example for RHEL 7 we only have:
```yaml
  labels:
    app.kubernetes.io/component: templating
    app.kubernetes.io/managed-by: ssp-operator
    app.kubernetes.io/name: common-instancetypes
    app.kubernetes.io/part-of: hyperconverged-cluster
    app.kubernetes.io/version: 4.14.3
    instancetype.kubevirt.io/common-instancetypes-version: v0.3.3
    instancetype.kubevirt.io/os-type: linux
    instancetype.kubevirt.io/vendor: redhat.com
```

pkg/controller/plan/adapter/ovirt/builder.go
```diff
@@ -120,6 +120,30 @@ var osMap = map[string]string{
...
+var osInstanceTypeMap = map[string]string{
+	"rhel_7x64":         "rhel.7",
+	"rhel_8x64":         "rhel.8",
+	"rhel_9x64":         "rhel.9",
+	"ubuntu_12_04":      "ubuntu",
+	"ubuntu_12_10":      "ubuntu",
+	"ubuntu_13_04":      "ubuntu",
+	"ubuntu_13_10":      "ubuntu",
+	"ubuntu_14_04":      "ubuntu",
+	"windows_10":        "windows.10", // windows also have virtio: e.g windows.10.virtio might be good to imports from ovirt?
```
ahadas, line 133: that would be a benefit of having this configurable - then we can default it to windows.10.virtio, and if customers have Windows guests in oVirt without virtio, then they can switch

pkg/controller/plan/adapter/openstack/builder.go
```diff
@@ -823,23 +824,92 @@ func (r *Builder) TemplateLabels(vmRef ref.Ref) (labels map[string]string, err e
...
+		os = UnknownOS
+	case RHEL, CentOS, Fedora, Ubuntu, OpenSUSE, Windows:
+		os = distro
+	case SLED:
+		os = OpenSUSE
+	case MSDOS:
+		os = Windows
+	default:
+		os = UnknownOS
+	}
```
ahadas, line 874: it's better after the split but there's still room for reducing code duplication

pkg/controller/plan/adapter/ova/builder.go
ahadas: ok I see, so it's not because it's built from configuration but since we do not take the operating system within the OVF into account, we can do it like we do in oVirt. CC @bkhizgiy

pkg/controller/plan/kubevirt.go
ahadas: but we may have an actual error here..
I see your point in saying that it's not necessarily a severe error
I'd check the type of the error - if it's a not-found error, print an info message that no corresponding instance type was found, falling back... and if the error is different, log an error

pkg/controller/plan/adapter/ocp/builder.go
ahadas: > Options would be:
> 
> * Clearing the `revisionName` of the instancetype or preference (Caution: Instancetype or preference on dest cluster might look different or are nonexistent at all)
> * Use the Expand endpoint of the KubeVirt API to expand the use instancetype/preference into the VM definition (you lose the abstraction, your VM definition becomes larger)

I don't mind expanding (yes, that's the correct term, I wrote "collapse" before but meant "expand") the configuration since the VM is already provisioned (so the abstraction doesn't give us much and I don't think the length of the VM definition would be important at this point).

but there's an argument saying that typical VMs in kubevirt are based on instancetype/preference and we don't want VMs that are imported from MTV to be different...

> * create new preference by forklift, copying the exact source preference.

the problem with this approach is that the instancetype/preference might already exist on the desintation - so we may end up with centos-from-cluster-x, centos-from-cluster-y, etc - which is not that great

@fabiand we talked about this in the context of importing appliance from OVA - the question is how important is it for you to avoid having VMs that are not based on instancetypes/preferences, because AFAIU it doesn't give us much value for VMs that are already provisioned in kubevirt

ahadas:
> > Good idea, let's get this straightened out! Did you consider using instancetypes (resource sizing) too?
> 
> Yes. I decided not to go with `instancetypes` because of:
> 
> * It's mainly for memory and CPU resources, which we override based on the source VM.
> * For "special" type(memory, compute) we can't know what the VM is intended to do, therefore it makes it difficult to choose which profile we need.

it's ok not to use `instancetypes` in this PR but limit it to template -> vm-preference
however, there has not been proper discussion about whether or not to use instancetypes

yes, when users migrate from vsphere/ovirt/openstack - it's likely that they'll want their VMs to be as similar as possible to how they are defined on the source, so properties like memory and cpu would typically be overridden
but instance types can potentially bring us value (besides having the VMs configured as typical VMs in kubevirt - as I wrote on a different thread):
1. there are properties that are not considered at all and thus get default values, like ioThreadsPolicy
2. forklift doesn't allow mass change of configuration of the migrated vms, e.g., changing the memory

if we enable users to select an instance type, we can improve those aspects

pkg/controller/plan/adapter/ovirt/builder.go
liranr23: ok, so let's do all ovirt windows vms with virtio

pkg/controller/plan/adapter/openstack/builder.go
0xFelix: Yep, there is no preference for CentOS 8 (anymore).

pkg/controller/plan/kubevirt.go
liranr23: it changed to not found error - info, otherwise it will return a real error.

pkg/controller/plan/adapter/ovirt/builder.go
liranr23: added config maps

pkg/controller/plan/kubevirt.go
```diff
@@ -1166,6 +1175,91 @@ func (r *KubeVirt) virtualMachine(vm *plan.VMStatus) (object *cnv.VirtualMachine
...
+	label := map[string]string{}
+	switch providerType {
+	case api.VSphere:
+		label["subject"] = "vsphere-osmap-config"
+	case api.OVirt:
+		label["subject"] = "ovirt-osmap-config"
+	default:
+		return
+	}
+	err = r.Destination.Client.List(
```
liranr23, line 1210: this part is tricky... it means that the configmaps on destination cluster, which we can't promise.
the workaround for it is to post the configmaps to the destination cluster, either by using our deployed ones or by hardcode it and post it...

pkg/controller/plan/adapter/ovirt/builder.go
0xFelix: Would adding an identifier similar to the ones used by osinfo-db help?

pkg/controller/plan/kubevirt.go
```diff
@@ -1166,6 +1175,91 @@ func (r *KubeVirt) virtualMachine(vm *plan.VMStatus) (object *cnv.VirtualMachine
...
+	virtualMachine.Spec.Preference = &cnv.PreferenceMatcher{Name: preferenceName, Kind: kind}
+	return
+}
+
+func (r *KubeVirt) getOsMapConfig(providerType v1beta1.ProviderType) (configMap *core.ConfigMap, err error) {
+	list := &core.ConfigMapList{}
+	label := map[string]string{}
+	switch providerType {
+	case api.VSphere:
+		label["subject"] = "vsphere-osmap-config"
```
0xFelix, line 1204: Might make sense to use consts here.

pkg/controller/plan/kubevirt.go
```diff
@@ -1166,6 +1175,91 @@ func (r *KubeVirt) virtualMachine(vm *plan.VMStatus) (object *cnv.VirtualMachine
...
+	case api.OVirt:
+		label["subject"] = "ovirt-osmap-config"
+	default:
+		return
+	}
+	err = r.Client.List(
```
0xFelix, line 1210: Will the names of config maps be randomized? Can't use a single name / Get here?

pkg/controller/plan/kubevirt.go
```diff
@@ -1166,6 +1175,91 @@ func (r *KubeVirt) virtualMachine(vm *plan.VMStatus) (object *cnv.VirtualMachine
...
+		},
+	)
+	if err != nil {
+		err = liberr.Wrap(err)
+		return
+	}
+	if len(list.Items) > 0 {
+		configMap = &list.Items[0]
+	}
+	return
```
0xFelix, line 1224: This might return a `nil` ptr and `err == nil`, which is unusual for golang.

0xFelix: Could return something like `NotFoundError` instead.

pkg/controller/plan/kubevirt.go
```diff
@@ -1166,6 +1175,91 @@ func (r *KubeVirt) virtualMachine(vm *plan.VMStatus) (object *cnv.VirtualMachine
...
+
+func (r *KubeVirt) getPreference(vm *plan.VMStatus, preferenceName string) (name, kind string, err error) {
+	virtualMachinePreference := &instancetype.VirtualMachinePreference{}
+	err = r.Destination.Client.Get(
+		context.TODO(),
+		client.ObjectKey{Name: preferenceName, Namespace: r.Plan.Spec.TargetNamespace},
+		virtualMachinePreference)
+	if err != nil {
+		if k8serr.IsNotFound(err) {
+			r.Log.Info("could not find a local instance type preference for destination VM. trying cluser wide",
```
0xFelix, line 1235: typo: `cluser` / `cluster`

pkg/controller/plan/kubevirt.go
```diff
@@ -1166,6 +1175,91 @@ func (r *KubeVirt) virtualMachine(vm *plan.VMStatus) (object *cnv.VirtualMachine
...
+		err = liberr.Wrap(err)
+		return
+	}
+	if len(list.Items) > 0 {
+		configMap = &list.Items[0]
+	}
+	return
+}
+
+func (r *KubeVirt) getPreference(vm *plan.VMStatus, preferenceName string) (name, kind string, err error) {
```
0xFelix, line 1227: If you refactor getting a `virtualMachinePreference` or `virtualMachineClusterPreference` into helper funcs you should get away without too many nested if/else statements and early returns instead.

pkg/controller/plan/adapter/openstack/builder.go
```diff
@@ -820,26 +819,73 @@ func (r *Builder) TemplateLabels(vmRef ref.Ref) (labels map[string]string, err e
 	default:
 		os = UnknownOS
 	}
+	return
+}
 
+func (r *Builder) getPreferenceOs(vm *model.Workload, os, version, distro string) string {
```
ahadas, line 825: how about adding unit tests for this?

pkg/controller/plan/adapter/ova/builder.go
```diff
@@ -440,6 +440,12 @@ func (r *Builder) Tasks(vmRef ref.Ref) (list []*plan.Task, err error) {
 	return
 }
 
+func (r *Builder) PreferenceName(vmRef ref.Ref, configMap *core.ConfigMap) (name string, err error) {
+	// The VM is built from configuration, we don't need the preference
+	err = liberr.New("preferences are not used by this provider")
```
ahadas, line 445: this builder should be similar to that of vSphere, no?

pkg/controller/plan/adapter/ovirt/builder.go
```diff
@@ -559,6 +559,28 @@ func (r *Builder) Tasks(vmRef ref.Ref) (list []*plan.Task, err error) {
...
+			err,
+			"VM lookup failed.",
+			"vm",
+			vmRef.String())
+		return
+	}
+	name, ok := configMap.Data[vm.OSType]
+	if !ok {
+		err = liberr.Wrap(err,
+			"nothing fits the input OS")
```
ahadas, line 579: did you check that this log looks alright? the signature of liberr.Wrap is `(err error, kvpair ...interface{})`

pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -695,6 +695,30 @@ func (r *Builder) Tasks(vmRef ref.Ref) (list []*plan.Task, err error) {
...
+			"VM lookup failed.",
+			"vm",
+			vmRef.String())
+		return
+	}
+
+	name, ok := configMap.Data[vm.GuestID]
+	if !ok {
+		err = liberr.Wrap(err,
+			"nothing fits the input OS")
```
ahadas, line 717: same here

pkg/controller/plan/kubevirt.go
```diff
@@ -1166,6 +1175,91 @@ func (r *KubeVirt) virtualMachine(vm *plan.VMStatus) (object *cnv.VirtualMachine
...
+		return
+	}
+
+	virtualMachine = r.emptyVm(vm)
+	virtualMachine.Spec.Preference = &cnv.PreferenceMatcher{Name: preferenceName, Kind: kind}
+	return
+}
+
+func (r *KubeVirt) getOsMapConfig(providerType v1beta1.ProviderType) (configMap *core.ConfigMap, err error) {
+	list := &core.ConfigMapList{}
```
ahadas, line 1200: listing the config maps here doesn't make much sense - it would be better to propagate the name from the operator (e.g., `vsphere_osmap_configmap_name`) and call 'get'

pkg/controller/plan/kubevirt.go
ahadas: +1
that's what I meant in my comment above

pkg/controller/plan/kubevirt.go
ahadas: +1
it also took me a minute to realize that we rely on PreferenceName to check whether the config map is nil - that's not standard...

operator/roles/forkliftcontroller/tasks/main.yml
```diff
@@ -62,6 +62,16 @@
       state : present
       definition: "{{ lookup('template', 'controller/configmap-controller.yml.j2') }}"
 
+  - name: "Setup controller ovirt operation systems config map"
```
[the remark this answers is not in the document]
liranr23: done

operator/roles/forkliftcontroller/tasks/main.yml
```diff
@@ -62,6 +62,16 @@
...
+    k8s:
+      state : present
+      definition: "{{ lookup('template', 'controller/configmap-osmap-ovirt.yml.j2') }}"
+
+  - name: "Setup controller vsphere operation systems config map"
```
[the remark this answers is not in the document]
liranr23: done

operator/roles/forkliftcontroller/templates/controller/configmap-osmap-ovirt.yml.j2
```diff
@@ -0,0 +1,31 @@
...
+  windows_10x64: "windows.10.virtio"
+  windows_2012R2x64: "win2k12.virtio"
+  windows_2012x64: "win2k12.virtio"
+  windows_2016x64: "win2k16.virtio"
+  windows_2019x64: "win2k19.virtio"
+  windows_2022: "win2k22.virtio"
+  windows_7: "windows.10.virtio"
+  windows_7x64: "windows.10.virtio"
+  windows_8: "windows.10.virtio"
+  windows_8x64: "windows.10.virtio"
```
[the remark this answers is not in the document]
liranr23: removed

operator/roles/forkliftcontroller/templates/controller/configmap-osmap-vsphere.yml.j2
```diff
@@ -0,0 +1,26 @@
...
+  centos8_64Guest: "centos.stream8"
+  centos8Guest: "centos.stream8"
+  fedora64Guest: "fedora"
+  fedoraGuest: "fedora"
+  rhel7_64Guest: "rhel.7"
+  rhel7Guest: "rhel.7"
+  rhel8_64Guest: "rhel.8"
+  ubuntu64Guest: "ubuntu"
+  ubuntuGuest: "ubuntu"
+  windows8Server64Guest: "windows.2k12"
```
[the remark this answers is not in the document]
liranr23: removed

operator/roles/forkliftcontroller/templates/controller/configmap-osmap-vsphere.yml.j2
```diff
@@ -0,0 +1,26 @@
...
+  windows9_64Guest: "windows.10"
+  windows9Guest: "windows.10"
+  windows9Server64Guest: "windows.2k19"
```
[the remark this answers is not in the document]
liranr23: yup, my bad.

pkg/controller/plan/adapter/base/doc.go
```diff
@@ -78,6 +78,8 @@ type Builder interface {
 	GetPopulatorTaskName(pvc *core.PersistentVolumeClaim) (taskName string, err error)
 	// Convert the PVC to a given format
 	ConvertPVCs(pvcs []core.PersistentVolumeClaim) (ready bool, err error)
+	// Get the preference name
```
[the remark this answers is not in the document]
liranr23: done

pkg/controller/plan/adapter/ova/builder.go
liranr23: i think you forgot about the previous comment regarding this one.. in OVA we always set to `unknown` the OS we read from the OVF, so it doesn't matter here.
Generally - yes, it should be like vsphere.

pkg/controller/plan/adapter/ovirt/builder.go
liranr23: added more data to align with the rest of the liberr wrap

pkg/controller/plan/adapter/vsphere/builder.go
liranr23: added more data to align with the rest of the liberr wrap

pkg/controller/plan/kubevirt.go
liranr23: changed to const name of the configmap

pkg/controller/plan/kubevirt.go
liranr23: done

pkg/controller/plan/kubevirt.go
liranr23: done

pkg/controller/plan/kubevirt.go
liranr23: i changed to `GET` now, so i think it's covered. i just removed this check.

pkg/controller/plan/kubevirt.go
liranr23: changed, it's still with nested `if` and `else` but it looks better.

pkg/controller/plan/kubevirt.go
liranr23: done

pkg/controller/plan/adapter/openstack/builder.go
liranr23: i'll try to do on another commit, i changed all the rest first

pkg/controller/plan/adapter/ovirt/builder.go
liranr23: yes, as you can already see in previous comment, we already using `os.template.kubevirt.io/<os>` when the `os` is mapped from vsphere/ovirt to the osinfo. that would've saved us from duplication (which we can remove later on and use the config map introduced here).

pkg/controller/plan/adapter/ova/builder.go
ahadas: ah yeah, I had Déjà-vu when I wrote it :) so let's add a comment?

pkg/controller/plan/adapter/openstack/builder.go
```diff
@@ -820,26 +819,75 @@ func (r *Builder) TemplateLabels(vmRef ref.Ref) (labels map[string]string, err e
...
 		os = UnknownOS
 	}
+	return
+}
 
+func getPreferenceOs(os, version, distro string) string {
+	if os != UnknownOS && version != "" {
+		if distro == CentOS && len(version) >= 1 && (version[:1] == "8" || version[:1] == "9") {
+			os = fmt.Sprintf("%s.stream%s", distro, version)
+		} else if os == Windows {
```
liranr23, line 829: i do wonder if we should set them to include `virtio`. @ahadas wdyt?

ahadas: I don't see much sense in having this different from the logic we have for ovirt - after all, both platforms are based on the same hypervisor..

pkg/controller/plan/adapter/openstack/builder_test.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			actualOs := getPreferenceOs(tt.os, tt.version, tt.distro)
+			if actualOs != tt.expectedOs {
+				t.Errorf("got (%s), want (%s)", actualOs, tt.expectedOs)
+			}
+		})
+	}
```
ahadas, line 59: how about using a Table like we do [here](https://github.com/kubev2v/forklift/blob/main/pkg/controller/provider/container/vsphere/collector_test.go#L29-L36)?

pkg/controller/plan/adapter/ova/builder.go
```diff
@@ -440,6 +440,12 @@ func (r *Builder) Tasks(vmRef ref.Ref) (list []*plan.Task, err error) {
 	return
 }
 
+func (r *Builder) PreferenceName(vmRef ref.Ref, configMap *core.ConfigMap) (name string, err error) {
+	// The VM is built from configuration, we don't need the preference
```
ahadas, line 444:
```suggestion
	// We currently set the operating systems for VMs from OVA to UNKNOWN so we cannot get the corresponding preference
```

pkg/controller/plan/adapter/ova/builder.go
ahadas: suggested a change to the comment

pkg/controller/plan/adapter/ovirt/builder.go
ahadas: ok, [filed an issue to fix it separately](https://github.com/kubev2v/forklift/issues/741)

pkg/controller/plan/adapter/ovirt/builder.go
```diff
@@ -559,6 +559,30 @@ func (r *Builder) Tasks(vmRef ref.Ref) (list []*plan.Task, err error) {
 	return
 }
 
+func (r *Builder) PreferenceName(vmRef ref.Ref, configMap *core.ConfigMap) (name string, err error) {
+	if configMap == nil {
+		err = liberr.New("can't map the OS, missing config map")
+		return
+	}
```
ahadas, line 566: forgot to remove this check?

pkg/controller/plan/kubevirt.go
```diff
@@ -1166,6 +1181,97 @@ func (r *KubeVirt) virtualMachine(vm *plan.VMStatus) (object *cnv.VirtualMachine
...
+	case api.VSphere:
+		configMapName = vsphereOsConfigMap
+	case api.OVirt:
+		configMapName = ovirtOsConfigMap
+	default:
+		return
+	}
+	err = r.Client.Get(
+		context.TODO(),
+		client.ObjectKey{Name: configMapName},
```
liranr23, line 1218: now the problem with get is to know the correct namespace :)

0xFelix: If the default ServiceAccount is present in the Pod, you can get the namespace the Pod is running in by reading `/var/run/secrets/kubernetes.io/serviceaccount/namespace`. This necessitates that the ConfigMap is located in the same namespace as the Pod however.

pkg/controller/plan/adapter/openstack/builder_test.go
0xFelix: +1, also this is using plain testing without Ginkgo/Gomega atm.

pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -695,6 +695,32 @@ func (r *Builder) Tasks(vmRef ref.Ref) (list []*plan.Task, err error) {
...
 }
 
+func (r *Builder) PreferenceName(vmRef ref.Ref, configMap *core.ConfigMap) (name string, err error) {
+	if configMap == nil {
+		err = liberr.New("can't map the OS, missing config map")
+		return
+	}
+	vm := &model.Workload{}
+	err = r.Source.Inventory.Find(vm, vmRef)
+	if err != nil {
```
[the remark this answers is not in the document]
liranr23: what was the suggestion? it tried to get the VM from the inventory container in order to use the OS set there and map it.

pkg/controller/plan/kubevirt.go
liranr23: we have envVar holding it within the pod, i used this approach.

pkg/controller/plan/adapter/vsphere/builder.go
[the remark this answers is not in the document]
liranr23: most likely forgotten, thanks. done now.

pkg/controller/plan/adapter/openstack/builder_test.go
```diff
@@ -0,0 +1,20 @@
...
+
+import (
+	. "github.com/onsi/ginkgo"
+	"github.com/onsi/ginkgo/extensions/table"
+	. "github.com/onsi/gomega"
+	gtypes "github.com/onsi/gomega/types"
+)
+
+var _ = Describe("vSphere builder", func() {
+	table.DescribeTable("should", func(os, version, distro string, matchPreferenceName gtypes.GomegaMatcher) {
```
ahadas, line 11: we talked offline about replacing matchPreferenceName with a string argument

pkg/controller/plan/kubevirt.go
```diff
@@ -85,6 +87,12 @@ const (
 	qemuGroup = int64(107)
 )
 
+// Preference configmaps names
+const (
+	vsphereOsConfigMap = "forklift-vsphere-osmap"
+	ovirtOsConfigMap   = "forklift-ovirt-osmap"
```
ahadas, line 93: how about getting these values from the operator using environment variables?

ahadas (commented):
two minor comments inside, overall it looks good
about making the built-in mappings configurable - we can follow-up on this, I do want to include this change in 2.6 even without that part

pkg/controller/plan/kubevirt.go
liranr23: i think this will resolve the build-in mappings to be configurable :)

operator/roles/forkliftcontroller/templates/controller/deployment-controller.yml.j2
```diff
@@ -92,6 +92,14 @@ spec:
         - name: FEATURE_OVIRT_WARM_MIGRATION
           value: "true"
 {% endif %}
+{% if ovirt_osmap_configmap_name %}
```
ahadas, line 95: don't you need `is defined` here?

ahadas (approved):

pkg/controller/plan/adapter/openstack/builder_test.go
```diff
@@ -0,0 +1,19 @@
+package openstack
+
+import (
+	. "github.com/onsi/ginkgo"
+	"github.com/onsi/ginkgo/extensions/table"
+	. "github.com/onsi/gomega"
```
0xFelix, line 6: Looks like you are using ginkgo v1 and v2 in this repo? When starting from scratch I would consider using v2.

0xFelix (commented):
One last comment :)

0xFelix (approved):
/lgtm

liranr23:
> /lgtm

@0xFelix Thank you for your reviews!

pkg/controller/plan/adapter/openstack/builder_test.go
ahadas: yeah, we recently added v2 without removing v1 - we should do that (separately)

PR #718, opened 2024-01-23
OVA: Consume firmware settings from the virt-v2v produced XML

bkhizgiy:
After the conversion process is completed, this pull request introduces functionality to extract the firmware variable from the OVF file produced by virt-v2v. This is accomplished by setting up an HTTP server that sends this configuration to be used by the VM builder


ahadas (commented):
there will be conflicts with my changes in the entrypoint script but this direction looks good

pkg/controller/plan/adapter/ova/builder.go
```diff
@@ -349,12 +354,26 @@ func (r *Builder) mapCPU(vm *model.VM, object *cnv.VirtualMachineSpec) {
...
 
 func (r *Builder) mapFirmware(vm *model.VM, object *cnv.VirtualMachineSpec) {
+	migrationAnnotations := r.Migration.GetAnnotations()
+	var vmFirmware string
+	if virtV2VFirmware, ok := migrationAnnotations[vm.ID]; ok {
+		err := json.Unmarshal([]byte(virtV2VFirmware), &vmFirmware)
+		if err != nil {
+			err = liberr.Wrap(
+				err,
+				"Failed to unmarshel VM virt-v2v fimware config",
```
liranr23, line 364:
```suggestion
				"Failed to unmarshel VM virt-v2v firmware config",
```

pkg/controller/plan/migration.go
```diff
@@ -1726,6 +1733,49 @@ func (r *Migration) setPopulatorPodsWithLabels(vm *plan.VMStatus, migrationID st
 	}
 }
 
+func (r *Migration) getVirtV2VDiskConfig(vm *plan.VMStatus, step *plan.Step) (err error) {
```
liranr23, line 1736: maybe there is a better placement to this function, the update of the migration part might fit kubevirt.go

virt-v2v/cold/entrypoint.go
```diff
@@ -0,0 +1,145 @@
...
+	OVA     = "ova"
+	vSphere = "vSphere"
+	DIR     = "/var/tmp/v2v"
+	FS      = "/mnt/disks/disk[0-9]*"
+	Block   = "/dev/block[0-9]*"
+	VDDK    = "/opt/vmware-vix-disklib-distrib"
+)
+
+func main() {
+	fmt.Println("Bella")
```
liranr23, line 22: :)

virt-v2v/cold/entrypoint.go
```diff
@@ -0,0 +1,145 @@
...
+	for _, v := range envVars {
+		if os.Getenv(v) == "" {
+			return false
+		}
+	}
+	return true
+}
+
+func genName(diskNum int) string {
+
```
liranr23, line 94: redundant line

virt-v2v/cold/entrypoint.go
```diff
@@ -0,0 +1,145 @@
...
+	requiredEnvVars := map[string][]string{
+		vSphere: {"V2V_libvirtURL", "V2V_secretKey", "V2V_vmName"},
+		OVA:     {"V2V_diskPath", "V2V_vmName"},
+	}
+
+	if envVars, ok := requiredEnvVars[source]; ok {
+		if !checkEnvVariablesSet(envVars...) {
+			fmt.Printf("Following environment variables need to be defined: %v\n", envVars)
+			//klog.Errorf("Following environment variables need to be defined: %v\n", envVars)
```
liranr23, line 34: klog.errorf? fmt.Printf? in the whole file. i think logger with error is a better fit

cmd/virt-v2v-monitor/virt-v2v-monitor.go
```diff
@@ -65,6 +157,9 @@ func main() {
 	// Start prometheus metrics HTTP handler
 	klog.Info("Setting up prometheus endpoint :2112/metrics")
 	http.Handle("/metrics", promhttp.Handler())
+	klog.Info("Setting up prometheus endpoint :2112/vm_config")
```
liranr23, line 160: i think it's not a prometheus endpoint

cmd/virt-v2v-monitor/virt-v2v-monitor.go
```diff
@@ -108,6 +204,15 @@ func main() {
...
 			err = updateProgress(progressCounter, diskNumber, progress)
+		} else if strings.Contains(string(line), XML_CONF) {
+			klog.Infof("Strating reading xml configuratoin")
+			inXML = true
+		} else if inXML {
+			xmlData.WriteString(string(line) + "\n")
+			if strings.Contains(string(line), XML_CONF_END) {
+				klog.Infof("Finished reading xml configuratoin")
+				inXML = false
+			}
```
liranr23, line 215: i find this part strange. we need to iterate this twice? once to open the XML and once to end it?

liranr23 (commented):
what's the whole benefit of serving the vm config in order to get its firmware? isn't there a better option? or we intend to serve the whole vm config XML and firmware is just a start?

pkg/controller/plan/migration.go
```diff
@@ -1726,6 +1733,49 @@ func (r *Migration) setPopulatorPodsWithLabels(vm *plan.VMStatus, migrationID st
...
+	pod, err := r.kubevirt.GetGuestConversionPod(vm)
+	switch {
+	case err != nil:
+		return liberr.Wrap(err)
+	case pod == nil:
+		step.MarkCompleted()
+		step.AddError("Guest conversion pod not found")
```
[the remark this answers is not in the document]
bkhizgiy: I took this logic from other steps, but i"ll double check its correct.

virt-v2v/cold/entrypoint.go
bkhizgiy: I want originally with klog but currently this dependency not working since virt-v2v set in a different workspace then the rest off the project, once i"ll be able to solve these issues i"ll update the logger as well.

cmd/virt-v2v-monitor/virt-v2v-monitor.go
bkhizgiy: nope, since we using virt-v2v with `-v -x` parameters set (for logging and debugging) the xml file generated is not being separately created but rather printed to the logs. so we need somehow read it and save it to a variable, since the metrics code already doing this for the progress I've decided to use it for reading this configuration as well to reduce code dups. so its basically checking for the xml header and once it detected writes all the xml lines into var until its encounters the last line of the xml where its always the same.

cmd/virt-v2v-monitor/virt-v2v-monitor.go
bkhizgiy: yeah I was thinking if created a new endpoint for that or using the existing one even though its not in the category

cmd/virt-v2v-monitor/virt-v2v-monitor.go
liranr23: wouldn't it be nicer, if possible, to make virt-v2v print it to a file, then just read the file?

cmd/virt-v2v-monitor/virt-v2v-monitor.go
```diff
@@ -57,6 +120,35 @@ func updateProgress(progressCounter *prometheus.CounterVec, disk, progress uint6
...
+	path := xmlConf.OS.Loader.Path
+	var firmware string
+	if strings.Contains(path, "OVMF") {
+		firmware = "uefi"
+	} else {
+		firmware = "bios"
+	}
+
+	if firmware != "" {
+		jsonData, err := json.Marshal(firmware)
```
bennyz, line 140: I think we can skip this, it's a simple string

cmd/virt-v2v-monitor/virt-v2v-monitor.go
```diff
@@ -57,6 +120,35 @@ func updateProgress(progressCounter *prometheus.CounterVec, disk, progress uint6
 	return
 }
 
+func xmlHandler(w http.ResponseWriter, r *http.Request) {
+	var xmlConf Domain
+	err := xml.Unmarshal([]byte(xmlData.String()), &xmlConf)
```
bennyz, line 125: I don't know if it's worth it, but since we only need one field, it may make sense to avoid loading the entire XML into the memory and instead iterate over it using `xml.Decoder` and get it directly

virt-v2v/cold/entrypoint.go
```diff
@@ -1,37 +1,92 @@
...
+	requestProcessed = make(chan bool)
 )
 
 func main() {
 	fmt.Println("Bella")
-	virtV2vArgs := []string{"virt-v2v", "-v", "-x"}
-	source := os.Getenv("V2V_source")
+	ovaPath := ovaDir + "/vm.ova"
+
+	if err := os.MkdirAll(ovaDir, 0755); err != nil {
```
bkhizgiy, line 68: @bennyz starting here

virt-v2v/cold/entrypoint.go
```diff
@@ -1,37 +1,92 @@
...
+		fmt.Printf("Failed to create directory /tmp/ova: %v\n", err)
+		return
+	}
+
+	if err := downloadOVA(ovaFileURL, ovaDir); err != nil {
+		fmt.Printf("Failed to download OVA file: %v\n", err)
+		return
```
bkhizgiy, line 75: @bennyz ends here

virt-v2v/cold/entrypoint.go
```diff
@@ -0,0 +1,274 @@
...
+	virtV2vCmd.Stderr = os.Stderr
+	if err = virtV2vCmd.Run(); err != nil {
+		fmt.Printf("Error executing command: %v\n", err)
+		return
+	}
+	return
+}
+
+func waitForXMLFile(dir, fileExtension string, timeoutMinutes, checkIntervalSeconds int) (string, error) {
+	deadline := time.Now().Add(time.Duration(timeoutMinutes) * time.Minute)
```
liranr23, line 208: on which step (area) is it generated by virt-v2v? are we sure 5 minutes is enough/too much?

virt-v2v/cold/entrypoint.go
```diff
@@ -0,0 +1,274 @@
...
+		if len(files) > 0 {
+			return files[0], nil
+		}
+		time.Sleep(time.Duration(checkIntervalSeconds) * time.Second)
+	}
+	return "", fmt.Errorf("timeout reached without finding an XML file")
+}
+
+func firmwareHandler(w http.ResponseWriter, r *http.Request) {
+
```
liranr23, line 223: redundant line

pkg/controller/plan/kubevirt.go
```diff
@@ -857,6 +859,48 @@ func (r *KubeVirt) GetGuestConversionPod(vm *plan.VMStatus) (pod *core.Pod, err
 	return
 }
 
+func (r *KubeVirt) GetVirtV2VDiskConfig(vm *plan.VMStatus, pod *core.Pod, step *plan.Step) (err error) {
+
```
liranr23, line 863: redundant line

virt-v2v/cold/entrypoint.go
```diff
@@ -0,0 +1,274 @@
...
+	for _, v := range envVars {
+		if os.Getenv(v) == "" {
+			return false
+		}
+	}
+	return true
+}
+
+func genName(diskNum int) string {
+
```
liranr23, line 158: redundant line

virt-v2v/cold/BUILD.bazel
```diff
@@ -89,15 +90,29 @@ container_image(
...
+)
+
 container_image(
     name = "forklift-virt-v2v",
     base = ":virt-v2v-layer",
     directory = "/usr/local/bin/",
     empty_dirs = ["/disks"],
-    entrypoint = ["/usr/local/bin/entrypoint"],
-    env = {"LIBGUESTFS_BACKEND": "direct"},
+    entrypoint = ["/usr/local/bin/cold"],
```
liranr23, line 111: i didn't line it's being called `cold`, here in the file, but that's only a matter of taste

liranr23 (commented):
> what's the whole benefit of serving the vm config in order to get its firmware? isn't there a better option? or we intend to serve the whole vm config XML and firmware is just a start?

same question

virt-v2v/cold/entrypoint.go
```diff
@@ -0,0 +1,274 @@
...
+		os.Exit(1)
+	}
+
+	if err := executeVirtV2v(virtV2vArgs); err != nil {
+		os.Exit(1)
+	}
+
+	if source == OVA {
+		var err error
+		xmlFilePath, err = waitForXMLFile(DIR, "xml", 5, 60)
```
bennyz, line 119: as discussed, if the XML is expected to be present when virt-v2v finishes there is no need to wait

pkg/controller/plan/adapter/ova/builder.go
```diff
@@ -31,6 +31,10 @@ const (
 	BIOS = "bios"
 )
 
+const (
+	UEFI = "uefi"
+)
```
ahadas, line 36: this should be placed next to BIOS above - you can replace `BIOS types` with `Firmware types`

pkg/controller/plan/adapter/ova/builder.go
```diff
@@ -349,19 +353,25 @@ func (r *Builder) mapCPU(vm *model.VM, object *cnv.VirtualMachineSpec) {
...
 		Serial: vm.UUID,
 	}
-	switch vm.Firmware {
+
+	secureBootEnabled := false
+	switch virtV2VFirmware {
 	case BIOS:
 		firmware.Bootloader = &cnv.Bootloader{BIOS: &cnv.BIOS{}}
+	case UEFI:
+		firmware.Bootloader = &cnv.Bootloader{EFI: &cnv.EFI{SecureBoot: &secureBootEnabled}}
```
ahadas, line 368: it's exactly the same as we set UEFI below, so why repeating it?..

pkg/controller/plan/adapter/ova/builder.go
```diff
@@ -349,19 +353,25 @@ func (r *Builder) mapCPU(vm *model.VM, object *cnv.VirtualMachineSpec) {
...
 func (r *Builder) mapFirmware(vm *model.VM, object *cnv.VirtualMachineSpec) {
+	migrationAnnotations := r.Migration.GetAnnotations()
+	virtV2VFirmware := migrationAnnotations[vm.ID]
+
 	firmware := &cnv.Firmware{
 		Serial: vm.UUID,
 	}
-	switch vm.Firmware {
+
+	secureBootEnabled := false
```
ahadas, line 363: we can use the ptr package to avoid declaring secureBootEnabled

pkg/controller/plan/kubevirt.go
```diff
@@ -857,6 +859,48 @@ func (r *KubeVirt) GetGuestConversionPod(vm *plan.VMStatus) (pod *core.Pod, err
...
+	if pod.Status.PodIP == "" {
+		//we need the IP for fetching the fimware info.
```
ahadas, line 865: did you consider creating a service instead?
and there's a typo:
```suggestion
		//we need the IP for fetching the firmware info.
```

pkg/controller/plan/kubevirt.go
```diff
@@ -857,6 +859,48 @@ func (r *KubeVirt) GetGuestConversionPod(vm *plan.VMStatus) (pod *core.Pod, err
...
+	resp, err := http.Get(url)
+
+	if err != nil {
+		if strings.Contains(err.Error(), "connection refused") {
+			return nil
+		}
+		return
+	} else {
+		defer resp.Body.Close()
+		firmaware, errFirrmware := io.ReadAll(resp.Body)
```
ahadas, line 879: I think it makes more sense to return the entire OVF from the pod and let the caller takes what it needs, in this case the firmare

pkg/controller/plan/kubevirt.go
```diff
@@ -857,6 +859,48 @@ func (r *KubeVirt) GetGuestConversionPod(vm *plan.VMStatus) (pod *core.Pod, err
...
+
+		currentAnnotations := r.Migration.Annotations
+		if currentAnnotations == nil {
+			currentAnnotations = make(map[string]string)
+		}
+		var vmId string
+		labels := pod.GetLabels()
+		if val, ok := labels["vmID"]; ok {
+			vmId = val
+		}
```
ahadas, line 893: else?

pkg/controller/plan/kubevirt.go
```diff
@@ -857,6 +859,48 @@ func (r *KubeVirt) GetGuestConversionPod(vm *plan.VMStatus) (pod *core.Pod, err
...
+		currentAnnotations[vmId] = string(firmaware)
+		r.Migration.SetAnnotations(currentAnnotations)
+		err = r.Destination.Client.Update(context.TODO(), r.Migration.DeepCopy())
```
ahadas, line 896: `patch` seems to better fit here
but why placing it in the annotations in the first place (and not in the status)?

pkg/controller/plan/kubevirt.go
```diff
@@ -857,6 +859,48 @@ func (r *KubeVirt) GetGuestConversionPod(vm *plan.VMStatus) (pod *core.Pod, err
 	return
 }
 
+func (r *KubeVirt) GetVirtV2VDiskConfig(vm *plan.VMStatus, pod *core.Pod, step *plan.Step) (err error) {
```
ahadas, line 862: it's not really disk config, in oVirt we called it ["converted VM"](https://github.com/oVirt/ovirt-engine/blob/master/backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/vdsbroker/GetConvertedOvfVDSCommand.java)

virt-v2v/cold/BUILD.bazel
```diff
@@ -89,15 +90,29 @@ container_image(
...
 
+go_library(
+    name = "cold_lib",
+    srcs = ["entrypoint.go"],
+    importpath = "github.com/konveyor/forklift-controller/virt-v2v/cold",
+    visibility = ["//visibility:private"],
+)
+
+go_binary(
+    name = "cold",
```
ahadas, line 101: yeah, 'cold' is not that great name for this as @liranr23 pointed out below - how about virt-v2v-wrapper?

virt-v2v/cold/entrypoint.go
ahadas: yeah, the OVF should be available when the conversion is done

bkhizgiy: I've added it to avoid any possibility of condition race, but I think I just check if the command really ended rather than if the file was created.

pkg/controller/plan/kubevirt.go
bkhizgiy: I did initially, and then I saw that the monitor part is using the same mechanism and its simpler this way, not sure if its worth the addition of the service creation and then adding its query just for getting a single value.

virt-v2v/cold/entrypoint.go
```diff
@@ -0,0 +1,274 @@
...
+		if err != nil {
+			fmt.Println("Error waiting for XML file:", err)
+			os.Exit(1)
+		}
+
+		http.HandleFunc("/firmware", firmwareHandler)
+		server := &http.Server{Addr: ":8080"}
+
+		go func() {
```
bennyz, line 128: I'd have the caller invoke the shutdown rather than the pod shut down itself

bkhizgiy:
> > what's the whole benefit of serving the vm config in order to get its firmware? isn't there a better option? or we intend to serve the whole vm config XML and firmware is just a start?
> 
> same question

we currently saw some mismatch in some of the ovf files about this info, and from the produced xml we get the most accurate values after the conversion, for now we take only the firmware but it implemented in such a way that all other fields can be easily fetched if needed and I think maybe we will need to take out the disk bus info as well.

virt-v2v/cold/entrypoint.go
ahadas: if we go this way, we can also drop the call to get the OVF but take it from the status instead and later call shutdown from the controller..

pkg/controller/plan/kubevirt.go
bkhizgiy: I understand your perspective, but I'm not entirely convinced. It seems unnecessary to send the entire file when some of its content is irrelevant. In my opinion, it's more logical to parse it on the "server" side and transmit only the required fields. If we need more configuration in the future, we can always include it in the response or consider sending the full XML. However, based on our current requirements, sending everything seams, IMO redundant.

pkg/controller/plan/kubevirt.go
ahadas: that's a fair point - creating a service would complicate the flow and we already do it the same way for the progress
but then we can get the OVF after the 'conversion' step is completed - at that point the pod is definitely running and has the OVF ready

bkhizgiy: That's another point, since we currently depend on pod completion to continue to the next step in the migration process, leaving it running to retrieve additional information later would require altering this flow as well. So I think at this point its better to leave it as it is and we can think on this improvement in a later stage.

virt-v2v/cold/entrypoint.go
bkhizgiy: @bennyz recommended adding an endpoint to shut down the server after confirming data receipt from the request, which also allows for retries if necessary. It's a minor adjustment that makes sense to ensure data fetching. We could also change this to your suggestion, but I'm not sure it's worth the effort right now since its involving changing the migration flow, As I  mentioned in other comment, shutting down the pod is required to mark the stage complete but we can always make these improvements later.

virt-v2v/cold/BUILD.bazel
bkhizgiy: changed.

virt-v2v/cold/entrypoint.go
bkhizgiy: I drooped this entirely, there another comment about that below.

pkg/controller/plan/adapter/ova/builder.go
bkhizgiy: //TODO

pkg/controller/plan/kubevirt.go
bkhizgiy: //TODO

virt-v2v/cold/entrypoint.go
bkhizgiy: I've added the change of another endpoint and shutting down from the client side.

bkhizgiy:
I"ve uploaded most of the changes, still need to address minor two comments (but I need to drop for a while so will continue later) and address the open discussions.

liranr23:
> > > what's the whole benefit of serving the vm config in order to get its firmware? isn't there a better option? or we intend to serve the whole vm config XML and firmware is just a start?
> > 
> > 
> > same question
> 
> we currently saw some mismatch in some of the ovf files about this info, and from the produced xml we get the most accurate values after the conversion, for now we take only the firmware but it implemented in such a way that all other fields can be easily fetched if needed and I think maybe we will need to take out the disk bus info as well.

OK, I see. So we are saying - the OVF may be corrupted, while the point of truth is virt-v2v as virt-v2v look within the disk guest and gather the info.
As you wrote in the future we may gather more data. Would we need to add a new endpoint for each value we are looking for? Would it be better to just have one single endpoint such as "xml", and get the whole xml to the controller, there we can get any detail we want?

pkg/controller/plan/kubevirt.go
liranr23: i commented below as a general comment and then i saw this.
it's true that for now only `firmware` and having single value sent by the server is enough - because that's the change and we don't need the whole XML.
but, if you thinking about the future, in case we need more fields, i think sending the whole XML and parsing it within the controller is better. then, instead of changing the server to add more single fields, parsing, see how you send it to the controller - you won't need to.
you will need only an additional parse on controller side.

bkhizgiy: ack, I'm getting your view on that, so i"ll change that to match your suggestion.

bennyz: we don't have to add an endpoint per attribute, we can make one that's generic and receives it as an arguments returning the value

ahadas: ideally, we should have taken the OVF that is produced by virt-v2v and adjust it according to network+storage mappings, vm name, etc. we could have also let virt-v2v to do that and even post the VM eventually (that was discussed with rjones long ago, but it's problematic in terms of permissions the pod needs to have and it breaks the abstraction we have in the migration flow..)
from that perspective, it's not like we should take "just a single or handful of properties" but the entire configuration. practically, we just need the firmware (and in oVirt [we also updated the display type](https://github.com/oVirt/ovirt-engine/blob/master/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/exportimport/UpdateConvertedVmCommand.java#L113-L116) which is irrelevant for forklift). yet, I'd just return the entire OVF - the conversion pod already does quite a bit of work, no need to have the parsing of the OVF there as well...

pkg/controller/plan/kubevirt.go
ahadas: @bkhizgiy  did you check whether this call to `Update` triggers a reconcile of the migration?

bkhizgiy: nope, I just saw it was correctly added, but now that I've moved this change to the vmStatus inside the migration I cant get it to update correctly, its not being saved.

pkg/controller/plan/kubevirt.go
bkhizgiy: the code changed so now i'n sending the xml file and getting the relevant data from the controller.

pkg/controller/plan/kubevirt.go
bkhizgiy: I'm not sure why but the status isnt updated in the CR, tried from both the migration and plan objects. for now I returned it to annotation, and once I"ll finish with the ova validation i"ll fix it.

I was thinking of adding a new map under the VM status, and there we can save all the needed config for this vm, so we wont need to have map of the vm id and the resource.

pkg/controller/plan/kubevirt.go
```diff
@@ -859,6 +891,105 @@ func (r *KubeVirt) GetGuestConversionPod(vm *plan.VMStatus) (pod *core.Pod, err
...
+	xmlConf, err := readConfFromXML(xmlConfig)
+	if err != nil {
+		return
+	}
+
+	path := xmlConf.OS.Loader.Path
+	if strings.Contains(path, "OVMF") {
+		return "uefi", nil
+	}
+	return "bios", nil
```
liranr23, line 990: you can use the consts of firmware

pkg/controller/plan/kubevirt.go
```diff
@@ -859,6 +891,105 @@ func (r *KubeVirt) GetGuestConversionPod(vm *plan.VMStatus) (pod *core.Pod, err
...
+	var domain Domain
+
+	reader := strings.NewReader(xmlData)
+	decoder := xml.NewDecoder(reader)
+
+	err := decoder.Decode(&domain)
+	if err != nil {
+		return &domain, err
+	}
+	fmt.Printf("this is domain %v", domain)
```
liranr23, line 976: forgotten? or it should be logged? if log, i would expect the usage of `r.Log` and more details (vm and so on)

pkg/controller/plan/kubevirt.go
```diff
@@ -859,6 +891,105 @@ func (r *KubeVirt) GetGuestConversionPod(vm *plan.VMStatus) (pod *core.Pod, err
...
+	currentAnnotations := r.Migration.Annotations
+	if currentAnnotations == nil {
+		currentAnnotations = make(map[string]string)
+	}
+	var vmId string
+	labels := pod.GetLabels()
+	if val, ok := labels["vmID"]; ok {
+		vmId = val
+	} else {
+		return errors.New("failed to get virt-v2v pod labels")
```
liranr23, line 946: liberr.New?

pkg/controller/plan/kubevirt.go
```diff
@@ -859,6 +891,105 @@ func (r *KubeVirt) GetGuestConversionPod(vm *plan.VMStatus) (pod *core.Pod, err
...
+		defer resp.Body.Close()
+	} else {
+		if strings.Contains(err.Error(), "connection refused") {
+			return nil
+		} else {
+			for i := 0; i < maxRetries; i++ {
+				resp, err = http.Get(url)
+				if err != nil {
+					if i < maxRetries-1 {
+						fmt.Printf("Retrying... (%d/%d)\n", i+1, maxRetries)
```
liranr23, line 915: r.Log

pkg/controller/plan/kubevirt.go
```diff
@@ -97,6 +100,35 @@ type KubeVirt struct {
 	Builder adapter.Builder
 }
 
+type Domain struct {
```
liranr23, line 103: since it's only OVA, i would prefer it to be somewhere in its adapter but let's see what others think

ahadas (commented):
besides the comments inside, I think it makes sense to extract the introduction of virt-v2v-wrapper in go to a separate PR that can get in quickly

pkg/controller/plan/kubevirt.go
ahadas: > I was thinking of adding a new map under the VM status, and there we can save all the needed config for this vm, so we wont need to have map of the vm id and the resource.

yeah, that would be a better approach
we don't even need a new map, we can have something like:
``` yaml
status:
  ...
  vms:
  - id: ...
    name: ...
    ovf: <ovf>
```

pkg/controller/plan/kubevirt.go
```diff
@@ -859,6 +891,105 @@ func (r *KubeVirt) GetGuestConversionPod(vm *plan.VMStatus) (pod *core.Pod, err
...
 }
 
+func (r *KubeVirt) GetVirtV2VConvertedVMConfig(vm *plan.VMStatus, pod *core.Pod, step *plan.Step) (err error) {
+	if pod.Status.PodIP == "" {
+		//we need the IP for fetching the firmware info.
+		return
+	}
+
+	url := fmt.Sprintf("http://%s:8080/xml", pod.Status.PodIP)
+	maxRetries := 3
```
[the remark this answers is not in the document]
bkhizgiy: we don't, its just for a case we will have some issue getting the xml file from the server, we wont imiditly fail, but try again. From all the runs I did its alleyways get it in the first try.

virt-v2v/cold/entrypoint.go
```diff
@@ -0,0 +1,207 @@
...
+
+	requiredEnvVars := map[string][]string{
+		vSphere: {"V2V_libvirtURL", "V2V_secretKey", "V2V_vmName"},
+		OVA:     {"V2V_diskPath", "V2V_vmName"},
+	}
+
+	if envVars, ok := requiredEnvVars[source]; ok {
+		if !checkEnvVariablesSet(envVars...) {
+			fmt.Printf("Following environment variables need to be defined: %v\n", envVars)
+			os.Exit(1)
```
ahadas, line 40: how about using `klog.Fatal` here instead?

virt-v2v/cold/entrypoint.go
```diff
@@ -0,0 +1,207 @@
...
+		}
+	}
+
+	switch source {
+	case vSphere:
+		virtV2vArgs = append(virtV2vArgs, "-i", "libvirt", "-ic", os.Getenv("V2V_libvirtURL"))
+		pwFilePath := fmt.Sprintf("%s/vmware.pw", DIR)
```
ahadas, line 47: the entrypoint script [created the DIR folder](https://github.com/kubev2v/forklift/blob/main/virt-v2v/cold/entrypoint#L44), don't we need to do that here?

virt-v2v/cold/entrypoint.go
```diff
@@ -0,0 +1,207 @@
...
+		if err := os.WriteFile(pwFilePath, []byte(os.Getenv("V2V_secretKey")), 0600); err != nil {
```
ahadas, line 48: I see that you create the DIR folder below, how come that this call doesn't fail then?..

virt-v2v/cold/entrypoint.go
```diff
@@ -0,0 +1,207 @@
...
+		virtV2vArgs = append(virtV2vArgs, "-ip", pwFilePath)
+
+		if _, err := os.Stat(VDDK); err == nil {
+			virtV2vArgs = append(virtV2vArgs,
+				"-it", "vddk",
+				"-io", fmt.Sprintf("vddk-libdir=%s", VDDK),
+				"-io", fmt.Sprintf("vddk-thumbprint=%s", os.Getenv("V2V_thumbprint")),
+			)
+		}
+		virtV2vArgs = append(virtV2vArgs, "--", os.Getenv("V2V_vmName"))
```
ahadas, line 61: I thought this should be the last argument for the command...

virt-v2v/cold/entrypoint.go
```diff
@@ -0,0 +1,207 @@
...
+	return
+}
+
+func executeVirtV2v(args []string) (err error) {
+	virtV2vCmd := exec.Command("bash", "-c", strings.Join(args, " "), "|& /usr/local/bin/virt-v2v-monitor")
+	virtV2vCmd.Stdout = os.Stdout
+	virtV2vCmd.Stderr = os.Stderr
+	if err = virtV2vCmd.Run(); err != nil {
+		fmt.Printf("Error executing command: %v\n", err)
+		return
```
ahadas, line 160: this `return` is redundant

virt-v2v/cold/entrypoint.go
```diff
@@ -0,0 +1,207 @@
...
+
+	xmlData, err := os.ReadFile(xmlFilePath)
+	if err != nil {
+		fmt.Printf("Error reading XML file: %v\n", err)
+		http.Error(w, "Error reading XML file", http.StatusInternalServerError)
+		return
+	}
+
+	w.Header().Set("Content-Type", "application/xml")
+	w.WriteHeader(http.StatusOK)
```
ahadas, line 191: wouldn't it make more sense to write xmlData and then write StatusOK if err == nil?

pkg/controller/plan/kubevirt.go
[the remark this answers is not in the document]
ahadas: I see, then it sounds like an unnecessary complication - let's drop it

ahadas:
the failure in vsphere tests is unrelated (strangely, it's probably related to the VDDK validation - maybe we should wait for more than one minute..) but the end-to-end tests for vSphere don't check virt-v2v

virt-v2v/cold/entrypoint.go
bkhizgiy: klog isnt working here because virt-v2v is set in a different  workspace so this dependencies are not recognized here, that the reason I went with simple print. We consider extracting this part to the main workspace like we do for the monitor but we agreed its to big of a change all at once, so we can do it later.

pkg/controller/plan/kubevirt.go
```diff
@@ -859,6 +895,104 @@ func (r *KubeVirt) GetGuestConversionPod(vm *plan.VMStatus) (pod *core.Pod, err
...
+		defer resp.Body.Close()
+	} else {
+		if strings.Contains(err.Error(), "connection refused") {
+			return nil
+		} else {
+			for i := 0; i < maxRetries; i++ {
+				resp, err = http.Get(url)
+				if err != nil {
+					if i < maxRetries-1 {
+						r.Log.Info("Retrying to get virt-v2v vm configuration after conversion... (%d/%d), vmID:%v\n", i+1, maxRetries, vm.ID)
```
liranr23, line 919:
```suggestion
						r.Log.Info("Retrying to get virt-v2v vm configuration after conversion... (%d/%d)", i+1, maxRetries, "vm", vm.String())
```

virt-v2v/cold/entrypoint.go
bkhizgiy: I checked again, I see this arg comes at the end of the flow: https://github.com/kubev2v/forklift/blob/0e672a487e399e2ca96790e75afa0c06898ab93d/virt-v2v/cold/entrypoint#L94
am I missing somthing?

virt-v2v/cold/entrypoint.go
bkhizgiy: I just noticed it override the secure changes, I removed this part and changed to the newer one (this is the older implementation)

virt-v2v/cold/entrypoint.go
ahadas: ok

virt-v2v/cold/entrypoint.go
ahadas: ah right, I also missed that part

virt-v2v/cold/entrypoint.go
ahadas: @bkhizgiy right but later we run `virtV2vArgs = append(virtV2vArgs, "-o", "local", "-os", DIR)`

bkhizgiy: @ahadas yeah the order here is reversed, I'm testing now the changes with live migration and uploading the final version.

pkg/controller/plan/migration.go
```diff
@@ -1111,6 +1111,20 @@ func (r *Migration) execute(vm *plan.VMStatus) (err error) {
 		if err != nil {
 			return
 		}
+		// only for OVA, fetch config from the conversion pod
+		if r.Source.Provider.Type() == v1beta1.Ova {
```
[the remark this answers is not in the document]
bkhizgiy: we cant move it to a different stage, because we relying on the pod completion to finish with that stage (we don't have other indication) .

pkg/controller/plan/kubevirt.go
```diff
@@ -850,6 +852,40 @@ func (r *KubeVirt) GetGuestConversionPod(vm *plan.VMStatus) (pod *core.Pod, err
...
+	if pod.Status.PodIP == "" {
+		//we need the IP for fetching the configuration of the convered VM.
+		return
+	}
+
+	url := fmt.Sprintf("http://%s:8080/xml", pod.Status.PodIP)
+
+	resp, err := http.Get(url)
+	if err != nil {
+		if strings.Contains(err.Error(), "connection refused") {
```
bennyz, line 865: I believe you can check if err is ErrUnexpectedEOF directly or something similar

operator/config/crd/bases/forklift.konveyor.io_migrations.yaml
```diff
@@ -293,6 +293,9 @@ spec:
                       description: The VM Namespace Only relevant for an openshift
                         source.
                       type: string
+                    ovfConfig:
```
ahadas, line 296: @bennyz is less happy after seeing this change :)
there are two similar things:
1. returning the OVF from the conversion pod - there are arguments in favor or and opposing to returning the entire OVF (already discussed as part of this PR). I'm in favor of returning the entire OVF as this PR currently does
2. storing the OVF in the status - I'm ok with storing only the firmware here, and possibly add more fields from the OVF when needed, as @bennyz suggested

pkg/controller/plan/migration.go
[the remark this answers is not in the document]
ahadas: I meant to replace this with:
```suggestion
		if vm.Phase == CopyDisksVirtV2V && r.Source.Provider.Type() == v1beta1.Ova {
```
because I think it may fail when we call it during ConvertGuest but if you never saw it failing..

and what about making sure we don't trigger this unnecessarily by checking if the firmware is already set?

virt-v2v/cold/entrypoint.go
```diff
@@ -94,6 +101,25 @@ func main() {
...
+
+	if source == OVA {
+		var err error
+		xmlFilePath, err = getXMLFile(DIR, "xml")
+		if err != nil {
+			fmt.Println("Error gettin XML file:", err)
+			os.Exit(1)
+		}
+
+		http.HandleFunc("/xml", xmlHandler)
```
ahadas, line 113:
```suggestion
		http.HandleFunc("/ovf", xmlHandler)
```

virt-v2v/cold/entrypoint.go
```diff
@@ -177,6 +203,50 @@ func executeVirtV2v(args []string) (err error) {
...
+	if err != nil {
+		return "", err
+	}
+	if len(files) > 0 {
+		return files[0], nil
+	}
+	return "", fmt.Errorf("XML file was not found.")
+}
+
+func xmlHandler(w http.ResponseWriter, r *http.Request) {
```
ahadas, line 217:
```suggestion
func ovfHandler(w http.ResponseWriter, r *http.Request) {
```

pkg/controller/plan/kubevirt.go
```diff
@@ -850,6 +852,40 @@ func (r *KubeVirt) GetGuestConversionPod(vm *plan.VMStatus) (pod *core.Pod, err
 	return
 }
 
+func (r *KubeVirt) GetVirtV2VConvertedVMConfig(vm *plan.VMStatus, pod *core.Pod, step *plan.Step) (err error) {
+	if pod.Status.PodIP == "" {
+		//we need the IP for fetching the configuration of the convered VM.
+		return
+	}
+
+	url := fmt.Sprintf("http://%s:8080/xml", pod.Status.PodIP)
```
ahadas, line 861:
```suggestion
	url := fmt.Sprintf("http://%s:8080/ovf", pod.Status.PodIP)
```

operator/config/crd/bases/forklift.konveyor.io_migrations.yaml
bkhizgiy: changed, now we save only the firmware on the controller side.

operator/config/crd/bases/forklift.konveyor.io_migrations.yaml
```diff
@@ -432,6 +432,9 @@ spec:
                     type:
                       description: Type used to qualify the name.
                       type: string
+                    vmFimware:
```
ahadas, line 435: there's a typo there (Fimware -> Firmware), I think just `firmware` would be alright

pkg/controller/plan/kubevirt.go
bkhizgiy: Its not catching it as this error not as `ErrUnexpectedEOF` and not ad ``EOF`, leaving it as is.

operator/config/crd/bases/forklift.konveyor.io_migrations.yaml
```diff
@@ -432,6 +432,9 @@ spec:
...
+                      description: virt-v2v ovf file
```
ahadas, line 436: need to update the description, here we can say that this is the firmware that was detected by virt-v2v while converting the guest

pkg/controller/plan/adapter/ova/builder.go
```diff
@@ -332,10 +334,22 @@ func (r *Builder) mapCPU(vm *model.VM, object *cnv.VirtualMachineSpec) {
 }
 
 func (r *Builder) mapFirmware(vm *model.VM, object *cnv.VirtualMachineSpec) {
+	var virtV2VFirmware string
+	if vm.Firmware == "" {
+		for _, vmConf := range r.Migration.Status.VMs {
+			if vmConf.ID == vm.ID {
+				virtV2VFirmware = vmConf.VmFimware
```
ahadas, line 341: we can also `break` here

pkg/controller/plan/kubevirt.go
```diff
@@ -850,6 +854,57 @@ func (r *KubeVirt) GetGuestConversionPod(vm *plan.VMStatus) (pod *core.Pod, err
...
+	}
+
+	url := fmt.Sprintf("http://%s:8080/ovf", pod.Status.PodIP)
+
+	/* Due to the virt-v2v operation, the ovf file is only available after the command's execution,
+	meaning it appears following the copydisks phase.
+	The server will be accessible via virt-v2v only after the command has finished.
+	Until then, attempts to connect will result in a 'connection refused' error.
+	Once the VM server is running, we can make a single call to obtain the OVF configuration,
+	followed by a shutdown request. This will complete the pod process, allowing us to move to the next phase.
```
ahadas, line 870: thanks!

pkg/controller/plan/kubevirt.go
```diff
@@ -850,6 +854,57 @@ func (r *KubeVirt) GetGuestConversionPod(vm *plan.VMStatus) (pod *core.Pod, err
...
+	*/
+	resp, err := http.Get(url)
+	if err != nil {
+		if err == syscall.ECONNREFUSED {
+			return nil
+		} else {
+			return
+		}
```
ahadas, line 878: how about
```golang
                if err == syscall.ECONNREFUSED {
			err = nil
		}
		return
```

pkg/controller/plan/kubevirt.go
bennyz: I'm not sure why I said EOF here, I think I meant to comment this on L#882, this is connection refused

here this should work: https://go.dev/play/p/CvgfnU8QdQ5

bkhizgiy: haha ok, this one I've changed already:) I thought you were talking about the error at the end of the func.

pkg/controller/plan/kubevirt.go
```diff
@@ -850,6 +854,57 @@ func (r *KubeVirt) GetGuestConversionPod(vm *plan.VMStatus) (pod *core.Pod, err
...
+	shutdownURL := fmt.Sprintf("http://%s:8080/shutdown", pod.Status.PodIP)
+	resp, err = http.Post(shutdownURL, "application/json", nil)
+	if err == nil {
+		defer resp.Body.Close()
+	} else {
+		// This error indicates that the server was shut down
+		if strings.Contains(err.Error(), "EOF") {
+			return nil
+		}
+		return
```
ahadas, line 903: also here we can have it more concise:
```golang
	} else {
		// This error indicates that the server was shut down
		if strings.Contains(err.Error(), "EOF") {
			err = nil
		}
	}
```

pkg/controller/plan/kubevirt.go
```diff
@@ -858,11 +860,18 @@ func (r *KubeVirt) GetVirtV2VConvertedVMConfig(vm *plan.VMStatus, pod *core.Pod,
...
+	meaning it appears following the copydisks phase.
+	The server will be accessible via virt-v2v only after the command has finished.
+	Until then, attempts to connect will result in a 'connection refused' error.
+	Once the VM server is running, we can make a single call to obtain the OVF configuration,
+	followed by a shutdown request. This will complete the pod process, allowing us to move to the next phase.
+	*/
 	resp, err := http.Get(url)
 	if err != nil {
-		if strings.Contains(err.Error(), "connection refused") {
+		if err == syscall.ECONNREFUSED {
```
bkhizgiy, line 874: @bennyz here it was changed.

pkg/controller/plan/migration.go
```diff
@@ -1107,10 +1107,26 @@ func (r *Migration) execute(vm *plan.VMStatus) (err error) {
...
 		step.Phase = Running
+		// only for OVA, fetch config from the conversion pod
+		if r.Source.Provider.Type() == v1beta1.Ova {
+			pod, err := r.kubevirt.GetGuestConversionPod(vm)
+			if err != nil {
+				return err
+			}
+
+			if pod.Status.Phase != core.PodSucceeded {
+				err := r.kubevirt.GetVirtV2VConvertedVMConfig(vm, pod, step)
```
ahadas, line 1118: that's a bit weird to have a function called Get... that returns only an error - how about renaming it to `updateVmByConvertedConfig`?

ahadas (commented):
looks good, minor comments for polishing the code

ahadas (approved):

PR #720, opened 2024-01-24
Enable secure migrations from vSphere

ahadas:
Since 2.4 we require users to provide fingerprint/thumbprint, also for insecure providers, and secure migrations practically don't work for vSphere providers.

This main contributions in this PR are:
1. We enable secure migrations from vSphere in case the required certificate doesn't exist in the system pool: the forklift-controller and virt-v2v pods use a certificate that is specified in the provider's secret.
2. We drop the fingerprint field for (secure/insecure) vSphere providers

This PR also introduces an additional service that returns the TLS certificate for a given URL. The idea is that clients, and specifically forklift-console-plugins, would use it to simplify the certificate gathering for users (by getting the certificate over an insecure channel, and asking the user to confirm that it's the right certificate to use)

### DIFF
--- a/pkg/controller/plan/adapter/vsphere/builder.go
+++ b/pkg/controller/plan/adapter/vsphere/builder.go
@@ -1,14 +1,9 @@
 package vsphere
 
 import (
-	"bytes"
 	"context"
-	"crypto/sha1"
-	"crypto/tls"
-	"crypto/x509"
 	"errors"
 	"fmt"
-	"net/url"
 	liburl "net/url"
 	"path"
 	"regexp"
@@ -192,14 +187,7 @@ func (r *Builder) PodEnvironment(vmRef ref.Ref, sourceSecret *core.Secret) (env 
 		return
 	}
 
-	libvirtURL, err := r.getLibvirtURL(vm, sourceSecret)
-	if err != nil {
-		return
-	}
-
-	// The fingerpint/thumbprint is not confidential since one can retrieve
-	// it from the server as we do, so we don't have to place it in a secret
-	fingerprint, err := r.getFingerprint(sourceSecret)
+	libvirtURL, fingerprint, err := r.getSourceDetails(vm, sourceSecret)
 	if err != nil {
 		return
 	}
@@ -218,6 +206,8 @@ func (r *Builder) PodEnvironment(vmRef ref.Ref, sourceSecret *core.Secret) (env 
 			Name:  "V2V_source",
 			Value: "vSphere",
 		},
+		// The fingerpint/thumbprint is not confidential since one can retrieve
+		// it from the server as we do, so we don't have to place it in a secret
 		core.EnvVar{
 			Name:  "V2V_fingerprint",
 			Value: fingerprint,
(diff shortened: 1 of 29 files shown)


pkg/forklift-api/services/tls_certificate.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/konveyor/forklift-controller/pkg/controller/provider/container"
+	core "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func serveTlsCertificate(resp http.ResponseWriter, req *http.Request, client client.Client) {
+	if url := req.URL.Query().Get("URL"); url != "" {
+		log.Info("received a request to retrieve certificate", "url", url)
+		secret := &core.Secret{
+			Data: map[string][]byte{"insecureSkipVerify": []byte("true")},
```
liranr23, line 16: why you set it to true instead of checking the secret? or we get here only if cert wasn't set - then we know it's insecure?

pkg/forklift-api/webhooks/validating-webhook/admitters/secret-admitter.go
```diff
@@ -72,6 +72,10 @@ func (admitter *SecretAdmitter) Admit(ar *admissionv1.AdmissionReview) *admissio
 }
 
 func (admitter *SecretAdmitter) validateProviderSecret() *admissionv1.AdmissionResponse {
+	if _, ok := admitter.secret.Data["cacert"]; ok && container.InsecureProvider(&admitter.secret) {
+		return webhookutils.ToAdmissionResponseError(fmt.Errorf("received a request to add insecure provider with a CA certificate"))
```
liranr23, line 76: can't we just ignore the cert in this case?

pkg/forklift-api/services/tls_certificate.go
ahadas: this code is executed before we have a secret - it's supposed to return a TLS certificate to clients over an insecure channel

pkg/forklift-api/webhooks/validating-webhook/admitters/secret-admitter.go
ahadas: we can but this assumes the user wanted to get an insecure provider which is not necessarily true. if the input is invalid, we better return an error than trying to interpret what the user wanted to achieve

virt-v2v/cold/entrypoint
```diff
@@ -72,9 +72,16 @@ for disk in /dev/block[0-9]* ; do
...
+  # password
+  args+=(-ip "/etc/secret/secretKey")
+
+  if [ -e "/etc/secret/cacert" ]; then
+    # use the specified certificate
+    ln -sf /etc/secret/cacert /opt/ca-bundle.crt
+  else
+    # otherwise, keep system pool certificates
+    ln -sf /etc/pki/tls/certs/ca-bundle.crt.bak /opt/ca-bundle.crt
+  fi
```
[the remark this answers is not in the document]
ahadas: changed, note that [we're in the process of replacing this script](https://github.com/kubev2v/forklift/pull/718) so those parts are going to change anyway

liranr23 (approved):

ahadas:
changed tests to migrate from a secure vSphere provider and take the fingerprint (thumbprint) of the host when running a cold migration from an ESXi host

liranr23 (approved):

pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -285,29 +294,24 @@ func (r *Builder) Secret(vmRef ref.Ref, in, object *core.Secret) (err error) {
...
 		return
 	}
-	thumbprint := string(in.Data["thumbprint"])
 	if hostDef, found := r.hosts[hostID]; found {
-		hostSecret, nErr := r.hostSecret(hostDef)
-		if nErr != nil {
-			err = nErr
-			return
+		if in, err := r.hostSecret(hostDef); err == nil {
+			object.Data = map[string][]byte{
```
bennyz, line 299: I think this can be simplified by moving this up:
```go
object.Data = map[string][]byte{
	"accessKeyId": in.Data["user"],
	"secretKey":   in.Data["password"],
}
```

then we can only check for the error

pkg/forklift-api/services/tls_certificate.go
```diff
@@ -0,0 +1,42 @@
...
+	"encoding/pem"
+	"net/http"
+
+	"github.com/konveyor/forklift-controller/pkg/controller/provider/container"
+	core "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func serveTlsCertificate(resp http.ResponseWriter, req *http.Request, client client.Client) {
+	if url := req.URL.Query().Get("URL"); url != "" {
```
bennyz, line 13: I think if we give up on the if with a statement here we can reduce some of the nesting (not sure if sonarcloud complains)

```go

url := req.URL.Query().Get("URL")
if url == "" {
	msg := "Required parameter is missing: URL"
	if _, err := resp.Write([]byte(msg)); err != nil {
		log.Error(err, "failed to write error message", "message", msg)
	}
	resp.WriteHeader(http.StatusBadRequest)
	return
}
...
```

Also, instead of using `resp.WriteHeader` I believe we can use `http.Error` (we already use it, but not everywhere):
```go
	http.Error(resp, msg, http.StatusBadRequest)
```

pkg/controller/plan/adapter/vsphere/builder.go
ahadas: but when there's a host specification defined, we want to get the user and password from there (it's not the same `in` secret)

bennyz: ack, missed the new declaration

pkg/forklift-api/services/tls_certificate.go
ahadas: changed to use http.Error - it's nice that it doesn't return anything, that's why that code got more complicated than it was at the beginning
but it's not complex enough for sonarcloud to complain about :)

bennyz (approved):

PR #722, opened 2024-01-30
Fix creation of settingless vSphere provider

ahadas:


ahadas:
that failure of sonarcloud is definitely unrelated...

liranr23 (approved):

PR #725, opened 2024-01-31
vsphere: consider parentless folder a terminal node

ahadas:
We used to break when encontering a datacenter while creating a path of a host. That was ok in most cases since Datacenters typically reside within the root folder named "Datacenters" that should not be part of the path that is passed to virt-v2v and libvirt.

However, it is possible to create nested datacenters, which reside within a folder under the aforementioned root folder, and in this case we need to include the parent folder of the datacenter in the path of the host. Therefore, we now don't stop traveling when encountering a datacenter but rather continue and break when we encounter a folder that has an empty parent instead, without adding this folder (the root folder) to the path.

### DIFF
--- a/pkg/controller/provider/web/vsphere/BUILD.bazel
+++ b/pkg/controller/provider/web/vsphere/BUILD.bazel
@@ -1,4 +1,4 @@
-load("@io_bazel_rules_go//go:def.bzl", "go_library")
+load("@io_bazel_rules_go//go:def.bzl", "go_library", "go_test")
 
 go_library(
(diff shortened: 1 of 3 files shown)


liranr23:
can it be with multiple level nesting (e.g `"/myfolder/myfolder1/mydc/mycluster/myhost"`)?

ahadas:
> can it be with multiple level nesting (e.g `"/myfolder/myfolder1/mydc/mycluster/myhost"`)?

yes

PR #726, opened 2024-01-31
ocp: fetch cacert for provider

bennyz:
- [ ] Also add in the controller
- [ ] Other places where we create a k8s client


pkg/apis/forklift/v1beta1/provider.go
```diff
@@ -134,11 +138,28 @@ func (p *Provider) RestCfg(secret *core.Secret) (cfg *rest.Config) {
 		cfg, _ = config.GetConfig()
 		return
 	}
+
+	_, ok := secret.Data[api.Insecure]
+	var insecure bool
```
bennyz, line 143: this method should move elsewhere

pkg/lib/util/util.go
```diff
@@ -1,4 +1,4 @@
-package container
```
bennyz, line 1: @ahadas heads up

pkg/apis/forklift/v1beta1/provider.go
```diff
@@ -56,10 +51,8 @@ func (t ProviderType) String() string {
 	return string(t)
 }
 
-// Secret fields.
 const (
-	Token    = "token"
-	Insecure = "insecureSkipVerify"
+	Insecure = "insecure"
```
ahadas, line 55: why changing this? we may have existing secret with insecureSkipVerify

pkg/lib/client/openshift/client.go
```diff
@@ -0,0 +1,69 @@
...
+	util "github.com/konveyor/forklift-controller/pkg/lib/util"
+	core "k8s.io/api/core/v1"
+	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/rest"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
+)
+
+const (
+	Token = "token"
```
ahadas, line 17: having it in the api package also serves as some kind of documentation of what setting we expect to have in specified secrets, why moving it here?

pkg/lib/util/util.go
ahadas: ack, that looks alright

pkg/lib/util/util.go
```diff
@@ -9,11 +9,12 @@ import (
 	liburl "net/url"
 	"strconv"
 
-	api "github.com/konveyor/forklift-controller/pkg/apis/forklift/v1beta1"
 	liberr "github.com/konveyor/forklift-controller/pkg/lib/error"
 	core "k8s.io/api/core/v1"
 )
 
+const Insecure = "insecureSkipVerify"
```
ahadas, line 16: same here, it would be nicer to place the settings we support in a single place - is there a reason not to import the api package still? seems you resolved the cyclic dependency, no?

pkg/apis/forklift/v1beta1/provider.go
bennyz: don't know, I tried to restore it and this is what happened

pkg/apis/forklift/v1beta1/provider.go
```diff
@@ -56,10 +51,9 @@ func (t ProviderType) String() string {
 	return string(t)
 }
 
-// Secret fields.
```
liranr23, line 59: you didn't like the comment? :)

liranr23 (approved):

pkg/lib/client/openshift/client.go
```diff
@@ -0,0 +1,63 @@
...
+		return nil
+	}
+
+	if p.IsHost() {
+		return cfg
+	}
+
+	insecure, _ := strconv.ParseBool(string(secret.Data[api.Insecure]))
+
+	// Default to insecure if parsing fails or explicitly set
```
ahadas, line 30: it's problematic to default to insecure when parsing fails - the user asks for secure connection and we do it differently...

bennyz: yes, but how did the user ask for a secure connection if we failed to parse it?

pkg/forklift-api/webhooks/validating-webhook/admitters/secret-admitter.go
```diff
@@ -72,7 +73,15 @@ func (admitter *SecretAdmitter) Admit(ar *admissionv1.AdmissionReview) *admissio
...
 
 func (admitter *SecretAdmitter) validateProviderSecret() *admissionv1.AdmissionResponse {
-	if _, ok := admitter.secret.Data["cacert"]; ok && container.InsecureProvider(&admitter.secret) {
+	var insecure bool
+	if val, ok := admitter.secret.Data["insecure"]; ok {
+		var err error
+		if insecure, err = strconv.ParseBool(string(val)); err != nil {
+			return webhookutils.ToAdmissionResponseError(fmt.Errorf("invalid value for 'insecure' field"))
+		}
+	}
```
ahadas, line 82: great stuff to have in a webhook!!

pkg/lib/client/openshift/client.go
```diff
@@ -0,0 +1,66 @@
...
+		return nil
+	}
+
+	cacert, hasCACert := secret.Data["cacert"]
+	cfg = &rest.Config{
+		Host:        p.Spec.URL,
+		BearerToken: string(secret.Data[api.Token]),
+		TLSClientConfig: rest.TLSClientConfig{
+			Insecure: insecure,
+			CAData:   cacert,
```
ahadas, line 40: why setting those fields here it we set them below?

pkg/lib/client/openshift/client.go
```diff
@@ -0,0 +1,66 @@
...
+		},
+	}
+	if !insecure && hasCACert {
+		cfg.TLSClientConfig.CAData = cacert
+	} else {
+		cfg.TLSClientConfig.Insecure = true
+	}
```
ahadas, line 47: this doesn't align with util.tlsConfig - we can't say that insecure==false and not to specify a CA (and in such case use the system pool)?

ahadas (approved):

PR #729, opened 2024-02-04
Validate VDDK

ahadas:
This PR introduces two-step validation of the VDDK init image:
1. The provider's validating webhook fails providers with empty VDDK init image setting. VDDK can be omitted when users don't want to use it (though it's highly recommended to use VDDK to accelerate migrations from vSphere) but it cannot be set with empty value now.
2. The validation of plans in forklift-controller creates a job on the target provider that checks the validity of the specified VDDK image by checking the existence of `/opt/vmware-vix-disklib-distrib/lib64/libvixDiskLib.so`. The job is set with `ActiveDeadlineSeconds` of 300  seconds (which is configurable) so that if the VDDK image cannot be pulled, the job would fail. There can be more than a single job for a plan in case the VDDK image for the provider changes and the jobs are deleted when archiving the plan.

### DIFF
--- a/pkg/controller/plan/migration.go
+++ b/pkg/controller/plan/migration.go
@@ -23,10 +23,13 @@ import (
 	liberr "github.com/konveyor/forklift-controller/pkg/lib/error"
 	libitr "github.com/konveyor/forklift-controller/pkg/lib/itinerary"
 	"github.com/konveyor/forklift-controller/pkg/settings"
+	batchv1 "k8s.io/api/batch/v1"
 	core "k8s.io/api/core/v1"
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
 	cdi "kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 // Requeue
@@ -343,6 +346,10 @@ func (r *Migration) Archive() {
 		if err := r.deletePvcPvForOva(); err != nil {
 			r.Log.Error(err, "Failed to clean up the PVC and PV for the OVA plan")
 		}
+	case v1beta1.VSphere:
+		if err := r.deleteValidateVddkJob(); err != nil {
+			r.Log.Error(err, "Failed to clean up validate-VDDK job(s)")
+		}
 	}
 
 	for _, vm := range r.Plan.Status.Migration.VMs {
@@ -541,6 +548,31 @@ func (r *Migration) deletePvcPvForOva() (err error) {
 	if err != nil {
 		r.Log.Error(err, "Failed to delete the plan PV")
 		return
+	}
+	return
+}
+
+func (r *Migration) deleteValidateVddkJob() (err error) {
+	selector := labels.SelectorFromSet(map[string]string{"plan": string(r.Plan.UID)})
+	jobs := &batchv1.JobList{}
+	err = r.Destination.Client.List(
+		context.TODO(),
+		jobs,
+		&client.ListOptions{
+			LabelSelector: selector,
+			Namespace:     r.Plan.Spec.TargetNamespace,
+		},
+	)
+	if err != nil {
+		return
+	}
+	for _, job := range jobs.Items {
+		background := meta.DeletePropagationBackground
+		opts := &client.DeleteOptions{PropagationPolicy: &background}
+		err = r.Destination.Client.Delete(context.TODO(), &job, opts)
+		if err != nil {
+			r.Log.Error(err, "Failed to delete validate-vddk job", "job", job)
+		}
 	}
 	return
 }
--- a/pkg/controller/plan/validation.go
+++ b/pkg/controller/plan/validation.go
@@ -2,6 +2,8 @@ package plan
 
 import (
 	"context"
+	"crypto/md5"
+	"encoding/hex"
 	"errors"
 	"fmt"
 	"path"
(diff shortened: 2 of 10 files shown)


pkg/controller/plan/validation.go
```diff
@@ -667,34 +677,187 @@ func (r *Reconciler) validateVddkImage(plan *api.Plan) (err error) {
...
+		Spec: batchv1.JobSpec{
+			ActiveDeadlineSeconds: ptr.To[int64](int64(settings.Settings.Migration.VddkJobActiveDeadline)),
+			BackoffLimit:          ptr.To[int32](2),
+			Completions:           ptr.To[int32](1),
+			Template: core.PodTemplateSpec{
+				Spec: core.PodSpec{
+					SecurityContext: &core.PodSecurityContext{
+						FSGroup:      &fsGroup,
+						RunAsUser:    &user,
+						RunAsNonRoot: &nonRoot,
```
bennyz, line 835: these can be replaced by `ptr.To` as well

pkg/controller/plan/validation.go
```diff
@@ -667,34 +677,183 @@ func (r *Reconciler) validateVddkImage(plan *api.Plan) (err error) {
...
+		if job, err = r.ensureVddkImageValidationJob(plan, el9, image); err != nil {
+			break
+		}
+		if len(job.Status.Conditions) == 0 {
+			err = liberr.New("validation of VDDK job is in progress")
+		}
+		for _, condition := range job.Status.Conditions {
+			switch condition.Type {
+			case batchv1.JobComplete:
+				r.Log.Info("validate VDDK job completed", "image", image)
```
liranr23, line 723: maybe worth adding to which plan/provider?

pkg/controller/plan/validation.go
```diff
@@ -667,34 +677,183 @@ func (r *Reconciler) validateVddkImage(plan *api.Plan) (err error) {
...
+	volumes := []core.Volume{
+		{
+			Name: VddkVolumeName,
+			VolumeSource: core.VolumeSource{
+				EmptyDir: &core.EmptyDirVolumeSource{},
+			},
+		},
+	}
+	return &batchv1.Job{
+		ObjectMeta: meta.ObjectMeta{
```
liranr23, line 812: how about adding ownerRef of the plan?

pkg/forklift-api/webhooks/validating-webhook/admitters/provider-admitter.go
```diff
@@ -84,6 +83,19 @@ func (admitter *ProviderAdmitter) validateVDDK() error {
 	return nil
 }
 
+func (admitter *ProviderAdmitter) validateVddkImage() error {
+	if image, found := admitter.provider.Spec.Settings[api.VDDK]; found {
+		if image == "" {
```
liranr23, line 88: is that because the UI? does the v2v-cold must have VDDK image? or it will provided with `nil` from UI and expected the same in CLI?

pkg/forklift-api/webhooks/validating-webhook/admitters/provider-admitter.go
```diff
@@ -84,6 +83,19 @@ func (admitter *ProviderAdmitter) validateVDDK() error {
...
+			err := liberr.New("The specified VDDK init image is empty")
+			log.Error(err, "empty image is invalid, failing",
```
liranr23, line 90: if the above it true, i would suggest to say only omitted or image value is accepted to hint what is it "empty"

pkg/controller/plan/migration.go
```diff
@@ -545,6 +552,31 @@ func (r *Migration) deletePvcPvForOva() (err error) {
 	return
 }
 
+func (r *Migration) deleteValidateVddkJob() (err error) {
```
liranr23, line 555: i would prefer to see it in destinationclient or kubevirt, but that's not that important

pkg/controller/plan/validation.go
liranr23: right, `r.Log` :)

pkg/controller/plan/validation.go
```diff
@@ -667,34 +677,183 @@ func (r *Reconciler) validateVddkImage(plan *api.Plan) (err error) {
...
 
-	el9, el9Err := plan.VSphereUsesEl9VirtV2v()
-	if el9Err != nil {
-		return el9Err
+func getVddkImageValidationJobLabels(plan *api.Plan) map[string]string {
+	image := plan.Referenced.Provider.Source.Spec.Settings[api.VDDK]
+	sum := md5.Sum([]byte(image))
+	return map[string]string{
+		"plan": string(plan.ObjectMeta.UID),
+		"vddk": hex.EncodeToString(sum[:]),
```
liranr23, line 771: what's the benefit of the md5? wouldn't the name as is and the plan will be enough?

pkg/controller/plan/migration.go
ahadas: yeah, we're inconsistent about where to place the cleanup pieces (e.g., deleteImporterPods is here and DeletePVCConsumerPod in kubevirt.go)

pkg/controller/plan/validation.go
ahadas: the name would be enough but it typically contains characters that are invalid in this context

liranr23: it will be also nicer to "debug" it without encoding on some sort. because then, we can look into the job and based on the label see the real path of the image. however if it requires too much change in terms of invalid chars... that troublesome

ahadas: the vddk image is set in the annotations for debugging (see line 815 below) - the labels are purely for matching

liranr23 (approved):

PR #736, opened 2024-02-07
openstack: fix linter issues

bennyz:
- Missing return
- Redundant casting

### DIFF
--- a/pkg/lib/client/openstack/client.go
+++ b/pkg/lib/client/openstack/client.go
@@ -403,6 +403,7 @@ func (c *Client) identityServiceAPI(object interface{}, opts interface{}) (err e
 	err = c.connectIdentityServiceAPI()
 	if err != nil {
 		err = liberr.Wrap(err)
+		return
 	}
 	switch object.(type) {
 	case *Region, *[]Region:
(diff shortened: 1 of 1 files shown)


liranr23 (approved):

pkg/lib/client/openstack/client.go
```diff
@@ -895,42 +876,38 @@ func (c *Client) volumeAPI(object interface{}, opts interface{}) (err error) {
...
-			opts := opts.(*VolumeUpdateOpts)
 			volume, err = volumes.Update(c.blockStorageService, object.ID, opts).Extract()
 			if err != nil {
 				return
 			}
 			*object = Volume{Volume: *volume}
 		case *DeleteOpts:
-			opts := volumes.DeleteOpts{Cascade: true}
-			err = volumes.Delete(c.blockStorageService, object.ID, opts).ExtractErr()
+			deleteOpts := volumes.DeleteOpts{Cascade: true}
```
ahadas, line 909: it was that way before so it's fine to keep it this way in this PR but wouldn't it make sense to rather set `Cascade` on `opts`?

ahadas (approved):

PR #742, opened 2024-02-12
Remove PV of OVA provider server

ahadas:
The way the PV for a pod that serves as an NFS server for an OVA provider is created doesn't allow its deletion by a CSI driver and therefore we need to take care of its cleanup when the OVA provider is deleted.

Here, we add a finalizer to OVA providers to prevent their deletion until we delete their PV.


liranr23:
> The way the PV for a pod that serves as an NFS server for an OVA provider is created doesn't allow its deletion by a CSI driver and therefore we need to take care of its cleanup when the OVA provider is deleted.

Why? and what can be done to do it on the creation? if possible.. i saw that we already set `PersistentVolumeReclaimPolicy` to delete.. but just wondering why it's stuck on this. maybe it holds finalizers within?

ahadas:
> > The way the PV for a pod that serves as an NFS server for an OVA provider is created doesn't allow its deletion by a CSI driver and therefore we need to take care of its cleanup when the OVA provider is deleted.
> 
> Why? and what can be done to do it on the creation? if possible.. i saw that we already set `PersistentVolumeReclaimPolicy` to delete.. but just wondering why it's stuck on this. maybe it holds finalizers within?

setting `PersistentVolumeReclaimPolicy` to delete has no meaning for a PV that is not allocated by a CSI driver that supports deletion.
what can be done to enable this? not much since the PV is not allocated dynamically, we allocated it manually to point to an NFS share that is provided by the user - in that case there's not much we can do other than deleting it "manually"

pkg/controller/provider/controller.go
```diff
@@ -234,6 +236,30 @@ func (r Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (r
 		}
 	}
 
+	if provider.DeletionTimestamp != nil && k8sutil.ContainsFinalizer(provider, api.OvaProviderFinalizer) {
```
bkhizgiy, line 239: I remembered another issue I had when looking into it, once the provider deleted the objects was no longer accessible, were you able to query its fields at this point?

ahadas: yeah, not that with a finalizer, an entity doesn't get removed until the finalizer is dropped - so when we get here, the provider still exists and has the DeletionTimestamp field set, so we can still get its properties

liranr23 (approved):

bkhizgiy (approved):

pkg/forklift-api/webhooks/mutating-webhook/mutators/provider-mutator.go
```diff
@@ -25,9 +26,25 @@ func (mutator *ProviderMutator) Mutate(ar *admissionv1.AdmissionReview) *admissi
 	}
 
 	specChanged := mutator.setSdkEndpointIfNeeded()
+	metadataChanged := mutator.ar.Request.Operation == admissionv1.Create && mutator.setFinalizers()
 
+	var patches []util.PatchOperation
 	if specChanged {
-		patches := mutator.patchPayload()
+		patches = append(patches, util.PatchOperation{
```
bennyz, line 33: not improtant but we can probably also use `jsonpatch.CreateMergePatch`

PR #743, opened 2024-02-13
ocp: validate source version > 4.13

bennyz:
We actually check the k8s version, where 1.26 corresponds to OCP 4.13. This is needed since CNV running on older version doesn't export the manifests in the VM export server


pkg/controller/plan/validation.go
```diff
@@ -710,3 +723,32 @@ func (r *Reconciler) setupSecret(plan *api.Plan) (err error) {
...
+	if err != nil {
+		return liberr.Wrap(err)
+	}
+
+	minor, err := strconv.Atoi(version.Minor)
+	if err != nil {
+		return liberr.Wrap(err)
+	}
+
+	if major < 1 && minor < 26 {
```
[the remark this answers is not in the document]
bennyz: yeah, I'm not good with boolean logic... and the test is still in progress

pkg/controller/plan/validation.go
```diff
@@ -87,6 +90,7 @@ const (
 // Validate the plan resource.
 func (r *Reconciler) validate(plan *api.Plan) error {
 	// Provider.
+	r.Log.Info("Validating plan.", "plan", plan.Name)
```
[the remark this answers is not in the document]
bennyz: yes, forgot to drop

pkg/controller/plan/validation.go
```diff
@@ -143,6 +146,28 @@ func (r *Reconciler) validate(plan *api.Plan) error {
 		return err
 	}
 
+	// Validate version only if target ocp-to-ocp
```
ahadas, line 149: what did you mean by "if target ocp-to-ocp", maybe "Validate version only in case of openshift-to-openshift"?

pkg/controller/plan/validation.go
```diff
@@ -143,6 +146,28 @@ func (r *Reconciler) validate(plan *api.Plan) error {
...
+	if plan.Provider.Source.Type() == api.OpenShift && plan.Provider.Destination.Type() == api.OpenShift {
+		restCfg := ocp.RestCfg(plan.Referenced.Provider.Source, plan.Referenced.Secret)
+		clientset, err := kubernetes.NewForConfig(restCfg)
+		if err != nil {
+			return liberr.Wrap(err)
+		}
+		err = r.checkOCPVersion(clientset)
+		if err != nil {
+			return err
+		}
```
liranr23, line 168: what about setting the condition?

pkg/controller/plan/validation.go
```diff
@@ -710,3 +736,27 @@ func (r *Reconciler) setupSecret(plan *api.Plan) (err error) {
...
+		return liberr.Wrap(err)
+	}
+
+	minor, err := strconv.Atoi(version.Minor)
+	if err != nil {
+		return liberr.Wrap(err)
+	}
+
+	if major < 1 || (major == 1 && minor < 26) {
+		return liberr.New("OpenShift version is not supported")
```
ahadas, line 758: it would display OpenShift also for k8s then which is not that great - how about "Version of source provider" instead?

pkg/controller/plan/validation_test.go
```diff
@@ -0,0 +1,56 @@
...
+				}
+			})
+
+			ginkgo.It("should return an error", func() {
+				err := reconciler.checkOCPVersion(fakeClientSet)
+				gomega.Expect(err).To(gomega.HaveOccurred())
+			})
+		})
+	})
+})
```
ahadas, line 56: did you consider to use a Table here?

pkg/controller/plan/validation_test.go
```diff
@@ -0,0 +1,56 @@
...
+
+	ginkgo.Describe("validateOCPVersion", func() {
+		ginkgo.Context("when the OpenShift version is supported", func() {
+			ginkgo.BeforeEach(func() {
+				fakeDiscovery, ok := fakeClientSet.Discovery().(*discovery.FakeDiscovery)
+				gomega.Expect(ok).To(gomega.BeTrue())
+				fakeDiscovery.FakedServerVersion = &version.Info{
+					Major: "1", Minor: "26",
+				}
+			})
```
liranr23, line 33: do we really need the before each on each case? is it possible to pull most of it out to the global before each? on each case maybe just set the fakeversion?

pkg/controller/plan/validation.go
```diff
@@ -421,7 +446,7 @@ func (r *Reconciler) validateVM(plan *api.Plan) error {
 		if err != nil {
 			return err
 		}
-		if plan.Referenced.Secret == nil {
+		if plan.Referenced.Secret == nil && plan.Referenced.Provider.Source.Type() != api.OpenShift {
```
liranr23, line 449: a comment regarding this?
we won't need the `plan.Referenced.Secret` anywhere if the source is OCP?

ahadas: > we won't need the `plan.Referenced.Secret` anywhere if the source is OCP?

yeah, that's a good point - we should check if the source is the 'host' cluster as the latest code does

ahadas (commented):
the changes look good but the title of the PR and of the first commit is "ocp: validate source version > 4.13", should be ".. >= 4.13", right? also we should use `openshift` more instead of `ocp` as MTV can now be deployed on OKE

bennyz:
> the changes look good but the title of the PR and of the first commit is "ocp: validate source version > 4.13", should be ".. >= 4.13", right? also we should use `openshift` more instead of `ocp` as MTV can now be deployed on OKE

I think ocp in our context stands for k8s distribution in general

pkg/controller/plan/validation.go
```diff
@@ -143,6 +147,39 @@ func (r *Reconciler) validate(plan *api.Plan) error {
...
+		restCfg := ocp.RestCfg(plan.Referenced.Provider.Source, plan.Referenced.Secret)
+		clientset, err := kubernetes.NewForConfig(restCfg)
+		if err != nil {
+			return liberr.Wrap(err)
+		}
+
+		err = r.checkOCPVersion(clientset)
+		if err != nil {
+			plan.Status.SetCondition(unsupportedVersion)
+			return err
```
liranr23, line 179: the error can be cause because multiple error even before getting the version. i guess it's a matter of taste of you wish to differentiate between those cases.

liranr23 (approved):

pkg/controller/plan/validation.go
bennyz: it's true but then it's a failure to validate rather than a validation failure

ahadas:
> > the changes look good but the title of the PR and of the first commit is "ocp: validate source version > 4.13", should be ".. >= 4.13", right? also we should use `openshift` more instead of `ocp` as MTV can now be deployed on OKE
> 
> I think ocp in our context stands for k8s distribution in general

we distinguish between openshift and k8s in the operator, but yeah - in the code "ocp" typically refers to both

@yaacov the table on oke vs ocp vs ocp plus can be found [here](https://docs.openshift.com/container-platform/4.14/welcome/oke_about.html)

ahadas:
@bennyz the commit still says > 4.13...

ahadas (approved):

PR #746, opened 2024-02-15
Change virt-v2v bash script to use go wrapper

bkhizgiy:
This PR updates the virt-v2v image, replacing the existing bash script with a Go wrapper. This enhancement simplifies control over code changes.


virt-v2v/cold/entrypoint.go
```diff
@@ -0,0 +1,215 @@
...
+		return
+	}
+
+	w.Header().Set("Content-Type", "application/xml")
+	_, err = w.Write(xmlData)
+	if err != nil {
+		fmt.Printf("Error writing response: %v\n", err)
+		http.Error(w, "Error writing response", http.StatusInternalServerError)
+	}
+	w.WriteHeader(http.StatusOK)
```
ahadas, line 203: this should run only when err == nil, otherwise it would override http.StatusInternalServerError in case of an error

virt-v2v/cold/entrypoint.go
```diff
@@ -0,0 +1,219 @@
...
+			return err
+		}
+	}
+	return
+}
+
+func executeVirtV2v(args []string) (err error) {
+	virtV2vCmd := exec.Command("bash", "-c", strings.Join(args, " "), "|& /usr/local/bin/virt-v2v-monitor")
+	virtV2vCmd.Stdout = os.Stdout
+	virtV2vCmd.Stderr = os.Stderr
```
liranr23, line 168: just making sure - because of these two, we will see the whole output of virt-v2v in the container log, correct?

virt-v2v/cold/entrypoint.go
```diff
@@ -0,0 +1,219 @@
...
+
+	if source == vSphere {
+		if _, err := os.Stat("/etc/secret/cacert"); err == nil {
+			err = os.Symlink("/etc/secret/cacert", "/opt/ca-bundle.crt")
+			if err != nil {
+				fmt.Println("Error creating ca cert link ", err)
+				os.Exit(1)
+			}
+		} else {
+			err := os.Symlink("/etc/pki/tls/certs/ca-bundle.crt.bak", "/opt/ca-bundle.crt")
```
ahadas, line 75: please keep the comment we had in the entrypoint script here ("otherwise, keep system pool certificates")

virt-v2v/cold/entrypoint.go
```diff
@@ -0,0 +1,219 @@
...
+)
+
+var (
+	xmlFilePath string
+	server      *http.Server
+)
+
+func main() {
+	virtV2vArgs := []string{"virt-v2v", "-v", "-x"}
+	source := os.Getenv("V2V_source")
```
ahadas, line 30: I miss a check of whether `source` is either ova or vSphere but I don't see it in the entrypoint script either...

ahadas (commented):
looks good, minor comments

virt-v2v/cold/entrypoint.go
bkhizgiy: Yes, that exactly that :)

virt-v2v/cold/entrypoint.go
bkhizgiy: I think we can assume its OK, since its being sent and verified from the controller side.

bkhizgiy:
After adding modification the correct command is executed now but it still taking long time (almost an hour) triggered the same conversion with the original image for time and output comparison.

virt-v2v/cold/entrypoint.go
liranr23: i think that without it we will just fail quickly..
we do have the assumption that the controller will provide it, but since we also have this assumption on other variables and we do validate them - we can also validate this

bkhizgiy:
> After adding modification the correct command is executed now but it still taking long time (almost an hour) triggered the same conversion with the original image for time and output comparison.

The issue is with the monitor code, it wasn't running properly, so both the conversion + copy disks steps were count under the conversion (and the time adds up similarly to the original code two steps), triggered another conversion to see if the monitor running correctly after the changes.

ahadas:
> The issue is with the monitor code, it wasn't running properly, so both the conversion + copy disks steps were count under the conversion (and the time adds up similarly to the original code two steps)

I think that I saw it once when I played with the VDDK validation - it doesn't seem to be a result of your changes
but that wouldn't explain the relatively long time, @bkhizgiy did you see that VDDK is now specified correctly?

bkhizgiy:
> > The issue is with the monitor code, it wasn't running properly, so both the conversion + copy disks steps were count under the conversion (and the time adds up similarly to the original code two steps)
> 
> I think that I saw it once when I played with the VDDK validation - it doesn't seem to be a result of your changes but that wouldn't explain the relatively long time, @bkhizgiy did you see that VDDK is now specified correctly?

Yes I see the command is set correctly now, and i also compered the outputs with my changes and the original script and it the same. Its just not running the monitor part correctly.

For the original script the conversion took 15 min and disk copy 43 min.
For my changes the conversion took 55 min and the copy none, so i just think its not recognize correctly when to move on to the next step.

liranr23:
> > > The issue is with the monitor code, it wasn't running properly, so both the conversion + copy disks steps were count under the conversion (and the time adds up similarly to the original code two steps)
> > 
> > 
> > I think that I saw it once when I played with the VDDK validation - it doesn't seem to be a result of your changes but that wouldn't explain the relatively long time, @bkhizgiy did you see that VDDK is now specified correctly?
> 
> Yes I see the command is set correctly now, and i also compered the outputs with my changes and the original script and it the same. Its just not running the monitor part correctly.
> 
> For the original script the conversion took 15 min and disk copy 43 min. For my changes the conversion took 55 min and the copy none, so i just think its not recognize correctly when to move on to the next step.

that might be because the last update of the patch (breaking the two commands apart)
previously it was made as `virt-v2v ... |& monitor`, so the the pipeline connected the output of virt-v2v to the monitor command input.
now it looks like it first executing `virt-v2v` and wait for it to finish, then it runs the monitor command

virt-v2v/cold/entrypoint.go
```diff
@@ -0,0 +1,236 @@
...
+	if err = virtV2vMonitorCmd.Start(); err != nil {
+		fmt.Printf("Error executing monitor command: %v\n", err)
+		return
+	}
+
+	if err = virtV2vCmd.Wait(); err != nil {
+		fmt.Printf("Error waiting for virt-v2v to finish : %v\n", err)
+		return
+	}
+	return
```
liranr23, line 189: how about trying this:
```
virtV2vCmd := exec.Command(args[0], args[1:]...)
virtV2vCmd.Stdout = os.Stdout
virtV2vCmd.Stderr = os.Stderr

virtV2vMonitorCmd := exec.Command("/usr/local/bin/virt-v2v-monitor")
virtV2vMonitorCmd.Stdin, _ = virtV2vCmd.StdoutPipe()
virtV2vMonitorCmd.Stdout = os.Stdout
virtV2vMonitorCmd.Stderr = os.Stderr

if err = virtV2vMonitorCmd.Start(); err != nil {
	fmt.Printf("Error executing monitor command: %v\n", err)
	return
}

if err = virtV2vCmd.Run(); err != nil {
	fmt.Printf("Error executing command: %v\n", err)
	return
}

if err = virtV2vCmd.Wait(); err != nil {
	fmt.Printf("Error waiting for virt-v2v to finish : %v\n", err)
	return
}
return
```

liranr23: and if we really wish to mimic the bash command we had, maybe we need to set 
virtV2vCmd.Stderr = os.Stdout
as `|&` is setting the stderr of virt-v2v to stdout and then passing the stdout to stdin of the monitor command

bennyz: the problem is that virt-v2v's stdout is consumed by the terminal so it can't be consumed by the monitor. It has to be either sent to both using `tee` or displayed by the monitor

bkhizgiy: yeah, it didn't really worked in the previous implementation as well, and in addition the command missed args. I'm trying to redirect the output now with `TeeReader` following @bennyz suggestion.

virt-v2v/cold/entrypoint.go
```diff
@@ -163,7 +171,13 @@ func LinkDisks(diskKind string, num int) (err error) {
...
 func executeVirtV2v(args []string) (err error) {
 	virtV2vCmd := exec.Command(args[0], args[1:]...)
-	virtV2vCmd.Stdout = os.Stdout
+	virtV2vStdoutPipe, err := virtV2vCmd.StdoutPipe()
+	if err != nil {
+		fmt.Printf("Error setting up stdout pipe: %v\n", err)
+		return
+	}
+
+	tee := io.TeeReader(virtV2vStdoutPipe, os.Stdout)
```
bkhizgiy, line 180: Fixed the monitor consuming issue, now the conversion phase work properly.

bkhizgiy: Screenshot from 2024-02-18 15-18-57

liranr23 (approved):

virt-v2v/cold/entrypoint.go
```diff
@@ -0,0 +1,260 @@
...
+	} else {
+		fmt.Println("Server shut down successfully.")
+	}
+}
+
+func isValidSource(source string) bool {
+	supportedSources := map[string]bool{
+		OVA:     true,
+		vSphere: true,
+	}
```
ahadas, line 256: that seems a bit cumbersome - why not to have something like the following then:
```golang
switch(source) {
    case OVA, vSphere:
        return true
    default:
        return false
}
```

virt-v2v/cold/entrypoint.go
```diff
@@ -0,0 +1,260 @@
...
+var (
+	xmlFilePath string
+	server      *http.Server
+)
+
+func main() {
+	virtV2vArgs := []string{"virt-v2v", "-v", "-x"}
+	source := os.Getenv("V2V_source")
+
+	if isValidSource(source) {
```
ahadas, line 32:
```suggestion
	if !isValidSource(source) {
```

virt-v2v/cold/entrypoint.go
```diff
@@ -0,0 +1,260 @@
...
+		}
+		virtV2vArgs = append(virtV2vArgs, "--", os.Getenv("V2V_vmName"))
+	}
+
+	if err := executeVirtV2v(virtV2vArgs); err != nil {
+		fmt.Println("Error executing virt-v2v command ", err)
+		os.Exit(1)
+	}
+
+	if source == OVA {
```
ahadas, line 105: without the controller part, the conversion pod would remain running, no? in that case, let's drop this part and add it along with the changes to forklift-controller

bkhizgiy: yeah, it wont complete the pod - removing.

ahadas (approved):

ahadas:
passes CI, validated locally - merging
and let's follow up with the fix for OVA which was the motivation for these changes

PR #747, opened 2024-02-15
Add VMs validation for OVA provider

bkhizgiy:
Adding VMs validations for OVA, since the OVA model is a subset of the vSphere one (using a similar structure and variable names in the inventory) the same validation were used, excluding the one that related to cluster/hosts or running state of the VM or any other configuration that are not relevant for OVA VMs.


ahadas:
@bkhizgiy can you please add to the description of this PR what validations were not taken from those for vSphere and which ones (if any) changed?

ahadas (approved):

PR #748, opened 2024-02-16
Fix calls to liberr.Wrap with even number of args

ahadas:
As the signature of the function is:
Wrap(err error, kvpair ...interface{})
It expects to get an error and key-value pairs. However, in some places we also pass a message after the error, which breaks this expectation. In this PR those messages are dropped to remain with an odd number of arguments.

### DIFF
--- a/pkg/controller/plan/adapter/openstack/builder.go
+++ b/pkg/controller/plan/adapter/openstack/builder.go
@@ -246,8 +246,6 @@ const (
 	Multus = "multus"
 )
 
-const VM_LOOKUP_FAILED = "VM lookup failed."
-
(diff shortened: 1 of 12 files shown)


liranr23 (approved):
That looks good, but did you consider adding another Wrap function with additional string?

ahadas:
> That looks good, but did you consider adding another Wrap function with additional string?

I don't see why adding another string - the returned error should be logged with more context (`..Error(err, msg, ...)`), the error should include the stacktrace and pairs. if one really insists on adding more context there, they could add another pair, e.g., `(err, "msg", <msg>)`

PR #749, opened 2024-02-18
Follow up on VDDK validation

ahadas:
1. Fix the provider-admitter to allow changing the VDDK init image for a vSphere provider that serves as the source provider for existing plans.
2. Ensure the target namespace exists in the target provider before creating the VDDK validation job.

### DIFF
--- a/pkg/controller/plan/validation.go
+++ b/pkg/controller/plan/validation.go
@@ -773,10 +773,18 @@ func (r *Reconciler) validateVddkImage(plan *api.Plan) (err error) {
 }
 
 func (r *Reconciler) ensureVddkImageValidationJob(plan *api.Plan, el9 bool, vddkImage string) (*batchv1.Job, error) {
-	jobLabels := getVddkImageValidationJobLabels(plan)
+	ctx, err := plancontext.New(r, plan, r.Log)
+	if err != nil {
+		return nil, err
+	}
+
+	if err = r.ensureNamespace(ctx); err != nil {
+		return nil, liberr.Wrap(err)
+	}
+
+	jobLabels := getVddkImageValidationJobLabels(ctx.Plan)
 	jobs := &batchv1.JobList{}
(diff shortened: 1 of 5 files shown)


pkg/controller/plan/validation.go
```diff
@@ -773,10 +773,18 @@ func (r *Reconciler) validateVddkImage(plan *api.Plan) (err error) {
 }
 
 func (r *Reconciler) ensureVddkImageValidationJob(plan *api.Plan, el9 bool, vddkImage string) (*batchv1.Job, error) {
-	jobLabels := getVddkImageValidationJobLabels(plan)
+	ctx, err := plancontext.New(r, plan, r.Log)
+	if err != nil {
+		return nil, err
+	}
+
+	if err = r.ensureNamespace(ctx); err != nil {
```
liranr23, line 781: it makes sense to do that, but that also have a downside.
if a user just created the plan - the namespace will be created on destination.. even without triggering the plan.
if i set the namespace and changed it, it will create 2 namespaces and on on, right?
as a user i wouldn't expect the namespace to be created before starting the migration operation

ahadas: > if i set the namespace and changed it, it will create 2 namespaces and on on, right?

right

> as a user i wouldn't expect the namespace to be created before starting the migration operation

that depends on the user I guess - we don't say when the namespace will be created, the user can pick an existing namespace or specify a name of a non-existing namespace that "will be created". some users would expect the namespace would be created when triggering the plan, others could expect that we'll check that we can create the namespace when validating the plan.. but I suspect that in a typical case, users would specify an existing namespace because they'll want to set permissions on it, add network attachment definition, etc

liranr23: looks like there is no way around it and this is the best we can do

liranr23 (approved):

PR #750, opened 2024-02-18
Fix 'make bazel-generate'

ahadas:
By upgrading the toolchain to Go 1.21.7

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -3295,8 +3295,8 @@ go_repository(
 
 go_rules_dependencies()
 
-# NOTE: Keep the version in sync with Go toolchain in GitHub action.
-go_register_toolchains(version = "1.20.10")
+# NOTE: Go version is also set in GitHub action
+go_register_toolchains(version = "1.21.7")
 
 # override rules_docker issue with this dependency
 # rules_docker 0.16 uses 0.1.4, bit since there the checksum changed, which is very weird, going with 0.1.4.1 to
--- a/cmd/forklift-api/BUILD.bazel
+++ b/cmd/forklift-api/BUILD.bazel
@@ -22,7 +22,6 @@ go_library(
 go_binary(
     name = "forklift-api",
     embed = [":forklift-api-lib"],
-    static = "on",
     visibility = ["//visibility:public"],
 )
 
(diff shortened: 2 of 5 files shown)


WORKSPACE
```diff
@@ -3295,8 +3295,8 @@ go_repository(
 
 go_rules_dependencies()
 
-# NOTE: Keep the version in sync with Go toolchain in GitHub action.
-go_register_toolchains(version = "1.20.10")
+# NOTE: Go version is also set in GitHub action
+go_register_toolchains(version = "1.21.7")
```
liranr23, line 3299: how about bumping it also in virt-v2v workspaces? and what about github actions in order to align with it?

ahadas: > how about bumping it also in virt-v2v workspaces? 

changed those workspaces as well

> and what about github actions in order to align with it?

we can do that once we branch out Forklift 2.6

liranr23 (approved):

ahadas:
there's a real failure with this change - forklift-api crashes with `SIGFPE: floating-point exception` so we may have to upgrade the version of Go we test with....

cmd/forklift-api/BUILD.bazel
```diff
@@ -22,7 +22,6 @@ go_library(
 go_binary(
     name = "forklift-api",
     embed = [":forklift-api-lib"],
-    static = "on",
```
bennyz, line 25: why was it on? also, isn't it the default

ahadas: there was an issue related to network with the version of Go we used in 2.4 and that was a workaround that issue

ahadas: > also, isn't it the default

ah and about the default settings, [it's disabled by default](https://github.com/bazelbuild/rules_go/blob/8c014fd0798b1c2965d4dc97e11089bc1d43490a/go/modes.rst#build-settings)

bennyz: hmm go's default is static unless CGO is enabled

bennyz: the network issue is resolved with netdns=go no?

ahadas: yeah, looking at git log, it wasn't part of the fix for that network issue but part of the original changes to create an image for forklift-controller using bazel - it was in 09/22, I don't recall why it was done this way
anyway, it's not needed anymore - without it the provider is added but now there's an issue with mounting the NFS volume to the server pod, I hope it's due to v4...

ahadas:
apparently `make bazel-generate` doesn't fail but produces new warnings
closing this PR as these changes break the ova-server pod

PR #752, opened 2024-02-18
WIP openstack,ovirt: modify PVC creation to work with CDI

bennyz:
Using forklift-controller from PR https://github.com/kubevirt/containerized-data-importer/pull/2983 we can now populate ovirt and openstack PVCs without our populator controller.

This requires scaling down the forklift-volume-populator-controller, and enabling the `WebhookPvcRendering` feature gate in CDI config


ahadas:
I don't think this would be an issue as we intend to have this in MTV 2.7, but just to be on the safe side - CDI on OpenShift 4.15 supports this, right? and 4.14?

bennyz:
> I don't think this would be an issue as we intend to have this in MTV 2.7, but just to be on the safe side - CDI on OpenShift 4.15 supports this, right? and 4.14?

No, it was just merged and it's unclear yet in which version it won't be behind a feature gate. It's just simpler for me to test the entire flow this way

bennyz:
the feature gate is unlikely to be enabled, closing

PR #756, opened 2024-02-19
Drop unused webhook-server-secret

ahadas:


ahadas:
backporting to the release-v2.5.5 branch which will be the source for 2.5.6

PR #757, opened 2024-02-20
OVA: generate random names for NFS pv and PVS used for migration

bkhizgiy:
An issue has been identified where manually constructed names for PVCs lead to migration failures, especially during restarts or when reusing the same plan names. This occurs because of conflicts with existing PVCs. To address this, the implementation now utilizes `GeneratedName` to ensure unique naming and prevent collisions. Correspondingly, the cleanup process has been changed to accommodate these changes.

### DIFF
--- a/pkg/controller/plan/kubevirt.go
+++ b/pkg/controller/plan/kubevirt.go
@@ -2104,7 +2104,7 @@ func (r *KubeVirt) EnsurePersistentVolume(vmRef ref.Ref, persistentVolumes []cor
 	return
 }
 
-func GetOvaPvNfs(dClient client.Client, planID string) (pvs *core.PersistentVolumeList, found bool, err error) {
+func GetOvaPvListNfs(dClient client.Client, planID string) (pvs *core.PersistentVolumeList, found bool, err error) {
 	pvs = &core.PersistentVolumeList{}
 	pvLabels := map[string]string{
 		"plan": planID,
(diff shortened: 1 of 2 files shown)


pkg/controller/plan/kubevirt.go
```diff
@@ -2186,28 +2186,18 @@ func (r *KubeVirt) CreatePvForNfs() (err error) {
...
 	sc := ""
-	pvName := getEntityName("pv", r.Source.Provider.Name, r.Plan.Name)
-	labels := map[string]string{"provider": r.Plan.Provider.Source.Name, "app": "forklift", "migration": r.Migration.Name, "plan": r.Plan.Name}
+	labels := map[string]string{"provider": r.Plan.Provider.Source.Name, "app": "forklift", "migration": r.Migration.Name, "plan": string(r.Plan.UID), "ova": "nfs-pvc"}
 	pvc := &core.PersistentVolumeClaim{
 		ObjectMeta: meta.ObjectMeta{
-			Name:      pvcName,
-			Namespace: r.Plan.Spec.TargetNamespace,
-			Labels:    labels,
+			GenerateName: pvcNamePrefix,
```
bennyz, line 2198: you can call `getEntityPrefixName` here directly I think

pkg/controller/plan/kubevirt.go
```diff
@@ -2102,16 +2104,20 @@ func (r *KubeVirt) EnsurePersistentVolume(vmRef ref.Ref, persistentVolumes []cor
 	return
 }
 
-func GetOvaPvNfs(client client.Client, planName string, providerName string) (pv *core.PersistentVolume, found bool, err error) {
-	pv = &core.PersistentVolume{}
-	err = client.Get(
+func GetOvaPvNfs(dClient client.Client, planID string) (pvs *core.PersistentVolumeList, found bool, err error) {
```
ahadas, line 2107: the name suggests that this function returns a single PV while we return a list - renaming the function makes more sense to me than returning the first result or something like that

ahadas (commented):
+1 for this approach
edit: there's a typo in the title of the commit, randon -> random

pkg/controller/plan/kubevirt.go
bkhizgiy: agreed, changed the name.

ahadas (approved):
it feels like we're still putting too much "effort" to generate the prefix which is less important when the postfix is generated but we can check that separately

PR #758, opened 2024-02-21
ovirt: fix connection test

bennyz:
Currently when the connection test returns 500 we have no error reported which makes the connection test succeed despite failure.

Return an error when the status returned by the connection test is not 200

### DIFF
--- a/pkg/controller/provider/container/ovirt/collector.go
+++ b/pkg/controller/provider/container/ovirt/collector.go
@@ -132,8 +132,7 @@ func (r *Collector) HasParity() bool {
 func (r *Collector) Test() (status int, err error) {
 	_, status, err = r.client.system()
 	if status != http.StatusOK {
-		err = liberr.New(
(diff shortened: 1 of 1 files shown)


bkhizgiy (approved):

pkg/controller/provider/container/ovirt/collector.go
```diff
@@ -130,6 +131,10 @@ func (r *Collector) HasParity() bool {
 // Test connect/logout.
 func (r *Collector) Test() (status int, err error) {
 	_, status, err = r.client.system()
+	if status != http.StatusOK {
+		err = liberr.New("got status != 200 from oVirt", "status", status))
```
ahadas, line 135: @bennyz how about that change in calling liberr.New?

bennyz: ack

ahadas (approved):

PR #765, opened 2024-02-22
Set resource requests for forklift-operator

ahadas:
The forklift-operator pod is not CPU intensive, so minimal CPU time should be sufficient. When it comes to memory, in some environments it consumes ~40Mi and in others ~100Mi, so asking for 64Mi seems reasonable.

see [this](https://github.com/redhat-openshift-ecosystem/community-operators-prod/pull/4079#issuecomment-1959939530)

### DIFF
--- a/operator/config/manager/manager.yaml
+++ b/operator/config/manager/manager.yaml
@@ -93,4 +93,7 @@ spec:
         image: ${OPERATOR_IMAGE}
         imagePullPolicy: Always
         name: forklift-operator
-        resources: {}
+        resources:
+          requests:
+            cpu: 10m
+            memory: 64Mi


liranr23:
I'm uncertain how we should know what is the memory and CPU resources that would fit the "golden zone"(more than minimal, but not exhausting them). we can start with your PR  and see if a problem appear?

liranr23 (approved):

ahadas:
> I'm uncertain how we should know what is the memory and CPU resources that would fit the "golden zone"(more than minimal, but not exhausting them). we can start with your PR and see if a problem appear?

it's unlikely that we'll see a problem because currently we don't set a request for memory so the memory consumption of the operator pod is not taken into consideration when scheduling the pod and it may end up running on a node that doesn't have enough memory available. so if the pod tries to allocate more than 64Mi and there's not enough memory available - we're at the same situation. and 64Mi is more or less what the pod takes, so we don't reserve significantly more than it needs.

and as for CPU, the operator pod is not CPU intensive - asking for 10% of what the main container in the forklift-controller pod asks for sounds reasonable

updated the commit message to have this documented

PR #767, opened 2024-02-28
openstack: use DataVolume to support WFFC

bennyz:
- Fix NPEs
- Fix permission issue
- Switch to DVs to support WFFC
- Fix inventory race

### DIFF
--- a/pkg/controller/plan/adapter/converter.go
+++ b/pkg/controller/plan/adapter/converter.go
@@ -142,6 +142,10 @@ func createConvertJob(pvc *v1.PersistentVolumeClaim, dv *cdi.DataVolume, srcForm
 							Type: v1.SeccompProfileTypeRuntimeDefault,
 						},
 					},
+					RestartPolicy: v1.RestartPolicyNever,
+					Containers: []v1.Container{
+						makeConversionContainer(pvc, srcFormat, dstFormat),
+					},
 					Volumes: []v1.Volume{
 						{
 							Name: "source",
--- a/pkg/controller/plan/adapter/converter.go
+++ b/pkg/controller/plan/adapter/converter.go
@@ -13,6 +13,7 @@ import (
 	k8serr "k8s.io/apimachinery/pkg/api/errors"
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/utils/ptr"
+	cdi "kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
@@ -46,26 +47,26 @@ func (c *Converter) ConvertPVCs(pvcs []*v1.PersistentVolumeClaim, srcFormat srcF
 			continue
 		}
 
-		scratchPVC, err := c.ensureScratchPVC(pvc)
+		scratchDV, err := c.ensureScratchDV(pvc)
 		if err != nil {
 			return false, err
 		}
 
-		switch scratchPVC.Status.Phase {
-		case v1.ClaimBound:
-			c.Log.Info("Scratch PVC bound", "pvc", scratchPVC.Name)
-		case v1.ClaimPending:
-			c.Log.Info("Scratch PVC pending", "pvc", scratchPVC.Name)
+		switch scratchDV.Status.Phase {
+		case cdi.ImportScheduled, cdi.Pending:
+			c.Log.Info("Scratch DV is not ready", "dv", scratchDV.Name, "status", scratchDV.Status.Phase)
 			return false, nil
-		case v1.ClaimLost:
-			c.Log.Info("Scratch PVC lost", "pvc", scratchPVC.Name)
-			return false, liberr.New("scratch pvc lost")
+		case cdi.ImportInProgress:
+			c.Log.Info("Scratch DV import in progress", "dv", scratchDV.Name)
+			return false, nil
+		case cdi.Succeeded:
+			c.Log.Info("Scratch DV is ready", "dv", scratchDV.Name)
 		default:
-			c.Log.Info("Scratch PVC status is unknown", "pvc", scratchPVC.Name, "status", scratchPVC.Status.Phase)
+			c.Log.Info("Scratch DV is not ready", "dv", scratchDV.Name, "status", scratchDV.Status.Phase)
 			return false, nil
 		}
 
-		convertJob, err := c.ensureJob(pvc, srcFormat(pvc), dstFormat)
+		convertJob, err := c.ensureJob(pvc, scratchDV, srcFormat(pvc), dstFormat)
 		if err != nil {
 			return false, err
 		}
@@ -77,9 +78,9 @@ func (c *Converter) ConvertPVCs(pvcs []*v1.PersistentVolumeClaim, srcFormat srcF
 				c.Log.Info("Convert job completed", "pvc", pvc.Name)
 
 				// Delete scrach PVC
-				err = c.Destination.Client.Delete(context.Background(), scratchPVC, &client.DeleteOptions{})
+				err = c.Destination.Client.Delete(context.Background(), scratchDV, &client.DeleteOptions{})
 				if err != nil {
-					c.Log.Error(err, "Failed to delete scratch PVC", "pvc", scratchPVC.Name)
+					c.Log.Error(err, "Failed to delete scratch DV", "dv", scratchDV.Name)
 				}
 
 				return true, nil
@@ -99,14 +100,14 @@ func (c *Converter) ConvertPVCs(pvcs []*v1.PersistentVolumeClaim, srcFormat srcF
 	return false, nil
 }
 
-func (c *Converter) ensureJob(pvc *v1.PersistentVolumeClaim, srcFormat, dstFormat string) (*batchv1.Job, error) {
+func (c *Converter) ensureJob(pvc *v1.PersistentVolumeClaim, dv *cdi.DataVolume, srcFormat, dstFormat string) (*batchv1.Job, error) {
 	jobName := getJobName(pvc, "convert")
 	job := &batchv1.Job{}
 	err := c.Destination.Client.Get(context.Background(), client.ObjectKey{Name: jobName, Namespace: pvc.Namespace}, job)
 	if err != nil {
 		if k8serr.IsNotFound(err) {
 			c.Log.Info("Converting PVC format", "pvc", pvc.Name, "srcFormat", srcFormat, "dstFormat", dstFormat)
-			job := createConvertJob(pvc, srcFormat, dstFormat, c.Labels)
+			job := createConvertJob(pvc, dv, srcFormat, dstFormat, c.Labels)
 			err = c.Destination.Client.Create(context.Background(), job, &client.CreateOptions{})
 			if err != nil {
 				return nil, err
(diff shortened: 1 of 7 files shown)


pkg/controller/plan/adapter/converter.go
```diff
@@ -134,7 +137,7 @@ func createConvertJob(pvc *v1.PersistentVolumeClaim, srcFormat, dstFormat string
 						SeccompProfile: &v1.SeccompProfile{
 							Type: v1.SeccompProfileTypeRuntimeDefault,
 						},
-						FSGroup: ptr.To(int64(107)),
+						RunAsNonRoot: ptr.To(true),
```
ahadas, line 140: I see that for the vddk-validator we get the following error when creating a plan that migrates to the default namespace:
```
Error: container has runAsNonRoot and image will run as root (pod: "vddk-validator-migplan-glhyj7m79-nk6zp_default(f028dd10-3c86-408f-a3a2-94ceb086a7f1)", container: validator)
```
and I see that this code changes the security context here to be similar to that of the vddk validator pod - do we need to also set a user?

bennyz:
```
...
spec.securityContext.fsGroup: Invalid value: []int64{107}: 107 is not an allowed group, provider restricted-v2: .containers[0].runAsUser: Invalid value: 107: must be in the ranges: [1000780000, 1000789999],
...
```

ahadas: posted a fix for that issue - #771, we can do it similarly here

bennyz: what's different?

ahadas: ah seems that I've missed the setting of `RunAsUser` below, so just to replace 107 with more expressive const

bennyz: uh this doesn't actually work with non-default namespaces

bennyz:
current code tested for default namespace and konveyor-forklift

pkg/controller/plan/adapter/converter.go
ahadas: interesting, yeah - there's a problem with restricted namespace
we had this issue with the ova server, but we don't have this issue with the v2v pod...

pkg/controller/plan/adapter/converter.go
```diff
@@ -134,7 +139,9 @@ func createConvertJob(pvc *v1.PersistentVolumeClaim, srcFormat, dstFormat string
...
+						RunAsUser:    ptr.To[int64](1000850000),
+						FSGroup:      ptr.To[int64](1000850000),
```
ahadas, line 144: that's temporary, right?

bennyz: we can use it... we can add are own range if it makes a difference, there's no reason in this context to specifically use the qemu numbers

pkg/controller/plan/adapter/converter_test.go
```diff
@@ -69,19 +62,58 @@ var _ = Describe("Converter tests", func() {
...
 				Type: batchv1.JobComplete,
 			})
-			converter = createFakeConverter(qcow2PVC, scratchPVC, convertJob)
+
+			dv := &cdi.DataVolume{
+				ObjectMeta: meta.ObjectMeta{
+					Name:      "test-dv",
+					Namespace: pvcNamespace,
+					Labels: map[string]string{
+						"forklift.konveyor.io/conversionSourcePVC": pvcName,
```
liranr23, line 71: can we reuse the const here?

pkg/controller/plan/adapter/converter.go
liranr23: this looks fishy. pretty strange that we could make other pods work and now we need this

liranr23 (commented):
did you check the cleanup?

pkg/controller/plan/adapter/converter.go
bennyz: we don't have forklift's securitycontext in this case because the pod is create from a job

pkg/controller/plan/adapter/converter.go
```diff
@@ -99,26 +96,37 @@ func (c *Converter) ConvertPVCs(pvcs []*v1.PersistentVolumeClaim, srcFormat srcF
 	return false, nil
 }
 
-func (c *Converter) ensureJob(pvc *v1.PersistentVolumeClaim, srcFormat, dstFormat string) (*batchv1.Job, error) {
+func (c *Converter) deleteScratchDV(scratchDV *cdi.DataVolume) {
+	// Delete scrach PVC
```
liranr23, line 100:
```suggestion
	// Delete scratch DV
```

pkg/controller/plan/adapter/converter.go
```diff
@@ -99,26 +96,37 @@ func (c *Converter) ConvertPVCs(pvcs []*v1.PersistentVolumeClaim, srcFormat srcF
...
+	err := c.Destination.Client.Delete(context.Background(), scratchDV, &client.DeleteOptions{})
+	if err != nil {
+		c.Log.Error(err, "Failed to delete scratch PVC", "pvc", scratchDV.Name)
```
liranr23, line 103:
```suggestion
		c.Log.Error(err, "Failed to delete scratch DV", "dv", scratchDV.Name)
```

pkg/controller/plan/adapter/converter.go
```diff
@@ -134,7 +141,6 @@ func createConvertJob(pvc *v1.PersistentVolumeClaim, srcFormat, dstFormat string
 						SeccompProfile: &v1.SeccompProfile{
 							Type: v1.SeccompProfileTypeRuntimeDefault,
 						},
-						FSGroup: ptr.To(int64(107)),
```
ahadas, line 137: ah-ah

pkg/controller/plan/adapter/converter.go
```diff
@@ -186,11 +192,6 @@ func makeConversionContainer(pvc *v1.PersistentVolumeClaim, srcFormat, dstFormat
 		Image: base.Settings.VirtV2vImageCold,
 		SecurityContext: &v1.SecurityContext{
 			AllowPrivilegeEscalation: ptr.To(false),
-			RunAsNonRoot:             ptr.To(true),
-			RunAsUser:                ptr.To(int64(107)),
```
ahadas, line 190: did you check that it works on k8s?

.github/workflows/pull-request.yml
```diff
@@ -53,12 +53,12 @@ jobs:
       - name: Checkout forkliftci
         uses: actions/checkout@v4
         with:
-          repository: kubev2v/forkliftci
+          repository: bennyz/forkliftci
```
ahadas, line 56: that would be excellent to have also an image-based vm in our CI

pkg/controller/plan/adapter/converter.go
ahadas: I've added Settings.OpenShift in #776  - that way we can differentiate between OpenShift and non-OpenShift clusters and specify RunAsNonRoot and RunAsUser on non-OpenShift clusters (I don't remember if I tried to set RunAsNonRoot=true without settings RunAsUser, I think I did and it failed)

ahadas: ok, this scenario wouldn't work on the default namespace because even though the pod starts, the conversion would fail to get write permissions on the PV so the settings of the VDDK validator is not the best reference for this

pkg/controller/plan/adapter/converter.go
```diff
@@ -230,42 +235,68 @@ func makeConversionContainer(pvc *v1.PersistentVolumeClaim, srcFormat, dstFormat
...
-}
+	if len(dvList.Items) == 1 {
+		c.Log.Info("Found DV", "dv", dvList.Items[0].Name)
+		return &dvList.Items[0], nil
+	} else if len(dvList.Items) > 1 {
+		return nil, liberr.New("multiple scratch DVs found for pvc", "pvc", sourcePVC.Name)
+	}
+
+	// DV doesn't exist, create it
+	scratchDV = makeScratchDV(sourcePVC)
```
ahadas, line 255:
```suggestion
	scratchDV := makeScratchDV(sourcePVC)
```
(and drop `scratchDV := &cdi.DataVolume{}` above)

pkg/controller/plan/adapter/converter.go
```diff
@@ -230,42 +235,68 @@ func makeConversionContainer(pvc *v1.PersistentVolumeClaim, srcFormat, dstFormat
...
+	c.Log.Info("DV doesn't exist, creating", "dv", scratchDV.Name)
+	err = c.Destination.Client.Create(context.Background(), scratchDV, &client.CreateOptions{})
```
ahadas, line 257: we can drop the third argument then, no?

ahadas (approved):

ahadas (commented):
my comments were so minor that I forgot I wrote them and approved - so changing to "Comment"

bennyz:
> my comments were so minor that I forgot I wrote them and approved - so changing to "Comment"

I need to make some additional changes, it seems I have not fixed the race like I though

.github/workflows/pull-request.yml
bennyz: unfortunately the current code isn't too friendly towards adding an additional vm, but this code path should be tested in e2e

pkg/controller/plan/adapter/converter.go
```diff
@@ -99,42 +95,62 @@ func (c *Converter) ConvertPVCs(pvcs []*v1.PersistentVolumeClaim, srcFormat srcF
...
+			Namespace:    pvc.Namespace,
+			Labels:       labels,
 		},
 		Spec: batchv1.JobSpec{
 			BackoffLimit: ptr.To(int32(3)),
 			Completions:  ptr.To(int32(1)),
 			Template: v1.PodTemplateSpec{
 				Spec: v1.PodSpec{
 					SecurityContext: &v1.PodSecurityContext{
+						RunAsNonRoot: ptr.To(true),
```
ahadas, line 150: how about adding a comment about the 'default' namespace, something like "specifying RunAsNonRoot=true without RunAsUser would fail the creation of the pod when migrating to the namespace 'default' in OpenShift but even if the pod starts, its execution would fail in this case"

pkg/controller/plan/adapter/converter.go
```diff
@@ -99,42 +95,62 @@ func (c *Converter) ConvertPVCs(pvcs []*v1.PersistentVolumeClaim, srcFormat srcF
 	return false, nil
 }
 
-func (c *Converter) ensureJob(pvc *v1.PersistentVolumeClaim, srcFormat, dstFormat string) (*batchv1.Job, error) {
-	jobName := getJobName(pvc, "convert")
-	job := &batchv1.Job{}
-	err := c.Destination.Client.Get(context.Background(), client.ObjectKey{Name: jobName, Namespace: pvc.Namespace}, job)
+func (c *Converter) deleteScratchDV(scratchDV *cdi.DataVolume) {
+	err := c.Destination.Client.Delete(context.Background(), scratchDV, &client.DeleteOptions{})
```
ahadas, line 99: can we drop the last argument here as well?
```suggestion
	err := c.Destination.Client.Delete(context.Background(), scratchDV)
```

pkg/controller/plan/adapter/converter.go
```diff
@@ -99,42 +95,62 @@ func (c *Converter) ConvertPVCs(pvcs []*v1.PersistentVolumeClaim, srcFormat srcF
...
+		c.Log.Info("Found convert job", "job", jobList.Items[0].Name)
+		return &jobList.Items[0], nil
+	} else if len(jobList.Items) > 1 {
+		return nil, liberr.New("multiple convert jobs found for pvc", "pvc", pvc.Name)
+	}
+
+	// Job doesn't exist, create it
+	job := createConvertJob(pvc, dv, srcFormat, dstFormat, c.Labels)
+	c.Log.Info("Creating convert job", "pvc", pvc.Name, "srcFormat", srcFormat, "dstFormat", dstFormat)
+	err = c.Destination.Client.Create(context.Background(), job, &client.CreateOptions{})
```
ahadas, line 124: and here?
```suggestion
	err = c.Destination.Client.Create(context.Background(), job)
```

pkg/controller/plan/adapter/converter.go
ahadas: cool thanks, filed https://issues.redhat.com/browse/MTV-964 to remember to document it as a known issue

ahadas (approved):

ahadas:
@bennyz please backport to release-v2.6.0

PR #770, opened 2024-03-03
ova: change timeout for PVC creation to wait until bind.

bkhizgiy:


pkg/controller/plan/kubevirt.go
```diff
@@ -2276,19 +2276,20 @@ func (r *KubeVirt) CreatePvcForNfs(pvcNamePrefix string, pvName string) (pvcName
...
-		if err != nil {
-			r.Log.Error(err, "Failed to get OVA plan PVC")
-			return false, err
-		}
-		return pvc.Status.Phase == core.ClaimBound, nil
-
-	}); err != nil {
-		r.Log.Error(err, "Failed to bind OVA PVC to PV ")
-		return
+	go func() {
```
bennyz, line 2279: I think this doesn't actually do anything, I mean, it checks for the status of the PVC asynchronously, but nothing is changed based on the result

bkhizgiy: you right, but I saw that I don't get any errors with that implementation and then I realize that the current code is redundant since the virt-v2v pod knows how to handle this behavior by itself.

I dink the best solution will be to remove entirely this check from the code but the downside will be we don't have the error in the case the pod stack on pending....

pkg/controller/plan/kubevirt.go
```diff
@@ -39,6 +38,7 @@ import (
 	libcnd "github.com/konveyor/forklift-controller/pkg/lib/condition"
 	liberr "github.com/konveyor/forklift-controller/pkg/lib/error"
 	core "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
liranr23, line 41: exact import as `core` one line above

ahadas, line 41: we already import that package one line above

pkg/controller/plan/kubevirt.go
```diff
@@ -834,6 +834,56 @@ func (r *KubeVirt) EnsureGuestConversionPod(vm *plan.VMStatus, vmCr *VirtualMach
 	return
 }
 
+func (r *KubeVirt) EnsureVirtV2VPVCStatus() (ready bool, err error) {
+	pvcs := &core.PersistentVolumeClaimList{}
+	pvcLabels := map[string]string{
+		"migration": r.Migration.Name,
+		"ova":       "nfs-pvc",
```
liranr23, line 841: `nfs-pvc` is used 3 times (including this change) in this file. how about making it a const?

pkg/controller/plan/kubevirt.go
```diff
@@ -834,6 +834,56 @@ func (r *KubeVirt) EnsureGuestConversionPod(vm *plan.VMStatus, vmCr *VirtualMach
...
+			Namespace:     r.Plan.Spec.TargetNamespace,
+		},
+	)
+	if err != nil {
+		return false, err
+	}
+
+	if len(pvcs.Items) == 0 {
+		return
+	}
```
liranr23, line 858: when can it happen? what does it mean if we have 0 pvcs associated with the vm we import?

pkg/controller/plan/kubevirt.go
```diff
@@ -834,6 +834,56 @@ func (r *KubeVirt) EnsureGuestConversionPod(vm *plan.VMStatus, vmCr *VirtualMach
...
+
+	pvc := &core.PersistentVolumeClaim{}
+	if len(pvcs.Items) > 1 {
+		for _, pvcVirtV2v := range pvcs.Items {
+			if pvcVirtV2v.CreationTimestamp.Time.After(r.Migration.CreationTimestamp.Time) {
+				pvc = &pvcVirtV2v
```
liranr23, line 864: what are we trying to achieve here? the last PVC that was made?

pkg/controller/plan/kubevirt.go
```diff
@@ -834,6 +834,56 @@ func (r *KubeVirt) EnsureGuestConversionPod(vm *plan.VMStatus, vmCr *VirtualMach
 	return
 }
 
+func (r *KubeVirt) EnsureVirtV2VPVCStatus() (ready bool, err error) {
```
ahadas, line 837: the name of this function is not specific to OVA but inside we set `"ova": "nfs-pvc"`  so the name should change to reflect that

pkg/controller/plan/kubevirt.go
```diff
@@ -834,6 +834,56 @@ func (r *KubeVirt) EnsureGuestConversionPod(vm *plan.VMStatus, vmCr *VirtualMach
...
+	err = r.Destination.Client.List(
+		context.TODO(),
+		pvcs,
+		&client.ListOptions{
+			LabelSelector: labels.SelectorFromSet(pvcLabels),
+			Namespace:     r.Plan.Spec.TargetNamespace,
+		},
+	)
+	if err != nil {
+		return false, err
```
ahadas, line 853: why not to simply return then? or shall we wrap the error first?

pkg/controller/plan/kubevirt.go
```diff
@@ -834,6 +834,56 @@ func (r *KubeVirt) EnsureGuestConversionPod(vm *plan.VMStatus, vmCr *VirtualMach
...
+	}
+
+	if len(pvcs.Items) == 0 {
+		return
+	}
+
+	pvc := &core.PersistentVolumeClaim{}
+	if len(pvcs.Items) > 1 {
+		for _, pvcVirtV2v := range pvcs.Items {
+			if pvcVirtV2v.CreationTimestamp.Time.After(r.Migration.CreationTimestamp.Time) {
```
ahadas, line 863: this code deserves a comment - is it because of leftovers we may have from previous migration attempts?

pkg/controller/plan/kubevirt.go
```diff
@@ -834,6 +834,56 @@ func (r *KubeVirt) EnsureGuestConversionPod(vm *plan.VMStatus, vmCr *VirtualMach
 	return
 }
 
+func (r *KubeVirt) EnsureVirtV2VPVCStatus() (ready bool, err error) {
+	pvcs := &core.PersistentVolumeClaimList{}
+	pvcLabels := map[string]string{
+		"migration": r.Migration.Name,
```
liranr23, line 840: don't we have also vmID label? that would help to minimize the PVC list and select the specific PVC of a VM we are now looking into. here we will get a list of all the PVCs associated with the migration.

ahadas (commented):
that's a better approach so generally it looks good but added few comments inside

pkg/controller/plan/kubevirt.go
ahadas: right, missed that one - shouldn't we create the PVC in this case?

bkhizgiy: maybe only for the first iteration when we sent the crate request for  the PVC but it didn't created yet

pkg/controller/plan/kubevirt.go
bkhizgiy: yes, for any case if we will have a list with more then one PVC filter the correct one by date, although it should not happen since we filter the list by migration name so we should have only one PVC at all times.

pkg/controller/plan/kubevirt.go
bkhizgiy: yep exactly, it should't happen because we filter the PVC by migration name but still for any case, I"ll add comment.

pkg/controller/plan/kubevirt.go
bkhizgiy: we didn't but I've added it now since it make more sense.

pkg/controller/plan/kubevirt.go
liranr23: but since you lookup for the all PVCs of the migration and not of the VM you may get a PVC that isn't related to your VM.
Also, it will always pick the last PVC that was made after the migration did? you don't compare the creation time of the PVCs between them.

`it should not happen since we filter the list by migration name so we should have only one PVC at all times.` - i didn't get that part.

pkg/controller/plan/kubevirt.go
ahadas: this thread makes me think that we better call `EnsureVirtV2VPVCStatus` from `EnsureGuestConversionPod`..

bkhizgiy: yeah, its sounds like a better solution and then we will query the pod status entirely rather than the PVC only. moved into `EnsureGuestConversionPod`.

pkg/controller/plan/kubevirt.go
bkhizgiy: yes the vmID was a good point so I've added that label to the PVC (I comment on that in one of the other threats:))
and after that we really will have only one PVC ( since we creating only one per vm per migration) and in case we don't - if some leftovers will be cough up - we will pick up the last one - which mean the newer one created.

liranr23: we don't support multiple disks VM when importing from OVA?

bkhizgiy: we do, this is the PVC that we used as nfs to get the ova/ovf source for the migration, not the disks of the VM. so from this type we have only one per migration attached to the virt-v2v pod.

pkg/controller/plan/kubevirt.go
```diff
@@ -2243,9 +2301,9 @@ func (r *KubeVirt) CreatePvForNfs() (pvName string, err error) {
 	return
 }
 
-func (r *KubeVirt) CreatePvcForNfs(pvcNamePrefix string, pvName string) (pvcName string, err error) {
+func (r *KubeVirt) CreatePvcForNfs(pvcNamePrefix, pvName, vmID string) (pvcName string, err error) {
 	sc := ""
-	labels := map[string]string{"provider": r.Plan.Provider.Source.Name, "app": "forklift", "migration": r.Migration.Name, "plan": string(r.Plan.UID), "ova": "nfs-pvc"}
+	labels := map[string]string{"provider": r.Plan.Provider.Source.Name, "app": "forklift", "migration": string(r.Migration.UID), "plan": string(r.Plan.UID), "ova": OvaPVCLabel, "vmID": vmID}
```
liranr23, line 2306: `kVM` is also possible here

liranr23 (approved):

pkg/controller/plan/migration.go
```diff
@@ -1095,12 +1095,16 @@ func (r *Migration) execute(vm *plan.VMStatus) (err error) {
...
-		err = r.ensureGuestConversionPod(vm)
+		ready, err := r.ensureGuestConversionPod(vm)
 		if err != nil {
 			step.AddError(err.Error())
 			err = nil
 			break
 		}
+		if !ready {
+			r.Log.Info("virt-v2v pod isn't ready yet")
+			return nil
```
ahadas, line 1106:
```suggestion
			return
```

bkhizgiy: the error var is not in the same scope here so we need to return nil manually.

bkhizgiy:
@ahadas lgtm

liranr23:
> @bkhizgiy @liranr23 please check the changes in the last commit

LGTM

PR #773, opened 2024-03-03
Fix validation of plans with an invalid source provider

ahadas:
When validating a plan we have specific logic that applies for source providers of type OpenShift but the check of the type of the source provider didn't take into consideration that the source provider may be invalid, leading to a sig fault. Here we extend the function that validates the source providers of type OpenShift with nil-check of the source provider as done in other validation function for a plan, and also nil-check of the destination provider.

### DIFF
--- a/pkg/controller/plan/validation.go
+++ b/pkg/controller/plan/validation.go
@@ -147,15 +147,19 @@ func (r *Reconciler) validate(plan *api.Plan) error {
 	}
 
 	// Validate version only if migration is OCP to OCP
-	err = r.validateOCPVersion(plan)
-	if err != nil {
+	if err := r.validateOpenShiftVersion(plan); err != nil {
 		return err
 	}
 
 	return nil
 }
 
-func (r *Reconciler) validateOCPVersion(plan *api.Plan) error {
+func (r *Reconciler) validateOpenShiftVersion(plan *api.Plan) error {
+	source := plan.Referenced.Provider.Source
+	if source == nil {
+		return nil
+	}
+
 	if plan.IsSourceProviderOCP() && plan.Provider.Destination.Type() == api.OpenShift {
 		unsupportedVersion := libcnd.Condition{
 			Type:     unsupportedVersion,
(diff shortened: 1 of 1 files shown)


liranr23:
`We not check the SourceProviderNotValid
condition to prevent this.`
did you mean: We did not... ?

pkg/controller/plan/validation.go
```diff
@@ -147,16 +147,15 @@ func (r *Reconciler) validate(plan *api.Plan) error {
...
 		return err
 	}
 
 	return nil
 }
 
-func (r *Reconciler) validateOCPVersion(plan *api.Plan) error {
-	if plan.IsSourceProviderOCP() && plan.Provider.Destination.Type() == api.OpenShift {
+func (r *Reconciler) validateOpenShiftVersion(plan *api.Plan) error {
+	if !plan.Status.HasCondition(validation.SourceProviderNotValid) && plan.IsSourceProviderOCP() && plan.Provider.Destination.Type() == api.OpenShift {
```
bennyz, line 158: This happens because we reference a provider that doesn't exist, why is it specific to ocp?
I think we shouldn't even reach here in this case

ahadas: you're right but this is the only verification that refers to the source provider

ahadas:
> `We not check the SourceProviderNotValid condition to prevent this.` did you mean: We did not... ?

no, I meant "We now check ..." - I'll fix that

pkg/controller/plan/validation.go
ahadas: it doesn't matter much but I've extracted that check to the `validate` function

ahadas: actually, changed it again to align this validation functions with the other ones that don't face this issue...

bennyz (approved):

liranr23 (approved):
i think the commit doesn't describes what really done here, but other than that LGTM

ahadas:
> i think the commit doesn't describes what really done here, but other than that LGTM

ah yeah, because the solution has changed, ack

PR #776, opened 2024-03-03
vSphere: Fix security context of VDDK validation pod

ahadas:
The previous attempt to fix the execution of the VDDK validation pod didn't succeed - it allowed to run it on the 'default' namespace but its execution failed on other restricted namespaces in which the QEMU user (107) is not included in the allowed range.

In order to fix this properly, we differentiate OpenShift from k8s: in case of OpenShift we don't specify RunAsNonRoot and RunAsUser in order to let the platform choose an appropriate user, and in case of k8s, we specify the QEMU user (107)


ahadas:
ok, apparently on k8s we can't avoid setting it so we'd need to differentiate k8s from OpenShift....

ahadas:
I don't expect the latest changes to pass our CI, need additional changes for k8s

PR #778, opened 2024-03-05
Change created virt-v2v conversion output to -o kubeVirt

bkhizgiy:
Currently, the output produced by virt-v2v is in libvirt XML format. However, since we are working with KubeVirt VMs and in light of recent changes by the virt-v2v team to add missing fields to the YAML output, we are switching the output format to -o kubevirt. This change will allow us to use this information to create the VM in OpenShift and capture some missing fields about the VMs. As part of this update, if a VM name is not valid, we will generate a new name and pass it to the conversion pod, where it will be updated as part of the virt-v2v command, as well as within the migration controller.


virt-v2v/cold/entrypoint.go
```diff
@@ -169,16 +175,26 @@ func LinkDisks(diskKind string, num int) (err error) {
...
-	tee := io.TeeReader(virtV2vStdoutPipe, os.Stdout)
-	virtV2vCmd.Stderr = os.Stderr
+	var teeErr io.Reader
+	if source == OVA {
+		virtV2vStderrPipe, err := virtV2vCmd.StderrPipe()
+		if err != nil {
+			fmt.Printf("Error setting up stdout pipe: %v\n", err)
+			return err
+		}
+		teeErr = io.TeeReader(virtV2vStderrPipe, os.Stderr)
```
liranr23, line 194: what's the benefit of it? will the stderr will still be shown in the logs?

i see you use it later on for scanning - do we expect the see the expected value in an error?

pkg/controller/plan/kubevirt.go
```diff
@@ -878,17 +877,15 @@ func (r *KubeVirt) UpdateVmByConvertedConfig(vm *plan.VMStatus, pod *core.Pod, s
...
 	}
 
-	firmware, err := ovfparser.GetFirmwareFromConfig(string(vmConfigXML))
-	if err != nil {
-		return
-	}
-
-	vm.Firmware = firmware
+	vm.Firmware = string(vmFirmware)
+	r.Log.Info("Settig the vm fimware",
```
liranr23, line 886:
```suggestion
	r.Log.Info("Setting the vm fimware",
```

virt-v2v/cold/entrypoint.go
bkhizgiy: Yes, all the extra detailed log of virt-v2v is directed to stdErr so we will see the firmware detection there, the virt-v2v standard output is only the stages of the conversion and the disk copy percentages.

bkhizgiy: here is an example of the command output to understand it better, all other information is redirected to `Stderr`

`[   0.0] Setting up the source: -i ova /home/bkhizgiy/local_code_test/rhel9-efi.ova
virt-v2v: warning: manifest has a checksum for non-existent file 1nisim-rhel9-efi-2.iso (ignored)
[   6.5] Opening the source
[  10.6] Inspecting the source
[  14.1] Checking for sufficient free disk space in the guest
[  14.1] Converting Red Hat Enterprise Linux 9.3 Beta (Plow) to run on KVM
virt-v2v: The QEMU Guest Agent will be installed for this guest at first boot.
virt-v2v: This guest has virtio drivers installed.
[  67.7] Mapping filesystem data to avoid copying unused and blank areas
[  68.3] Closing the overlay
[  68.5] Assigning disks to buses
[  68.5] Checking if the guest needs BIOS or UEFI to boot
[  68.5] Setting up the destination: -o disk -os /var/tmp/v2v
[  69.6] Copying disk 1/1
[  86.6] Creating output metadata
[  86.6] Finishing off
`

liranr23: this is an example output after the change?
we do wish to have the whole logs

bkhizgiy: nono, only for the Stdout, so it would be more clear why we need to scan the Stderr (since all the information is there :)), the pod output will remain the same as it was.

pkg/controller/plan/kubevirt.go
```diff
@@ -860,7 +859,7 @@ func (r *KubeVirt) UpdateVmByConvertedConfig(vm *plan.VMStatus, pod *core.Pod, s
 		return
 	}
 
-	url := fmt.Sprintf("http://%s:8080/ovf", pod.Status.PodIP)
+	url := fmt.Sprintf("http://%s:8080/firmware", pod.Status.PodIP)
```
ahadas, line 862: why changing it back to the firmware?
let's get @rwmjones's ack on our intention and if we're aligned with how it will be reported by virt-v2v in future versions, we could minimize the changes here

@rwmjones [when migrating to RHV, virt-v2v sets the bios-type that includes the firmware type in the generated ovf](https://github.com/libguestfs/virt-v2v/commit/c4d2800d90bf97118eef9a091a461126fad979a8), but MTV uses `-o local` so the generated OVF doesn't include it. would it make sense to add the firmware/bios-type also to `-o local`?

rwmjones: Sighs .. we actually have an `-o kubevirt` mode already which is really what you should use, and fix if it doesn't work.

However about this question, `-o local` doesn't set the firmware but it probably should do.  As always, we need a bug to track this.

bkhizgiy: @rwmjones Thank you for the feedback. I will open an issue to address this.

Initially, we explored the `-o kubevirt` option but found that it provides less information than the `-o local` option. Therefore, instead of using the generated YAML directly, we thought of extracting the missing information from the generated OVF file to use during VM creation.

Additionally, it's worth mentioning that our tests with the kubevirt option revealed it also does not include the firmware type field.

rwmjones: The way we want this to work if that you tell us what's missing from `-o kubevirt` and we'll add it.  That makes virt-v2v better too because it means people can use it directly for small conversions.  What do you need and what is missing?

bkhizgiy: I'm not entirely sure of the specifics, but @nyoxi  investigated this [issue](https://issues.redhat.com/browse/MTV-513) a while ago and found a gap between the provided configuration and our requirements.

With the proposed changes to the -o kubevirt fields, this could be a really interesting direction to explore. We can begin by adding the firmware type, as it's our most urgent need, and use it as a baseline for incorporating other missing configurations in the future.

rwmjones: Thanks - please file bug(s) in Jira or upstream about any features you need.

bkhizgiy: Thanks, just created this issue for beginning https://github.com/libguestfs/virt-v2v/issues/45

rwmjones: Actually I should have said in Jira **and** upstream, since I guess you will want this to be in RHEL and without a Jira bug it won't get there.

ahadas: @bkhizgiy @rwmjones let me try to scope this down:
1. On the one extreme, letting virt-v2v to post the VM (that's an idea we had long time ago) is complicated because we need to provide a service account, possibly on a remote cluster and breaks the generalization in Forklift.
2. On the other extreme, we cannot ignore the output of virt-v2v as Forklift did so far - here we have a concrete case we need this information.

Forklift currently build the entire VM configuration and with this change take the firmware produced by virt-v2v - I think we should still return the entire configuration that virt-v2v produces. There are two options:
1. Get the OVF virt-v2v produces with `-o local`
2. Get a YAML virt-v2v produces with `-o kubevirt`

The OVF includes more properties but to solve this issue we only need the firmware.
I suggest that we'll change virt-v2v command to include `-o kubevirt` and return that YAML:
1. Logically, we migrate to KubeVirt...
2. We can simplify parsing in Forklift by creating a VirtualMachine from that input (rather than parsing OVF)
3. As we go, we can add stuff to the YAML (in virt-v2v) until we take that as a baseline and adjust it, if needed, based on input from the user

If you agree with me so far, how about letting the conversion pod in Forklift add the firmware configuration to the YAML until we upgrade virt-v2v to a version in which it does it by its own? There won't be much room for mistake because we'll follow the format that KubeVirt stores this property within the YAML (with `cnv.Bootloader{BIOS: &cnv.BIOS{}}` or `cnv.Bootloader{ EFI: &cnv.EFI{ SecureBoot: ptr.To(false)}}`)

bkhizgiy: @ahadas Sounds good to me. It makes more sense to use the `-o kubevirt` option as we discussed yesterday, so if it's agreed, I'll make the suggested changes.

rwmjones: Sounds good here too.

virt-v2v/cold/entrypoint.go
```diff
@@ -299,3 +296,67 @@ func isValidSource(source string) bool {
 		return false
 	}
 }
+
+func addFirmwareToYaml(filePath string) (err error) {
```
bkhizgiy, line 300: Since virt-v2v currently set in a different workspace I was not able to use the yaml package so did a simple file write instead.

pkg/controller/plan/adapter/ova/kubevirtvmparser.go
```diff
@@ -0,0 +1,56 @@
...
+	EFI  *EFI  `yaml:"efi,omitempty"`
+}
+
+type Bios struct{}
+
+type EFI struct {
+	SecureBoot bool `yaml:"secureBoot"`
+}
+
+func ReadConfFromYaml(yamlData []byte) (firmware string, err error) {
```
ahadas, line 41: the name of this function is not that great since it returns firmware and not vm-configuration

virt-v2v/cold/entrypoint.go
```diff
@@ -196,47 +212,72 @@ func executeVirtV2v(args []string) (err error) {
...
+	if yamlFilePath == "" {
+		fmt.Println("Error: YAML file path is empty.")
+		http.Error(w, "YAML file path is empty", http.StatusInternalServerError)
 		return
 	}
 
-	xmlData, err := os.ReadFile(xmlFilePath)
+	err := addFirmwareToYaml(yamlFilePath)
+	if err != nil {
+		fmt.Printf("Error setting yaml file: %v\n", err)
```
ahadas, line 262: setting yaml file? you mean modify/change the yaml file (or vm configuration), no?

virt-v2v/cold/entrypoint.go
ahadas: yeah, another reason to move it to the main workspace - but we'll do that as a separate change later on

pkg/controller/plan/kubevirt.go
```diff
@@ -958,24 +957,24 @@ func (r *KubeVirt) UpdateVmByConvertedConfig(vm *plan.VMStatus, pod *core.Pod, s
...
-
 	switch r.Source.Provider.Type() {
 	case api.Ova:
-		if vm.Firmware, err = ovfparser.GetFirmwareFromConfig(vmConfigXML); err != nil {
+		if vm.Firmware, err = util.GetFirmwareFromYaml(vmConf); err != nil {
 			err = liberr.Wrap(err)
 			return
 		}
+		r.Log.Info("Setting the vm firmware ", vm.Firmware, "vmId", vm.ID)
 	case api.VSphere:
```
bkhizgiy, line 972: @ahadas @liranr23 Shouldn't this be applicable for OVA as well?

bkhizgiy: yep, from testing with kubevirt this filed exist so adding for OVA as well.

liranr23: then it will be nice to have also the mapping and setting the OS for OVA, as we don't do that now from what i remember

operator/config/crd/bases/forklift.konveyor.io_migrations.yaml
```diff
@@ -380,6 +380,9 @@ spec:
                     operatingSystem:
                       description: The Operating System detected by virt-v2v.
                       type: string
+                    newName:
+                      description: The new name of the VM after matching DNS1123 requirements.
+                      type: string
```
liranr23, line 385: did you automatically generated it? seems CI fails on this.

pkg/controller/plan/kubevirt.go
```diff
@@ -958,24 +957,24 @@ func (r *KubeVirt) UpdateVmByConvertedConfig(vm *plan.VMStatus, pod *core.Pod, s
...
-		if vm.OperatingSystem, err = ovfparser.GetOperationSystemFromConfig(vmConfigXML); err != nil {
+		if vm.OperatingSystem, err = util.GetOperationSystemFromYaml(vmConf); err != nil {
```
liranr23, line 973: so we don't get it for OVA?

pkg/controller/plan/util/kubevirtvmparser.go
```diff
@@ -0,0 +1,135 @@
...
+	"fmt"
+	"strings"
+
+	"github.com/konveyor/forklift-controller/pkg/lib/logging"
+	"gopkg.in/yaml.v2"
+)
+
+const (
+	// Name.
+	Name = "virt-v2v-parser"
```
liranr23, line 13: up to you - aligning with the file name

virt-v2v/cold/entrypoint.go
```diff
@@ -114,7 +116,14 @@ func buildCommand() []string {
 		fmt.Println("Error creating directory  ", err)
 		os.Exit(1)
 	}
-	virtV2vArgs = append(virtV2vArgs, "-o", "local", "-os", DIR)
+	virtV2vArgs = append(virtV2vArgs, "-o", "kubevirt")
+
+	if checkEnvVariablesSet("V2V_NewName") {
+		virtV2vArgs = append(virtV2vArgs, "-on", os.Getenv("V2V_NewName"))
+		nameChanged = true
+	}
```
liranr23, line 124: what's the benefit letting virt-v2v change the VM name during conversion?

virt-v2v/cold/entrypoint.go
```diff
@@ -114,7 +116,14 @@ func buildCommand() []string {
...
+
+	virtV2vArgs = append(virtV2vArgs, "-os", DIR)
```
liranr23, line 126: i find the split of these args from above unnecessary, but i'm not strongly against it.

liranr23 (commented):
Please add/fix the commits messages, I can't tell why the `newName` is in use.

virt-v2v/cold/entrypoint.go
rwmjones: It doesn't especially make any different for `-o kubevirt` mode, but if you're interested in what it actually affects, search for `output_name` in these two files:
https://github.com/libguestfs/virt-v2v/blob/master/output/create_kubevirt_yaml.ml
https://github.com/libguestfs/virt-v2v/blob/master/output/output_kubevirt.ml

liranr23: Thanks @rwmjones.
Based on what I see, the main reason for it is when we have a name of the RFC1123 standard which cause an error in the YAML creation(the same reason we change it on our end).
@bkhizgiy can you please add a comment on this and within the commit?

bkhizgiy: Yes, as @liranr23 mentioned, since the output is in kubevirt format, it should meet the k8s DNS1123 RFC requirements. The creation of the produced VM YAML fails when using an unsupported name. We are already changing the name in that case, and additionally propagating the value to virt-v2v to overcome this.

And sure, I'll add a comment.

virt-v2v/cold/entrypoint.go
bkhizgiy: the order meters in this case IIRC, so it was divided due to that.

virt-v2v/cold/entrypoint.go
rwmjones: Is that a bug in virt-v2v, eg is it generating the yaml quoting wrongly?  Here's that code: https://github.com/libguestfs/virt-v2v/blob/e9c0d63b22bcb4c31cfaac00fc1271713c8bd3b3/lib/YAML.ml#L126

virt-v2v/cold/entrypoint.go
rwmjones: Order doesn't matter (except obviously things like `-os` needs to have the directory parameter after).  On the other hand, having the order `-i ... -o ...` does make it easier for humans to read.

virt-v2v/cold/entrypoint.go
bkhizgiy: @rwmjones Nope, it's the desired behavior. :) There is a validation on the name before the YAML creation. Otherwise, what will happen is that once we try to apply the YAML with the original name, the VM creation will fail because the name is illegal.

pkg/controller/plan/kubevirt.go
bkhizgiy: right, removing the if.

operator/config/crd/bases/forklift.konveyor.io_migrations.yaml
bkhizgiy: yes it automatically generated, I"ll look into it

liranr23 (approved):

pkg/controller/plan/kubevirt.go
```diff
@@ -958,24 +957,24 @@ func (r *KubeVirt) UpdateVmByConvertedConfig(vm *plan.VMStatus, pod *core.Pod, s
...
-	vmConfigXML := string(vmConfigBytes)
-
 	switch r.Source.Provider.Type() {
 	case api.Ova:
-		if vm.Firmware, err = ovfparser.GetFirmwareFromConfig(vmConfigXML); err != nil {
+		if vm.Firmware, err = util.GetFirmwareFromYaml(vmConf); err != nil {
 			err = liberr.Wrap(err)
 			return
 		}
-	case api.VSphere:
```
ahadas, line 974: so wouldn't it be missing for vSphere?

pkg/controller/plan/kubevirt.go
```diff
@@ -1721,6 +1717,13 @@ func (r *KubeVirt) guestConversionPod(vm *plan.VMStatus, vmVolumes []cnv.Volume,
 				Value: vm.RootDisk,
 			})
 	}
+
+	environment = append(environment,
+		core.EnvVar{
+			Name:  "V2V_NewName",
+			Value: vm.NewName,
```
ahadas, line 1724: shouldn't we set it only when a new name is set? you check if V2V_NewName is set in the entrypoint of the conversion pod

ahadas (commented):
looks good to me in general

bkhizgiy:
@mnecas @yaacov There was a conflicts with some of the recent changes. Can you please verify that nothing is missing with the switch to the YAML file? Thanks!

yaacov:
> Can you please verify that nothing is missing with the switch to the YAML file? Thanks!

The xml to yaml converion in entrypoint file, looks good to me :+1:

mnecas:
Continuing the PR in the https://github.com/kubev2v/forklift/pull/1020
I have rebased it with the latest virt-v2v wrapper changes and made some additional changes
Closing

PR #779, opened 2024-03-05
extract secret setup

bennyz:
Move secret setup to the top, as VM validation is skipped when the plan is executing which can lead to an NPE in checks that rely on it


pkg/controller/plan/validation.go
```diff
@@ -322,6 +331,7 @@ func (r *Reconciler) validateStorageMap(plan *api.Plan) (err error) {
 // Validate listed VMs.
 func (r *Reconciler) validateVM(plan *api.Plan) error {
 	if plan.Status.HasCondition(Executing) {
+
```
liranr23, line 334: redundant

pkg/controller/plan/validation.go
```diff
@@ -457,14 +467,7 @@ func (r *Reconciler) validateVM(plan *api.Plan) error {
...
 		}
-		if plan.Referenced.Secret == nil &&
-			// local cluster has no secret
-			!plan.Referenced.Provider.Source.IsHost() {
-			err = r.setupSecret(plan)
-			if err != nil {
-				return err
-			}
-		}
+
```
liranr23, line 470: redundant

liranr23 (approved):

pkg/controller/plan/validation.go
```diff
@@ -114,6 +114,15 @@ func (r *Reconciler) validate(plan *api.Plan) error {
 	plan.Referenced.Provider.Source = pv.Referenced.Source
 	plan.Referenced.Provider.Destination = pv.Referenced.Destination
 
+	if plan.Referenced.Secret == nil &&
```
ahadas, line 117: wouldn't plan.Referenced.Secret always be nil when we get here?
additionally, if we don't use the secret in the validations (we shouldn't - the provider's state should cover the connectivity with the credentials within the secret) then why not placing it in the execute function? we'd save a call in case some validation fails..

bennyz: what execute function? we need this for the validation

ahadas: ah for the validation of the openshift cluster?

ahadas (approved):

pkg/controller/plan/validation.go
ahadas: ok, we still trigger a possibly redundant call here but at least we fixed a possible sig fault when the source provider is not set here - that's good enough

pkg/controller/plan/validation.go
```diff
@@ -191,6 +197,19 @@ func (r *Reconciler) validateOpenShiftVersion(plan *api.Plan) error {
 	return nil
 }
 
+func (r *Reconciler) ensureSecretForProvider(plan *api.Plan) error {
+	if plan.Referenced.Provider.Source != nil &&
+		plan.Referenced.Secret == nil &&
```
ahadas, line 202: unrelated to this PR but it's really not expressive enough to have a single secret on plan.Referenced since it's hard to say whether it refers to the source or the destination provider...

PR #780, opened 2024-03-05
OVA: match the correct vm for the firmware value

bkhizgiy:
currently for some cases the vm is not being match correctly in the builder which leads to not fetching the firmware value set by virt-v2v, which lead to that the default firmware value being set for bios VMs.

* need to be verified -  the bug is currently randomly reproducing.

### DIFF
--- a/pkg/controller/plan/adapter/ova/builder.go
+++ b/pkg/controller/plan/adapter/ova/builder.go
@@ -342,7 +342,7 @@ func (r *Builder) mapFirmware(vm *model.VM, vmRef ref.Ref, object *cnv.VirtualMa
 			}
 		}
 		if virtV2VFirmware == "" {
-			r.Log.Info("failed to match the vm, model VMid is %s, vmRef ID is %s", vm.ID, vmRef.ID)
+			r.Log.Info("failed to match the vm", "model ID", vm.ID, "vmRef ID", vmRef.ID)
 		}
 	} else {
 		virtV2VFirmware = vm.Firmware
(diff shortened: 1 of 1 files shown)


pkg/controller/plan/adapter/ova/builder.go
```diff
@@ -332,15 +332,18 @@ func (r *Builder) mapCPU(vm *model.VM, object *cnv.VirtualMachineSpec) {
...
 	if vm.Firmware == "" {
 		for _, vmConf := range r.Migration.Status.VMs {
-			if vmConf.ID == vm.ID {
+			if vmConf.ID == vmRef.ID {
 				virtV2VFirmware = vmConf.Firmware
 				break
 			}
 		}
+		if virtV2VFirmware == "" {
+			r.Log.Info("failed to match the vm, model VMid is %s, vmRef ID is %s", vm.ID, vmRef.ID)
```
liranr23, line 345:
```suggestion
			r.Log.Info("failed to match the vm, model vmID is %s, vmRef ID is %s", vm.ID, vmRef.ID)
```
or `vm id`? this type of `VMid` is unusual

liranr23 (commented):
i find it very strange that the model vm id differs from the reference.
do you have a clue how or why it happened?

pkg/controller/plan/adapter/ova/builder.go
ahadas: yep, should be vmID o vmId

ahadas: yep, should be vmID o vmId
```suggestion
			r.Log.Info("failed to match the vm", "model vmId", vm.ID, "vmRef ID", vmRef.ID)
```

ahadas (approved):

PR #787, opened 2024-03-06
OVA: map vm configuration from extraConfig in ovf file

bkhizgiy:
In some of the OVF file the configuration can be found under `ExtraConfig` rather that `Config`, currently the first option isnt supported in the OVA server and we may miss some impotent information,  this PR adds it.

example:
`<vmw:Config ovf:required="false" vmw:key="cpuHotRemoveEnabled" vmw:value="false"/>`
`<vmw:ExtraConfig ovf:required="false" vmw:key="firmware" vmw:value="efi"/>`

### DIFF
--- a/cmd/ova-provider-server/ova-provider-server.go
+++ b/cmd/ova-provider-server/ova-provider-server.go
@@ -513,33 +513,28 @@ func convertToVmStruct(envelope []Envelope, ovaPath []string) ([]VM, error) {
 	return vms, nil
 }
 
-func applyConfiguration(newVM *VM, confs []VirtualConfig) {
-	for _, conf := range confs {
-		switch conf.Key {
-		case "firmware":
-			newVM.Firmware = conf.Value
-		case "memoryHotAddEnabled":
-			newVM.MemoryHotAddEnabled, _ = strconv.ParseBool(conf.Value)
-		case "cpuHotAddEnabled":
-			newVM.CpuHotAddEnabled, _ = strconv.ParseBool(conf.Value)
-		case "cpuHotRemoveEnabled":
-			newVM.CpuHotRemoveEnabled, _ = strconv.ParseBool(conf.Value)
-		}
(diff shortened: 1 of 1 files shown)


cmd/ova-provider-server/ova-provider-server.go
```diff
@@ -514,6 +513,36 @@ func convertToVmStruct(envelope []Envelope, ovaPath []string) ([]VM, error) {
...
+			newVM.MemoryHotAddEnabled, _ = strconv.ParseBool(conf.Value)
+		case "cpuHotAddEnabled":
+			newVM.CpuHotAddEnabled, _ = strconv.ParseBool(conf.Value)
+		case "cpuHotRemoveEnabled":
+			newVM.CpuHotRemoveEnabled, _ = strconv.ParseBool(conf.Value)
+		}
+	}
+}
+
+func applyExtraConfiguration(newVM *VM, confs []ExtraVirtualConfig) {
```
ahadas, line 531: I tried to find the difference between `applyExtraConfiguration` and `applyConfiguration` and I don't see any - if these functions are identical then let's keep `applyConfiguration` and call it twice

bkhizgiy: the only difference is the the object that sent to the func, I initially wanted to use the same object and func since they identical but we need a different xml annotation to get the data correctly.
I though of using an interface for both objects to reduce code duplication but it will be more added code than just this func, so i'm not sure what will be better....

ahadas: I see, ok - I didn't notice the type of the second argument..
I posted another commit to extract the common handling then

bkhizgiy:
@ahadas reviewed the changes you posted, looking good to me.

liranr23 (commented):
The configs can be set either in the `Config` or in the `ExtraConfig`. I think it depends on the vsphere version

bkhizgiy:
@ahadas created https://github.com/kubev2v/forklift/pull/809

PR #789, opened 2024-03-10
Do not override description of wrapped error

ahadas:
The liberr.Wrap function should get an odd number of arguments, otherwise the description of the specified error would be overridden (follow-up on #772).


liranr23 (approved):
I did get the error, but yeah the `"couldn't build the PVC"` is useless and don't align with the wrap.

PR #796, opened 2024-03-11
Add root=first to cold v2v command

liranr23:
When importing from vSphere a multiple-boot operation system v2v will ask the user which one to use. In our case virt-v2v isn't interactively. Therefore, we will default to the first boot device as we done previsouly.

### DIFF
--- a/virt-v2v/cold/entrypoint.go
+++ b/virt-v2v/cold/entrypoint.go
@@ -49,7 +49,7 @@ func main() {
 	fmt.Println("Preparing virt-v2v")
 	switch source {
 	case vSphere:
-		virtV2vArgs = append(virtV2vArgs, "-i", "libvirt", "-ic", os.Getenv("V2V_libvirtURL"))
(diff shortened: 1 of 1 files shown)


virt-v2v/cold/entrypoint.go
```diff
@@ -49,7 +49,7 @@ func main() {
 	fmt.Println("Preparing virt-v2v")
 	switch source {
 	case vSphere:
-		virtV2vArgs = append(virtV2vArgs, "-i", "libvirt", "-ic", os.Getenv("V2V_libvirtURL"))
+		virtV2vArgs = append(virtV2vArgs, "--root=first", "-i", "libvirt", "-ic", os.Getenv("V2V_libvirtURL"))
```
ahadas, line 52: shouldn't it be
```suggestion
		virtV2vArgs = append(virtV2vArgs, "--root first", "-i", "libvirt", "-ic", os.Getenv("V2V_libvirtURL"))
```

ahadas: I'm pretty sure it will work, but can you please run a real migration with this change? (the CI doesn't test a real conversion)

ahadas (approved):

PR #802, opened 2024-03-12
OVA: workaround for consuming firmware from virt-v2v

bkhizgiy:
The change to use `-o kubvirt` requires too many adjustments to the virt-v2v pod. For version 2.6, we will continue using -o local and add the firmware type to the created XML after it is detected from the virt-v2v output. Then, this value will be consumed in the controller to determine the missing firmware.

### DIFF
--- a/pkg/controller/plan/kubevirt.go
+++ b/pkg/controller/plan/kubevirt.go
@@ -941,7 +941,7 @@ func (r *KubeVirt) UpdateVmByConvertedConfig(vm *plan.VMStatus, pod *core.Pod, s
 		return
 	}
 
-	r.Log.Info("setting the vm firmware to ", firmware, "vmId", vm.ID)
+	r.Log.Info("setting vm firmware", "vmId", vm.ID, "firmware", firmware)
 
 	vm.Firmware = firmware
 
--- a/pkg/controller/plan/kubevirt.go
+++ b/pkg/controller/plan/kubevirt.go
@@ -913,7 +913,7 @@ func (r *KubeVirt) UpdateVmByConvertedConfig(vm *plan.VMStatus, pod *core.Pod, s
 		return
 	}
 
-	url := fmt.Sprintf("http://%s:8080/ovf", pod.Status.PodIP)
+	url := fmt.Sprintf("http://%s:8080/vm", pod.Status.PodIP)
 
 	/* Due to the virt-v2v operation, the ovf file is only available after the command's execution,
 	meaning it appears following the copydisks phase.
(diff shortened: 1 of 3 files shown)


pkg/controller/plan/kubevirt.go
```diff
@@ -941,6 +941,8 @@ func (r *KubeVirt) UpdateVmByConvertedConfig(vm *plan.VMStatus, pod *core.Pod, s
 		return
 	}
 
+	r.Log.Info("setting the vm firmware to ", firmware, "vmId", vm.ID)
```
ahadas, line 944: the number of parameters should be odd (message and then key-value pairs)

virt-v2v/cold/entrypoint.go
```diff
@@ -196,6 +212,26 @@ func executeVirtV2v(args []string) (err error) {
...
+		scanner := bufio.NewScanner(teeErr)
+		const maxCapacity = 1024 * 1024
+		buf := make([]byte, 0, 64*1024)
+		scanner.Buffer(buf, maxCapacity)
+
+		for scanner.Scan() {
+			line := scanner.Bytes()
+			if match := UEFI_RE.FindSubmatch(line); match != nil {
+				fmt.Println("UEFI firmware detected")
+				firmware = "efi"
```
ahadas, line 225: we can break here, right?

virt-v2v/cold/entrypoint.go
```diff
@@ -255,3 +297,66 @@ func isValidSource(source string) bool {
...
+	tempFilePath := filePath + ".tmp"
+
+	tempFile, err := os.Create(tempFilePath)
+	if err != nil {
+		return
+	}
+	defer tempFile.Close()
+
+	scanner := bufio.NewScanner(file)
+	domainFound := false
```
ahadas, line 328: this variable seems to be redundant

ahadas (commented):
minor comments, I'll post the changes

ahadas:
@bkhizgiy please review the changes in the second commit

virt-v2v/cold/entrypoint.go
```diff
@@ -255,3 +298,60 @@ func isValidSource(source string) bool {
...
+		return
+	}
+	defer tempFile.Close()
+
+	scanner := bufio.NewScanner(file)
+
+	for scanner.Scan() {
+		line := scanner.Text()
+
+		_, err = tempFile.WriteString(line + "\n")
```
bennyz, line 333:
```go
if _, err = tempFile.WriteString(line + "\n"); err != nil {
		return err
}
```

also below

virt-v2v/cold/entrypoint.go
```diff
@@ -255,3 +298,60 @@ func isValidSource(source string) bool {
 		return false
 	}
 }
+
+func addFirmwareToXml(filePath string) (err error) {
+	var newFirmwareData string
```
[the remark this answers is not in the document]
bkhizgiy: this can be simplified even more, i change based on your suggestion.

bkhizgiy:
@ahadas Looking good, I've added another small cleanup based on @bennyz suggestions.

ahadas (approved):

PR #808, opened 2024-03-14
OVA: Fix workload construction for validations and match vars names.

bkhizgiy:
Fix the workload construction for passing VM information to validations.

### DIFF
--- a/pkg/controller/provider/web/ova/workload.go
+++ b/pkg/controller/provider/web/ova/workload.go
@@ -8,6 +8,7 @@ import (
 	api "github.com/konveyor/forklift-controller/pkg/apis/forklift/v1beta1"
 	model "github.com/konveyor/forklift-controller/pkg/controller/provider/model/ova"
 	"github.com/konveyor/forklift-controller/pkg/controller/provider/web/base"
+	libmodel "github.com/konveyor/forklift-controller/pkg/lib/inventory/model"
 )
 
 // Routes.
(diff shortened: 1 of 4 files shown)


liranr23 (commented):
looks like the DB field doesn't use camel case (first letter without capital). wouldn't it make sense to check it there instead?

bkhizgiy:
@liranr23 the tags addition sort it out :)

liranr23 (approved):

pkg/controller/provider/web/ova/workload.go
```diff
@@ -60,6 +65,10 @@ func (h WorkloadHandler) Get(ctx *gin.Context) {
 	}
 	r := Workload{}
 	r.With(m)
+	err = r.Expand(db)
+	if err != nil {
+		return
```
ahadas, line 70: shouldn't we to set the status to http.StatusInternalServerError?

pkg/controller/provider/web/ova/workload.go
```diff
@@ -85,3 +94,15 @@ func (r *Workload) Link(p *api.Provider) {
...
 		})
 }
+
+// Expand the resource.
+func (r *Workload) Expand(db libmodel.DB) (err error) {
+	vm := &model.VM{
+		Base: model.Base{ID: r.ID},
+	}
+	err = db.Get(vm)
+	if err != nil {
```
ahadas, line 104: there's no reason to check this if we'll anyway return

pkg/controller/provider/web/ova/workload.go
bkhizgiy: isn't that covered by the defer inside this func? since we using the same `err` variable in the same scope I think it should be catched there.

ahadas: ah ok, didn't notice that part when reviewing the changes - so that's fine

ahadas (approved):

ahadas:
@bkhizgiy please backport to the release-v2.6.0 branch

bkhizgiy:
@ahadas created https://github.com/kubev2v/forklift/pull/810

PR #813, opened 2024-03-19
converter: add progress

bennyz:


pkg/controller/plan/util/openstack.go
```diff
@@ -28,3 +36,39 @@ func OpenstackVolumePopulator(image *openstack.Image, sourceUrl *url.URL, transf
...
+func CreateConversionTask(inventory web.Client, vmRef ref.Ref) (*plan.Task, error) {
+	workload := &model.Workload{}
+	err := inventory.Find(workload, vmRef)
+	if err != nil {
+		err = liberr.Wrap(err, "vm", vmRef.String())
+		return nil, err
+	}
+
+	if workload.ImageID == "" {
+		//nolint:nilnil
```
bennyz, line 49: perhaps I should use a sentinel error?

yaacov:
@bennyz @mnecas hi, this looks important, due to @bennyz capacity, do we want to update and merge this PR, or do a new PR, because updating this one is too time consuming, when we have more capacity ?

bennyz:
given the focus shifted away from openstack I would not consider this important

PR #815, opened 2024-03-19
Always bind WFFC for data volumes (backport 2.5)

liranr23:
When importing and using DVs, always set the annotation to bind WFFC. This commit removes the consumer pod for vsphere flows as we can use the annotation for it.


ahadas (commented):
backport of #811

PR #821, opened 2024-03-24
openstack: fix bootOrder setting

bennyz:
Currently, because bootOrder was declared outside the loop, it would keep the same value from a previous iteration, being set on a disk that shouldn't be set.


ahadas:
backport of #820

ahadas (approved):

PR #822, opened 2024-03-24
ovirt: Respect storage mapping

ahadas:
Previously, when migrating to the cluster Forklift is deployed on using a cold migration, all disks of the migrated VM were transferred to the same storage class, which is one of the target storage class that appear in the storage mapping that is set for the migration plan.

With these changes, each disk will be mapped to a target storage class based on its original storage domain and the storage mapping that is defined for the migration plan.

### DIFF
--- a/pkg/controller/plan/adapter/ovirt/builder.go
+++ b/pkg/controller/plan/adapter/ovirt/builder.go
@@ -192,9 +192,7 @@ func (r *Builder) DataVolumes(vmRef ref.Ref, secret *core.Secret, configMap *cor
 		mapped := &dsMapIn[i]
 		ref := mapped.Source
 		sd := &model.StorageDomain{}
-		fErr := r.Source.Inventory.Find(sd, ref)
-		if fErr != nil {
-			err = fErr
+		if err = r.Source.Inventory.Find(sd, ref); err != nil {
 			return
 		}
 		for _, da := range vm.DiskAttachments {
@@ -715,6 +713,7 @@ func (r *Builder) PopulatorVolumes(vmRef ref.Ref, annotations map[string]string,
 		return
 	}
 
+	var sdToStorageClass map[string]string
 	for _, diskAttachment := range workload.DiskAttachments {
 		if diskAttachment.Disk.StorageType == "lun" {
 			continue
@@ -732,7 +731,12 @@ func (r *Builder) PopulatorVolumes(vmRef ref.Ref, annotations map[string]string,
 				return
 			}
 			var pvc *core.PersistentVolumeClaim
-			storageClassName := r.Context.Map.Storage.Spec.Map[0].Destination.StorageClass
+			if sdToStorageClass == nil {
+				if sdToStorageClass, err = r.mapStorageDomainToStorageClass(); err != nil {
+					return
+				}
+			}
+			storageClassName := sdToStorageClass[diskAttachment.Disk.StorageDomain]
 			pvc, err = r.persistentVolumeClaimWithSourceRef(diskAttachment, storageClassName, populatorName, annotations, vmRef.ID)
 			if err != nil {
 				if !k8serr.IsAlreadyExists(err) {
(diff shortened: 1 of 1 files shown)


ahadas:
@bennyz we have [the same issue with OpenStack](https://github.com/kubev2v/forklift/blob/main/pkg/controller/plan/adapter/openstack/builder.go#L1013), could you please post a fix for that as well?

bennyz:
> @bennyz we have [the same issue with OpenStack](https://github.com/kubev2v/forklift/blob/main/pkg/controller/plan/adapter/openstack/builder.go?rgh-link-date=2024-03-24T13%3A57%3A18Z#L1013), could you please post a fix for that as well?

it looks (somewhat) fine, the next lines override the SC based on the volume type

ahadas:
> > @bennyz we have [the same issue with OpenStack](https://github.com/kubev2v/forklift/blob/main/pkg/controller/plan/adapter/openstack/builder.go?rgh-link-date=2024-03-24T13%3A57%3A18Z#L1013), could you please post a fix for that as well?
> 
> it looks (somewhat) fine, the next lines override the SC based on the volume type

ok, so we don't use the first storage class that we find in the mapping for all the volumes but why do we use it at all?

ahadas:
filed https://github.com/kubev2v/forklift/issues/823 for OpenStack, it's indeed a bit different and less important than the issue this fix is intended for

pkg/controller/plan/adapter/ovirt/builder.go
```diff
@@ -732,7 +731,12 @@ func (r *Builder) PopulatorVolumes(vmRef ref.Ref, annotations map[string]string,
 				return
 			}
 			var pvc *core.PersistentVolumeClaim
-			storageClassName := r.Context.Map.Storage.Spec.Map[0].Destination.StorageClass
+			if sdToStorageClass == nil {
+				if sdToStorageClass, err = r.mapStorageDomainToStorageClass(); err != nil {
+					return
+				}
```
liranr23, line 737: why not pulling it out before the loop?

ahadas: because it could be that we won't need this, e.g., for VMs with a single LUN

pkg/controller/plan/adapter/ovirt/builder.go
```diff
@@ -748,6 +752,21 @@ func (r *Builder) PopulatorVolumes(vmRef ref.Ref, annotations map[string]string,
 	return
 }
 
+func (r *Builder) mapStorageDomainToStorageClass() (map[string]string, error) {
+	sdToStorageClass := make(map[string]string)
+	dsMapIn := r.Context.Map.Storage.Spec.Map
+	for i := range dsMapIn {
+		mapped := &dsMapIn[i]
+		ref := mapped.Source
```
bennyz, line 760: do we need this var?

pkg/controller/plan/adapter/ovirt/builder.go
```diff
@@ -748,6 +752,21 @@ func (r *Builder) PopulatorVolumes(vmRef ref.Ref, annotations map[string]string,
 	return
 }
 
+func (r *Builder) mapStorageDomainToStorageClass() (map[string]string, error) {
+	sdToStorageClass := make(map[string]string)
+	dsMapIn := r.Context.Map.Storage.Spec.Map
+	for i := range dsMapIn {
+		mapped := &dsMapIn[i]
```
bennyz, line 759: I think we can just use mapped in the for statement directly, we don't take the address anyway so there wouldn't be loopvar issues

pkg/controller/plan/adapter/ovirt/builder.go
ahadas: I've just sort of copy-pasted the code above, sure - there is a room for improvement there. done

pkg/controller/plan/adapter/ovirt/builder.go
ahadas: yep, copy-pasted, done

PR #824, opened 2024-03-24
Add support for instance types

liranr23:
Setting the `instanceType` to the VM will set the specific VM in
the plan with the selected instance type available in the destination
cluster (if found). This means, the VM won't be set with their original
CPUs and memory they had in the source environment.

### DIFF
--- a/operator/config/crd/bases/forklift.konveyor.io_plans.yaml
+++ b/operator/config/crd/bases/forklift.konveyor.io_plans.yaml
@@ -373,6 +373,10 @@ spec:
                         vsphere:
                           The managed object ID.
                       type: string
+                    instanceType:
+                      description: Selected InstanceType that will override the VM
+                        properties.
+                      type: string
                     luks:
                       description: Disk decryption LUKS keys
                       properties:
@@ -824,6 +828,10 @@ spec:
                             The object ID.
                             vsphere:
                               The managed object ID.
+                          type: string
+                        instanceType:
+                          description: Selected InstanceType that will override the
+                            VM properties.
                           type: string
                         luks:
                           description: Disk decryption LUKS keys
--- a/pkg/controller/plan/adapter/openstack/builder.go
+++ b/pkg/controller/plan/adapter/openstack/builder.go
@@ -259,7 +259,7 @@ var DefaultProperties = map[string]string{
 }
 
 // Create the destination Kubevirt VM.
-func (r *Builder) VirtualMachine(vmRef ref.Ref, vmSpec *cnv.VirtualMachineSpec, persistentVolumeClaims []*core.PersistentVolumeClaim) (err error) {
+func (r *Builder) VirtualMachine(vmRef ref.Ref, vmSpec *cnv.VirtualMachineSpec, persistentVolumeClaims []*core.PersistentVolumeClaim, usesInstanceType bool) (err error) {
 	vm := &model.Workload{}
 	err = r.Source.Inventory.Find(vm, vmRef)
 	if err != nil {
@@ -283,7 +283,7 @@ func (r *Builder) VirtualMachine(vmRef ref.Ref, vmSpec *cnv.VirtualMachineSpec, 
 	}
 
 	r.mapFirmware(vm, vmSpec)
-	r.mapResources(vm, vmSpec)
+	r.mapResources(vm, vmSpec, usesInstanceType)
 	r.mapHardwareRng(vm, vmSpec)
 	r.mapInput(vm, vmSpec)
 	r.mapVideo(vm, vmSpec)
@@ -406,10 +406,12 @@ func (r *Builder) mapInput(vm *model.Workload, object *cnv.VirtualMachineSpec) {
 	}
 }
 
-func (r *Builder) mapResources(vm *model.Workload, object *cnv.VirtualMachineSpec) {
-
+func (r *Builder) mapResources(vm *model.Workload, object *cnv.VirtualMachineSpec, usesInstanceType bool) {
 	// KubeVirt supports Q35 or PC-Q35 machine types only.
 	object.Template.Spec.Domain.Machine = &cnv.Machine{Type: Q35}
+	if usesInstanceType {
+		return
+	}
 	object.Template.Spec.Domain.CPU = &cnv.CPU{}
 
 	// Set CPU Policy
--- a/pkg/controller/plan/adapter/ova/builder.go
+++ b/pkg/controller/plan/adapter/ova/builder.go
@@ -204,7 +204,7 @@ func updateDataVolumeAnnotations(dv *cdi.DataVolume, disk *ova.Disk) {
 }
 
 // Create the destination Kubevirt VM.
-func (r *Builder) VirtualMachine(vmRef ref.Ref, object *cnv.VirtualMachineSpec, persistentVolumeClaims []*core.PersistentVolumeClaim) (err error) {
+func (r *Builder) VirtualMachine(vmRef ref.Ref, object *cnv.VirtualMachineSpec, persistentVolumeClaims []*core.PersistentVolumeClaim, usesInstanceType bool) (err error) {
 	vm := &model.VM{}
 	err = r.Source.Inventory.Find(vm, vmRef)
 	if err != nil {
@@ -228,11 +228,14 @@ func (r *Builder) VirtualMachine(vmRef ref.Ref, object *cnv.VirtualMachineSpec, 
 	}
 	r.mapDisks(vm, persistentVolumeClaims, object)
 	r.mapFirmware(vm, vmRef, object)
-	r.mapCPU(vm, object)
+	r.setMachine(object)
 	r.mapInput(object)
-	err = r.mapMemory(vm, object)
-	if err != nil {
-		return
+	if !usesInstanceType {
+		r.mapCPU(vm, object)
+		err = r.mapMemory(vm, object)
+		if err != nil {
+			return
+		}
 	}
 	err = r.mapNetworks(vm, object)
 	if err != nil {
@@ -322,11 +325,15 @@ func (r *Builder) mapMemory(vm *model.VM, object *cnv.VirtualMachineSpec) error 
 	return nil
 }
 
+func (r *Builder) setMachine(object *cnv.VirtualMachineSpec) {
+	object.Template.Spec.Domain.Machine = &cnv.Machine{Type: "q35"}
+}
+
 func (r *Builder) mapCPU(vm *model.VM, object *cnv.VirtualMachineSpec) {
 	if vm.CoresPerSocket == 0 {
 		vm.CoresPerSocket = 1
 	}
-	object.Template.Spec.Domain.Machine = &cnv.Machine{Type: "q35"}
+
 	object.Template.Spec.Domain.CPU = &cnv.CPU{
 		Sockets: uint32(vm.CpuCount / vm.CoresPerSocket),
 		Cores:   uint32(vm.CoresPerSocket),
--- a/pkg/controller/plan/adapter/ovirt/builder.go
+++ b/pkg/controller/plan/adapter/ovirt/builder.go
@@ -244,7 +244,7 @@ func (r *Builder) DataVolumes(vmRef ref.Ref, secret *core.Secret, configMap *cor
 }
 
 // Create the destination Kubevirt VM.
-func (r *Builder) VirtualMachine(vmRef ref.Ref, object *cnv.VirtualMachineSpec, persistentVolumeClaims []*core.PersistentVolumeClaim) (err error) {
+func (r *Builder) VirtualMachine(vmRef ref.Ref, object *cnv.VirtualMachineSpec, persistentVolumeClaims []*core.PersistentVolumeClaim, usesInstanceType bool) (err error) {
 	vm := &model.Workload{}
 	err = r.Source.Inventory.Find(vm, vmRef)
 	if err != nil {
@@ -268,8 +268,11 @@ func (r *Builder) VirtualMachine(vmRef ref.Ref, object *cnv.VirtualMachineSpec, 
 	}
 	r.mapDisks(vm, persistentVolumeClaims, object)
 	r.mapFirmware(vm, &vm.Cluster, object)
-	r.mapCPU(vm, object)
-	r.mapMemory(vm, object)
+	r.setMachine(object)
+	if !usesInstanceType {
+		r.mapCPU(vm, object)
+		r.mapMemory(vm, object)
+	}
 	r.mapClock(vm, object)
 	r.mapInput(object)
 	r.mapTpm(vm, object)
@@ -366,13 +369,18 @@ func (r *Builder) mapMemory(vm *model.Workload, object *cnv.VirtualMachineSpec) 
 	}
 	object.Template.Spec.Domain.Memory = &cnv.Memory{Guest: reservation}
 }
-func (r *Builder) mapCPU(vm *model.Workload, object *cnv.VirtualMachineSpec) {
+
+func (r *Builder) setMachine(object *cnv.VirtualMachineSpec) {
 	object.Template.Spec.Domain.Machine = &cnv.Machine{Type: "q35"}
+}
+
+func (r *Builder) mapCPU(vm *model.Workload, object *cnv.VirtualMachineSpec) {
 	object.Template.Spec.Domain.CPU = &cnv.CPU{
 		Sockets: uint32(vm.CpuSockets),
 		Cores:   uint32(vm.CpuCores),
 		Threads: uint32(vm.CpuThreads),
 	}
+
 	if vm.CpuPinningPolicy == model.Dedicated {
 		object.Template.Spec.Domain.CPU.DedicatedCPUPlacement = true
 	}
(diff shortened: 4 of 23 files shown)


liranr23:
Based on https://github.com/kubev2v/forklift-console-plugin/issues/1017#issuecomment-2016819275, do we wish to query and add the instance types to the inventory in order to let UI do a selection box instead of string?

operator/config/crd/bases/forklift.konveyor.io_plans.yaml
```diff
@@ -222,6 +222,9 @@ spec:
                 - destination
                 - source
                 type: object
+              selectedInstanceType:
```
ahadas, line 225:
```suggestion
              instanceType:
```
(it's obvious that it's selected, like the target namespace that is selected for the plan)

operator/config/crd/bases/forklift.konveyor.io_plans.yaml
```diff
@@ -222,6 +222,9 @@ spec:
...
+                description: Selected InstanceType that will override the VM properties.
```
ahadas, line 226: we should state that it's optional

pkg/controller/plan/adapter/openstack/builder.go
```diff
@@ -429,6 +429,9 @@ func (r *Builder) mapResources(vm *model.Workload, object *cnv.VirtualMachineSpe
 			object.Template.Spec.Domain.CPU.IsolateEmulatorThread = true
 		}
 	}
+	if object.Instancetype != nil {
+		return
```
ahadas, line 433: I think early return doesn't really fit here - it's easy to miss this and add something at the bottom of this function, I'd rather prefer to have a block that runs only when the instance type is not set instead and still get to the bottom of this function

pkg/controller/plan/adapter/ova/builder.go
```diff
@@ -326,6 +329,9 @@ func (r *Builder) mapCPU(vm *model.VM, object *cnv.VirtualMachineSpec) {
 		vm.CoresPerSocket = 1
 	}
 	object.Template.Spec.Domain.Machine = &cnv.Machine{Type: "q35"}
```
ahadas, line 331: that's an indication that this part does not belong here
CC @bkhizgiy 
@liranr23 could you please clean it up and make the changes here similar to those in the builder of openstack?

pkg/controller/plan/adapter/ovirt/builder.go
```diff
@@ -384,6 +382,14 @@ func (r *Builder) mapCPU(vm *model.Workload, object *cnv.VirtualMachineSpec) {
 	} else if r.Plan.Spec.PreserveClusterCPUModel {
 		r.setCpuFlags(r.getClusterCpu(vm), object)
 	}
+	if object.Instancetype != nil {
+		return
+	}
```
ahadas, line 387: ok, here it's a bit more tricky since PreserveClusterCPUModel is related to CPU, so here just to replace the early return

pkg/controller/plan/adapter/vsphere/builder.go
ahadas, line 1: same as the ova builder

pkg/controller/plan/kubevirt.go
```diff
@@ -1279,6 +1284,67 @@ func (r *KubeVirt) vmPreference(vm *plan.VMStatus) (virtualMachine *cnv.VirtualM
...
+	virtualMachine = object
+	return
+}
+
+// Attempt to find a suitable instance type
+func (r *KubeVirt) getInstancetype(vm *plan.VMStatus, instancetypeName string) (name, kind string, err error) {
+	name, kind, err = r.getVirtualMachineInstancetype(instancetypeName)
+	if err != nil {
+		if k8serr.IsNotFound(err) {
+			r.Log.Info("could not find a local instance type for destination VM. trying cluster wide",
```
ahadas, line 1305: this logic is a bit weird to me - first, since it assumes the instance type resides in the target namespace and second, because we can't distinguish between namespaced and cluster instancetype with the same name
we should probably add also a namespace to the instance type on the plan

liranr23: it's clear that the namespace will be the target namespace. we can't use anything else if the vm is planned to exist there

ahadas: I'm not sure - did you try to use an instance type from a different namespace that you have permissions on?

ahadas: @0xFelix, I see that indeed [InstacetypeMatcher has no namespace](https://github.com/kubevirt/kubevirt/blob/74f5cb9bcfee98661548ae5dfe900430d8cee9f9/staging/src/kubevirt.io/api/core/v1/types.go#L2835-L2868), so users can use instacetypes they have on other namespaces they have permissions on?

0xFelix: @ahadas Right now it is possible to use either use cluster wide instancetypes or instancetypes in the same namespace as the VM. Cross-namespace access to instancetypes is not possible.

pkg/controller/plan/adapter/openstack/builder.go
```diff
@@ -429,6 +429,9 @@ func (r *Builder) mapResources(vm *model.Workload, object *cnv.VirtualMachineSpe
 			object.Template.Spec.Domain.CPU.IsolateEmulatorThread = true
 		}
 	}
+	if object.Instancetype != nil {
```
0xFelix, line 432: The `DedicatedCPUPlacement` and `IsolateEmulatorThread` above also collide with instancetypes.

pkg/controller/plan/adapter/ovirt/builder.go
0xFelix: Same, `DedicatedCPUPlacement` above collides with instancetype.

pkg/controller/plan/kubevirt.go
```diff
@@ -1279,6 +1284,67 @@ func (r *KubeVirt) vmPreference(vm *plan.VMStatus) (virtualMachine *cnv.VirtualM
 	return
 }
 
+func (r *KubeVirt) setIntanceType(vm *plan.VMStatus, object *cnv.VirtualMachine) (virtualMachine *cnv.VirtualMachine, err error) {
```
0xFelix, line 1287:
```suggestion
func (r *KubeVirt) setInstanceType(vm *plan.VMStatus, object *cnv.VirtualMachine) (virtualMachine *cnv.VirtualMachine, err error) {
```

pkg/controller/plan/kubevirt.go
```diff
@@ -1279,6 +1284,67 @@ func (r *KubeVirt) vmPreference(vm *plan.VMStatus) (virtualMachine *cnv.VirtualM
...
+	if r.Plan.Spec.SelectedInstanceType == "" {
+		return object, nil
+	}
+	intancetypeName, kind, err := r.getInstancetype(vm, r.Plan.Spec.SelectedInstanceType)
+	if err != nil {
+		return
+	}
+	object.Spec.Instancetype = &cnv.InstancetypeMatcher{Name: intancetypeName, Kind: kind}
+	virtualMachine = object
+	return
```
0xFelix, line 1297: Naked return / named returns necessary?

pkg/controller/plan/kubevirt.go
```diff
@@ -1279,6 +1284,67 @@ func (r *KubeVirt) vmPreference(vm *plan.VMStatus) (virtualMachine *cnv.VirtualM
...
+}
+
+// Attempt to find a suitable instance type
+func (r *KubeVirt) getInstancetype(vm *plan.VMStatus, instancetypeName string) (name, kind string, err error) {
```
0xFelix, line 1301: Spell it either `InstanceType` or `Instancetype`?

pkg/controller/plan/kubevirt.go
```diff
@@ -1279,6 +1284,67 @@ func (r *KubeVirt) vmPreference(vm *plan.VMStatus) (virtualMachine *cnv.VirtualM
...
+func (r *KubeVirt) getVirtualMachineInstancetype(instancetypeName string) (name, kind string, err error) {
+	virtualMachineClusterInstancetype := &instancetype.VirtualMachineInstancetype{}
+	err = r.Destination.Client.Get(
+		context.TODO(),
+		client.ObjectKey{Name: instancetypeName, Namespace: r.Plan.Spec.TargetNamespace},
+		virtualMachineClusterInstancetype)
+	if err != nil {
+		return
+	}
+	return instancetypeName, "VirtualMachineInstancetype", nil
```
0xFelix, line 1326: There could be a constant for `VirtualMachineInstancetype` in the KubeVirt API. (not 100% sure)

ahadas: yeah, I also suspect such constant would exist in the kubevirt API but I saw that such a string is used all over the place..

pkg/controller/plan/adapter/ovirt/builder.go
liranr23: yeah i saw. also hugepages and such for memory. i will leave them clean (nothing in CPU or memory will be set). thanks @0xFelix

pkg/controller/plan/kubevirt.go
ahadas: ack
@liranr23 we should keep this in mind when we get to document it then because I suppose ordinary users won't be provided with permissions to create cluster-level instance type so in order to use a custom instance type, users would have to create a namespace and write or copy an instance type such that it would exist in the target namespace..

pkg/controller/plan/kubevirt.go
0xFelix: https://github.com/kubevirt/kubevirt/blob/74f5cb9bcfee98661548ae5dfe900430d8cee9f9/staging/src/kubevirt.io/api/instancetype/register.go#L26

pkg/controller/plan/adapter/ova/builder.go
liranr23: i might have made it quite the opposite, please revisit

pkg/controller/plan/kubevirt.go
```diff
@@ -1287,6 +1293,69 @@ func (r *KubeVirt) vmPreference(vm *plan.VMStatus) (virtualMachine *cnv.VirtualM
 	return
 }
 
+func (r *KubeVirt) setInstanceType(vm *plan.VMStatus, object *cnv.VirtualMachine) (err error) {
+	if r.Plan.Spec.InstanceType == "" {
+		return
+	}
+	intancetypeName, kind, err := r.getInstanceType(vm, r.Plan.Spec.InstanceType)
```
ahadas, line 1300:
```suggestion
	instanceTypeName, kind, err := r.getInstanceType(vm, r.Plan.Spec.InstanceType)
```

pkg/controller/plan/kubevirt.go
```diff
@@ -1287,6 +1293,69 @@ func (r *KubeVirt) vmPreference(vm *plan.VMStatus) (virtualMachine *cnv.VirtualM
...
+	object.Spec.Instancetype = &cnv.InstancetypeMatcher{Name: intancetypeName, Kind: kind}
+	return
+}
+
+// Attempt to find a suitable instance type
+func (r *KubeVirt) getInstanceType(vm *plan.VMStatus, instancetypeName string) (name, kind string, err error) {
+	name, kind, err = r.getVirtualMachineInstanceType(instancetypeName)
+	if err != nil {
+		if k8serr.IsNotFound(err) {
+			r.Log.Info("could not find a local instance type for destination VM. trying cluster wide",
```
ahadas, line 1313:
```suggestion
			r.Log.Info("could not find a namespaced instance type for destination VM. trying cluster wide",
```

pkg/controller/plan/kubevirt.go
```diff
@@ -1287,6 +1293,69 @@ func (r *KubeVirt) vmPreference(vm *plan.VMStatus) (virtualMachine *cnv.VirtualM
...
+				"vm",
+				vm.String())
+		} else {
+			r.Log.Error(err, "could not fetch a local instance type for destination VM. trying cluster wide",
```
ahadas, line 1317:
```suggestion
			r.Log.Error(err, "could not fetch a namespaced instance type for destination VM. trying cluster wide",
```

pkg/controller/plan/kubevirt.go
```diff
@@ -1287,6 +1293,69 @@ func (r *KubeVirt) vmPreference(vm *plan.VMStatus) (virtualMachine *cnv.VirtualM
...
+	intancetypeName, kind, err := r.getInstanceType(vm, r.Plan.Spec.InstanceType)
+	if err != nil {
+		return
+	}
+	object.Spec.Instancetype = &cnv.InstancetypeMatcher{Name: intancetypeName, Kind: kind}
+	return
+}
+
+// Attempt to find a suitable instance type
+func (r *KubeVirt) getInstanceType(vm *plan.VMStatus, instancetypeName string) (name, kind string, err error) {
```
ahadas, line 1309: will `name` ever be different than `instancetypeName` (which should be `instanceTypeName`)? if not, then please drop `name`

pkg/controller/plan/kubevirt.go
```diff
@@ -1287,6 +1293,69 @@ func (r *KubeVirt) vmPreference(vm *plan.VMStatus) (virtualMachine *cnv.VirtualM
...
+				"vm",
+				vm.String())
+		}
+		name, kind, err = r.getVirtualMachineClusterInstanceType(vm, instancetypeName)
+	}
+
+	return
+}
+
+func (r *KubeVirt) getVirtualMachineInstanceType(instancetypeName string) (name, kind string, err error) {
```
ahadas, line 1327: same comment here about `name`

pkg/controller/plan/kubevirt.go
```diff
@@ -1355,11 +1424,15 @@ func (r *KubeVirt) getVirtualMachineClusterPreference(vm *plan.VMStatus, prefere
...
 		return
 	}
-	return preferenceName, "VirtualMachineClusterPreference", nil
+	return preferenceName, instancetypeapi.ClusterSingularPreferenceResourceName, nil
 }
 
 // Attempt to find a suitable template and extract a VirtualMachine definition from it.
 func (r *KubeVirt) vmTemplate(vm *plan.VMStatus) (virtualMachine *cnv.VirtualMachine, ok bool) {
+	if r.Plan.Spec.InstanceType != "" {
+		r.Log.Info("InstanceType is set, passing the usage of template", "vm", vm.String())
```
ahadas, line 1433:
```suggestion
		r.Log.Info("InstanceType is set, not setting a template", "vm", vm.String())
```

pkg/controller/provider/container/ocp/collection.go
```diff
@@ -334,7 +350,12 @@ func (r *VM) Reconcile(ctx context.Context) (err error) {
 		err = liberr.Wrap(err)
 		return
 	}
-	defer tx.End()
+	defer func() {
+		err = tx.End()
+		if err != nil {
+			err = liberr.Wrap(err)
+		}
+	}()
```
ahadas, line 358: wouldn't it make sense to extract this to a separate function?

pkg/controller/provider/container/ocp/collection.go
```diff
@@ -403,3 +424,207 @@ func (r *VM) Delete(e event.DeleteEvent) bool {
...
+		}
+		m := &model.InstanceType{}
+		m.With(&resource)
+		r.Collector.UpdateThreshold(m)
+		r.log.Info("Create", libref.ToKind(m), m.String())
+		err = tx.Insert(m)
+		if err != nil {
+			err = liberr.Wrap(err)
+			return
+		}
```
ahadas, line 475: why not to place this block within `default` above?

pkg/controller/provider/container/ocp/collection.go
```diff
@@ -403,3 +424,207 @@ func (r *VM) Delete(e event.DeleteEvent) bool {
...
+		m := &model.ClusterInstanceType{}
+		m.With(&resource)
+		r.Collector.UpdateThreshold(m)
+		r.log.Info("Create", libref.ToKind(m), m.String())
+		err = tx.Insert(m)
+		if err != nil {
+			err = liberr.Wrap(err)
+			return
+		}
```
ahadas, line 577: same here

pkg/controller/provider/web/ocp/client.go
```diff
@@ -289,3 +361,41 @@ func (r *Finder) Host(ref *base.Ref) (object interface{}, err error) {
...
+//	ProviderNotSupportedErr
+//	ProviderNotReadyErr
+//	NotFoundErr
+//	RefNotUniqueErr
+func (r *Finder) ClusterInstanceType(ref *base.Ref) (object interface{}, err error) {
+	it := &ClusterInstanceType{}
+	err = r.ByRef(it, *ref)
+	if err == nil {
+		ref.ID = it.UID
+		ref.Name = path.Join(it.Namespace, it.Name)
```
ahadas, line 396: why does the name need to include the namespace in this case?..

pkg/controller/provider/container/ocp/collection.go
ahadas: yeah, that's what I meant (even though I'm not sure why you placed the function in a different place rather than having it here as a "private" function)

ahadas (approved):

PR #825, opened 2024-03-24
openstack: avoid NPE when there's only an image attached

bennyz:
Since it's possible to create a plan without storage maps, image based VMs that have no volumes attached will crash the controller with an NPE when attempting to fetch the storage map

### DIFF
--- a/pkg/controller/plan/adapter/openstack/builder.go
+++ b/pkg/controller/plan/adapter/openstack/builder.go
@@ -1020,13 +1020,14 @@ func (r *Builder) ensureVolumePopulatorPVC(workload *model.Workload, image *mode
 
 		// VM is image based, look for a glance key in the mapping
 		if workload.ImageID != "" {
+			// At this point the StorageMap has been validated, and the VM has to be fully mapped
 			for _, storageMap := range mapList {
 				if storageMap.Source.Name == api.GlanceSource {
 					storageClassName = storageMap.Destination.StorageClass
 				}
 			}
 		} else {
-			// VM is volume based, look for the volume type in the mapping
+			// VM has a volume, look for the volume type in the mapping
 			if volumeType := r.getVolumeType(workload, originalVolumeDiskId); volumeType != "" {
 				storageClassName, err = r.getStorageClassName(workload, volumeType)
 				if err != nil {
--- a/pkg/controller/plan/adapter/openstack/builder.go
+++ b/pkg/controller/plan/adapter/openstack/builder.go
@@ -31,9 +31,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-// GlanceSource represents an image of which the source is Glance, to be used in storage mapping
-const GlanceSource = "glance"
-
 // Openstack builder.
 type Builder struct {
 	*plancontext.Context
@@ -1024,7 +1021,7 @@ func (r *Builder) ensureVolumePopulatorPVC(workload *model.Workload, image *mode
 		// VM is image based, look for a glance key in the mapping
 		if workload.ImageID != "" {
 			for _, storageMap := range mapList {
-				if storageMap.Source.Name == "glance" {
+				if storageMap.Source.Name == api.GlanceSource {
 					storageClassName = storageMap.Destination.StorageClass
 				}
 			}
--- a/pkg/controller/plan/adapter/openstack/builder.go
+++ b/pkg/controller/plan/adapter/openstack/builder.go
@@ -31,6 +31,9 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+// GlanceSource represents an image of which the source is Glance, to be used in storage mapping
+const GlanceSource = "glance"
+
 // Openstack builder.
 type Builder struct {
 	*plancontext.Context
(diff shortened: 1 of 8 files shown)


pkg/controller/plan/adapter/openstack/builder.go
```diff
@@ -1008,6 +1008,12 @@ func (r *Builder) ensureVolumePopulatorPVC(workload *model.Workload, image *mode
 			originalVolumeDiskId = imageProperty.(string)
 		}
 
+		// Check if there's a storage map available
+		if len(r.Context.Map.Storage.Spec.Map) == 0 {
```
liranr23, line 1012: how did we reach to this state at all? the user created the plan without storage mapping?

liranr23 (commented):
> It's possible to not create a storage map when only an image is attached to a VM, this will lead to an NPE later on

I find the commit message sophisticated

pkg/controller/plan/adapter/openstack/builder.go
bennyz: yes, it's possible to do that when the VM only has an image attached (vs. having a volume attached)

liranr23:
> > It's possible to not create a storage map when only an image is attached to a VM, this will lead to an NPE later on
> 
> I find the commit message sophisticated

How about:
When we don't have a VM that only has image attached, we need to make sure that a storage map exists. Otherwise, we will result in NPE.

pkg/controller/plan/adapter/openstack/builder.go
```diff
@@ -1008,6 +1008,12 @@ func (r *Builder) ensureVolumePopulatorPVC(workload *model.Workload, image *mode
...
+			err = liberr.New("no storage map found in the migration plan")
```
ahadas, line 1013: this doesn't necessarily mean that there's no storage map, there could be an empty storage map, no?
I guess the proper way to handle it would be using a validation on the storage map but if that's complicated, how about replacing the code that takes an arbitrary storage class from the storage map with an arbitrary storage class from the inventory?

bennyz: I can try to ensure there is an actual destination available, but I don't we should pick an arbitrary one from the inventory because in the current approach there's an actual way (albeit slightly hacky) for the user ensure the image ends up on a storage class they pick rather than a random one

ahadas: yeah, maybe I was carried away by this suggestion - there's not much difference between picking an arbitrary storage class from the storage map or from the source environments but I agree that the latter is not better. so maybe we should just pick the default storage class for virtualization?..

ahadas: and maybe I jumped too quickly to the solution - my concern with this approach of raising an error here (rather than when validating the storage map) is that I don't see how it would work from the UI: I create a plan with a single VM that is based on an image, so when I get to the network mapping dialog, the UI won't suggest any volume type to map so I'll proceed with an empty mapping and we'll get to this code, right?

bennyz: currently UI suggests all volume types so I'm not entirely sure how it would look

ahadas: > currently UI suggests all volume types so I'm not entirely sure how it would look

specifically for OpenStack?..

ahadas: ok, as we spoke offline - it's ok to keep this code when the UI would specify a "dummy" volume type for images in OpenStack

pkg/controller/plan/adapter/openstack/builder.go
```diff
@@ -1008,14 +1011,34 @@ func (r *Builder) ensureVolumePopulatorPVC(workload *model.Workload, image *mode
...
+			return
+		}
+
+		var storageClassName string
+
+		// VM is image based, look for a glance key in the mapping
+		if workload.ImageID != "" {
+			for _, storageMap := range mapList {
+				if storageMap.Source.Name == "glance" {
```
ahadas, line 1027:
```suggestion
				if storageMap.Source.Name == GlanceSource {
```

pkg/controller/plan/adapter/openstack/builder_test.go
```diff
@@ -16,3 +16,9 @@ var _ = Describe("vSphere builder", func() {
 		Entry("ubuntu 22", Ubuntu, "22.04.3", Ubuntu, "ubuntu"),
 	)
 })
+
+var _ = Describe("OpenStack Glance const test", func() {
+	It("GlanceSource should be glance, changing it may break the UI", func() {
+		Expect(GlanceSource).Should(Equal("glance"))
```
ahadas, line 22:
```suggestion
		Expect(GlanceSource).Should(Equal(GlanceSource))
```

pkg/controller/plan/adapter/openstack/builder.go
```diff
@@ -1008,14 +1011,34 @@ func (r *Builder) ensureVolumePopulatorPVC(workload *model.Workload, image *mode
...
+					storageClassName = storageMap.Destination.StorageClass
+				}
+			}
+		} else {
+			// VM is volume based, look for the volume type in the mapping
+			if volumeType := r.getVolumeType(workload, originalVolumeDiskId); volumeType != "" {
+				storageClassName, err = r.getStorageClassName(workload, volumeType)
+				if err != nil {
+					err = liberr.Wrap(err)
```
ahadas, line 1036: I'm not sure why we need this check (shouldn't we always a corresponding storage class since otherwise we would have set the storage mapping as invalid?) but since we have it here, do we need it for glance above as well?

pkg/controller/provider/web/openstack/client.go
```diff
@@ -396,6 +396,10 @@ func (r *Finder) Network(ref *base.Ref) (object interface{}, err error) {
 //	NotFoundErr
 //	RefNotUniqueErr
 func (r *Finder) Storage(ref *base.Ref) (object interface{}, err error) {
+	if ref.Name == "glance" {
```
ahadas, line 399: use the const?

pkg/controller/plan/adapter/openstack/builder_test.go
bennyz: isn't it always true?

pkg/controller/plan/adapter/openstack/builder.go
bennyz: I think it's just there, the mapping has to be valid at this point

pkg/controller/plan/adapter/openstack/builder_test.go
ahadas: ah ok, I didn't pay much attention to the first part :)

pkg/controller/plan/adapter/openstack/builder.go
ahadas: so please add a comment that reflects this

tests/suit/utils/storagemap.go
```diff
@@ -47,6 +47,15 @@ func NewStorageMap(namespace string, providerIdentifier forkliftv1.Provider, sto
 		sdPairs = append(sdPairs, pair)
 	}
 
+	if providerIdentifier.Type() == api.OpenStack {
```
ahadas, line 50: why not setting it as part of the storage mapping that is created in openstack_test.go?

ahadas (commented):
looks good overall

tests/suit/utils/storagemap.go
bennyz: I don't see that we manipulate it in the specific provider tests, OVA also has a specific branch above

ahadas: I mean to define the mapping of glance to the same target storage class [here](https://github.com/kubev2v/forklift/blob/main/tests/suit/openstack_test.go#L64) to reflect what clients are expected to provide

bennyz: I understood, what I meant is that we don't provider specific manipulations there

ahadas (approved):

ahadas:
@bennyz please backport

PR #826, opened 2024-03-25
vsphere: transfer updated disk state on remote migrations

ahadas:
The problem these changes address is as follow: when initiating a cold migration from vSphere to a remote provider of a VM with a snapshot(s), the transferred disk didn't include the most updated state of the disk because the very first volume of the disk, that doesn't include later changes, was transferred.

Unlike warm migrations, in which we intend to transfer the first (base) volume of the disk initially and then copy the changes using CBT, in cold migrations we need to transfer the state of the very last volume to include the most updated state of the disk.

In Forklift 2.3 and below, this issue affected all cold migrations. As from 2.4, it only affects cold migrations to a remote cluster, in which we still use CDI to transfer the disks from vSphere.

### DIFF
--- a/pkg/controller/plan/adapter/vsphere/builder.go
+++ b/pkg/controller/plan/adapter/vsphere/builder.go
@@ -365,7 +365,7 @@ func (r *Builder) DataVolumes(vmRef ref.Ref, secret *core.Secret, _ *core.Config
 					// Let CDI do the copying
 					dvSource = cdi.DataVolumeSource{
 						VDDK: &cdi.DataVolumeSourceVDDK{
-							BackingFile:  trimBackingFileName(disk.File),
+							BackingFile:  r.baseVolume(disk.File),
 							UUID:         vm.UUID,
 							URL:          url,
 							SecretRef:    secret.Name,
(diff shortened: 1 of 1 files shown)


pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -826,6 +826,22 @@ func trimBackingFileName(fileName string) string {
 	return backingFilePattern.ReplaceAllString(fileName, ".vmdk")
 }
 
+func (r *Builder) baseVolume(fileName string) string {
+	if r.Plan.Spec.Warm {
+		// for warm migrations, we return the very first volume of the disk
+		// as the base volume and CBT will be used to transfer later changes
+		return backingFilePattern.ReplaceAllString(fileName, ".vmdk")
```
liranr23, line 833: can we remove `trimBackingFileName` then?

ahadas: no, it's used elsewhere

ahadas: I changed this part though to call trimBackingFileName rather than duplicating its content

liranr23 (approved):

ahadas:
the changes here are specific to vSphere, the problem with OpenStack suite is unrelated..

PR #828, opened 2024-03-26
ovirt: Respect storage mapping

ahadas:
Previously, when migrating to the cluster Forklift is deployed on using a cold migration, all disks of the migrated VM were transferred to the same storage class, which is one of the target storage class that appear in the storage mapping that is set for the migration plan.

With these changes, each disk will be mapped to a target storage class based on its original storage domain and the storage mapping that is defined for the migration plan.


ahadas:
backport of #822

PR #829, opened 2024-03-26
openstack: attempt to fix race

bennyz:
The status of the snapshot might not be synced in the inventory. Check the status is up-to-date before proceeding

### DIFF
--- a/pkg/controller/plan/adapter/openstack/client.go
+++ b/pkg/controller/plan/adapter/openstack/client.go
@@ -804,9 +804,15 @@ func (r *Client) ensureVmSnapshot(vm *libclient.VM) (ready bool, err error) {
 			"vm", vm.Name, "image", vmSnapshotImage.Name, "imageID", vmSnapshotImage.ID)
 		ready = true
 		if _, ok := vm.Image["id"]; ok {
+			r.Log.Info("the VM is image based, checking the image properties", "vm", vm.Name, "snapshot", vmSnapshotImage.Name)
 			ready, err = r.ensureImageUpToDate(vm, vmSnapshotImage, vmTypeImageBased)
 			if err != nil {
 				r.Log.Error(err, "checking the VM snapshot image properties", "vm", vm.Name, "image", vmSnapshotImage.Name)
+				return
+			}
+
+			if !ready {
+				r.Log.Info("the VM snapshot image properties are not in sync, skipping...", "vm", vm.Name, "image", vmSnapshotImage.Name)
 				return
 			}
 		}
@@ -855,6 +861,8 @@ func (r *Client) ensureImagesFromVolumesReady(vm *libclient.VM) (ready bool, err
 			if !ready {
 				return false, nil
 			}
+			r.Log.Info("the image is ready in the inventory",
+				"vm", vm.Name, "image", image.Name, "properties", image.Properties)
 
 			go func() {
 				// executing this in a non-blocking mode
@@ -940,6 +948,12 @@ func (r *Client) ensureImageUpToDate(vm *libclient.VM, image *libclient.Image, v
 			r.Log.Info("the image does not exist in the inventory, waiting...",
 				"vm", vm.Name, "image", image.Name, "properties", image.Properties)
 		}
+		return
+	}
+
+	if inventoryImage.Status != string(image.Status) {
+		r.Log.Info("image status is not in sync, waiting...",
+			"vm", vm.Name, "image", inventoryImage.Name, "status", inventoryImage.Status)
 		return
 	}
 


liranr23 (approved):

pkg/controller/plan/adapter/openstack/client.go
```diff
@@ -804,11 +804,17 @@ func (r *Client) ensureVmSnapshot(vm *libclient.VM) (ready bool, err error) {
...
+			r.Log.Info("the VM is image based, checking the image properties", "vm", vm.Name, "snapshot", vmSnapshotImage.Name)
 			ready, err = r.ensureImageUpToDate(vm, vmSnapshotImage, vmTypeImageBased)
 			if err != nil {
 				r.Log.Error(err, "checking the VM snapshot image properties", "vm", vm.Name, "image", vmSnapshotImage.Name)
 				return
 			}
+
+			if !ready {
+				r.Log.Info("the VM snapshot image properties are not in sync, skipping...", "vm", vm.Name, "image", vmSnapshotImage.Name)
+				return false, nil
```
ahadas, line 816:
```suggestion
				return
```
?

pkg/controller/plan/adapter/openstack/client.go
```diff
@@ -943,6 +951,12 @@ func (r *Client) ensureImageUpToDate(vm *libclient.VM, image *libclient.Image, v
 		return
 	}
 
+	if inventoryImage.Status != string(image.Status) {
+		r.Log.Info("image status is not in sync, waiting...",
+			"vm", vm.Name, "image", inventoryImage.Name, "status", inventoryImage.Status)
```
ahadas, line 956: don't we need to set `ready` to false?

bennyz: set where?

pkg/controller/plan/adapter/openstack/client.go
bennyz: yes, I felt I need to be more explicit here for readability... I don't like naked returns

pkg/controller/plan/adapter/openstack/client.go
ahadas: ah the definitions in lines 938-941 are not within a function (why don't we have them at the top?..), so upToDate is necessarily false when we get here..

pkg/controller/plan/adapter/openstack/client.go
ahadas: ok, but we have one above and all over this module - so maybe for consistency?

pkg/controller/plan/adapter/openstack/client.go
bennyz: because it's not possible to do type aliasing within a function

pkg/controller/plan/adapter/openstack/client.go
bennyz: ok

pkg/controller/plan/adapter/openstack/client.go
ahadas: > because it's not possible to do type aliasing within a function

I meant at the top of the file instead of having them between the functions

pkg/controller/plan/adapter/openstack/client.go
```diff
@@ -940,7 +948,13 @@ func (r *Client) ensureImageUpToDate(vm *libclient.VM, image *libclient.Image, v
 			r.Log.Info("the image does not exist in the inventory, waiting...",
 				"vm", vm.Name, "image", image.Name, "properties", image.Properties)
 		}
-		return
+		return false, nil
```
ahadas, line 951: this change looks incorrect since we do want to return an error when it's not of type NotFoundError, right? let's minimize the changes here so we can quickly backport it to release-v2.6.0 and post cosmetic changes separately?

pkg/controller/plan/adapter/openstack/client.go
ahadas: let's consider doing it separately

pkg/controller/plan/adapter/openstack/client.go
```diff
@@ -955,7 +969,7 @@ func (r *Client) ensureImageUpToDate(vm *libclient.VM, image *libclient.Image, v
 			"vm", vm.Name, "image", inventoryImage.Name, "properties", inventoryImage.Properties)
 	}
 
-	return
+	return upToDate, nil
```
ahadas, line 972: I'd suggest to drop the named out arguments then, but again - let's minimize the changes here so please post it separately

ahadas (approved):

PR #832, opened 2024-03-27
Do not deploy OVA server pod when the provider is being deleted

bkhizgiy:
If the OVA provider server fails to be created for any reason, we keep hitting the timeout check in the reconcile and return from there, so we do not reach the cleanup. To solve this issue, a check was added for the provider query to perform only when the provider has not started the deletion process.

### DIFF
--- a/pkg/controller/provider/controller.go
+++ b/pkg/controller/provider/controller.go
@@ -190,7 +190,7 @@ func (r Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (r
 		}
 	}
 
-	if provider.Type() == api.Ova {
+	if provider.Type() == api.Ova && provider.DeletionTimestamp == nil {
 
 		deploymentName := fmt.Sprintf("%s-deployment-%s", ovaServer, provider.Name)
 


pkg/controller/provider/controller.go
```diff
@@ -190,6 +190,31 @@ func (r Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (r
 		}
 	}
 
+	if provider.DeletionTimestamp != nil && k8sutil.ContainsFinalizer(provider, api.OvaProviderFinalizer) {
+		fmt.Println("we are here at finalizer")
```
liranr23, line 194: forgot to delete?

ahadas: yes, looks like it was used for debugging, @bkhizgiy  please drop

pkg/controller/provider/controller.go
```diff
@@ -190,6 +190,31 @@ func (r Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (r
...
+			}
+			clonedProvider := provider.DeepCopy()
+			k8sutil.RemoveFinalizer(provider, api.OvaProviderFinalizer)
+			if err := r.Patch(context.TODO(), provider, client.MergeFrom(clonedProvider)); err != nil {
+				r.Log.Error(err, "Failed to remove finalizer", "provider", provider)
+			}
+		}
+	}
+
 	if provider.Type() == api.Ova {
```
ahadas, line 218: we shouldn't run this block when removing the provider
how about checking `provider.DeletionTimestamp == nil` and then we can keep the previous order?

ahadas:
@bkhizgiy thanks for making the previous changes per our offline conversation, I don't care much about the order of the blocks, since they should be mutual exclusive - we should ensure they really are mutual exclusive and by doing that we can minimize the changes here

pkg/controller/provider/controller.go
bkhizgiy: sure it will provide the same outcome, so i'm fine with both ways.

pkg/controller/provider/controller.go
```diff
@@ -437,3 +420,28 @@ func (r *Catalog) get(request reconcile.Request) (p *api.Provider, found bool) {
 	p, found = r.content[request]
 	return
 }
+
+func (r *Reconciler) CleanupProviderOnDelete(provider *api.Provider) (err error) {
```
ahadas, line 424: we should change the visibility of this function and the name should reflect that it only handles OVA providers
```suggestion
func (r *Reconciler) removeVolumeOfOVAServer(provider *api.Provider) (err error) {
```

ahadas (commented):
looks good, I'd suggest to separate out the refactoring to a separate PR - we should minimize the changes for a backport

ahadas:
> looks good, I'd suggest to separate out the refactoring to a different PR - we should minimize the changes for a backport

@bkhizgiy and also please update the title and the description of the PR (and commits)

pkg/controller/provider/controller.go
bkhizgiy: @ahadas open a separate PR for the refactoring https://github.com/kubev2v/forklift/pull/834.

bkhizgiy:
backport to v2.6  https://github.com/kubev2v/forklift/pull/835

ahadas:
@bkhizgiy please squash the two commits and update the title and description

ahadas (approved):

PR #833, opened 2024-03-27
Get the proper importer pods

liranr23:
The name of importer pods starts with `importer-<pvc name>`. This patch will change the way we find the importer pods for a particular PVC name.

### DIFF
--- a/pkg/controller/plan/kubevirt.go
+++ b/pkg/controller/plan/kubevirt.go
@@ -239,7 +239,7 @@ func (r *KubeVirt) getImporterPods(pvc *core.PersistentVolumeClaim) (pods []core
 		return
 	}
 	for _, pod := range podList.Items {
-		if strings.Contains(pod.Name, fmt.Sprintf("importer-%s-%s", r.Plan.Name, pvc.Annotations[kVM])) {
+		if strings.Contains(pod.Name, fmt.Sprintf("importer-%s", pvc.Name)) {
 			pods = append(pods, pod)
 		}
 	}


pkg/controller/plan/kubevirt.go
```diff
@@ -239,7 +239,7 @@ func (r *KubeVirt) getImporterPods(pvc *core.PersistentVolumeClaim) (pods []core
 		return
 	}
 	for _, pod := range podList.Items {
-		if strings.Contains(pod.Name, fmt.Sprintf("importer-%s-%s", r.Plan.Name, pvc.Annotations[kVM])) {
+		if strings.Contains(pod.Name, fmt.Sprintf("importer-%s", pvc.Name)) {
```
bennyz, line 242: I think there should be an annotation on the pvc `cdi.kubevirt.io/storage.import.importPodName`, but I looked in some envs and I don't see it, so perhaps it's only set on the prime PVC or something

bennyz: uhh, in L#224 we do check it:
```go
	if _, ok := pvc.Annotations[AnnImporterPodName]; !ok {
		return
	}
```
so I'm not sure what's going on

liranr23: it exists in a single importer pod. when using warm migration it won't be there, as we create importer pod per checkpoint

bennyz: so what does the if in L#224 do?

ahadas: and isn't there a label we can query instead?

ahadas:
> The importer pods name contains only 17 chars of the VM ID. But, it's prefix is the same as the PVC name. This patch will change the way we select the importer pods to be based on the PVC name.

I'm not sure why the 
The name of importer pods contains only 17 chars of the VM ID. But its prefix is the same as the PVC name. This patch changes the way we select importer pods such that we'll filter them by the PVC name.

pkg/controller/plan/kubevirt.go
liranr23: > so what does the if in L#224 do?

I think making sure our PVC is legit (part of import). We may drop it if you like.



> and isn't there a label we can query instead?

I think there isn't one.

ahadas: > > and isn't there a label we can query instead?
> 
> I think there isn't one.

ok, please check - if there is, it would be better

bennyz: >I think making sure our PVC is legit (part of import). We may drop it if you like.

you mean the annotation is set but it's an empty string?

liranr23: > > > and isn't there a label we can query instead?
> > 
> > 
> > I think there isn't one.
> 
> ok, please check - if there is, it would be better

there isn't.



> > I think making sure our PVC is legit (part of import). We may drop it if you like.
> 
> you mean the annotation is set but the value is an empty string?

once we create warm migration the VM has multiple PVCs (original and scratch), the original contains our annotation while the scratch isn't. but, looking on `getPVCs` function, we anyway get the PVC with the annotation. in my opinion we can drop this `if` at the start.

ahadas (commented):
the change looks good but the description of the PR is not phrased well, and it's too complex - shouldn't it just say that the name of importer pods starts with `importer-<pvc name>` so we fix the way we find importer pod for a particular pvc accordingly?

liranr23:
> the change looks good but the description of the PR is not phrased well, and it's too complex - shouldn't it just say that the name of importer pods starts with `importer-<pvc name>` so we fix the way we find importer pod for a particular pvc accordingly?

done

ahadas (approved):

PR #834, opened 2024-03-28
OVA: refactor PV clean up flow into a function

bkhizgiy:


liranr23 (approved):

pkg/controller/provider/controller.go
```diff
@@ -437,3 +420,28 @@ func (r *Catalog) get(request reconcile.Request) (p *api.Provider, found bool) {
...
+	} else {
+		for _, pv := range pvList.Items {
+			if err = r.Client.Delete(context.TODO(), &pv); err != nil {
+				r.Log.Error(err, "Failed to delete PV", "PV", pv)
+				return
+			}
+		}
+		clonedProvider := provider.DeepCopy()
+		k8sutil.RemoveFinalizer(provider, api.OvaProviderFinalizer)
+		if err := r.Patch(context.TODO(), provider, client.MergeFrom(clonedProvider)); err != nil {
```
ahadas, line 442: @bkhizgiy this was kind of a hack to avoid introducing another variable for an error that is not returned (in this case we only log the error) but now that it's split to a separate function we can improve this - how about remove the named output argument (i.e., change the signature to `func (r *Reconciler) removeVolumeOfOVAServer(provider *api.Provider) error`) and return an error when we used to?

ahadas (approved):

PR #835, opened 2024-03-28
Backport: Do not deploy OVA server pod when the provider is being deleted

bkhizgiy:
If the OVA provider server fails to be created for any reason, we keep hitting the timeout check in the reconcile and return from there, so we do not reach the cleanup. To solve this issue, a check was added for the provider query to perform only when the provider has not started the deletion process.

### DIFF
--- a/pkg/controller/provider/controller.go
+++ b/pkg/controller/provider/controller.go
@@ -190,7 +190,7 @@ func (r Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (r
 		}
 	}
 
-	if provider.Type() == api.Ova {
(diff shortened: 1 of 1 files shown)


ahadas:
@bkhizgiy please update this PR (keeping the title and description of the PR for the main branch)

liranr23:
> backport of #832

it's not only 832, there is also the virt-v2v break here. please mention it as well (probably in another commit/PR)

bkhizgiy:
@liranr23 yeah I accidentally pushed it in to the same brunch, i"ll fix it now

ahadas (approved):

PR #836, opened 2024-03-28
Set guest memory

ahadas:
Previously we didn't set the guest memory but only memory request, since the guest memory property is optional and according to the documentation it "Defaults to the requested memory in the resources section if not specified.". However, if we don't set it and the migrated VM is based on a template/instancetypepreference, the value of this field would be retrieved from the template/instancetypepreference, which may be different than what the VM had on the source and this could lead to various issues, e.g., starting with insufficient memory for booting.

### DIFF
--- a/pkg/controller/plan/adapter/openstack/builder.go
+++ b/pkg/controller/plan/adapter/openstack/builder.go
@@ -440,6 +440,7 @@ func (r *Builder) mapResources(vm *model.Workload, object *cnv.VirtualMachineSpe
 	memory := resource.NewQuantity(int64(vm.Flavor.RAM)*1024*1024, resource.BinarySI)
 	resourceRequests := map[core.ResourceName]resource.Quantity{}
 	resourceRequests[core.ResourceMemory] = *memory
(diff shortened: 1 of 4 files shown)


pkg/controller/plan/adapter/openstack/builder.go
```diff
@@ -440,6 +440,7 @@ func (r *Builder) mapResources(vm *model.Workload, object *cnv.VirtualMachineSpe
 	memory := resource.NewQuantity(int64(vm.Flavor.RAM)*1024*1024, resource.BinarySI)
 	resourceRequests := map[core.ResourceName]resource.Quantity{}
 	resourceRequests[core.ResourceMemory] = *memory
+	object.Template.Spec.Domain.Memory.Guest = memory
```
liranr23, line 443: would be safer to create initiate the memory as it might be nil

liranr23 (commented):
please notice my comment, i think CI failing because of it (and most likely every migration will fail)

liranr23 (approved):

PR #839, opened 2024-03-28
Set guest memory

ahadas:
Previously we didn't set the guest memory but only memory request, since the guest memory property is optional and according to the documentation it "Defaults to the requested memory in the resources section if not specified.". However, if we don't set it and the migrated VM is based on a template, the value of this field would be retrieved from the template which may be different than what the VM had on the source and this could lead to issues, e.g., starting with insufficient memory for booting.


ahadas:
backport of #836

PR #841, opened 2024-04-01
Fix local preference

liranr23:
We always looked up for cluster wide preference even when we found a local one.

### DIFF
--- a/pkg/controller/plan/kubevirt.go
+++ b/pkg/controller/plan/kubevirt.go
@@ -1310,11 +1310,12 @@ func (r *KubeVirt) getPreference(vm *plan.VMStatus, preferenceName string) (name
 				"vm",
 				vm.String())
 		} else {
-			return
+			r.Log.Error(err, "could not fetch a local instance type preference for destination VM. trying cluster wide",
+				"vm",
+				vm.String())
 		}
+		name, kind, err = r.getVirtualMachineClusterPreference(vm, preferenceName)
 	}
-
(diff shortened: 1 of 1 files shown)


pkg/controller/plan/kubevirt.go
```diff
@@ -1304,17 +1304,12 @@ func (r *KubeVirt) getOsMapConfig(providerType v1beta1.ProviderType) (configMap
...
 	name, kind, err = r.getVirtualMachinePreference(preferenceName)
-	if err != nil {
-		if k8serr.IsNotFound(err) {
-			r.Log.Info("could not find a local instance type preference for destination VM. trying cluster wide",
-				"vm",
-				vm.String())
-		} else {
-			return
-		}
+	if err != nil && k8serr.IsNotFound(err) {
```
ahadas, line 1307: wouldn't it make sense to always fall back to the cluster's preference in case of an error? (sure, when there's no error we shouldn't proceed to getting the cluster's preference)

liranr23: i think that once we see an error we should return, clearly something bad happened.
we can also say: preferences aren't so important to our point of view, then try our best and fallback.

i prefer the first option

ahadas: > ..., clearly something bad happened

this "bad" thing might be lack of permissions for getting the preference in the target namespace..
if we say that we prioritized namespaced preferences first, then cluster preferences and then templates - the question is why to fall back from namespaced preferences to templates when we face an issue in querying the namespaced preference

ahadas (approved):

PR #843, opened 2024-04-01
Let the os maps be fully customized

liranr23:
Until now, if the user wished to create is own OS map for preferences the operator would think his new config map is managed by himself. This means it will populate the default values if they are not existing.


ahadas (approved):

ahadas:
> Until now, if the user wished to create is own OS map for preferences the operator would think his new config map is managed by himself. This means it will populate the default values if they are not existing.

his own*
by itself
if they don't exist* (@liranr23  what did you mean by that btw?)

ahadas:
> > > Until now, if the user wished to create is own OS map for preferences the operator would think his new config map is managed by himself. This means it will populate the default values if they are not existing.
> > 
> > 
> > his own* by itself* if they don't exist* (@liranr23 what did you mean by that btw?)
> 
> it means that for example you set a single OS in your new mapping. then the operator would populate all the other OS mappings that you didn't have

ok so you mean that the operator would add mappings for which the keys don't exist?

liranr23:
> > > > Until now, if the user wished to create is own OS map for preferences the operator would think his new config map is managed by himself. This means it will populate the default values if they are not existing.
> > > 
> > > 
> > > his own* by itself* if they don't exist* (@liranr23 what did you mean by that btw?)
> > 
> > 
> > it means that for example you set a single OS in your new mapping. then the operator would populate all the other OS mappings that you didn't have
> 
> ok so you mean that the operator would add mappings for which the keys don't exist?

exactly

ahadas:
> > > > > Until now, if the user wished to create is own OS map for preferences the operator would think his new config map is managed by himself. This means it will populate the default values if they are not existing.
> > > > 
> > > > 
> > > > his own* by itself* if they don't exist* (@liranr23 what did you mean by that btw?)
> > > 
> > > 
> > > it means that for example you set a single OS in your new mapping. then the operator would populate all the other OS mappings that you didn't have
> > 
> > 
> > ok so you mean that the operator would add mappings for which the keys don't exist?
> 
> exactly

ack, so please rephrase that part and we'll get it in

Issue #751, opened 2024-02-18
Bump the go version to 1.21


liranr23:
See more in #750

PR #847, opened 2024-04-02
Bump Go to 1.21

liranr23:
Fixes: #751

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -109,7 +109,7 @@ jobs:
         uses: actions/setup-go@v4
         with:
           # NOTE: Keep the version in sync with Go toolchain in WORKSPACE.
-          go-version: '1.20.10'
+          go-version: '1.21.9'
 
       # Setup the run environment and run CI test suite
       - name: Run test suite
--- a/.github/workflows/release-branch-pull-request.yml
+++ b/.github/workflows/release-branch-pull-request.yml
@@ -30,7 +30,7 @@ jobs:
             uses: actions/setup-go@v4
             with:
               # NOTE: Keep the version in sync with Go toolchain in WORKSPACE.
-              go-version: '1.20.10'
+              go-version: '1.21.9'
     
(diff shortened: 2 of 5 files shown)


liranr23:
@ahadas based on the UUID change, it may happen that if a user had an OVA plan previously and will upgrade it to 2.6.1 containing this commit, he will need to change the IDs in the plan.

bkhizgiy:
@liranr23 I think from the UI at least we should get a warning, since the ID will change the value in the inventory will also change and we should get `VM wasn't found` or something of a sort in the plan validations, otherwise yes its need to be manually changed.

liranr23:
> @liranr23 I think from the UI at least we should get a warning, since the ID will change the value in the inventory will also change and we should get `VM wasn't found` or something of a sort in the plan validations, it still need to be change but we will have indication for that.

I think that will be tough to do.
1. it's possible that really something change and the plan isn't ready.
2. even if we drop the point above, we will need to to see which plan was ready and changed state while comparing to when did we upgrade.

i think the best solution will be documenting it. if an existing plan was in ready status and after the upgrade it changed, new plan creation(or editing) with the new ids required.

tests/suit/framework/ova.go
```diff
@@ -17,7 +17,7 @@ func (r *OvaClient) LoadSourceDetails() (vm *OvaVM, err error) {
 		r.storageClass = "nfs-csi"
 	}
 
-	r.vmData.testVMId = "c5686650854d1e69b4123f4bf2e70fe1ed2a"
+	r.vmData.testVMId = "fdb4680df9160052f35e66c932e3fe5b62f5"
```
ahadas, line 20: need to look at our options and see if we cannot preserve those IDs

liranr23: that might be done if we are going to make them in another way. which means, it might change them anyway

ahadas: handled in #884

ahadas (approved):

PR #848, opened 2024-04-03
vSphere: Get static IP addresses

ahadas:


ahadas:
@liranr23 FYI

liranr23:
Based on the commit - each type of disk can be without datastore?

ahadas:
> Based on the commit - each type of disk can be without datastore?

yes

liranr23 (approved):

PR #849, opened 2024-04-03
Sanitize datastore identifiers from ESXi providers

ahadas:
when we get datastores from the ESXi SDK, their identifier may be in the form of '10.11.12.13:/vol/virtv2v/function' which leads to invalid endpoints in the inventory so we sanitize such identifiers


yaacov:
lgtm

liranr23 (approved):

PR #852, opened 2024-04-09
Map vSphere static IPs

liranr23:
This patch will get the static IPs configurations from the vSphere provider to each guest and will propogate it to virt-v2v to set it up while converting the VM.

### DIFF
--- a/operator/config/crd/bases/forklift.konveyor.io_plans.yaml
+++ b/operator/config/crd/bases/forklift.konveyor.io_plans.yaml
@@ -141,6 +141,9 @@ spec:
                 description: Preserve the CPU model and flags the VM runs with in
                   its oVirt cluster.
                 type: boolean
+              preserveStaticIPs:
+                description: Preserve static IPs of VMs in vSphere (Windows only)
+                type: boolean
               provider:
                 description: Providers.
                 properties:
--- a/pkg/controller/plan/adapter/base/doc.go
+++ b/pkg/controller/plan/adapter/base/doc.go
@@ -135,6 +135,8 @@ type Validator interface {
 	WarmMigration() bool
 	// Validate that no more than one of a VM's networks is mapped to the pod network.
 	PodNetwork(vmRef ref.Ref) (bool, error)
+	// Validate that we have information about static IPs for every virtual NIC
+	StaticIPs(vmRef ref.Ref) (bool, error)
 }
 
 // DestinationClient API.
--- a/pkg/controller/plan/adapter/ova/validator.go
+++ b/pkg/controller/plan/adapter/ova/validator.go
@@ -100,3 +100,8 @@ func (r *Validator) MaintenanceMode(vmRef ref.Ref) (ok bool, err error) {
 func (r *Validator) DirectStorage(vmRef ref.Ref) (bool, error) {
 	return true, nil
 }
+
+// NO-OP
+func (r *Validator) StaticIPs(vmRef ref.Ref) (bool, error) {
+	return true, nil
+}
--- a/pkg/controller/plan/adapter/ovirt/validator.go
+++ b/pkg/controller/plan/adapter/ovirt/validator.go
@@ -165,3 +165,9 @@ func (r *Validator) MaintenanceMode(_ ref.Ref) (ok bool, err error) {
 	ok = true
 	return
 }
+
+// NO-OP
+func (r *Validator) StaticIPs(vmRef ref.Ref) (bool, error) {
+	// the guest operating system is not modified during the migration so static IPs should be preserved
+	return true, nil
+}
--- a/pkg/controller/plan/adapter/vsphere/builder.go
+++ b/pkg/controller/plan/adapter/vsphere/builder.go
@@ -183,6 +183,8 @@ func (r *Builder) PodEnvironment(vmRef ref.Ref, sourceSecret *core.Secret) (env 
 		return
 	}
 
+	macsToIps := r.mapMacStaticIps(vm)
+
 	libvirtURL, fingerprint, err := r.getSourceDetails(vm, sourceSecret)
 	if err != nil {
 		return
@@ -209,7 +211,23 @@ func (r *Builder) PodEnvironment(vmRef ref.Ref, sourceSecret *core.Secret) (env 
 			Value: fingerprint,
 		},
 	)
+	if macsToIps != "" {
+		env = append(env, core.EnvVar{
+			Name:  "V2V_staticIPs",
+			Value: macsToIps,
+		})
+	}
 	return
+}
+
+func (r *Builder) mapMacStaticIps(vm *model.VM) string {
+	configurations := []string{}
+	for _, guestNetwork := range vm.GuestNetworks {
+		if guestNetwork.Origin == string(types.NetIpConfigInfoIpAddressOriginManual) {
+			configurations = append(configurations, fmt.Sprintf("%s:ip:%s", guestNetwork.MAC, guestNetwork.IP))
+		}
+	}
+	return strings.Join(configurations, "_")
 }
 
 func (r *Builder) getSourceDetails(vm *model.VM, sourceSecret *core.Secret) (libvirtURL liburl.URL, fingerprint string, err error) {
--- a/pkg/controller/plan/adapter/vsphere/builder.go
+++ b/pkg/controller/plan/adapter/vsphere/builder.go
@@ -66,6 +66,7 @@ const (
 	DefaultWindows = "win10"
 	DefaultLinux   = "rhel8.1"
 	Unknown        = "unknown"
+	WindowsPrefix  = "win"
 )
 
 // Annotations
@@ -224,6 +225,9 @@ func (r *Builder) PodEnvironment(vmRef ref.Ref, sourceSecret *core.Secret) (env 
 }
 
 func (r *Builder) mapMacStaticIps(vm *model.VM) string {
+	if !isWindows(vm) {
+		return ""
+	}
 	configurations := []string{}
 	for _, guestNetwork := range vm.GuestNetworks {
 		if guestNetwork.Origin == string(types.NetIpConfigInfoIpAddressOriginManual) {
@@ -231,6 +235,10 @@ func (r *Builder) mapMacStaticIps(vm *model.VM) string {
 		}
 	}
 	return strings.Join(configurations, "_")
+}
+
+func isWindows(vm *model.VM) bool {
+	return strings.Contains(vm.GuestID, WindowsPrefix) || strings.Contains(vm.GuestName, WindowsPrefix)
 }
 
 func (r *Builder) getSourceDetails(vm *model.VM, sourceSecret *core.Secret) (libvirtURL liburl.URL, fingerprint string, err error) {
@@ -721,7 +729,7 @@ func (r *Builder) TemplateLabels(vmRef ref.Ref) (labels map[string]string, err e
 		os = osMap[vm.GuestID]
 	} else if strings.Contains(vm.GuestName, "linux") || strings.Contains(vm.GuestName, "rhel") {
 		os = DefaultLinux
-	} else if strings.Contains(vm.GuestName, "win") {
+	} else if strings.Contains(vm.GuestName, WindowsPrefix) {
 		os = DefaultWindows
 	} else {
 		os = Unknown
--- a/pkg/controller/plan/adapter/vsphere/builder.go
+++ b/pkg/controller/plan/adapter/vsphere/builder.go
@@ -183,7 +183,10 @@ func (r *Builder) PodEnvironment(vmRef ref.Ref, sourceSecret *core.Secret) (env 
 		return
 	}
 
-	macsToIps := r.mapMacStaticIps(vm)
+	macsToIps := ""
+	if r.Plan.Spec.PreserveStaticIPs {
+		macsToIps = r.mapMacStaticIps(vm)
+	}
 
 	libvirtURL, fingerprint, err := r.getSourceDetails(vm, sourceSecret)
 	if err != nil {
--- a/pkg/controller/plan/adapter/vsphere/validator.go
+++ b/pkg/controller/plan/adapter/vsphere/validator.go
@@ -127,3 +127,30 @@ func (r *Validator) MaintenanceMode(vmRef ref.Ref) (ok bool, err error) {
 func (r *Validator) DirectStorage(vmRef ref.Ref) (bool, error) {
 	return true, nil
 }
+
+// Validate that we have information about static IPs for every virtual NIC
+func (r *Validator) StaticIPs(vmRef ref.Ref) (ok bool, err error) {
+	if !r.plan.Spec.PreserveStaticIPs {
+		return true, nil
+	}
+	vm := &model.Workload{}
+	err = r.inventory.Find(vm, vmRef)
+	if err != nil {
+		err = liberr.Wrap(err, "vm", vmRef)
+		return
+	}
+	for _, nic := range vm.NICs {
+		found := false
+		for _, guestNetwork := range vm.GuestNetworks {
+			if nic.MAC == guestNetwork.MAC {
+				found = true
+				break
+			}
+		}
+		if !found {
+			return
+		}
+	}
+	ok = true
+	return
+}
--- a/pkg/controller/plan/adapter/vsphere/validator.go
+++ b/pkg/controller/plan/adapter/vsphere/validator.go
@@ -139,6 +139,10 @@ func (r *Validator) StaticIPs(vmRef ref.Ref) (ok bool, err error) {
 		err = liberr.Wrap(err, "vm", vmRef)
 		return
 	}
+	if !isWindows(&vm.VM) {
+		return true, nil
+	}
+
 	for _, nic := range vm.NICs {
 		found := false
 		for _, guestNetwork := range vm.GuestNetworks {
--- a/pkg/controller/plan/adapter/vsphere/validator_test.go
+++ b/pkg/controller/plan/adapter/vsphere/validator_test.go
@@ -27,10 +27,14 @@ func (m *mockInventory) Find(resource interface{}, ref ref.Ref) error {
 				},
 				GuestNetworks: []vsphere.GuestNetwork{
 					{MAC: "mac1"},
-				}},
+				},
+				GuestID: "windows7Guest"},
 		}
 		if ref.Name == "full_guest_network" {
 			res.VM.GuestNetworks = append(res.VM.GuestNetworks, vsphere.GuestNetwork{MAC: "mac2"})
+		}
+		if ref.Name == "not_windows_guest" {
+			res.VM.GuestID = "rhel8_64Guest"
 		}
 	}
 	return nil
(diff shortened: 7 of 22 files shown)


pkg/controller/provider/container/vsphere/model.go
```diff
@@ -613,6 +613,26 @@ func (v *VmAdapter) Apply(u types.ObjectUpdate) {
...
+						ipConfigIp := &info.IpConfig.IpAddress
+						for _, ip := range *ipConfigIp {
+							guestNetworksList = append(guestNetworksList, model.GuestNetwork{
+								MAC:    info.MacAddress,
+								IP:     ip.IpAddress,
+								Origin: ip.Origin,
+							})
+						}
+					}
+					if len(guestNetworksList) > 0 {
```
liranr23, line 632: this condition is because once the vm is shutdown, we get an update of empty values.
maybe we can make a better condition. as now if the user remove all his networks in the vm, we won't let it here.

pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -209,9 +211,26 @@ func (r *Builder) PodEnvironment(vmRef ref.Ref, sourceSecret *core.Secret) (env
 			Value: fingerprint,
 		},
 	)
+	if macsToIps != "" {
+		env = append(env, core.EnvVar{
+			Name:  "V2V_staticips",
```
ahadas, line 216:
```suggestion
			Name:  "V2V_staticIPs",
```

pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -209,9 +211,26 @@ func (r *Builder) PodEnvironment(vmRef ref.Ref, sourceSecret *core.Secret) (env
...
+			Value: macsToIps,
+		})
+	}
 	return
 }
 
+func (r *Builder) mapMacStaticIps(vm *model.VM) string {
+	configurations := []string{}
+	for _, guestNetwork := range vm.GuestNetworks {
+		if guestNetwork.Origin != string(types.NetIpConfigInfoIpAddressOriginManual) {
```
ahadas, line 226: how about being more positive :)
```suggestion
		if guestNetwork.Origin == string(types.NetIpConfigInfoIpAddressOriginManual) {
```
and then call `append`?

pkg/controller/provider/container/vsphere/model.go
```diff
@@ -613,6 +613,26 @@ func (v *VmAdapter) Apply(u types.ObjectUpdate) {
...
 					}
 				}
+			case fGuestNet:
+				if nics, cast := p.Val.(types.ArrayOfGuestNicInfo); cast {
+					guestNetworksList := []model.GuestNetwork{}
+					for _, info := range nics.GuestNicInfo {
+						if info.IpConfig == nil {
+							continue
+						}
+						ipConfigIp := &info.IpConfig.IpAddress
```
ahadas, line 623: why taking the reference here?

pkg/controller/provider/container/vsphere/model.go
ahadas: it's not that likely that someone would remove all the networks but even if that happens, I don't see an issue in keeping the mac to ip-address set, do you see a problem with this?

and if it was important enough to add this comment here, then it should also be on the source code

ahadas (commented):
I see that this is a work in progress - but commented with early feedback

pkg/controller/provider/container/vsphere/model.go
liranr23: i didn't try to run virt-v2v with non-existing MAC on the guest, i think it will just ignore it, in that case we are good.

liranr23: tested - works

virt-v2v/cold/entrypoint.go
```diff
@@ -86,6 +87,12 @@ func main() {
 			}
 		}
 		virtV2vArgs = append(virtV2vArgs, "-ip", "/etc/secret/secretKey")
+		macsToIps := os.Getenv("V2V_staticIPs")
```
bkhizgiy, line 90: Since we only appending this var to the pod only when its its not empty, you can use `checkEnvVariablesSet(V2V_staticIPs)` for the IF check,  and then get its value directly in the for loop.

operator/config/crd/bases/forklift.konveyor.io_plans.yaml
```diff
@@ -141,6 +141,9 @@ spec:
                 description: Preserve the CPU model and flags the VM runs with in
                   its oVirt cluster.
                 type: boolean
+              preserveStaticIPs:
+                description: Preserve static IPs of VMs in vSphere cluster
```
ahadas, line 145:
```suggestion
                description: Preserve static IPs of VMs in vSphere
```
and how about setting the expectation properly by saying it applies to Windows guests only?

pkg/controller/plan/adapter/base/doc.go
```diff
@@ -135,6 +135,8 @@ type Validator interface {
 	WarmMigration() bool
 	// Validate that no more than one of a VM's networks is mapped to the pod network.
 	PodNetwork(vmRef ref.Ref) (bool, error)
+	// Validate that every VM NIC has it's data from the guest to get static IPs
```
ahadas, line 138:
```suggestion
	// Validate that we have information about static IPs for every virtual NIC
```

pkg/controller/plan/adapter/ovirt/validator.go
```diff
@@ -165,3 +165,8 @@ func (r *Validator) MaintenanceMode(_ ref.Ref) (ok bool, err error) {
 	ok = true
 	return
 }
+
+// NO-OP
+func (r *Validator) StaticIPs(vmRef ref.Ref) (bool, error) {
+	return true, nil
```
ahadas, line 171: here we can comment that the guest operating system is not modified during the migration so static IPs should be preserved (also for OpenStack above)

pkg/controller/plan/adapter/ova/validator.go
```diff
@@ -100,3 +100,8 @@ func (r *Validator) MaintenanceMode(vmRef ref.Ref) (ok bool, err error) {
 func (r *Validator) DirectStorage(vmRef ref.Ref) (bool, error) {
 	return true, nil
 }
+
+// NO-OP
+func (r *Validator) StaticIPs(vmRef ref.Ref) (bool, error) {
+	return true, nil
```
ahadas, line 106: I wonder if there's any indication of static IPs in the OVF, even though I doubt it, but that's something for another day

pkg/controller/plan/adapter/vsphere/validator.go
```diff
@@ -127,3 +127,29 @@ func (r *Validator) MaintenanceMode(vmRef ref.Ref) (ok bool, err error) {
...
+
+// Validate that every VM NIC has it's data from the guest to get static IPs
+func (r *Validator) StaticIPs(vmRef ref.Ref) (ok bool, err error) {
+	if !r.plan.Spec.PreserveStaticIPs {
+		return true, nil
+	}
+	vm := &model.Workload{}
+	err = r.inventory.Find(vm, vmRef)
+	if err != nil {
+		err = liberr.Wrap(err, "vm", vmRef.String())
```
ahadas, line 139:
```suggestion
		err = liberr.Wrap(err, "vm", vmRef)
```

pkg/controller/plan/adapter/vsphere/validator.go
```diff
@@ -127,3 +127,29 @@ func (r *Validator) MaintenanceMode(vmRef ref.Ref) (ok bool, err error) {
...
+		return
+	}
+	for _, nic := range vm.NICs {
+		found := false
```
ahadas, line 143: I think the code within this for-loop would be [better with a label](https://relistan.com/continue-statement-with-labels-in-go), but if you have something against labels then at least we can break when we find a corresponding guestNetwork for the NIC..

pkg/controller/plan/validation.go
```diff
@@ -432,6 +434,14 @@ func (r *Reconciler) validateVM(plan *api.Plan) error {
 		Message:  "VM has more than one interface mapped to the pod network.",
 		Items:    []string{},
 	}
+	missingStaticIPs := libcnd.Condition{
+		Type:     VMMissingGuestIPs,
+		Status:   True,
+		Reason:   MissingGuestInfo,
+		Category: Critical,
+		Message:  "VM is missing Static IPs, please try to boot the VM first.",
```
ahadas, line 442: "VM is missing static IPs" doesn't reflect the problem good enough (it sounds like the VM is not configured with static IPs which is ok), how about
```suggestion
		Message:  "Guest information on vNICs is missing, cannot preserve static IPs. Make sure VMware tools are installed and the VM is running.",
```

pkg/controller/provider/container/vsphere/model.go
```diff
@@ -613,6 +613,26 @@ func (v *VmAdapter) Apply(u types.ObjectUpdate) {
...
+						}
+						for _, ip := range info.IpConfig.IpAddress {
+							guestNetworksList = append(guestNetworksList, model.GuestNetwork{
+								MAC:    info.MacAddress,
+								IP:     ip.IpAddress,
+								Origin: ip.Origin,
+							})
+						}
+					}
+					// once the vm is shutdown, we get an update of empty values - this condition allows us to keep them.
```
ahadas, line 631:
```suggestion
					// when the vm goes down, we get an update with empty values - the following check keeps the previously reported data.
```

pkg/lib/condition/condition.go
```diff
@@ -313,6 +317,11 @@ func (r *Conditions) HasBlockerCondition() bool {
 	return r.HasConditionCategory(Critical, Error)
 }
 
+// The collection contains a blocker conditions that needs the plan to keep reconcile.
```
ahadas, line 320:
```suggestion
// The collection contains blocker conditions that keep the plan reconciling.
```

pkg/controller/plan/adapter/vsphere/validator.go
liranr23: do we ever used the labels in the project? i'm unsure about my opinion to add `goto` in the code

operator/config/crd/bases/forklift.konveyor.io_plans.yaml
liranr23: added, please re-check

pkg/controller/plan/adapter/ova/validator.go
liranr23: i doubt it as well, maybe @bkhizgiy saw something in a OVF recently. as far i can remember, the option is to save MACs.

pkg/controller/plan/adapter/vsphere/validator.go
ahadas: apparently, [we do](https://github.com/kubev2v/forklift/blob/main/pkg/controller/provider/container/vsphere/collector.go#L300)

operator/config/crd/bases/forklift.konveyor.io_plans.yaml
ahadas: looks good now

pkg/controller/plan/adapter/ova/validator.go
bkhizgiy: @ahadas @liranr23 Haha, I checked that after reviewing this PR yesterday, was having the same question:)
So yes, we can have a static IP address in the OVF, but it's not part of the basic configuration. It should be added manually as a property by the user (alongside other available configurations). The MAC, on the other hand, can be specified directly in the network configurations (and we are fetching this data). So, given it's not a standard configuration, I'm not sure if we want (or can) to support it..

bkhizgiy (approved):

pkg/controller/plan/adapter/vsphere/validator.go
liranr23: i prefer it as is, but i might need to verify again the PR after the last change

ahadas (approved):

ahadas:
@liranr23 since we have more time for getting this PR in, could you please check where it makes sense to add more tests? this PR indeed adds quite a few lines without tests...

yaacov:
https://github.com/kubev2v/forklift-console-plugin/pull/1138 - UI for save IPs flag

pkg/controller/plan/adapter/vsphere/validator_test.go
```diff
@@ -0,0 +1,121 @@
...
+
+func (m *mockInventory) Finder() web.Finder {
+	return nil
+}
+
+func (m *mockInventory) Get(resource interface{}, id string) error {
+	return nil
+}
+
+//nolint:nilnil
```
bennyz, line 47: I think you can just put it at the top of the file

liranr23: looks like i don't :)

liranr23:
as i talked with Sharon today. i didn't address that currently it's for Windows VMs only.

pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -209,9 +215,35 @@ func (r *Builder) PodEnvironment(vmRef ref.Ref, sourceSecret *core.Secret) (env
...
+	for _, guestNetwork := range vm.GuestNetworks {
+		if guestNetwork.Origin == string(types.NetIpConfigInfoIpAddressOriginManual) {
+			configurations = append(configurations, fmt.Sprintf("%s:ip:%s", guestNetwork.MAC, guestNetwork.IP))
+		}
+	}
+	return strings.Join(configurations, "_")
+}
+
+func IsWindows(vm *model.VM) bool {
+	if vm.GuestID != "" && strings.Contains(vm.GuestID, WindowsPrefix) || strings.Contains(vm.GuestName, WindowsPrefix) {
```
ahadas, line 241: why do we need to check both GuestID and GuestName here? in `TemplateLabels` we only check GuestName

pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -209,9 +215,35 @@ func (r *Builder) PodEnvironment(vmRef ref.Ref, sourceSecret *core.Secret) (env
...
+		return true
+	}
+	return false
```
ahadas, line 244: we can simply return the result of the condition above then

pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -209,9 +215,35 @@ func (r *Builder) PodEnvironment(vmRef ref.Ref, sourceSecret *core.Secret) (env
...
+	configurations := []string{}
+	for _, guestNetwork := range vm.GuestNetworks {
+		if guestNetwork.Origin == string(types.NetIpConfigInfoIpAddressOriginManual) {
+			configurations = append(configurations, fmt.Sprintf("%s:ip:%s", guestNetwork.MAC, guestNetwork.IP))
+		}
+	}
+	return strings.Join(configurations, "_")
+}
+
+func IsWindows(vm *model.VM) bool {
```
ahadas, line 240: do we call this outside of the package?

virt-v2v/cold/cold_test.go
```diff
@@ -27,3 +28,32 @@ func TestGenName(t *testing.T) {
...
+	}
+
+	for _, c := range cases {
+		if c.inputConfig != "" {
+			t.Setenv("V2V_staticIPs", c.inputConfig)
+		}
+		command := strings.Join(buildCommand(), " ")
+		for _, outputArg := range c.outputArgs {
+			if !strings.Contains(command, outputArg) {
+				t.Errorf("The command is: %s. Excpected to have '%s' in it", command, outputArg)
```
ahadas, line 55:
```suggestion
				t.Errorf("The command is: %s. Excpected to contain '%s'", command, outputArg)
```

virt-v2v/cold/entrypoint.go
```diff
@@ -73,23 +101,31 @@ func main() {
...
-				fmt.Println("Error creating ca cert link ", err)
-				os.Exit(1)
-			}
-		} else {
-			// otherwise, keep system pool certificates
-			err := os.Symlink("/etc/pki/tls/certs/ca-bundle.crt.bak", "/opt/ca-bundle.crt")
-			if err != nil {
-				fmt.Println("Error creating ca cert link ", err)
-				os.Exit(1)
+		if os.Getenv("V2V_test") != "true" {
```
ahadas, line 104: ah?..

pkg/controller/plan/adapter/vsphere/builder.go
liranr23: we check both with priority to `GuestID`.
```
if vm.GuestID != "" {
		os = osMap[vm.GuestID]
	} else if strings.Contains(vm.GuestName, "linux") || strings.Contains(vm.GuestName, "rhel") {
		os = DefaultLinux
	} else if strings.Contains(vm.GuestName, "win") {
		os = DefaultWindows
```

virt-v2v/cold/entrypoint.go
liranr23: yeah.. in the tests we can't do the symbol linking.. i tried to overcome it in a creative way

i'm up for suggestions

liranr23: maybe just pulling it out of this function will do :)

pkg/controller/plan/adapter/vsphere/builder.go
ahadas: yeah, ok, I missed the checks of GuestName below somehow

pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -209,9 +215,32 @@ func (r *Builder) PodEnvironment(vmRef ref.Ref, sourceSecret *core.Secret) (env
...
+	for _, guestNetwork := range vm.GuestNetworks {
+		if guestNetwork.Origin == string(types.NetIpConfigInfoIpAddressOriginManual) {
+			configurations = append(configurations, fmt.Sprintf("%s:ip:%s", guestNetwork.MAC, guestNetwork.IP))
+		}
+	}
+	return strings.Join(configurations, "_")
+}
+
+func isWindows(vm *model.VM) bool {
+	return vm.GuestID != "" && strings.Contains(vm.GuestID, WindowsPrefix) || strings.Contains(vm.GuestName, WindowsPrefix)
```
ahadas, line 241: seems we can simplify this even further with
```suggestion
	return strings.Contains(vm.GuestID, WindowsPrefix) || strings.Contains(vm.GuestName, WindowsPrefix)
```
right?

virt-v2v/cold/entrypoint.go
```diff
@@ -73,22 +119,13 @@ func main() {
...
-			}
-		} else {
-			// otherwise, keep system pool certificates
-			err := os.Symlink("/etc/pki/tls/certs/ca-bundle.crt.bak", "/opt/ca-bundle.crt")
-			if err != nil {
-				fmt.Println("Error creating ca cert link ", err)
-				os.Exit(1)
+		virtV2vArgs = append(virtV2vArgs, "-ip", "/etc/secret/secretKey")
+
+		if checkEnvVariablesSet("V2V_staticIPs") {
```
ahadas, line 124: really, what's the point in calling os.Getenv twice? you need the value of V2V_staticIPs here, so why not getting it and then check it to avoid another system call?..

virt-v2v/cold/entrypoint.go
ahadas: > maybe just pulling it out of this function will do :)

yep, that's a better approach - it was an indication of having too many responsibilities within the same function

ahadas (approved):

PR #853, opened 2024-04-09
Don't set time within clock

liranr23:
When using preferences the timer configuration may conflict with the empty timer we used to set. That would lead windows VMs to fail running.

### DIFF
--- a/pkg/controller/plan/adapter/ovirt/builder.go
+++ b/pkg/controller/plan/adapter/ovirt/builder.go
@@ -350,9 +350,7 @@ func (r *Builder) mapInput(object *cnv.VirtualMachineSpec) {
 
 func (r *Builder) mapClock(vm *model.Workload, object *cnv.VirtualMachineSpec) {
 	if object.Template.Spec.Domain.Clock == nil {
-		object.Template.Spec.Domain.Clock = &cnv.Clock{
-			Timer: &cnv.Timer{},
-		}
+		object.Template.Spec.Domain.Clock = &cnv.Clock{}
 	}
 
 	timezone := cnv.ClockOffsetTimezone(vm.Timezone)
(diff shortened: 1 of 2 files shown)


pkg/controller/plan/adapter/ovirt/builder.go
```diff
@@ -350,9 +350,10 @@ func (r *Builder) mapInput(object *cnv.VirtualMachineSpec) {
 
 func (r *Builder) mapClock(vm *model.Workload, object *cnv.VirtualMachineSpec) {
 	if object.Template.Spec.Domain.Clock == nil {
-		object.Template.Spec.Domain.Clock = &cnv.Clock{
-			Timer: &cnv.Timer{},
-		}
+		object.Template.Spec.Domain.Clock = &cnv.Clock{}
+	}
+	if object.Preference == nil && object.Template.Spec.Domain.Clock.Timer == nil {
+		object.Template.Spec.Domain.Clock.Timer = &cnv.Timer{}
```
ahadas, line 356: can't we just avoid setting the Timer? it doesn't seem we do anything with it..

liranr23: i am afraid of the case of empty vm (no template/preference at all). but i think we can drop it. do you recall why we added it?

ahadas: it's there [for quite some time](https://github.com/kubev2v/forklift-controller/commit/4b417a59b184bbacd46240419e0fb6c31fb751c6#diff-08ba6abe786bb864c38a0ada1520a41bfcf8f5c5f8a7b491e8b97b4637159a42), if we don't set anything inside then passing an empty struct is redundant - let's drop it

ahadas (approved):

liranr23:
failures in checks unrelated to the PR

PR #854, opened 2024-04-09
vSphere: warn when independent disks are detected

ahadas:


validation/policies/io/konveyor/forklift/vmware/disk_mode.rego
```diff
@@ -0,0 +1,16 @@
...
+
+independent_disk {
+    some i
+    input.disks[i].mode in ["independent_persistent", "independent_nonpersistent"]
+}
+
+concerns[flag] {
+    independent_disk
+    flag := {
+        "category": "Warning",
```
liranr23, line 12: is it just a warning? or critical?
can we migrate these kind of disks?

seeing the message - we can't, i think it should be critical. once a VDDK will support it, we can switch to warning.

ahadas: the category is set to Warning because it is possible to migrate such disks without VDDK (which will be significantly slower) or presumably with older versions of VDDK

liranr23: maybe it worth a mention somewhere(not only OVA workaround)? even a link to docs?

ahadas: that's already mentioned implicitly: Independent disks cannot be transferred using **recent** versions of VDDK
about link to docs, I'm not sure which docs you refer to - but the idea here is to recommend changing the disks to Dependent, that's the best resolution

liranr23 (approved):

PR #858, opened 2024-04-11
Drop retainAfterCompletion

liranr23:
Removed the retainAfterCompletion annotation from DataVolumes.
This is still configurable by setting `FEATURE_RETAIN_PRECOPY_IMPORTER_PODS`.


ahadas:
> Removed the retainAfterCompletion annotation from DataVolumes for populator flows. When using CDI, use the previous condition.

but we currently don't create data volumes with volume populators, no?

liranr23:
> > Removed the retainAfterCompletion annotation from DataVolumes for populator flows. When using CDI, use the previous condition.
> 
> but we currently don't create data volumes with volume populators, no?

yeah, so maybe we just need to drop the warm condition and set the default value of this config to false..

ahadas (approved):

PR #859, opened 2024-04-14
vSphere: skip immediate binding when using virt-v2v on el9

ahadas:
In b0d0a68, we incorrectly dropped a check that skips setting the "cdi.kubevirt.io/storage.bind.immediate.requested" annotation when using virt-v2v on el9 - in that case, the conversion pod acts as the first-consumer of the data volume and thus, there is no need in artificial first-consumer pod.

See https://issues.redhat.com/browse/CNV-40266

### DIFF
--- a/pkg/controller/plan/kubevirt.go
+++ b/pkg/controller/plan/kubevirt.go
@@ -1159,8 +1159,12 @@ func (r *KubeVirt) dataVolumes(vm *plan.VMStatus, secret *core.Secret, configMap
 		annotations[AnnDefaultNetwork] = path.Join(
 			r.Plan.Spec.TransferNetwork.Namespace, r.Plan.Spec.TransferNetwork.Name)
 	}
-	// Set annotation for WFFC storage classes
-	annotations[planbase.AnnBindImmediate] = "true"
+	if r.Plan.Spec.Warm || !r.Destination.Provider.IsHost() || r.Plan.IsSourceProviderOCP() {
+		// Set annotation for WFFC storage classes. Note that we create data volumes while
+		// running a cold migration to the local cluster only when the source is either OpenShift
+		// or vSphere, and in the latter case the conversion pod acts as the first-consumer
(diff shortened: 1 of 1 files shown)


liranr23:
i don't see the benefit of this change since CDI fixed the bug on their end. why do we care if the PVC binds when the virt-v2v pod starts or when creating the PVC?

ahadas:
> i don't see the benefit of this change since CDI fixed the bug on their end. why do we care if the PVC binds when the virt-v2v pod starts or when creating the PVC?

the benefit is twofold:
1. imports would keep working with earlier versions of OpenShift Virtualization that don't include the fix
2. it removes redundant phase of creating another pod, this accelerating the migration

pkg/controller/plan/kubevirt.go
```diff
@@ -1159,8 +1159,12 @@ func (r *KubeVirt) dataVolumes(vm *plan.VMStatus, secret *core.Secret, configMap
 		annotations[AnnDefaultNetwork] = path.Join(
 			r.Plan.Spec.TransferNetwork.Namespace, r.Plan.Spec.TransferNetwork.Name)
 	}
-	// Set annotation for WFFC storage classes
-	annotations[planbase.AnnBindImmediate] = "true"
+	if r.Plan.Spec.Warm || !r.Destination.Provider.IsHost() || r.Plan.IsSourceProviderOCP() {
```
liranr23, line 1162: maybe a comment we should drop this condition once 4.14 EOL?

liranr23 (approved):

pkg/controller/plan/kubevirt.go
ahadas: @liranr23 even then it would be better not to request CDI to immediately bind but to wait for the conversion pod
in the long run, we should see how it would work with volume populators that run by CDI and virt-v2v on el9 should be started by CDI in a similar way in cold migrations to a remote cluster

bkhizgiy (approved):

PR #862, opened 2024-04-14
Skip querying preference if OS is empty

liranr23:
We may have VM without OS set. In such case we get an empty string from the map. This patch will prevent querying and erroring when we hit this case.

### DIFF
--- a/pkg/controller/plan/kubevirt.go
+++ b/pkg/controller/plan/kubevirt.go
@@ -1272,6 +1272,10 @@ func (r *KubeVirt) vmPreference(vm *plan.VMStatus) (virtualMachine *cnv.VirtualM
 	if err != nil {
 		return
 	}
+	if preferenceName == "" {
+		err = liberr.New("couldn't find a corresponding preference", "vm", vm)
+		return
+	}
 
 	preferenceName, kind, err := r.getPreference(vm, preferenceName)
 	if err != nil {


pkg/controller/plan/kubevirt.go
```diff
@@ -1272,6 +1272,10 @@ func (r *KubeVirt) vmPreference(vm *plan.VMStatus) (virtualMachine *cnv.VirtualM
 	if err != nil {
 		return
 	}
+	if preferenceName == "" {
+		err = liberr.New("couldn't find the right OS.", "vm", vm.String())
```
ahadas, line 1276: if we replace `vm.String()` with `vm`, would we get all its properties?

pkg/controller/plan/kubevirt.go
```diff
@@ -1272,6 +1272,10 @@ func (r *KubeVirt) vmPreference(vm *plan.VMStatus) (virtualMachine *cnv.VirtualM
 	if err != nil {
 		return
 	}
+	if preferenceName == "" {
+		err = liberr.New("couldn't find the right OS.", "vm", vm)
```
ahadas, line 1276: please rephrase that message, it should probably say something like "couldn't find a corresponding preference"

liranr23: that semi-true. we saw it when the incoming VM had the OS field missing, probably template will fail as well.
since it might be on different cases as well (os that is not in the config map), it does make sense to switch it into preference case message :)

ahadas (approved):

PR #863, opened 2024-04-16
ocp: increase export timeout to 12 hours

bennyz:
Reduce the amount of timeouts for larger PVCs

### DIFF
--- a/pkg/settings/migration.go
+++ b/pkg/settings/migration.go
@@ -98,7 +98,9 @@ func (r *Migration) Load() (err error) {
 		return liberr.Wrap(fmt.Errorf("failed to find environment variable %s", VirtV2vImage))
 	}
 	r.VirtV2vDontRequestKVM = getEnvBool(VirtV2vDontRequestKVM, false)
-	if r.CDIExportTokenTTL, err = getPositiveEnvLimit(CDIExportTokenTTL, 0); err != nil {
+
+	// Set timeout to 12 hours instead of the default 2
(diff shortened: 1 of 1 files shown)


pkg/settings/migration.go
```diff
@@ -98,7 +98,9 @@ func (r *Migration) Load() (err error) {
 		return liberr.Wrap(fmt.Errorf("failed to find environment variable %s", VirtV2vImage))
 	}
 	r.VirtV2vDontRequestKVM = getEnvBool(VirtV2vDontRequestKVM, false)
-	if r.CDIExportTokenTTL, err = getPositiveEnvLimit(CDIExportTokenTTL, 0); err != nil {
+
+	// Set timeout to 8 hours instead of the default 2
```
liranr23, line 102: i think it worth more details, why we needed the increase of time?(on the commit as well)

pkg/settings/migration.go
```diff
@@ -98,7 +98,9 @@ func (r *Migration) Load() (err error) {
...
+	if r.CDIExportTokenTTL, err = getPositiveEnvLimit(CDIExportTokenTTL, 720); err != nil {
```
bkhizgiy, line 103: What does this number mean, is it minutes? Maybe it's worth using a constant instead to make it easier to read

bkhizgiy (approved):
@bennyz you still need to update the PR title

ahadas (approved):

PR #864, opened 2024-04-16
ocp: do not return error when setting condition

bennyz:


pkg/controller/plan/util/ocp.go
ahadas, line 1: did you intend to include this file?

bennyz: no 😅

liranr23 (approved):

ahadas:
@bennyz can you double check if the commit exists on the `main` branch?..

PR #865, opened 2024-04-16
Don't cleanup forklift entities on remote cluster 

liranr23:
When MTV is not installed on the remote OCP cluster and we try to
cleanup we won't succeed. This patch won't try to remove or edit forklift
entities, as if MTV doesn't exist there, it will fail - causing the plan
to fail.


ahadas (commented):
> When MTV is not installed on the remote OCP cluster and we try to cleanup we won't succeed. This patch tries to remove them, if MTV exists it will work. Otherwise, it won't. But we won't show an error about it.

that's one way to look at it but I'd say that we should be about to perform cleanup but without removing forklift entities - how about skip removing the "populator data sources" when migrating to a remote cluster so that the cleanup would succeed?

ahadas (approved):

PR #313, opened 2023-04-27
Operator: Use original control-plane name

mnecas:
Issue: The control-plane field is immutable so once the operator tried to deploy the 2.4 it failed.
Fixes: https://github.com/kubev2v/forklift/issues/314

PR #866, opened 2024-04-16
operator: change matchLabels of operator deployment

ahadas:
The matcher includes the label "name: controller-manager". This differs from what we set in MTV (there 'name' is mapped to 'forklift-operator') and this can lead to errors when creating a new release of MTV.

Here we change the aforementioned label to match that of MTV. This change would break upgradea to Forklift 2.7.0, however, there is an easy workaround for this by removing the custom resource 'forklift-controller' of type ForkliftController from the namespace the operator is deployed on.

### DIFF
--- a/operator/config/manager/manager.yaml
+++ b/operator/config/manager/manager.yaml
@@ -10,19 +10,17 @@ kind: Deployment
 metadata:
   name: forklift-operator
   namespace: system
-  labels:
-    control-plane: controller-manager
 spec:
   replicas: 1
   selector:
     matchLabels:
       app: forklift
-      name: controller-manager
+      name: forklift-operator
   template:
     metadata:
       labels:
         app: forklift
-        name: controller-manager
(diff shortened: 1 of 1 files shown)


liranr23:
why not revertin https://github.com/kubev2v/forklift/pull/313 if we go in this direction?

ahadas:
> why not reverting #313 if we go in this direction?

it's possible to change the other labels (even though some of them are better be dropped..) but that's a different kind of change, this change solves the misalignment between Forklift and MTV

mnecas:
Is it the only misalignment between the forklift and MTV which requires the forklift-controller recreation?
If we want to unify it and force users to recreate the forklift-controller we should do all possible changes at once.

liranr23:
typo in the commit `upgradea` -> `upgrade`

ahadas:
I think this change is still needed but I have no time to work on this anymore, therefore closing

PR #867, opened 2024-04-17
Skip querying preference if OS is empty (backport)

liranr23:
We may have VM without OS set. In such case we get an empty string from the map. This patch will prevent querying and erroring when we hit this case.


ahadas:
backport of #862

ahadas (approved):

PR #872, opened 2024-04-24
Fix virt-v2v disk name generation

bkhizgiy:
Currently, the function generating the disk names didn't work as expected, and instead of matching the disk number with the corresponding letter, it caused the letters to be attached from the previous disk. This resulted in only the first disk being properly copied due to incorrect linking. With these changes, the disk name generation works as expected.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -83,7 +83,7 @@ all: test forklift-controller
 
 # Run tests
 test: generate fmt vet manifests validation-test
-	go test -coverprofile=cover.out ./pkg/... ./cmd/...
+	go test -coverprofile=cover.out ./pkg/... ./cmd/... ./virt-v2v/cold/...
 
 # Experimental e2e target
 e2e-sanity: e2e-sanity-ovirt e2e-sanity-vsphere
@@ -113,6 +113,7 @@ e2e-sanity-ova:
 # Build forklift-controller binary
 forklift-controller: generate fmt vet
 	go build -o bin/forklift-controller github.com/konveyor/forklift-controller/cmd/forklift-controller
+
 
 # Build manager binary with compiler optimizations disabled
 debug: generate fmt vet
--- a/virt-v2v/cold/entrypoint.go
+++ b/virt-v2v/cold/entrypoint.go
@@ -136,11 +136,15 @@ func genName(diskNum int) string {
 		return ""
 	}
 
-	letters := []int32("abcdefghijklmnopqrstuvwxyz")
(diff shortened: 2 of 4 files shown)


bennyz (approved):

virt-v2v/cold/entrypoint.go
```diff
@@ -135,12 +135,7 @@ func genName(diskNum int) string {
 	if diskNum <= 0 {
 		return ""
 	}
-
-	letters := []int32("abcdefghijklmnopqrstuvwxyz")
-	index := (diskNum - 1) % len(letters)
-	cycels := (diskNum - 1) % len(letters)
-
-	return genName(cycels) + string(letters[index])
+	return fmt.Sprintf("sd%c", 'a'+(diskNum-1)%26)
```
liranr23, line 138: i would put `26` as a const

liranr23 (approved):

ahadas (commented):
I've made changes in the description of the PR, please update the commit (which is currently empty) accordingly
and also please add unit tests, there's a concrete case we can take from https://issues.redhat.com/browse/MTV-1098

germanovm:
Won't this break at the 27th disk now by both the 1st and 27th disk being named sda instead of sdaa for the 27th?

bkhizgiy:
@germanovm ahh yes, it will, we need to add the cycle back here like we had in the original code (but without the bug:) )

ahadas:
> @germanovm ahh yes, it will, we need to add the cycle back here like we had in the original code (but without the bug:) )

another case to cover in unit tests ;)

Makefile
```diff
@@ -84,6 +84,9 @@ all: test forklift-controller
 # Run tests
 test: generate fmt vet manifests validation-test
 	go test -coverprofile=cover.out ./pkg/... ./cmd/...
+	go test ./virt-v2v/cold/...
+
+test1: test virt-v2v
```
liranr23, line 89: leftover?

Makefile
```diff
@@ -84,6 +84,9 @@ all: test forklift-controller
 # Run tests
 test: generate fmt vet manifests validation-test
 	go test -coverprofile=cover.out ./pkg/... ./cmd/...
+	go test ./virt-v2v/cold/...
```
liranr23, line 87: `go test -coverprofile=cover.out ./pkg/... ./cmd/... ./virt-v2v/...`
or with `cold` path if you wish to be more specific

bkhizgiy: @liranr23 added.

ahadas (approved):

PR #873, opened 2024-04-24
WIP: let the user choose the boot disk for multiboot

bkhizgiy:


bkhizgiy:
@bennyz the linking part is still missing, I'm not sure what would be the best way to do it....

bennyz:
> @bennyz the linking part is still missing, I'm not sure what would be the best way to do it....

As discussed shouldn't be an issue, we only care about passing the device representing the selected disk to `--root`

operator/config/crd/bases/forklift.konveyor.io_plans.yaml
```diff
@@ -730,6 +733,9 @@ spec:
                           description: 'The object ID. vsphere: The managed object
                             ID.'
                           type: string
+                        multiBootDiskNumber:
```
bennyz, line 736: do we need these?

I though we only need the one in the `VM` struct

pkg/apis/forklift/v1beta1/plan/vm.go
```diff
@@ -30,6 +30,8 @@ type VM struct {
 	ref.Ref `json:",inline"`
 	// Enable hooks.
 	Hooks []HookRef `json:"hooks,omitempty"`
+	// Choose the bootable disk number for praimery
+	MultiBootDiskNumber int `json:"multiBootDiskNumber,omitempty"`
```
bennyz, line 34: Should be just name I think, the number is an implementation detail for us. I think `RootDisk` or something would better represent this

pkg/controller/plan/kubevirt.go
```diff
@@ -1584,6 +1584,15 @@ func (r *KubeVirt) guestConversionPod(vm *plan.VMStatus, vmVolumes []cnv.Volume,
 	if err != nil {
 		return
 	}
+	if vm.VM.MultiBootDiskNumber != 0 {
```
bennyz, line 1587: I suppose this should change, we need to match the selected disk name to the number it will be assigned

operator/config/crd/bases/forklift.konveyor.io_plans.yaml
bkhizgiy: @bennyz probably right, it was automatically generated.

operator/config/crd/bases/forklift.konveyor.io_migrations.yaml
```diff
@@ -429,6 +429,9 @@ spec:
                     restorePowerState:
                       description: Source VM power state before migration.
                       type: string
+                    rootDisk:
+                      description: Choose the bootable disk number for praimery
```
ahadas, line 433: primary?
```suggestion
                      description: Choose the primary disk the VM boots from
```

pkg/apis/forklift/v1beta1/plan/vm.go
```diff
@@ -30,6 +30,8 @@ type VM struct {
 	ref.Ref `json:",inline"`
 	// Enable hooks.
 	Hooks []HookRef `json:"hooks,omitempty"`
+	// Choose the bootable disk number for praimery
```
ahadas, line 33: primary? we should also say that this field is optional and what happens when it's not specified
```suggestion
                      description: Choose the primary disk the VM boots from
```

pkg/apis/forklift/v1beta1/plan/vm.go
```diff
@@ -30,6 +30,8 @@ type VM struct {
...
+	RootDisk string `json:"rootDisk,omitempty"`
```
ahadas, line 34: that's an interesting decision to place this field at the vm level
on the one hand, we do the same for hooks but as it goes against out desire to have the UI designed for scale migrations, the UI allows users to specify a hook per plan and then sets it per each vm in the plan
on the other hand, [we considered this for encryption keys](https://github.com/kubev2v/forklift/pull/676) and decided to place it at the plan-level

we better be consistent here
I don't mind going with this approach of specifying the primary disk per-VM but then we should also do that for the encryption keys and implement this field at the UI level as hooks
@liranr23 WDYT?

bkhizgiy: @ahadas I do think its better to keep it in the VM level (for all cases) its giving more flexibility when setting up planes for case we want to have more that one VM like this in a plan, otherwise the user will need to have separate plans for each VM.

pkg/controller/plan/kubevirt.go
```diff
@@ -1584,6 +1585,29 @@ func (r *KubeVirt) guestConversionPod(vm *plan.VMStatus, vmVolumes []cnv.Volume,
...
+		vmInventory := &model.VM{}
+		err = r.Source.Inventory.Find(vmInventory, vm.Ref)
+		if err != nil {
+			err = liberr.Wrap(err, "vm", vm.Ref.String())
+			return
+		}
+
+		var rootDisk string
+		for diskNum, disk := range vmInventory.Disks {
+			if disk.File == vm.RootDisk {
```
bkhizgiy, line 1598: @liranr23 do you remember if the order of the disk here is consistent? I remember we talked about soothing similar when working on the OVA feature, but the question if we can rely on the disk order inside the VM or we should sort it beforehand?

pkg/apis/forklift/v1beta1/plan/vm.go
liranr23: we do have something different.
here, we need to provide per VM the boot device.
in LUKS, we also need the pass phrase per device, but since the device id is unique, we can put all the ids and their pass phrases to every virt-v2v command. it will just ignore the non-existing devices. therefore, we can ease on the process, and generalize it to the plan (make it more mass migration state of mind).

i do think the boot disk makes more sense to be per-vm, but i don't think we need to change the LUKs approach, especially we will have the `all` devices option from virt-v2v.

pkg/controller/plan/kubevirt.go
liranr23: i wouldn't count on it to be consistent. it depends on the inventory which depends on the way we get it from the source provider and that depends on the API itself. there are many places that may change.
however, we do use the VM CR in `podVolumeMounts` function above, there it should be consistent, if you find it helpful.

bkhizgiy: @liranr23 cool thanks, will look into that.

pkg/controller/plan/kubevirt.go
```diff
@@ -1584,6 +1585,29 @@ func (r *KubeVirt) guestConversionPod(vm *plan.VMStatus, vmVolumes []cnv.Volume,
 	if err != nil {
 		return
 	}
+	if vm.RootDisk != "" {
+		vmInventory := &model.VM{}
```
liranr23, line 1589: since we might query the inventory also in other functions (like the `PodEnvironment`), maybe we can query it once at the beginning and send it to the function instead querying it multiple times?

pkg/controller/plan/kubevirt.go
```diff
@@ -1584,6 +1585,29 @@ func (r *KubeVirt) guestConversionPod(vm *plan.VMStatus, vmVolumes []cnv.Volume,
...
+		err = r.Source.Inventory.Find(vmInventory, vm.Ref)
+		if err != nil {
+			err = liberr.Wrap(err, "vm", vm.Ref.String())
+			return
+		}
+
+		var rootDisk string
+		for diskNum, disk := range vmInventory.Disks {
+			if disk.File == vm.RootDisk {
+				rootDisk = fmt.Sprintf("/dev/sd%c", 'a'+(diskNum)%26)
```
liranr23, line 1599: consts?
also, i saw germano's comment on the other PR (https://github.com/kubev2v/forklift/pull/872) - he is right regarding the need to have more than 26 disks.
in that case maybe we can put the path generation into a single place and re-use it in both?

liranr23 (commented):
any reason it's PR straight to `release-v2.6.0` branch and not master (in any case it suppose to be 2.6.1 for backport)?

pkg/apis/forklift/v1beta1/plan/vm.go
ahadas: @liranr23 ok, so I commented on the encryption keys part in #676 - we can continue that there
so if we're in agreement that it's better to place the primary disk configurations like hooks, let's go with that then

pkg/controller/plan/kubevirt.go
bkhizgiy: yes it's not an offical PR, just a draft we worked on last week, wanted to test it on the the customer version and share it with Benny, but will open this on the main once its ready.

pkg/controller/plan/kubevirt.go
yaacov: > i wouldn't count on it to be consistent.

if the disk order is not constant, a vm may have some disk as 'sda' on one boot, and then it will be 'sdc' on a second boot ? can this work, what am i missing ?

liranr23: @yaacov do you mean on the vm itself? if the vm itself is already posted nothing is gonna change there.

yaacov: > do you mean on the vm itself?

Thanks, I talked with @bkhizgiy and the conclusion was that the UI will set a number "1", "2" and mtv will do the logic to make sure "1" will map to the first disk "2" to the second, etc.

PR #884, opened 2024-05-05
ova: fix the version of encoding/gob

ahadas:
Unfortunately, we rely on 'encoding/gob' to determine the UUID of VMs from OVAs and this package has changed in Go 1.21 in a way we get a different UUID compared to the one we get with Go 1.20 for the same input, which could break existing OVA providers.

Therefore, this package integrates the relevant pieces from 'encoding/gob' of Go 1.20 so that we can bump Go to 1.21 without breaking compatibility with OVA providers that were defined in lower versions of Forklift.


bkhizgiy:
@ahadas So, does that mean we will now need to manually change this package with each Go version update? Or is this just a workaround until we have something more permanent, like adding hashing to the OVF file itself, as we discussed?

ahadas:
@liranr23 it's less likely that we'll see security fixes in a package for encoding but generally, we won't get improvements and fixes by rebasing for this code

@bkhizgiy the code that was added in this PR was taken from Go 1.20 - we don't need to change it until we replace this with a better solution. we should introduce a different mechanism for the hash, that's for sure. we can introduce it in a y-version and say that we break backward compatibility or a better approach would be to apply the new mechanism for newly created OVA providers, and keep the existing mechanism for existing OVA providers

PR #889, opened 2024-05-09
Add PendingPopulation DataVolume status

liranr23:
With the new CDI, the usage of volume populator added a new status to the DataVolume: `PendingPopulation`. In case we use WaitForFirstConsumer StorageClass, we need act as the PVC is ready to use, which in these flows the conversion pod will be the consumer.


vendor/modules.txt
```diff
@@ -1044,8 +1044,8 @@ kubevirt.io/api/export
 kubevirt.io/api/export/v1alpha1
 kubevirt.io/api/instancetype
 kubevirt.io/api/instancetype/v1beta1
-# kubevirt.io/containerized-data-importer-api v1.57.0-alpha1
-## explicit; go 1.19
+# kubevirt.io/containerized-data-importer-api v1.59.0
+## explicit; go 1.21
```
ahadas, line 1048: this means that we have to upgrade to RHEL 9.4 when releasing this fix, right?

pkg/controller/plan/migration.go
```diff
@@ -1501,7 +1501,7 @@ func (r *Migration) updateCopyProgress(vm *plan.VMStatus, step *plan.Step) (err
 			}
 			conditions := dv.Conditions()
 			switch dv.Status.Phase {
-			case cdi.Succeeded, cdi.Paused:
+			case cdi.Succeeded, cdi.Paused, cdi.PendingPopulation:
```
ahadas, line 1504: this should hold only when migrating from vSphere, is it really limited to vSphere this way?

vendor/modules.txt
liranr23: we already did upgrade to 9.4. what is the concern?

ahadas: ok, so if we create a new branch for 2.6.2, we need to keep in mind that we should change the base image to RHEL 9.4 there as well then

pkg/controller/plan/migration.go
```diff
@@ -1502,22 +1520,16 @@ func (r *Migration) updateCopyProgress(vm *plan.VMStatus, step *plan.Step) (err
...
-				cnd := conditions.FindCondition("Bound")
-				if cnd != nil && cnd.Status == True {
-					cnd = conditions.FindCondition("Running")
+				r.setTaskCompleted(task, completed)
+			case cdi.PendingPopulation:
+				switch r.Source.Provider.Type() {
+				case v1beta1.VSphere:
+					r.setTaskCompleted(task, completed)
+				default:
+					pendingReason = r.setTaskPending(task, pending, conditions)
```
ahadas, line 1529: wouldn't it be simpler to replace PendingPopulation with Succeeded in case the source provider is of type vSphere before getting here? something like:
```golang
if dv.Status.Phase == cdi.PendingPopulation && r.Source.Provider.Type() == v1beta1.VSphere {
  dv.Status.Phase = cdi.Succeeded
}
```

ahadas (approved):

PR #890, opened 2024-05-09
v2v: add extra args

ahadas:
Add an argument for extending the virt-v2v command with additional arguments that are taken from ForkliftController. This allows to modify the virt-v2v command, without creating new images.

### DIFF
--- a/pkg/controller/plan/adapter/vsphere/builder.go
+++ b/pkg/controller/plan/adapter/vsphere/builder.go
@@ -24,6 +24,7 @@ import (
 	libcnd "github.com/konveyor/forklift-controller/pkg/lib/condition"
 	liberr "github.com/konveyor/forklift-controller/pkg/lib/error"
 	libitr "github.com/konveyor/forklift-controller/pkg/lib/itinerary"
+	"github.com/konveyor/forklift-controller/pkg/settings"
 	"github.com/vmware/govmomi/vim25"
 	"github.com/vmware/govmomi/vim25/types"
 	core "k8s.io/api/core/v1"
@@ -207,6 +208,10 @@ func (r *Builder) PodEnvironment(vmRef ref.Ref, sourceSecret *core.Secret) (env 
 		core.EnvVar{
 			Name:  "V2V_fingerprint",
 			Value: fingerprint,
+		},
+		core.EnvVar{
+			Name:  "V2V_extra_args",
+			Value: settings.Settings.Migration.VirtV2vExtraArgs,
 		},
 	)
 	return
--- a/pkg/settings/migration.go
+++ b/pkg/settings/migration.go
@@ -1,6 +1,7 @@
 package settings
 
 import (
+	"encoding/json"
 	"fmt"
 	"os"
 	"strings"
@@ -26,6 +27,7 @@ const (
 	OvirtOsConfigMap        = "OVIRT_OS_MAP"
 	VsphereOsConfigMap      = "VSPHERE_OS_MAP"
 	VddkJobActiveDeadline   = "VDDK_JOB_ACTIVE_DEADLINE"
+	VirtV2vExtraArgs        = "VIRT_V2V_EXTRA_ARGS"
 )
 
 // Migration settings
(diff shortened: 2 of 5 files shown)


virt-v2v/cold/entrypoint.go
```diff
@@ -97,6 +97,10 @@ func main() {
 				"-io", fmt.Sprintf("vddk-thumbprint=%s", os.Getenv("V2V_fingerprint")),
 			)
 		}
+		if extraArgs := os.Getenv("V2V_extra_args"); len(extraArgs) > 0 {
+			virtV2vArgs = append(virtV2vArgs, extraArgs)
```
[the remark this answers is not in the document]
liranr23: i agree with benny.
i think we should have some decision on a delimiter. then we can send a long string containing many arguments and split them.

virt-v2v/cold/entrypoint.go
```diff
@@ -97,6 +97,10 @@ func main() {
 				"-io", fmt.Sprintf("vddk-thumbprint=%s", os.Getenv("V2V_fingerprint")),
 			)
 		}
+		if extraArgs := os.Getenv("V2V_extra_args"); len(extraArgs) > 0 {
```
liranr23, line 100: `checkEnvVariablesSet`

virt-v2v/cold/entrypoint.go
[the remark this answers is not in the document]
ahadas: we currently need a single argument but the idea behind naming it "extra_args" was to enable specifying multiple arguments, I'll check if we have to split it or maybe with go 1.21 we can pass a string with whitespaces

pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -208,6 +208,10 @@ func (r *Builder) PodEnvironment(vmRef ref.Ref, sourceSecret *core.Secret) (env
 			Name:  "V2V_fingerprint",
 			Value: fingerprint,
 		},
+		core.EnvVar{
+			Name:  "V2V_extra_args",
+			Value: settings.Settings.Migration.Virt,
+		},
```
liranr23, line 214: you can take https://github.com/kubev2v/forklift/pull/852/files#diff-c14563539716a81d759e3204377a4a3f07328ee75d3fb4d9bb5bdc74f7791538R218 as a reference, i think this is why it fails.

virt-v2v/cold/entrypoint.go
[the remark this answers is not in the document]
ahadas: done, the arguments are now split

virt-v2v/cold/entrypoint.go
ahadas: > `checkEnvVariablesSet`

that function returns a boolean while I need the value of the environment variable

pkg/controller/plan/adapter/vsphere/builder.go
ahadas: yeah, that was problematic as `settings.Settings.Migration.Virt` doesn't exist - changed

pkg/settings/migration.go
```diff
@@ -126,6 +130,14 @@ func (r *Migration) Load() (err error) {
 	if r.VddkJobActiveDeadline, err = getPositiveEnvLimit(VddkJobActiveDeadline, 300); err != nil {
 		return liberr.Wrap(err)
 	}
+	r.VirtV2vExtraArgs = "[]"
+	if val, found := os.LookupEnv(VirtV2vExtraArgs); found && len(val) > 0 {
```
liranr23, line 134: i would prefer to check if the env var exists, if it does, i wouldn't check the length of it
in case we don't have this environmental var, i wouldn't send it to the conversion pod and within the conversion pod if it doesn't exist - do nothing.

in my opinion it will save troubleshooting and the extra marshal/unmarshal. but it also a matter of taste

liranr23 (approved):

bkhizgiy (approved):

pkg/settings/migration.go
ahadas: > i would prefer to check if the env var exists, if it does, i wouldn't check the length of it

the reason for checking the length there is also not to fail when the user sets the extra arguments to an empty string

PR #892, opened 2024-05-15
CDI integration

bennyz:
With https://github.com/kubevirt/containerized-data-importer/pull/3261 being merged, adjust forklift to work with the external controller, this includes, but not limited to:
- Adjust the secret name
- Scale down our populator controller


yaacov:
@mnecas hi,

This PR is about supporting the move to use CDI populators while supporting OCP pre 4.16, so both APIs are supported,
`forklift.cdi.konveyor.io` and `forklift.konveyor.io`.

AFAIU:
a. We are currently moving to use forklift peculators, so no need to move to CDI based populators 
b. We wil not supporting OCP pre 4.16

Does it make sense to close this PR ?

mnecas:
It's not needed but I would like to move to the CDI at some point aswell, but not right now as this would require lot of testing and additional effort. Closing for now

PR #893, opened 2024-05-15
Sanitize forklift-api input

liranr23:


pkg/forklift-api/services/tls-certificate.go
```diff
@@ -21,6 +22,7 @@ func serveTlsCertificate(resp http.ResponseWriter, req *http.Request, client cli
 				Type:  "CERTIFICATE",
 				Bytes: cacert.Raw,
 			})
+			encoded = bluemonday.UGCPolicy().SanitizeBytes(encoded)
```
ahadas, line 25: can't we sanitize the certificate using x509.ParseCertificate instead?

liranr23: but using `x509.ParseCertificate` will take e.g the `encoded` and generate a new cert from it, where i need to re-encode?

ahadas: right, I didn't mean to parse the encoded block but cacert.Raw - IIUC the risk is in getting malicious code there, no?

pkg/forklift-api/services/tls-certificate.go
```diff
@@ -17,6 +18,11 @@ func serveTlsCertificate(resp http.ResponseWriter, req *http.Request, client cli
 			Data: map[string][]byte{"insecureSkipVerify": []byte("true")},
 		}
 		if cacert, err := util.GetTlsCertificate(url, secret); err == nil {
+			cacert, err := x509.ParseCertificate(cacert.Raw)
+			if err != nil {
+				msg := fmt.Sprintf("failed to parse certificate: %v", err)
+				http.Error(resp, msg, http.StatusInternalServerError)
```
ahadas, line 24: and we should break here without proceeding to `resp.Write` below
but as we anyway need to change this PR, let me move it to the utl.GetTlsCertificate function, it will simplify the code a bit

ahadas (approved):

PR #897, opened 2024-05-16
Fix direct LUN migrations

liranr23:
Since
https://github.com/kubev2v/forklift/commit/0e9edd7f45ebe54aac80fa35278e359d144da4e0, the way we get PVCs associated with VMs has changed. This patch adds the relevant labels to the LUN PVCs, so we can get them correctly.

Also, there is a difference of versioning between ovirt to rhv. Now we can use them both in the version validation.

### DIFF
--- a/pkg/controller/provider/container/ovirt/collector.go
+++ b/pkg/controller/provider/container/ovirt/collector.go
@@ -146,15 +146,16 @@ func (r *Collector) Version() (major, minor, build, revision string, err error) 
 	version := strings.Split(system.Product.Version.FullVersion, ".")
 	major = version[0]
 	minor = version[1]
-	build = version[2]
-	revision = "0"
-	if strings.Contains(build, "-") {
-		build = strings.Split(version[2], "-")[0]
-		revision = strings.Split(version[2], "-")[1]
-	} else {
-		if len(version) > 3 {
-			revision = strings.Split(version[3], "-")[0]
-		}
+
+	split := strings.SplitN(version[2], "-", 2)
+	build = split[0]
+	switch {
+	case len(split) > 1:
+		revision = split[1]
+	case len(version) > 3:
+		revision = strings.Split(version[3], "-")[0]
+	default:
+		revision = "0"
 	}
 
 	return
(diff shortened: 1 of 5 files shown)


pkg/controller/provider/container/ovirt/collector.go
```diff
@@ -148,9 +148,15 @@ func (r *Collector) Version() (major, minor, build, revision string, err error)
...
-		revision = strings.Split(version[3], "-")[0]
+	if strings.Contains(build, "-") {
+		build = strings.Split(version[2], "-")[0]
+		revision = strings.Split(version[2], "-")[1]
+	} else {
+		if len(version) > 3 {
+			revision = strings.Split(version[3], "-")[0]
+		}
 	}
+
```
ahadas, line 159: could you please add unit tests for this? it will also serve as documentation which was missing for this part in the commit message

pkg/controller/provider/container/ovirt/collector_suite_test.go
```diff
@@ -0,0 +1,22 @@
...
+	if len(callerSkip) > 0 {
+		callerSkip[0]++
+	}
+	ginkgo.Fail(message, callerSkip...)
+}
+
+func TestTests(t *testing.T) {
+	defer ginkgo.GinkgoRecover()
+	RegisterFailHandler(forkliftFailHandler)
+	ginkgo.RunSpecs(t, "ovirt controller")
```
ahadas, line 21:
```suggestion
	ginkgo.RunSpecs(t, "ovirt collector")
```

pkg/controller/provider/container/ovirt/collector.go
```diff
@@ -143,13 +143,24 @@ func (r *Collector) Version() (major, minor, build, revision string, err error)
 	if err != nil {
 		return
 	}
-	version := strings.Split(system.Product.Version.FullVersion, ".")
+	major, minor, build, revision = r.parseVersion(system.Product.Version.FullVersion)
+	return
+}
+
+func (r *Collector) parseVersion(fullVersion string) (major, minor, build, revision string) {
```
ahadas, line 150: it's kind of a utility function and as such we can separate it out from the Collector struct, which will simplify the tests a little bit
```suggestion
func parseVersion(fullVersion string) (major, minor, build, revision string) {
```

ahadas (approved):

PR #901, opened 2024-05-19
Fix virt-v2v progress after using go wrapper

bkhizgiy:
After switching to use a Go wrapper, not all the information was redirected to the virt-v2v monitor command. Additionally, with changes done in the virt-v2v progress display, the progress was not reported to the controller. This patch fixes the piping to the virt-v2v monitoring command and the regex used to determine the copy disk progress of the migration.

PR #903, opened 2024-05-20
Fix virt-v2v progress update

bkhizgiy:
After switching to use a Go wrapper, not all the information was redirected to the virt-v2v monitor command. Additionally, with changes done in the virt-v2v progress display, the progress was not reported to the controller. This patch fixes the piping to the virt-v2v monitoring command and the regex used to determine the copy disk progress of the migration.
https://github.com/kubev2v/forklift/pull/901

### DIFF
--- a/cmd/virt-v2v-monitor/virt-v2v-monitor.go
+++ b/cmd/virt-v2v-monitor/virt-v2v-monitor.go
@@ -2,7 +2,7 @@ package main
 
 import (
 	"bufio"
-	"flag"
+	"fmt"
 	"net/http"
 	"os"
 	"regexp"
@@ -11,11 +11,10 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	dto "github.com/prometheus/client_model/go"
-	"k8s.io/klog/v2"
 )
 
 var COPY_DISK_RE = regexp.MustCompile(`^.*Copying disk (\d+)/(\d+)`)
-var DISK_PROGRESS_RE = regexp.MustCompile(`^..\s*(\d+)% \[.*\]`)
+var DISK_PROGRESS_RE = regexp.MustCompile(`.+ (\d+)% \[[*-]+\]`)
 var FINISHED_RE = regexp.MustCompile(`^\[[ .0-9]*\] Finishing off`)
 
 // Here is a scan function that imposes limit on returned line length. virt-v2v
@@ -51,19 +50,15 @@ func updateProgress(progressCounter *prometheus.CounterVec, disk, progress uint6
 
 	change := float64(progress) - previous_progress
 	if change > 0 {
-		klog.Infof("Progress changed for disk %d about %v", disk, change)
+		fmt.Printf("virt-v2v monitoring: Progress changed for disk %d about %v\n", disk, change)
 		progressCounter.WithLabelValues(label).Add(change)
 	}
 	return
 }
 
 func main() {
-	klog.InitFlags(nil)
-	defer klog.Flush()
-	flag.Parse()
-
 	// Start prometheus metrics HTTP handler
-	klog.Info("Setting up prometheus endpoint :2112/metrics")
+	fmt.Println("virt-v2v monitoring: Setting up prometheus endpoint :2112/metrics")
 	http.Handle("/metrics", promhttp.Handler())
 	go http.ListenAndServe(":2112", nil)
 
@@ -78,10 +73,10 @@ func main() {
 	if err := prometheus.Register(progressCounter); err != nil {
 		// Exit gracefully if we fail here. We don't need monitoring
 		// failures to hinder guest conversion.
-		klog.Error("Prometheus progress counter not registered:", err)
+		fmt.Println("virt-v2v monitoring: Prometheus progress counter not registered:", err)
 		return
 	}
-	klog.Info("Prometheus progress counter registered.")
+	fmt.Println("virt-v2v monitoring: Prometheus progress counter registered.")
 
 	var diskNumber uint64 = 0
 	var disks uint64 = 0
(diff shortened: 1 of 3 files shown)


cmd/virt-v2v-monitor/virt-v2v-monitor.go
```diff
@@ -11,11 +11,10 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	dto "github.com/prometheus/client_model/go"
-	"k8s.io/klog/v2"
 )
 
 var COPY_DISK_RE = regexp.MustCompile(`^.*Copying disk (\d+)/(\d+)`)
-var DISK_PROGRESS_RE = regexp.MustCompile(`^..\s*(\d+)% \[.*\]`)
+var DISK_PROGRESS_RE = regexp.MustCompile(`.+ (\d+)% \[[*-]+\]`)
```
liranr23, line 17: are we sure we need to change the regex? i looked on this regex and the progress given in the log and it looked correct.

liranr23 (commented):
maybe some more info in the commit (and even the PR) should be nicer.

something like:
After refactoring the bash script into golang we missed using stderr as stdin to the v2v-monitor. Because of that, we didn't have the data progress reported.....

cmd/virt-v2v-monitor/virt-v2v-monitor.go
ahadas: @liranr23 this change was inspired from https://github.com/oVirt/vdsm/pull/402
the problem with the current regex is that it catches lines that are unrelated to the progress

liranr23:
please run `make bazel-generate`

cmd/virt-v2v-monitor/virt-v2v-monitor.go
```diff
@@ -95,37 +90,37 @@ func main() {
 		os.Stdout.Write([]byte("\n"))
 		err := scanner.Err()
 		if err != nil {
-			klog.Fatal("Output monitoring failed! ", err)
+			fmt.Println("virt-v2v monitoring: Output monitoring failed! ", err)
```
ahadas, line 93: klog.Fatal also calls OsExit which we lose with this change

cmd/virt-v2v-monitor/virt-v2v-monitor.go
```diff
@@ -95,37 +90,37 @@ func main() {
...
 			// Make sure we flag conversion as finished. This is
 			// just in case we miss the last progress update for some reason.
-			klog.Infof("Finished")
+			fmt.Println("virt-v2v monitoring: Finished")
 			for disk := uint64(0); disk < disks; disk++ {
 				err = updateProgress(progressCounter, disk, 100)
 			}
 		} else {
-			klog.V(1).Info("Ignoring line: ", string(line))
+			fmt.Println("virt-v2v monitoring: Ignoring line: ", string(line))
```
ahadas, line 114: need to drop/change this - it creates too much noise

cmd/virt-v2v-monitor/virt-v2v-monitor.go
```diff
@@ -95,37 +90,37 @@ func main() {
...
 			// Don't make processing errors fatal.
-			klog.Error("Error updating progress: ", err)
+			fmt.Println("virt-v2v monitoring: Error updating progress: ", err)
 			err = nil
 		}
 	}
 	err := scanner.Err()
 	if err != nil {
-		klog.Fatal("Output monitoring failed! ", err)
+		fmt.Println("virt-v2v monitoring: Output monitoring failed! ", err)
```
ahadas, line 124: same comment as above for replacing klog.Fatal with a simple print

cmd/virt-v2v-monitor/virt-v2v-monitor.go
bkhizgiy: added os.Exit in addition.

cmd/virt-v2v-monitor/virt-v2v-monitor.go
bkhizgiy: same.

bkhizgiy:
@liranr23 ACK, I've updated the PR description and commit with more info.

liranr23:
> After switching to use a Go wrapper, not all the information was redirected to the virt-v2v monitor command. Additionally, with changes done in the virt-v2v progress display, the progress was not reported to the controller. This patch fixes the piping to the virt-v2v monitoring command and the regex used to determine the copy disk progress of the migration. #901

great! but you changed it only on the PR and not on the commit?

bkhizgiy:
> > After switching to use a Go wrapper, not all the information was redirected to the virt-v2v monitor command. Additionally, with changes done in the virt-v2v progress display, the progress was not reported to the controller. This patch fixes the piping to the virt-v2v monitoring command and the regex used to determine the copy disk progress of the migration. #901
> 
> great! but you changed it only on the PR and not on the commit?

why, I've added `after switching to go wrapper` :) , I'm usually not adding to mush info into the commits, add some more details?

liranr23:
> > > After switching to use a Go wrapper, not all the information was redirected to the virt-v2v monitor command. Additionally, with changes done in the virt-v2v progress display, the progress was not reported to the controller. This patch fixes the piping to the virt-v2v monitoring command and the regex used to determine the copy disk progress of the migration. #901
> > 
> > 
> > great! but you changed it only on the PR and not on the commit?
> 
> why, I've added `after switching to go wrapper` :) , I'm usually not adding to mush info into the commits, add some more details?

:)
That's OK. If you add it into the commit as a paragraph, it will allow others to look into git log and see it. When you don't do that, it will require to open the PR itself.

bkhizgiy:
> > > > After switching to use a Go wrapper, not all the information was redirected to the virt-v2v monitor command. Additionally, with changes done in the virt-v2v progress display, the progress was not reported to the controller. This patch fixes the piping to the virt-v2v monitoring command and the regex used to determine the copy disk progress of the migration. #901
> > > 
> > > 
> > > great! but you changed it only on the PR and not on the commit?
> > 
> > 
> > why, I've added `after switching to go wrapper` :) , I'm usually not adding to mush info into the commits, add some more details?
> 
> :) That's OK. If you add it into the commit as a paragraph, it will allow others to look into git log and see it. When you don't do that, it will require to open the PR itself.

Added to the commit as well:)

liranr23 (approved):

virt-v2v/cold/entrypoint.go
```diff
@@ -213,9 +208,8 @@ func executeVirtV2v(args []string) (err error) {
 	}
 
 	virtV2vMonitorCmd := exec.Command("/usr/local/bin/virt-v2v-monitor")
-	virtV2vMonitorCmd.Stdin = tee
+	virtV2vMonitorCmd.Stdin = r
 	virtV2vMonitorCmd.Stdout = os.Stdout
-	virtV2vMonitorCmd.Stderr = os.Stderr
```
ahadas, line 218: was it done intentionally?

ahadas (approved):

PR #904, opened 2024-05-21
Refactor v2v's entrypoint.go

ahadas:


virt-v2v/cold/entrypoint.go
```diff
@@ -195,33 +195,38 @@ func LinkDisks(diskKind string, num int) (err error) {
...
+	if err := v2vCmd.Run(); err != nil {
+		fmt.Printf("Error executing v2v command: %v\n", err)
+		return err
 	}
 
-	if err = virtV2vCmd.Wait(); err != nil {
-		fmt.Printf("Error waiting for virt-v2v to finish: %v\n", err)
-		return
+	// virt-v2v is done, we can close the pipe to virt-v2v-monitor
+	writer.Close()
```
liranr23, line 222: used in defer above

virt-v2v/cold/entrypoint.go
```diff
@@ -195,33 +195,38 @@ func LinkDisks(diskKind string, num int) (err error) {
...
+	monitorCmd.Stderr = os.Stderr
+
+	var writer *io.PipeWriter
+	monitorCmd.Stdin, writer = io.Pipe()
+	v2vCmd.Stdout = writer
+	v2vCmd.Stderr = writer
+	defer writer.Close()
+
+	if err := monitorCmd.Start(); err != nil {
+		fmt.Printf("Error executing monitor command: %v\n", err)
```
bennyz, line 211: We can use `"log/slog"`

virt-v2v/cold/entrypoint.go
```diff
@@ -195,33 +195,38 @@ func LinkDisks(diskKind string, num int) (err error) {
...
+		return err
 	}
 
-	if err = virtV2vCmd.Wait(); err != nil {
-		fmt.Printf("Error waiting for virt-v2v to finish: %v\n", err)
-		return
+	// virt-v2v is done, we can close the pipe to virt-v2v-monitor
+	writer.Close()
+
+	if err := monitorCmd.Wait(); err != nil {
```
bkhizgiy, line 224: Do we really care if the monitor finished? once virt-v2v done it should be enough of indication the conversion is done successfully, and we also have the mechanism in the controller that know its done even if some of the metrics were missing or wasn't properly sent. I just not sure if it worth to fail or block the migration In case the monitor command facing some issues.

virt-v2v/cold/entrypoint.go
ahadas: right but code that is being deferred, runs when we return from the function and we need to close the writer before that for the monitor command to finish

virt-v2v/cold/entrypoint.go
ahadas: we can call Wait on a best-effort basis but we need to call it in order to release resources

virt-v2v/cold/entrypoint.go
ahadas: @bennyz do you mind posting this change is separately?

ahadas: resolving, this can be posted separately

virt-v2v/cold/entrypoint.go
ahadas: > we can call Wait on a best-effort basis but we need to call it in order to release resources

I don't remember the details, but we had a case in which the monitoring command didn't finish successfully and it really was an indication of an issue - and as this code exists in the release-v2.6.2 branch and works well, I don't see a reason to ignore the error (on the contrary, if we would start getting an error here, we should rather look into it to see the root cause)

PR #906, opened 2024-05-21
OVA: Change connection refuse check for virt-v2v vm configuration server

bkhizgiy:
Change connection refuse check for virt-v2v vm configuration server, and change the order of the functions so it will fully complete the progress and only then fetch the vm configuration.

backport https://github.com/kubev2v/forklift/pull/907


ahadas (commented):
ok, let's align this with the changes in #907

PR #919, opened 2024-06-04
Always take fingerprint from ESXi provider

ahadas:
When migrating from an ESXi provider, always take the fingerprint from the status of the provider, even when there is a transfer network configured for the host since the SDK of ESXi doesn't return the fingerprint of the host as vCenter does.

### DIFF
--- a/pkg/controller/plan/adapter/vsphere/builder.go
+++ b/pkg/controller/plan/adapter/vsphere/builder.go
@@ -270,7 +270,13 @@ func (r *Builder) getSourceDetails(vm *model.VM, sourceSecret *core.Secret) (lib
 			Path:     "",
 			RawQuery: sslVerify,
 		}
-		fingerprint = host.Thumbprint
+		if r.Source.Provider.Spec.Settings[api.SDK] == api.ESXI {
+			// the SDK of ESXi doesn't return a fingerprint/thumbprint for the host
+			// so we take it from the provider instead
+			fingerprint = r.Source.Provider.Status.Fingerprint
+		} else {
+			fingerprint = host.Thumbprint
+		}
 	} else if r.Source.Provider.Spec.Settings[api.SDK] == api.ESXI {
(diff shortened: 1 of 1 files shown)


pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -270,7 +270,11 @@ func (r *Builder) getSourceDetails(vm *model.VM, sourceSecret *core.Secret) (lib
 			Path:     "",
 			RawQuery: sslVerify,
 		}
-		fingerprint = host.Thumbprint
+		if r.Source.Provider.Spec.Settings[api.SDK] == api.ESXI {
```
liranr23, line 273: we have this condition below.
what's the difference between `hostDef, found := r.hosts[host.ID]; found` to `r.Source.Provider.Spec.Settings[api.SDK] == api.ESXI`?
if here we can get the same condition?
also we use `fingerprint = r.Source.Provider.Status.Fingerprint` in every if case, the only different case is when `r.Source.Provider.Spec.Settings[api.SDK] != api.ESXI`. maybe we can set it in a single place and then only change it in this case?

ahadas: > what's the difference between `hostDef, found := r.hosts[host.ID]; found` to `r.Source.Provider.Spec.Settings[api.SDK] == api.ESXI`? if here we can get the same condition?

see the `Host` part in the URL - when there's an ESXi network, we set it to `hostDef.Spec.IpAddress` and when there's no ESXi network, we set it to `host.Name`, and the source of the credentials also changes

> also we use `fingerprint = r.Source.Provider.Status.Fingerprint` in every if case, the only different case is when `r.Source.Provider.Spec.Settings[api.SDK] != api.ESXI`. maybe we can set it in a single place and then only change it in this case?

that makes less sense - it would make the code less expressive

liranr23 (approved):

PR #926, opened 2024-06-09
vsphere: virt-v2v transfers disks sequentially

ahadas:
when scheduling migrations to accommodate the controller_max_vm_inflight setting for vSphere, we didn't take into account that disks are transferred sequentially by virt-v2v (as opposed to CDI that transfers them in parallel by different pods). This lead to performance degregeration since could have triggered more migrations, in case of migrations of multi-disk VMs, simultaneously without exceeding the value of controller_max_vm_inflight. This is fixed by setting the cost of each VM for which the disks are transferred by virt-v2v to 1.

### DIFF
--- a/pkg/controller/plan/scheduler/vsphere/scheduler.go
+++ b/pkg/controller/plan/scheduler/vsphere/scheduler.go
@@ -107,7 +107,7 @@ func (r *Scheduler) buildInFlight() (err error) {
 			return
 		}
 		if vmStatus.Running() {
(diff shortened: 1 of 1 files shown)


liranr23 (approved):

bkhizgiy (approved):

liranr23 (commented):
I think you should also consider it in `buildInFlight()`

ahadas:
> I think you should also consider it in `buildInFlight()`

correct, changed

liranr23 (approved):

PR #929, opened 2024-06-13
Use virt-v2v XML's Operation System output on vSphere migrations

liranr23:
When migrating from vSphere,  the VM configuration that is generated by virt-v2v will be used to determine the guest operation system. It will be used when building and creating the VM.


pkg/controller/plan/kubevirt.go
```diff
@@ -925,12 +925,21 @@ func (r *KubeVirt) UpdateVmByConvertedConfig(vm *plan.VMStatus, pod *core.Pod, s
...
+	case api.Ova:
+		firmware, erro := ovfparser.GetFirmwareFromConfig(string(vmConfigXML))
+		if erro != nil {
+			return
+		}
 
-	vm.Firmware = firmware
+		vm.Firmware = firmware
+	case api.VSphere:
+		os, erro := ovfparser.GetOperationSystemFromConfig(string(vmConfigXML))
```
ahadas, line 937: how about mapping the value of the operating system here instead of adding another step (`MapVirtV2vOs`)?

pkg/controller/plan/migration.go
```diff
@@ -1125,7 +1125,8 @@ func (r *Migration) execute(vm *plan.VMStatus) (err error) {
 		}
 
 		// only for OVA, fetch config from the conversion pod
-		if r.Source.Provider.Type() == v1beta1.Ova {
+		source := r.Source.Provider.Type()
+		if source == v1beta1.Ova || source == v1beta1.VSphere {
```
ahadas, line 1129: please update the comment above
and how about using switch-case? it would be a bit more elegant in my opinion

ahadas (commented):
besides the comments inside, I also changed the description of the PR - please check it out

pkg/controller/plan/adapter/ova/ovfparser.go
```diff
@@ -6,9 +6,10 @@ import (
 )
 
 type OvaVmconfig struct {
-	XMLName xml.Name `xml:"domain"`
-	Name    string   `xml:"name"`
-	OS      OS       `xml:"os"`
+	XMLName  xml.Name `xml:"domain"`
+	Name     string   `xml:"name"`
+	OS       OS       `xml:"os"`
+	Metadata Metadata `xml:"metadata"`
```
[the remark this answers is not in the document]
bkhizgiy: IIRC we talked briefly about adding this to OVA as well,  so it make sense to update it for both providers.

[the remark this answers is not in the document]
ahadas: it would make sense to do that also for OVA, but the description of the PR talks specifically about vSphere ("When migrating from vSphere...") and we need to minimize the changes in 2.6 - I'm ok with doing the changes for OVA in a separate PR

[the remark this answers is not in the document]
liranr23: yes, but the parser sits here, unless i wish to pull it out entirely and make a bigger change - i used it where it is now.

pkg/controller/plan/kubevirt.go
liranr23: indeed it looks like i forgot to call the builder here :)

pkg/controller/plan/adapter/ova/ovfparser.go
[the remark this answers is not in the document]
ahadas: ah I see, it's placed incorrectly in the `ova` package even though it's now used also for vsphere providers
it needs to move elsewhere but let's do that separately

pkg/controller/plan/kubevirt.go
ahadas: > but please look again how it's done now. i think you meant we should put the whole map here? IMO currently it's how it suppose to be done

no, I didn't mean to the location of the code but to the timing of mapping the value given from virt-v2v - currently we get the value of `os` and then call the builder to map it but this requires adding it to the interface of the Builder that affects all providers. I'd prefer to get the value already converted from the VM status

pkg/controller/plan/migration.go
```diff
@@ -1124,8 +1124,10 @@ func (r *Migration) execute(vm *plan.VMStatus) (err error) {
 			return
 		}
 
-		// only for OVA, fetch config from the conversion pod
-		if r.Source.Provider.Type() == v1beta1.Ova {
+		// fetch config from the conversion pod
+		switch r.Source.Provider.Type() {
+		case v1beta1.Ova:
+		case v1beta1.VSphere:
```
ahadas, line 1130:
```suggestion
		case v1beta1.Ova, v1beta1.VSphere:
```

pkg/controller/plan/kubevirt.go
liranr23: making it in the interface does make sense to me (OVA later on?). do you mean to have it within the parser?

pkg/apis/forklift/v1beta1/plan/vm.go
```diff
@@ -64,6 +64,8 @@ type VMStatus struct {
 	RestorePowerState VMPowerState `json:"restorePowerState,omitempty"`
 	// The firmware type detected from the OVF file produced by virt-v2v.
 	Firmware string `json:"firmware,omitempty"`
+	// The Operation System detected by virt-v2v.
+	OperationSystem string `json:"operationSystem,omitempty"`
```
ahadas, line 68:
```suggestion
	OperatingSystem string `json:"operationSystem,omitempty"`
```

pkg/controller/plan/kubevirt.go
ahadas: yes, currently from the parser and later on from a utility function that returns the operating system from the vm configuration

pkg/apis/forklift/v1beta1/plan/vm.go
```diff
@@ -64,6 +64,8 @@ type VMStatus struct {
 	RestorePowerState VMPowerState `json:"restorePowerState,omitempty"`
 	// The firmware type detected from the OVF file produced by virt-v2v.
 	Firmware string `json:"firmware,omitempty"`
+	// The Operation System detected by virt-v2v.
```
ahadas, line 67:
```suggestion
	// The Operating System detected by virt-v2v.
```
and please re-generate the yaml files

pkg/apis/forklift/v1beta1/plan/vm.go
```diff
@@ -64,6 +64,8 @@ type VMStatus struct {
...
+	OperatingSystem string `json:"operationSystem,omitempty"`
```
ahadas, line 68:
```suggestion
	OperatingSystem string `json:"operatingSystem,omitempty"`
```

pkg/controller/plan/adapter/ova/ovfparser.go
```diff
@@ -60,3 +108,26 @@ func GetFirmwareFromConfig(vmConfigXML string) (firmware string, err error) {
...
+	if err != nil {
+		return
+	}
+	return mapOs(xmlConf.Metadata.LibOsInfo.V2VOS.ID), nil
+}
+
+func mapOs(xmlOs string) (os string) {
+	split := strings.Split(xmlOs, "/")
+	distro := split[3]
+	if distro == "rocky" || distro == "opensuse" || distro == "ubuntu" || distro == "fedora" {
```
ahadas, line 123: how about changing this to switch-case as well? it would be a bit more elegant

pkg/controller/plan/adapter/ova/ovfparser.go
```diff
@@ -60,3 +108,26 @@ func GetFirmwareFromConfig(vmConfigXML string) (firmware string, err error) {
...
+		os = distro
+	} else {
+		os = split[3] + strings.Split(split[4], ".")[0]
+	}
+	os, ok := osV2VMap[os]
+	if !ok {
+		os = "otherGuest64"
```
ahadas, line 130: shall we log the `os` that was received?

pkg/controller/plan/kubevirt.go
ahadas: much better, thanks

pkg/controller/plan/adapter/ova/ovfparser.go
liranr23: we can see it in the conversion pod and on the VM status the result. but yeah that maybe can help debugging.

virt-v2v/cold/entrypoint.go
```diff
@@ -56,23 +56,21 @@ func main() {
...
-		var err error
-		xmlFilePath, err = getXMLFile(DIR, "xml")
-		if err != nil {
-			fmt.Println("Error gettin XML file:", err)
-			os.Exit(1)
-		}
+	var err error
+	xmlFilePath, err = getXMLFile(DIR, "xml")
+	if err != nil {
+		fmt.Println("Error gettin XML file:", err)
```
ahadas, line 62:
```suggestion
		fmt.Println("Error getting XML file:", err)
```

pkg/controller/plan/migration.go
```diff
@@ -1124,8 +1124,9 @@ func (r *Migration) execute(vm *plan.VMStatus) (err error) {
 			return
 		}
 
-		// only for OVA, fetch config from the conversion pod
-		if r.Source.Provider.Type() == v1beta1.Ova {
+		// fetch config from the conversion pod
```
ahadas, line 1127: please move this below the `case v1beta1.Ova, v1beta1.VSphere:`

pkg/controller/plan/kubevirt.go
```diff
@@ -925,12 +925,21 @@ func (r *KubeVirt) UpdateVmByConvertedConfig(vm *plan.VMStatus, pod *core.Pod, s
...
 
-	firmware, err := ovfparser.GetFirmwareFromConfig(string(vmConfigXML))
-	if err != nil {
-		return
-	}
+	switch r.Source.Provider.Type() {
+	case api.Ova:
+		firmware, erro := ovfparser.GetFirmwareFromConfig(string(vmConfigXML))
+		if erro != nil {
+			return
```
ahadas, line 932: don't we need to set `err` here?

pkg/controller/plan/kubevirt.go
```diff
@@ -925,12 +925,21 @@ func (r *KubeVirt) UpdateVmByConvertedConfig(vm *plan.VMStatus, pod *core.Pod, s
...
+		}
 
-	vm.Firmware = firmware
+		vm.Firmware = firmware
+	case api.VSphere:
+		os, erro := ovfparser.GetOperationSystemFromConfig(string(vmConfigXML))
+		if erro != nil {
+			return
```
ahadas, line 939: and here

pkg/controller/plan/adapter/ova/ovfparser.go
ahadas: right, I meant only when we get to line 142 above so that if this happens, it will be more visible to us

ahadas (commented):
looks good overall

ahadas (approved):
I pushed another commit with cleanup in kubevirt.go, @liranr23 please check it out

PR #930, opened 2024-06-13
Let user specify root disk for multiboot VMs

bkhizgiy:
When migrating a VM with multiple bootable disks, virt-v2v automatically chooses the first disk it recognizes with an OS. This feature allows the user to specify which disk should be used as the root disk and pass that directly to virt-v2v for conversion. Additionally, the boot order will be adjusted so that this specified disk will be the first one.


pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -675,6 +686,9 @@ func (r *Builder) mapDisks(vm *model.VM, persistentVolumeClaims []*core.Persiste
 				},
 			},
 		}
+		if bootDisk == i+1 {
```
liranr23, line 689: comment? why this is a special case?

pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -926,3 +940,16 @@ func (r *Builder) GetPopulatorTaskName(pvc *core.PersistentVolumeClaim) (taskNam
...
+	if len(deviceString) == 0 {
+		return 0
+	}
+
+	for i := len(deviceString) - 1; i >= 0; i-- {
+		if unicode.IsLetter(rune(deviceString[i])) {
+			return int(deviceString[i] - 'a' + 1)
+		}
+	}
+	return 0
```
liranr23, line 954: is it possible to reach here? if so, does it mean we didn't find what we need? info log in such case? or it's just for making a return statement then this is enough.

pkg/apis/forklift/v1beta1/plan/vm.go
```diff
@@ -64,7 +64,8 @@ type VMStatus struct {
 	RestorePowerState VMPowerState `json:"restorePowerState,omitempty"`
 	// The firmware type detected from the OVF file produced by virt-v2v.
 	Firmware string `json:"firmware,omitempty"`
-
+	//Choose the primary disk the VM boots from
```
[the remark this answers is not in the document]
ahadas: this field needs to be the `VM` struct as it's part of the spec of the plan

pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -654,6 +655,16 @@ func (r *Builder) mapDisks(vm *model.VM, persistentVolumeClaims []*core.Persiste
...
 	}
+
+	var bootDisk int
+	for _, vmConf := range r.Migration.Status.VMs {
+		if vmConf.ID == vmRef.ID {
+			bootDisk = getDeviceNumber(vmConf.RootDisk)
+			break
+		}
+	}
+	bootOrder := func(order uint) *uint { return &order }
```
ahadas, line 666: please use `"k8s.io/utils/ptr"` instead

pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -926,3 +940,16 @@ func (r *Builder) GetPopulatorTaskName(pvc *core.PersistentVolumeClaim) (taskNam
...
+
+func getDeviceNumber(deviceString string) int {
+	if len(deviceString) == 0 {
+		return 0
+	}
+
+	for i := len(deviceString) - 1; i >= 0; i-- {
+		if unicode.IsLetter(rune(deviceString[i])) {
+			return int(deviceString[i] - 'a' + 1)
+		}
```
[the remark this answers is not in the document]
ahadas: a better way than documenting it would be to add unit tests ;)

pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -654,6 +655,16 @@ func (r *Builder) mapDisks(vm *model.VM, persistentVolumeClaims []*core.Persiste
 			pvcMap[pvc.Annotations[AnnImportBackingFile]] = pvc
 		}
 	}
+
+	var bootDisk int
```
ahadas, line 659: @bkhizgiy I don't remember - did you check whether virt-v2v sets the boot order accordingly in the generated VM configuration. if it does, it would be great to take it from there

virt-v2v/cold/entrypoint.go
```diff
@@ -98,9 +98,17 @@ func buildCommand() []string {
 	}
 
 	fmt.Println("Preparing virt-v2v")
+
+	if checkEnvVariablesSet("V2V_RootDisk") {
+		rootDisk := fmt.Sprintf("/dev/%s", os.Getenv("V2V_RootDisk"))
+		virtV2vArgs = append(virtV2vArgs, "--root", rootDisk)
+	} else {
+		virtV2vArgs = append(virtV2vArgs, "--root", "first")
+	}
```
ahadas, line 107: ok, this way it also applies to OVA but is it intentional? (it's relevant also for OVA but did you make sure nothing breaks this way?)
and please just determine the root disk (<device>/first) and append the argument in a single place

pkg/apis/forklift/v1beta1/plan/vm.go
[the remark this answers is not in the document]
bkhizgiy: @ahadas I'm not sure I understood what you mean, in the UI we specify it under the  plan once but then @yaacov manually coping it to be under each VM, so we expect to get it in the backend under the VM filed.

pkg/controller/plan/adapter/vsphere/builder.go
bkhizgiy: yes I looked, the order in virt-v2v is not sets the boot order, its just presents the disks in there lexicographical order, so we have to to the match ourselves.

pkg/controller/plan/adapter/vsphere/builder.go
bkhizgiy: in this case the root disk was changed we need to change the boot order, so this finds the new root disk location and sets its as the VM bootable disk.

pkg/controller/plan/adapter/vsphere/builder.go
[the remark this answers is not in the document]
bkhizgiy: ack, adding unit tests.

pkg/controller/plan/adapter/vsphere/builder.go
bkhizgiy: umm basically not really, because the virt-v2v will fail if the string is wrong, but its just for any case, we dont want to fail the migration because of boot order, ill add a log here.

virt-v2v/cold/entrypoint.go
bkhizgiy: From the tests I've did I saw its applied for ova as well, I'm not sure if its really doing anything but from virt-v2v at least its seams to be supported.

pkg/apis/forklift/v1beta1/plan/vm.go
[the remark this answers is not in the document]
liranr23: i think the meaning of it is as i wrote, moving it where LUKS is(VM struct above in the file).
there - we know it is configured by the user (or UI) and won't be edited by the controller. in the status - at least for me it tells me - it might changed by the controller.

pkg/controller/plan/adapter/vsphere/builder.go
liranr23: when writing how it's done it makes sense.
it still will be great to have a comment such as:
`if the selected boot device is the current disk, set it as the first in the boot order`

pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -926,3 +940,18 @@ func (r *Builder) GetPopulatorTaskName(pvc *core.PersistentVolumeClaim) (taskNam
 	err = planbase.VolumePopulatorNotSupportedError
 	return
 }
+
+func getDeviceNumber(deviceString string) int {
+	if !strings.HasPrefix(deviceString, "/dev/sd") || len(deviceString) < 8 {
```
liranr23, line 945: i really don't like the magic numbers here.
how about making a variable such as:
`diskPrefix := "/dev/sd"`
and then use it's length? 
`if !strings.HasPrefix(deviceString, diskPrefix) || len(deviceString) <= len(diskPrefix) {`?

pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -926,3 +940,18 @@ func (r *Builder) GetPopulatorTaskName(pvc *core.PersistentVolumeClaim) (taskNam
...
+		// In case we encounter an issue detecting the root disk order,
+		// we will return zero to avoid failing the migration due to boot orde
+		return 0
+	}
+
+	for i := 7; i < len(deviceString); i++ {
```
liranr23, line 951: same as above regarding magic numbers:
`for i := len(diskPrefix); i < len(deviceString); i++ {`

pkg/apis/forklift/v1beta1/plan/vm.go
ahadas, line 67: @liranr23: please add a leading space

bkhizgiy:
I didn't run full migration after the changes since its taking more than 24 hours with our env, but i checked each part separately.

virt-v2v/cold/entrypoint.go
bkhizgiy: @ahadas I've tested it for OVA and its not really considering the input so I returned it to vsphere only.

liranr23 (approved):

liranr23:
@bkhizgiy you need to rebase

pkg/apis/forklift/v1beta1/plan/vm.go
```diff
@@ -64,9 +67,14 @@ type VMStatus struct {
 	RestorePowerState VMPowerState `json:"restorePowerState,omitempty"`
 	// The firmware type detected from the OVF file produced by virt-v2v.
 	Firmware string `json:"firmware,omitempty"`
+<<<<<<< HEAD
 	// The Operating System detected by virt-v2v.
 	OperatingSystem string `json:"operatingSystem,omitempty"`
+	//Choose the primary disk the VM boots from
+	RootDisk string `json:"rootDisk,omitempty"`
```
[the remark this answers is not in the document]
bkhizgiy: I'm not sure something after the rebase was wrong, I added it back now.

operator/config/crd/bases/forklift.konveyor.io_migrations.yaml
```diff
@@ -508,6 +508,9 @@ spec:
                     restorePowerState:
                       description: Source VM power state before migration.
                       type: string
+                    rootDisk:
```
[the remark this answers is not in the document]
bkhizgiy: yes, I also checkout the files to main and regenerated them again using `make manifest` to be sure.
I also compered it to the `luks` key and it was in the same place.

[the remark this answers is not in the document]
ahadas: ah I see, VM status includes VM so it appears in two places - I thought it's a leftover...

pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -679,6 +690,11 @@ func (r *Builder) mapDisks(vm *model.VM, persistentVolumeClaims []*core.Persiste
 				},
 			},
 		}
+		// For multiboot VMs, if the selected boot device is the current disk,
+		// set it as the first in the boot order.
+		if bootDisk == i+1 {
+			kubevirtDisk.BootOrder = ptr.To(uint(1))
```
ahadas, line 696: isn't the boot order start from 0?

pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -679,6 +690,11 @@ func (r *Builder) mapDisks(vm *model.VM, persistentVolumeClaims []*core.Persiste
 				},
 			},
 		}
+		// For multiboot VMs, if the selected boot device is the current disk,
+		// set it as the first in the boot order.
```
ahadas, line 694: `For multiboot VMs, if ...`
otherwise bootDisk is zero, right?

pkg/controller/plan/util/utils.go
```diff
@@ -33,4 +40,19 @@ func CalculateSpaceWithOverhead(requestedSpace int64, volumeMode *core.Persisten
 	return spaceWithOverhead
 }
 
+func GetDeviceNumber(deviceString string) int {
+	if !strings.HasPrefix(deviceString, diskPrefix) || len(deviceString) <= len(diskPrefix) {
```
ahadas, line 44: isn't it enough to check `!strings.HasPrefix(deviceString, diskPrefix)`, why do we need the second part?

pkg/controller/plan/adapter/vsphere/vsphere_suite_test.go
```diff
@@ -11,3 +12,26 @@ func TestVsphere(t *testing.T) {
 	RegisterFailHandler(Fail)
 	RunSpecs(t, "vSphere Suite")
 }
+
+func TestGetDeviceNumber(t *testing.T) {
```
ahadas, line 16: this test should rather be in the 'util' package but ok, we can move and improve it separately

pkg/controller/plan/util/utils.go
bkhizgiy: ahh i think the `=` is redundant in the second check, it was for making sure we are in the correct format, for example if we will get only `/dev/sd` we missing the disk or partition so we will return 0.

pkg/controller/plan/adapter/vsphere/builder.go
bkhizgiy: from the CNV VM yaml its starts with one, so this marks the disk as the bootable one.

pkg/controller/plan/adapter/vsphere/builder.go
bkhizgiy: yes, otherwise it will keep the order same as we creating and attaching the disks, which means disk 0.

pkg/controller/plan/util/utils.go
ahadas: ok then:
```suggestion
	if !strings.HasPrefix(deviceString, diskPrefix) || len(deviceString) > len(diskPrefix) {
```

ahadas (approved):

PR #931, opened 2024-06-16
vsphere: perform CBT operations on hosts when needed

ahadas:
Previously, when running warm migrations from a vSphere provider with vCenter as its SDK endpoint, we ran create snapshot, set-checkpoints and remove-snapshots on vCenter, even when there was a network defined for an ESXi host. That cause CDI to fail: the identifier of a snapshot that was recievd from vCenter was different than the identifier of the same snapshot in ESXi and therefore CDI, that interacts with ESXi hosts in this scenario, failed to find the snapshot.

Now, when running CBT operations we check whether the disk(s) will be transferred from vCenter or ESXi, and perform the aforementioned CBT operations on the corresponding component.

### DIFF
--- a/pkg/controller/plan/migration.go
+++ b/pkg/controller/plan/migration.go
@@ -19,11 +19,11 @@ import (
 	plancontext "github.com/konveyor/forklift-controller/pkg/controller/plan/context"
 	"github.com/konveyor/forklift-controller/pkg/controller/plan/scheduler"
 	"github.com/konveyor/forklift-controller/pkg/controller/provider/web"
-	model "github.com/konveyor/forklift-controller/pkg/controller/provider/web/vsphere"
+
 	libcnd "github.com/konveyor/forklift-controller/pkg/lib/condition"
 	liberr "github.com/konveyor/forklift-controller/pkg/lib/error"
 	libitr "github.com/konveyor/forklift-controller/pkg/lib/itinerary"
-	libref "github.com/konveyor/forklift-controller/pkg/lib/ref"
+
 	"github.com/konveyor/forklift-controller/pkg/settings"
 	batchv1 "k8s.io/api/batch/v1"
 	core "k8s.io/api/core/v1"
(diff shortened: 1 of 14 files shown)


pkg/controller/plan/migration.go
```diff
@@ -1200,6 +1201,53 @@ func (r *Migration) execute(vm *plan.VMStatus) (err error) {
 	return
 }
 
+// Load host CRs.
+func (r *Migration) loadHosts() (hosts map[string]*v1beta1.Host, err error) {
```
liranr23, line 1205: move to kubevirt?

liranr23 (commented):
instead of calling `r.loadHosts` multiple times, can we "throw" it into the CR once and use it by querying a single CR later on??

ahadas:
> instead of calling `r.loadHosts` multiple times, can we "throw" it into the CR once and use it by querying a single CR later on??

which CR? the idea is to query the latest state from etcd in a lazy way

pkg/controller/plan/migration.go
ahadas: it's unrelated to kubevirt but to the source provider, but I don't mind moving it

liranr23:
> > instead of calling `r.loadHosts` multiple times, can we "throw" it into the CR once and use it by querying a single CR later on??
> 
> which CR? the idea is to query the latest state from etcd in a lazy way

ok if you wish to get always the latest state it's irrelevant. but if you are ok with doing it once, then maybe we an use the plan or migration CR to push the data into. once we need it just query the CR / have it available already. irrelevant since you want the latest.

ahadas:
> > > instead of calling `r.loadHosts` multiple times, can we "throw" it into the CR once and use it by querying a single CR later on??
> > 
> > 
> > which CR? the idea is to query the latest state from etcd in a lazy way
> 
> ok if you wish to get always the latest state it's irrelevant. but if you are ok with doing it once, then maybe we an use the plan or migration CR to push the data into. once we need it just query the CR / have it available already. irrelevant since you want the latest.

yeah, we need to query etcd since the host configuration may change

liranr23 (approved):

PR #935, opened 2024-06-19
configmap for user-provided v2v configuration files

ahadas:
Add an optional configuration for a config map in which users can specify configuration files, that should be accessible from within the conversion pod, for virt-v2v.

### DIFF
--- a/pkg/controller/plan/kubevirt.go
+++ b/pkg/controller/plan/kubevirt.go
@@ -98,6 +98,8 @@ const (
 	OvaPVLabel  = "nfs-pv"
 )
 
+const ExtraV2vConf = "extra-v2v-conf"
+
 // Map of VirtualMachines keyed by vmID.
 type VirtualMachineMap map[string]VirtualMachine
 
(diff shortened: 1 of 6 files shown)


pkg/controller/plan/kubevirt.go
```diff
@@ -1848,6 +1893,14 @@ func (r *KubeVirt) podVolumeMounts(vmVolumes []cnv.Volume, configMap *core.Confi
 				MountPath: "/opt",
 			},
 		)
+		if extraConfigMapExists {
+			mounts = append(mounts,
+				core.VolumeMount{
+					Name:      "extra-v2v-conf",
```
liranr23, line 1899: how about pulling `extra-v2v-conf` which is used here 3 times to a const?

ahadas: the names of the volume and the config map don't have to align, but ok, it could be that the code analysis tool would complain about that so made it a const

liranr23 (approved):

PR #939, opened 2024-06-19
Switch CentOS stream8 repos

liranr23:
Since we still use it and the mirrorlist isn't pointing to the `vault` repos. A command added to switch the repos. Now they will point `vault`, letting us use `dnf` commands later on.


ahadas (approved):

liranr23:
this still being tested and does not work

PR #943, opened 2024-06-19
Move the tests of GetDeviceNumber to the right package

ahadas:


pkg/controller/plan/util/util_suite_test.go
```diff
@@ -0,0 +1,13 @@
+package util
+
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+func TestVsphere(t *testing.T) {
```
bkhizgiy, line 10: isn't this redundant here?

ahadas: nope, without it the tests won't run
but changed it to TestUtil

PR #948, opened 2024-07-07
Parse defaultTransferNetwork annotation whith namespace

yaacov:
**Description**
The pull request introduces a conditional check to determine the format of the network key. The key can be in two formats:

  - `<namespace>/<name>`
  - `<name>`
  
**Explanation**
The updated code checks if the network key contains a `/`.
  - If it does, it splits the network key into namespace and name using `strings.SplitN(network, "/", 2)`.
  - If it doesn't, it assumes the key is just the name and assigns the Namespace using `mutator.plan.Spec.TargetNamespace`.
  
**Summary**
  - Before: The code always used `mutator.plan.Spec.TargetNamespace` for the Namespace.
  - After: The code checks the format of the network key. If it includes a namespace (`<namespace>/<name>`), it splits and uses the specified namespace. Otherwise, it defaults to using `mutator.plan.Spec.TargetNamespace`.

This enhancement ensures that the correct namespace is used based on the format of the network key, providing greater flexibility and correctness in handling network annotations.


yaacov:
@ahadas @bennyz please review

pkg/forklift-api/webhooks/mutating-webhook/mutators/plan-mutator.go
```diff
@@ -78,9 +79,21 @@ func (mutator *PlanMutator) setTransferNetworkIfNotSet() (bool, error) {
 		}
 
 		if network := targetProvider.Annotations["forklift.konveyor.io/defaultTransferNetwork"]; network != "" {
-			key := client.ObjectKey{
-				Namespace: mutator.plan.Spec.TargetNamespace,
-				Name:      network,
+			var key client.ObjectKey
+
+			// Netwoek key format <namespace>/<name> or <name>, in case the network key does not
```
ahadas, line 84:
```suggestion
			// Network key format <namespace>/<name> or <name>, in case the network key does not
```

ahadas (commented):
the logic that was introduced in https://github.com/kubev2v/forklift/pull/608 seems to be duplicated here, it would be better to reduce the code duplication, but the change itself looks good

tests/suit/forklift_test.go
```diff
@@ -140,4 +140,73 @@ var _ = Describe("Forklift", func() {
 			Expect(plan.Annotations[AnnPopulatorLabels]).To(Equal("True"))
 		})
 	})
+
+	Context("Plan with provider that is set with transfer network that include namespace", func() {
```
ahadas, line 144:
```suggestion
	Context("Plan with provider that is set with transfer network that includes namespace", func() {
```

ahadas (approved):

PR #969, opened 2024-07-23
Distinguish between different VDDK validation errors

jonner:
There are multiple cases that can lead to a "VDDK Init image is invalid" error message for a migration plan. They are currently handled with a single VDDKInvalid condition.

This patch adds a new error condition VDDKImageNotFound (and a new advisory condition VDDKImageNotReady) to help diagnose an issue of a missing image or incorrect image url.

Resolves: https://issues.redhat.com/browse/MTV-1150

### DIFF
--- a/pkg/controller/plan/validation.go
+++ b/pkg/controller/plan/validation.go
@@ -935,7 +935,7 @@ func createVddkCheckJob(plan *api.Plan) *batchv1.Job {
 			Template: core.PodTemplateSpec{
 				Spec: core.PodSpec{
 					SecurityContext: psc,
-					RestartPolicy:   core.RestartPolicyOnFailure,
+					RestartPolicy:   core.RestartPolicyNever,
 					InitContainers:  initContainers,
 					Containers: []core.Container{
 						{
--- a/pkg/controller/plan/validation.go
+++ b/pkg/controller/plan/validation.go
@@ -789,11 +789,12 @@ func (r *Reconciler) validateVddkImage(plan *api.Plan) (err error) {
 		return
 	}
 
-	if image, found := source.Spec.Settings[api.VDDK]; found {
+	if _, found := source.Spec.Settings[api.VDDK]; found {
 		var job *batchv1.Job
-		if job, err = r.ensureVddkImageValidationJob(plan, image); err != nil {
+		if job, err = r.ensureVddkImageValidationJob(plan); err != nil {
 			return
 		}
+		image := plan.Referenced.Provider.Source.Spec.Settings[api.VDDK]
 		if len(job.Status.Conditions) == 0 {
 			r.Log.Info("validation of VDDK job is in progress", "image", image)
 			plan.Status.SetCondition(vddkValidationInProgress)
@@ -817,7 +818,7 @@ func (r *Reconciler) validateVddkImage(plan *api.Plan) (err error) {
 	return
 }
 
-func (r *Reconciler) ensureVddkImageValidationJob(plan *api.Plan, vddkImage string) (*batchv1.Job, error) {
+func (r *Reconciler) ensureVddkImageValidationJob(plan *api.Plan) (*batchv1.Job, error) {
 	ctx, err := plancontext.New(r, plan, r.Log)
 	if err != nil {
 		return nil, err
@@ -841,7 +842,7 @@ func (r *Reconciler) ensureVddkImageValidationJob(plan *api.Plan, vddkImage stri
 	case err != nil:
 		return nil, err
 	case len(jobs.Items) == 0:
-		job := createVddkCheckJob(ctx.Plan, jobLabels, vddkImage)
+		job := createVddkCheckJob(ctx.Plan, jobLabels)
 		err = ctx.Destination.Client.Create(context.Background(), job)
 		if err != nil {
 			return nil, err
@@ -872,7 +873,9 @@ func getVddkImageValidationJobLabels(plan *api.Plan) map[string]string {
 	}
 }
 
-func createVddkCheckJob(plan *api.Plan, labels map[string]string, vddkImage string) *batchv1.Job {
+func createVddkCheckJob(plan *api.Plan, labels map[string]string) *batchv1.Job {
+	vddkImage := plan.Referenced.Provider.Source.Spec.Settings[api.VDDK]
+
 	mount := core.VolumeMount{
 		Name:      VddkVolumeName,
 		MountPath: "/opt",
--- a/pkg/controller/plan/validation.go
+++ b/pkg/controller/plan/validation.go
@@ -762,11 +762,11 @@ func (r *Reconciler) validateHooks(plan *api.Plan) (err error) {
 func (r *Reconciler) validateVddkImage(plan *api.Plan) (err error) {
 	source := plan.Referenced.Provider.Source
 	if source == nil {
-		return
+		return liberr.New("source provider is not set")
 	}
 	destination := plan.Referenced.Provider.Destination
 	if destination == nil {
-		return
+		return liberr.New("destination provider is not set")
 	}
 
 	if source.Type() != api.VSphere {
--- a/pkg/controller/plan/validation.go
+++ b/pkg/controller/plan/validation.go
@@ -70,6 +70,8 @@ const (
 	unsupportedVersion            = "UnsupportedVersion"
 	VDDKInvalid                   = "VDDKInvalid"
 	ValidatingVDDK                = "ValidatingVDDK"
+	VDDKInitImageNotReady         = "VDDKInitImageNotReady"
+	VDDKInitImageUnavailable      = "VDDKInitImageUnavailable"
 )
 
 // Categories
@@ -785,6 +787,15 @@ func (r *Reconciler) validateVddkImage(plan *api.Plan) (err error) {
 	return
 }
 
+func jobExceedsDeadline(job *batchv1.Job) bool {
+	ActiveDeadlineSeconds := settings.Settings.Migration.VddkJobActiveDeadline
+
+	if job.Status.StartTime == nil {
+		return false
+	}
+	return meta.Now().Sub(job.Status.StartTime.Time).Seconds() > float64(ActiveDeadlineSeconds)
+}
+
 func (r *Reconciler) validateVddkImageJob(job *batchv1.Job, plan *api.Plan) (err error) {
 	image := plan.Referenced.Provider.Source.Spec.Settings[api.VDDK]
 	vddkInvalid := libcnd.Condition{
@@ -806,6 +817,51 @@ func (r *Reconciler) validateVddkImageJob(job *batchv1.Job, plan *api.Plan) (err
 		r.Log.Info("validation of VDDK job is in progress", "image", image)
 		plan.Status.SetCondition(vddkValidationInProgress)
 	}
+	var ctx *plancontext.Context
+	ctx, err = plancontext.New(r, plan, r.Log)
+	if err != nil {
+		return
+	}
+	// check if a pod exists for the job
+	pods := &core.PodList{}
+	if err = ctx.Destination.Client.List(context.TODO(), pods, &client.ListOptions{
+		Namespace:     plan.Spec.TargetNamespace,
+		LabelSelector: labels.SelectorFromSet(map[string]string{"job-name": job.Name}),
+	}); err != nil {
+		return
+	}
+	if len(pods.Items) > 0 {
+		pod := pods.Items[0]
+		if len(pod.Status.InitContainerStatuses) == 0 {
+			return liberr.New("Validation pod doesn't contain expected init container", "pod", pod)
+		}
+		waiting := pod.Status.InitContainerStatuses[0].State.Waiting
+		if waiting != nil {
+			if jobExceedsDeadline(job) {
+				// If we've exceeded the deadline, set a `warning` condition to increase
+				// severity. Don't set it as `critical` because the job will continue retrying
+				// indefinitely until the pull succeeds or the provider's vddk init image URL is
+				// updated.
+				plan.Status.SetCondition(libcnd.Condition{
+					Type:     VDDKInitImageUnavailable,
+					Status:   True,
+					Reason:   waiting.Reason,
+					Category: Critical,
+					Message:  "Unable to Pull VDDK init image. Check that the image URL is correct.",
+				})
+			} else {
+				plan.Status.SetCondition(libcnd.Condition{
+					Type:     VDDKInitImageNotReady,
+					Status:   True,
+					Reason:   waiting.Reason,
+					Category: Advisory,
+					Message:  waiting.Message,
+				})
+			}
+		} else {
+			plan.Status.DeleteCondition(VDDKInitImageNotReady)
+		}
+	}
 	for _, condition := range job.Status.Conditions {
 		switch condition.Type {
 		case batchv1.JobComplete:
@@ -820,7 +876,52 @@ func (r *Reconciler) validateVddkImageJob(job *batchv1.Job, plan *api.Plan) (err
 			err = liberr.New("validation of VDDK job has an unexpected condition", "type", condition.Type)
 		}
 	}
+
 	return
+}
+
+// Cancel all other vddk validation jobs that are currently running for the
+// plan. This is necessary because validation jobs do not have a deadline,
+// so they will keep trying indefinitely if they can't pull the image. If the
+// VDDK URL is later changed, we will launch a new validation job, and the old
+// validation job is no longer relevant, so we can just kill it.
+func (r *Reconciler) cancelOtherActiveVddkCheckJobs(plan *api.Plan) (err error) {
+	ctx, err := plancontext.New(r, plan, r.Log)
+	if err != nil {
+		return
+	}
+	jobLabels := getVddkImageValidationJobLabels(ctx.Plan)
+
+	queryLabels := make(map[string]string, 1)
+	queryLabels["plan"] = jobLabels["plan"]
+	delete(queryLabels, "vddk")
+
+	jobs := &batchv1.JobList{}
+	if err = ctx.Destination.Client.List(
+		context.TODO(),
+		jobs,
+		&client.ListOptions{
+			LabelSelector: labels.SelectorFromSet(queryLabels),
+			Namespace:     plan.Spec.TargetNamespace,
+		},
+	); err != nil {
+		return
+	}
+
+	for _, job := range jobs.Items {
+		if job.Status.Active > 0 && job.Labels["vddk"] != jobLabels["vddk"] {
+			r.Log.Info("Another validation job is active for this plan. Stopping...", "job", job)
+			// make sure to delete the pod associated with this job so that it doesn't
+			// become orphaned while trying to pull its image indefinitely
+			fg := meta.DeletePropagationForeground
+			opts := &client.DeleteOptions{PropagationPolicy: &fg}
+			if err = ctx.Destination.Client.Delete(context.TODO(), &job, opts); err != nil {
+				return
+			}
+		}
+	}
+
+	return nil
 }
 
 func (r *Reconciler) ensureVddkImageValidationJob(plan *api.Plan) (*batchv1.Job, error) {
@@ -832,6 +933,8 @@ func (r *Reconciler) ensureVddkImageValidationJob(plan *api.Plan) (*batchv1.Job,
 	if err = r.ensureNamespace(ctx); err != nil {
 		return nil, liberr.Wrap(err)
 	}
+
+	r.cancelOtherActiveVddkCheckJobs(ctx.Plan)
 
 	jobLabels := getVddkImageValidationJobLabels(ctx.Plan)
 	jobs := &batchv1.JobList{}
@@ -929,9 +1032,8 @@ func createVddkCheckJob(plan *api.Plan) *batchv1.Job {
 			},
 		},
 		Spec: batchv1.JobSpec{
-			ActiveDeadlineSeconds: ptr.To[int64](int64(settings.Settings.Migration.VddkJobActiveDeadline)),
-			BackoffLimit:          ptr.To[int32](2),
-			Completions:           ptr.To[int32](1),
+			BackoffLimit: ptr.To[int32](2),
+			Completions:  ptr.To[int32](1),
 			Template: core.PodTemplateSpec{
 				Spec: core.PodSpec{
 					SecurityContext: psc,
--- a/pkg/controller/plan/validation.go
+++ b/pkg/controller/plan/validation.go
@@ -847,7 +847,7 @@ func (r *Reconciler) ensureVddkImageValidationJob(plan *api.Plan) (*batchv1.Job,
 	case err != nil:
 		return nil, err
 	case len(jobs.Items) == 0:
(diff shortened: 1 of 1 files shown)


pkg/controller/plan/validation.go
```diff
@@ -792,14 +814,37 @@ func (r *Reconciler) validateVddkImage(plan *api.Plan) (err error) {
 			r.Log.Info("validation of VDDK job is in progress", "image", image)
 			plan.Status.SetCondition(vddkValidationInProgress)
 		}
+		pods := &core.PodList{}
+		listerr := r.List(context.TODO(), pods, &client.ListOptions{
```
ahadas, line 818: note that the validation job may run on a separate k8s/OpenShift cluster and in that case, we need to use a client to the "destination cluster" (see use of `ctx.Destination.Client` below) - I think that this way we use the client to the cluster we run on

pkg/controller/plan/validation.go
```diff
@@ -792,14 +814,37 @@ func (r *Reconciler) validateVddkImage(plan *api.Plan) (err error) {
...
 			switch condition.Type {
 			case batchv1.JobComplete:
 				r.Log.Info("validate VDDK job completed", "image", image)
 				err = nil
 				return
 			case batchv1.JobFailed:
-				plan.Status.SetCondition(vddkInvalid)
+				if plan.Status.FindCondition(VDDKImageNotReady) != nil {
+					plan.Status.DeleteCondition(VDDKImageNotReady)
+					plan.Status.SetCondition(vddkImageNotFound)
```
ahadas, line 844: do I get this correctly that in this scenario (in which the VDDK image is not found), the pod will be set with `initStatus.State.Waiting != nil`?

ahadas (commented):
thanks @jonner for taking that issue and posting this fix!

pkg/controller/plan/validation.go
jsakil14: This is much needed!! I am struggling to understand what caused this sudden invalidvddkimage error on my vcenter running 7.x and I am using vddk 8.0.2 from broadcom - I queried the controller inventory / main logs - it just tells me InvalidVDDKImage - Steps tried

1. Deleted and recreated vsphere provider with diff. images of vddk - from quay.io/kubevirt/vddk , from broadcom site - vddk 7.03.8.0.2,8.0.3/// 
2. Deleted plans and recreated them
3. Deleted controller pods and updated deployment to have more cpu & mem. resources... below error from provider with - vddk:8.0.2-22388865

<h2 class="co-section-heading " style="padding: 0px; margin: 0px 0px 20px; box-sizing: border-box; font-size: var(--pf-v5-global--FontSize--xl); font-weight: var(--pf-v5-global--FontWeight--normal); font-family: var(--pf-v5-global--FontFamily--heading); line-height: var(--pf-v5-global--LineHeight--md); word-break: break-word; display: flex; justify-content: space-between; color: rgb(255, 255, 255); font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><span style="box-sizing: border-box;">Conditions</span></h2>
Type | Status | Updated | Reason | Message


```
--
{"level":"info","ts":"2024-07-24 07:37:30.842","logger":"plan|vvqrj","msg":"Condition added.","plan":{"name":"testmigration1","namespace":"konveyor-forklift"},"condition":{"type":"VDDKInvalid","status":"True","reason":"NotSet","category":"Critical","message":"VDDK init image is invalid","lastTransitionTime":null}}
{"level":"info","ts":"2024-07-24 07:37:30.842","logger":"plan|vvqrj","msg":"Condition deleted.","plan":{"name":"testmigration1","namespace":"konveyor-forklift"},"condition":{"type":"ValidatingVDDK","status":"True","reason":"Started","category":"Advisory","message":"Validating VDDK init image","lastTransitionTime":"2024-07-24T07:32:30Z"}}
{"level":"debug","ts":"2024-07-24 07:37:30.842","logger":"events","msg":"VDDK init image is invalid","type":"Warning","object":{"kind":"Plan","namespace":"konveyor-forklift","name":"testmigration1","uid":"a92ed2c6-1e2f-4725-8b34-306a6ad3e87d","apiVersion":"forklift.konveyor.io/v1beta1","resourceVersion":"136041944"},"reason":"VDDKInvalid"}
{"level":"debug","ts":"2024-07-24 07:37:30.842","logger":"events","msg":"Validating VDDK init image","type":"Normal","object":{"kind":"Plan","namespace":"konveyor-forklift","name":"testmigration1","uid":"a92ed2c6-1e2f-4725-8b34-306a6ad3e87d","apiVersion":"forklift.konveyor.io/v1beta1","resourceVersion":"136041944"},"reason":"ValidatingVDDK"}
{"level":"info","ts":"2024-07-24 07:37:30.851","logger":"plan|vvqrj","msg":"Reconcile ended.","plan":{"name":"testmigration1","namespace":"konveyor-forklift"},"reQ":0}
```
-- | -- | -- | -- | --
TargetNameNotValid | True | Jul 24, 2024, 1:02 PM | NotValid | Target VM name does not comply with DNS1123 RFC, will be automatically changed.
VDDKInvalid | True | Jul 24, 2024, 1:07 PM | NotSet | VDDK init image is invalid

jsakil14: How can I use this file or the updates you have made, i am running forklift operator 2.7 version

jonner:
I just realized that this is probably not a full fix.  

Example scenario:
1. I enter a URL `A` for the vddk image and create a migration plan with this provider. 
   - A validation job is started for this vddk init image
   - the image cannot be pulled (for whatever reason)
   - with the patch above, the validation job will indicate a `VDDKImageNotFound` error. 
2. I change the vddk image in the provider to a different url `B`
   - Due to the vddk url change, we kick off another validation job
   - Results the same as above
3. Now I change the URL back to `A`
   -  due to the vddk url change, we try to validate the vddk init image again, so we try to ensure a validation job exists
   - we find the cached job from step 1 that has the same labels, we don't create a new one and thus no validator pod is started and we don't try to re-pull the init image
   - since no new job is started, there are no pods to query about `InitContainerStatuses`, and the only thing we know is that the job failed
   - we now report `VDDKInvalid` instead of `VDDKImageNotFound`

jonner:
The new patch I just pushed does not yet address the issue I discussed in this comment: https://github.com/kubev2v/forklift/pull/969#issuecomment-2248506437

pkg/controller/plan/validation.go
ahadas: @jonner I'm not absolutely sure about that but I think that the error may also be related to scheduling (when the pod cannot be scheduled to a node) but as we don't set resource requests, it practically shouldn't happen. so I'm good with your approach

ahadas: @jsakil14 Forklift 2.7 has not been released yet so if you're running 2.7 it means you use "latest" images and those will be updated when we merge this PR

ahadas:
@jonner yeah and there's even more basic scenario that we don't handle so well - 
1. we try to validate URL `A`
2. the job fails on pulling the image
3. we sort out the issue with pulling the image from URL `A`
-> the job needs to be recreated in order to validate A again

honestly, we were not really sure whether it makes sense to add this job back then because the only case we've heard of about the VDDK image was when someone created it on a filesystem that doesn't support symbolic links.. but apparently, we see that this job actually detects issues in the field and in that case it makes less sense to expect from users to clone the plan or archive-and-recreate the plan in order to initiate another job...
but more specifically about what you wrote above, `since no new job is started, there are no pods to query about InitContainerStatuses` - wouldn't we still have the pod that ran before?

pkg/controller/plan/validation.go
```diff
@@ -792,14 +814,44 @@ func (r *Reconciler) validateVddkImage(plan *api.Plan) (err error) {
...
 		}
+		var ctx *plancontext.Context
+		ctx, err = plancontext.New(r, plan, r.Log)
+		if err != nil {
+			break
+		}
+		pods := &core.PodList{}
+		if err = ctx.Destination.Client.List(context.TODO(), pods, &client.ListOptions{
+			Namespace:     plan.Spec.TargetNamespace,
+			LabelSelector: k8slabels.SelectorFromSet(map[string]string{"job-name": job.Name}),
```
ahadas, line 825: cool, now this call would work in all scenarios but I think it should be placed elsewhere since in many reconcile-cycles we'll find the job in `JobComplete` and won't need the information from the pod

jonner:
> @jonner yeah and there's even more basic scenario that we don't handle so well -
> 
>     1. we try to validate URL `A`
> 
>     2. the job fails on pulling the image
> 
>     3. we sort out the issue with pulling the image from URL `A`
>        -> the job needs to be recreated in order to validate A again
> 
> 
> honestly, we were not really sure whether it makes sense to add this job back then because the only case we've heard of about the VDDK image was when someone created it on a filesystem that doesn't support symbolic links.. but apparently, we see that this job actually detects issues in the field and in that case it makes less sense to expect from users to clone the plan or archive-and-recreate the plan in order to initiate another job...

It's true that the more basic scenario has those issues, but at least it was consistent before: It failed with `VDDKInvalid` on the first run, and any subsequent changes didn't change that. But in the scenario I described, it would fail with `VDDKImageNotFound` on the first run and then any subsequent checks would report `VDDKInvalid`, which adds confusion, I think.

 
> specifically about what you wrote above, `since no new job is started, there are no pods to query about InitContainerStatuses` - wouldn't we still have the pod that ran before?


Well, you know the code better than I do, but in my testing, the pod related to the job doesn't seem to be retained after the job completes. Specifically, after the job completes, the call to `cxt.Destination.Client.List()` returns no pods. Should the pod still be available somewhere?

pkg/controller/plan/validation.go
jonner: > @jonner I'm not absolutely sure about that but I think that the error may also be related to scheduling (when the pod cannot be scheduled to a node) but as we don't set resource requests, it practically shouldn't happen. so I'm good with your approach

Well, it's easy enough to check the reason for ErrImagePull to differentiate those failure cases if you think it would be useful.

jsakil14: > @jsakil14 Forklift 2.7 has not been released yet so if you're running 2.7 it means you use "latest" images and those will be updated when we merge this PR

I meant operator version - v2.7.0 - Also , openshift console - installed operator gives below : Forklift Operator
2.7.0 provided by Konveyor @ahadas

ahadas: > > @jsakil14 Forklift 2.7 has not been released yet so if you're running 2.7 it means you use "latest" images and those will be updated when we merge this PR
> 
> I meant operator version - v2.7.0 - Also , openshift console - installed operator gives below : Forklift Operator 2.7.0 provided by Konveyor @ahadas

yeah, but you can see [here](https://github.com/redhat-openshift-ecosystem/community-operators-prod/tree/main/operators/forklift-operator) that we didn't publish 2.7.0 in the operator hub so if you managed to install it, I assume you've added another catalog source and the 'latest' tag of the images is used (you can check that by looking at the image of any pod in the konveyor-forklift namespace) - and then what I wrote about would hold

ahadas: > > @jonner I'm not absolutely sure about that but I think that the error may also be related to scheduling (when the pod cannot be scheduled to a node) but as we don't set resource requests, it practically shouldn't happen. so I'm good with your approach
> 
> Well, it's easy enough to check the reason for ErrImagePull to differentiate those failure cases if you think it would be useful.

ok, so let's do that then

ahadas:
> > @jonner yeah and there's even more basic scenario that we don't handle so well -
> > ```
> > 1. we try to validate URL `A`
> > 
> > 2. the job fails on pulling the image
> > 
> > 3. we sort out the issue with pulling the image from URL `A`
> >    -> the job needs to be recreated in order to validate A again
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > honestly, we were not really sure whether it makes sense to add this job back then because the only case we've heard of about the VDDK image was when someone created it on a filesystem that doesn't support symbolic links.. but apparently, we see that this job actually detects issues in the field and in that case it makes less sense to expect from users to clone the plan or archive-and-recreate the plan in order to initiate another job...
> 
> It's true that the more basic scenario has those issues, but at least it was consistent before: It failed with `VDDKInvalid` on the first run, and any subsequent changes didn't change that. But in the scenario I described, it would fail with `VDDKImageNotFound` on the first run and then any subsequent checks would report `VDDKInvalid`, which adds confusion, I think.

yeah, I agree

> 
> > specifically about what you wrote above, `since no new job is started, there are no pods to query about InitContainerStatuses` - wouldn't we still have the pod that ran before?
> 
> Well, you know the code better than I do, but in my testing, the pod related to the job doesn't seem to be retained after the job completes. Specifically, after the job completes, the call to `cxt.Destination.Client.List()` returns no pods. Should the pod still be available somewhere

unless I'm missing something, I read the changes in this PR as introducing logic that says: "once the job completes, we check the status of the pod that was triggered for the job to figure out the reason for a failure", so it assumes the pod exists also when the job is completed with a failure, right?
it could be that there's cleanup of the job's completed pods at some point (not by us but by k8s/OpenShift), I think I also saw it happening (having a completed job without associates pods), but I was puzzled by that part that suggests it's always the case that completed jobs have no associated pods

pkg/controller/plan/validation.go
jsakil14: > > > @jsakil14 Forklift 2.7 has not been released yet so if you're running 2.7 it means you use "latest" images and those will be updated when we merge this PR
> > 
> > 
> > I meant operator version - v2.7.0 - Also , openshift console - installed operator gives below : Forklift Operator 2.7.0 provided by Konveyor @ahadas
> 
> yeah, but you can see [here](https://github.com/redhat-openshift-ecosystem/community-operators-prod/tree/main/operators/forklift-operator) that we didn't publish 2.7.0 in the operator hub so if you managed to install it, I assume you've added another catalog source that uses the 'latest' tag (you can check that by looking at the image of any pod in the konveyor-forklift namespace) - and then what I wrote about would hold

Looks like I do use **latest**...Do I have to change it? @ahadas 

```
 env:
    - name: WATCH_NAMESPACE
      valueFrom:
        fieldRef:
          apiVersion: v1
          fieldPath: metadata.annotations['olm.targetNamespaces']
    - name: ANSIBLE_GATHERING
      value: explicit
    - name: CONTROLLER_IMAGE
      value: quay.io/kubev2v/forklift-controller:latest
    - name: API_IMAGE
      value: quay.io/kubev2v/forklift-api:latest
    - name: MUST_GATHER_IMAGE
      value: quay.io/kubev2v/forklift-must-gather:latest
    - name: UI_PLUGIN_IMAGE
      value: quay.io/kubev2v/forklift-console-plugin:latest
    - name: VALIDATION_IMAGE
      value: quay.io/kubev2v/forklift-validation:latest
    - name: VIRT_V2V_IMAGE
      value: quay.io/kubev2v/forklift-virt-v2v:latest
    - name: VIRT_V2V_WARM_IMAGE
      value: quay.io/kubev2v/forklift-virt-v2v-warm:latest
    - name: VIRT_V2V_DONT_REQUEST_KVM
    - name: SNAPSHOT_REMOVAL_TIMEOUT
    - name: SNAPSHOT_STATUS_CHECK_RATE
    - name: CLEANUP_RETRIES
    - name: CDI_EXPORT_TOKEN_TTL
    - name: FILESYSTEM_OVERHEAD
    - name: BLOCK_OVERHEAD
    - name: POPULATOR_CONTROLLER_IMAGE
      value: quay.io/kubev2v/populator-controller:latest
    - name: OVIRT_POPULATOR_IMAGE
      value: quay.io/kubev2v/ovirt-populator:latest
    - name: OPENSTACK_POPULATOR_IMAGE
      value: quay.io/kubev2v/openstack-populator:latest
    - name: OVA_PROVIDER_SERVER_IMAGE
      value: quay.io/kubev2v/forklift-ova-provider-server:latest
    - name: OVIRT_OS_MAP
    - name: VSPHERE_OS_MAP
    - name: OPERATOR_CONDITION_NAME
      **value: forklift-operator.v2.7.0**
    **image: quay.io/kubev2v/forklift-operator:latest**
    imagePullPolicy: Always
    livenessProbe:
      failureThreshold: 3
      httpGet:
        path: /healthz
        port: 6789
        scheme: HTTP
      initialDelaySeconds: 15
      periodSeconds: 20
      successThreshold: 1
      timeoutSeconds: 1
    name: forklift-operator
```

jonner:
> unless I'm missing something, I read the changes in this PR as introducing logic that says: "once the job completes, we check the status of the pod that was triggered for the job to figure out the reason for a failure", so it assumes the pod exists also when the job is completed with a failure, right?

Not quite. What happens is that while the job is running, `validateVddkImage()` checks the `vddk-validator-*` pod associated with the job and sets the condition to `VDDKImageNotReady` if the pod is waiting. 

At some point, the image will either be pulled (and the pod will start running), or the pod will time out waiting for the image. In the first scenario, the job should run to completion and either succeed or fail based on the validation result. In the second scenario, the job will fail.

Immediately after the job finishes (either successfully or due to timing out), the next call to `validateVddkImage()` will **not** find a pod associated with that job (in my  testing `cxt.Destination.Client.List()` returns no pods at this point). If the job has a `JobFailed` condition, the only indication about *why* the job failed is checking whether I have set a `VDDKImageNotReady` status condition on the plan during the previous reconcile (this is the reason that I made the `VDDKImageNotReady`  condition durable: so after the pod disappears, the next reconcile still has a hint about what might have caused the job to fail).

But the above description does not apply to a cached job. In my experience, there are no pods returned for a cached job, so there's no way to determine the failure reason in this case.

But maybe I am simply doing something wrong and there **is** a way to get pods for a completed job that I don't know about?

pkg/controller/plan/validation.go
jsakil14: #971

bkhizgiy:
@jonner I've been investigating the issue with pod deletion, and it turns out that it happens only when the pod is not fully created. If the VDDK image is invalid (the pod is unable to pull the image) and the job reaches the `VDDK_JOB_ACTIVE_DEADLINE` limit, the job fails, and the pod is deleted. However, if the job is successful, the pod remains as it is. Adding additional configuration to the job spec won't assist in this case.

After discussing this with @ahadas , we believe the best approach is to remove the `VDDK_JOB_ACTIVE_DEADLINE` limit from the job and handle it at the plan level. We'll add a condition to the plan controller to check if the job is still running and if the time exceeds the `VDDK_JOB_ACTIVE_DEADLINE` parameter. If it does, we will propagate the pod status to the plan without failing the job and notify the user of the issue. This approach also addresses another issue: since the job remains running, if the user updates the VDDK URL with a correct one, the job should successfully complete under the same job and plan.

What do you think of this solution?

jonner:
Interesting idea. There are some details of your proposed solution that aren't yet 100% clear in my mind, but I will investigate. It sounds promising.

pkg/controller/plan/validation.go
```diff
@@ -736,6 +741,38 @@ func (r *Reconciler) validateHooks(plan *api.Plan) (err error) {
 }
 
 func (r *Reconciler) validateVddkImage(plan *api.Plan) (err error) {
+	source := plan.Referenced.Provider.Source
+	if source == nil {
+		return
```
mnecas, line 746: We want to return nil only if everything passes, would it make sense to return error here?
I know it is just an edge case

pkg/controller/plan/validation.go
```diff
@@ -736,6 +741,38 @@ func (r *Reconciler) validateHooks(plan *api.Plan) (err error) {
...
+	}
+	destination := plan.Referenced.Provider.Destination
+	if destination == nil {
+		return
```
mnecas, line 750: Same here

pkg/controller/plan/validation.go
mnecas: + think that this should be already coverd in previous validation but don't mind having extra protection

pkg/controller/plan/validation.go
```diff
@@ -736,6 +741,38 @@ func (r *Reconciler) validateHooks(plan *api.Plan) (err error) {
...
+	}
+
+	if source.Type() != api.VSphere {
+		// VDDK is not used for other provider types
+		return
```
mnecas, line 755: Here the noop makes sense to me as we dont care about other providers.

pkg/controller/plan/validation.go
```diff
@@ -751,49 +788,98 @@ func (r *Reconciler) validateVddkImage(plan *api.Plan) (err error) {
...
+				if job.Annotations == nil {
+					job.Annotations = make(map[string]string)
+				}
+				var bytes []byte
+				if bytes, err = json.Marshal(initStatus.State.Waiting); err != nil {
+					return
+				}
+				job.Annotations[AnnVddkInitImageWaiting] = string(bytes)
+			} else {
+				delete(job.Annotations, AnnVddkInitImageWaiting)
```
mnecas, line 836: Please add log with the state

mnecas: Do we know that the job always have annotations at this point?

pkg/controller/plan/validation.go
```diff
@@ -751,49 +788,98 @@ func (r *Reconciler) validateVddkImage(plan *api.Plan) (err error) {
...
-		var job *batchv1.Job
-		if job, err = r.ensureVddkImageValidationJob(plan, image); err != nil {
-			return
-		}
-		if len(job.Status.Conditions) == 0 {
-			r.Log.Info("validation of VDDK job is in progress", "image", image)
-			plan.Status.SetCondition(vddkValidationInProgress)
+				// If a job fails due to exceeding the deadline, the pods will be deleted and
+				// we won't be able to determine what caused the failure. So persist any
+				// initStatus errors via job annotations so that they we can use them later.
```
mnecas, line 826: So we will be constatnly failing until we get the image?
What is the difference between this and removing the `activeDeadlineSeconds`?

mnecas: Do we know if we start another pull when the job restarts?
If we will restart the job multiple times we might start more pulls

pkg/controller/plan/validation.go
```diff
@@ -775,49 +813,118 @@ func (r *Reconciler) validateVddkImage(plan *api.Plan) (err error) {
...
+	if len(pods.Items) > 0 {
+		pod := pods.Items[0]
+		if len(pod.Status.InitContainerStatuses) == 0 {
+			return liberr.New("Validation pod doesn't contain expected init container", "pod", pod)
+		}
+		waiting := pod.Status.InitContainerStatuses[0].State.Waiting
+		if waiting != nil {
+			if jobExceedsDeadline(job) {
+				// If we've exceeded the deadline, set a `warning` condition to increase
+				// severity. Don't set it as `critical` because the job will continue retrying
```
mnecas, line 842: Think it should be cricital so we won't allow the users to start the migration.

mnecas (approved):
Drastic improvements of the VDDK validations!
Thank you so much for your help, LGTM!

PR #975, opened 2024-07-29
Add subnet mask, gateway and DNS to static IP preservation

liranr23:
Collect the prefix length from vSphere and send it to virt-v2v to set it on destination.

### DIFF
--- a/pkg/controller/plan/adapter/vsphere/builder.go
+++ b/pkg/controller/plan/adapter/vsphere/builder.go
@@ -242,6 +242,9 @@ func (r *Builder) mapMacStaticIps(vm *model.VM) string {
 	if !isWindows(vm) {
 		return ""
 	}
+	if vm.GuestNetworks == nil || len(vm.GuestNetworks) < 1 {
+		return ""
+	}
 	configurations := []string{}
 	sort.Slice(vm.GuestNetworks, func(i, j int) bool { return vm.GuestNetworks[i].DeviceId < vm.GuestNetworks[j].DeviceId })
 	firstId := vm.GuestNetworks[0].DeviceId
--- a/pkg/controller/plan/adapter/vsphere/builder.go
+++ b/pkg/controller/plan/adapter/vsphere/builder.go
@@ -243,7 +243,7 @@ func (r *Builder) mapMacStaticIps(vm *model.VM) string {
 	configurations := []string{}
 	for _, guestNetwork := range vm.GuestNetworks {
 		if guestNetwork.Origin == string(types.NetIpConfigInfoIpAddressOriginManual) {
-			configurations = append(configurations, fmt.Sprintf("%s:ip:%s", guestNetwork.MAC, guestNetwork.IP))
+			configurations = append(configurations, fmt.Sprintf("%s:ip:%s,,%d", guestNetwork.MAC, guestNetwork.IP, guestNetwork.PrefixLength))
 		}
 	}
 	return strings.Join(configurations, "_")
--- a/pkg/controller/plan/adapter/vsphere/builder.go
+++ b/pkg/controller/plan/adapter/vsphere/builder.go
@@ -4,10 +4,12 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"net"
 	liburl "net/url"
 	"path"
 	"regexp"
 	"sort"
+	"strconv"
 	"strings"
 
 	api "github.com/konveyor/forklift-controller/pkg/apis/forklift/v1beta1"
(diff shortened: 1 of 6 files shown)


pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -240,10 +242,35 @@ func (r *Builder) mapMacStaticIps(vm *model.VM) string {
...
 	configurations := []string{}
+	sort.Slice(vm.GuestNetworks, func(i, j int) bool { return vm.GuestNetworks[i].DeviceId < vm.GuestNetworks[j].DeviceId })
+	firstId := vm.GuestNetworks[0].DeviceId
 	for _, guestNetwork := range vm.GuestNetworks {
 		if guestNetwork.Origin == string(types.NetIpConfigInfoIpAddressOriginManual) {
-			configurations = append(configurations, fmt.Sprintf("%s:ip:%s", guestNetwork.MAC, guestNetwork.IP))
+			isIpv4 := false
+			if net.IP.To4(net.ParseIP(guestNetwork.IP)) != nil {
+				isIpv4 = true
+			}
```
bennyz, line 256: small nit, I think you can do:
```go
isIpv4 := net.IP.To4(net.ParseIP(guestNetwork.IP)) != nil
```

pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -241,9 +243,34 @@ func (r *Builder) mapMacStaticIps(vm *model.VM) string {
...
-			configurations = append(configurations, fmt.Sprintf("%s:ip:%s", guestNetwork.MAC, guestNetwork.IP))
+			gateway := ""
+			if !gatewaySet {
+				isIpv4 := net.IP.To4(net.ParseIP(guestNetwork.IP)) != nil
+				for _, ipStack := range vm.GuestIpStacks {
+					if net.IP.To4(net.ParseIP(ipStack.Gateway)) != nil && !isIpv4 || net.IP.To4(net.ParseIP(ipStack.Gateway)) == nil && isIpv4 {
+						// not the right IPv4 / IPv6 correlation
+						continue
+					}
+					network, err := netip.ParsePrefix(fmt.Sprintf("%s/%d", guestNetwork.IP, guestNetwork.PrefixLength))
```
liranr23, line 257: i think i can also check the DNS if that matters to have a better fit

liranr23: after discussion with tzahi sounds irrelevant

pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -241,9 +243,34 @@ func (r *Builder) mapMacStaticIps(vm *model.VM) string {
...
+					}
+					if network.Contains(gw) {
+						// checks if the gateway in the right network subnet. we set gateway only once as it is suppose to be system wide.
+						gateway = ipStack.Gateway
+						gatewaySet = true
+					}
+				}
+			}
+			dnsString := strings.Join(guestNetwork.DNS, ",")
+			configurations = append(configurations, fmt.Sprintf("%s:ip:%s,%s,%d,%s", guestNetwork.MAC, guestNetwork.IP, gateway, guestNetwork.PrefixLength, dnsString))
```
ahadas, line 273: do we know what happens with dual stack? as we spoke offline, I'm not sure if that's supported in kubevirt but as far as I remember it is supported in vSphere and if I'm right on this, then could it happen that we set the GW for IPv4 and then override it when specifying IPv6 for the same MAC address?

pkg/controller/provider/container/vsphere/model.go
```diff
@@ -332,7 +332,7 @@ func (v *HostAdapter) Apply(u types.ObjectUpdate) {
 					network := &v.model.Network
 					for _, nic := range array.HostVirtualNic {
 						dGroup := func() (key string) {
-							dp := nic.Spec.DistributedVirtualPort
+							dp := nic.Spec.DistributedVirtualPort //here
```
ahadas, line 335: ah?

liranr23: self-comment when debugging, i'll remove :)

pkg/controller/plan/adapter/vsphere/builder.go
liranr23: From what I've tested the VM will result in 2 gateways, one for IPv4 and one for IPv6. The DNS list will be combined for both (IPv4 and IPv6). Checking into details of the NIC per version it is set 1:1 to the source. Maybe someone in CNI can answer if it's allowed?

liranr23: @phoracek

liranr23: I had offline discussion with @ahadas, it looks alright.

ahadas (approved):

liranr23:
I'm merging and backporting this PR.
A note to remember is to update the virt-v2v deps to consume a new virtio-win which supports win2k22.

PR #979, opened 2024-08-13
Update virt-v2v bazel rpm deps

yaacov:
Update virt-v2v bazel rpm deps


yaacov:
@liranr23 @ahadas hi, without this update a new user can't run the image build scripts, please review.

bennyz (approved):

PR #980, opened 2024-08-13
[MTV-595] Add V2V_preserveStaticIPs env var for virt-v2v pod

yaacov:

**Issue**:
When migrating a simple RHEL8 VM from VMware 7 to OCPv 4.13 the name of the network interfaces changes and the static IP configuration for the VM no longer works

**Why we need this change**:
in MTV-595 we plan to introduce code that will support static IP defined using RHEL7 and above NetworkManager, we will need an indicator for user selecting preserveStaticIPs

**Why this is safe:**
We add a new env var, no other change needed

### DIFF
--- a/pkg/controller/plan/adapter/vsphere/builder.go
+++ b/pkg/controller/plan/adapter/vsphere/builder.go
@@ -200,6 +200,11 @@ func (r *Builder) PodEnvironment(vmRef ref.Ref, sourceSecret *core.Secret) (env 
 		if err != nil {
 			return
 		}
+
+		env = append(env, core.EnvVar{
+			Name:  "V2V_preserveStaticIPs",
+			Value: "true",
+		})
 	}
 
 	libvirtURL, fingerprint, err := r.getSourceDetails(vm, sourceSecret)


bennyz (approved):

pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -237,9 +237,6 @@ func (r *Builder) PodEnvironment(vmRef ref.Ref, sourceSecret *core.Secret) (env
 }
 
 func (r *Builder) mapMacStaticIps(vm *model.VM) string {
-	if !isWindows(vm) {
```
ahadas, line 240: we also need to remove [the check in the validator](https://github.com/kubev2v/forklift/blob/254f0093b0771b6177b50005ba0e08030bd9ccdc/pkg/controller/plan/adapter/vsphere/validator.go#L142) and remove the `isWindows` function then

yaacov: Owww, thanks, missed that

ahadas: on a second thought, if you're not going to use this for RHEL VMs, why do we need to pass this information to the conversion pod?..

yaacov: @liranr23 is using this env var to indicate `PreserveStaticIPs` , the ceck is:
https://github.com/kubev2v/forklift/blob/5010fb4b00e51fce6102ed0c2b02fc2dab2c398c/virt-v2v/cold/entrypoint.go#L131

yaacov: I can add an explicit `PreserveStaticIPs` but I think @liranr23 way is safer because he also check that the VM actually set the IPs manually and not just ask for static IPs

liranr23: i'm ok with adding the `isWindows` and the check in the validator. it's true we check if that's not an empty string in virt-v2v. it makes a bit nonsense to send it but if that help and covers all scenarios that's ok.

pkg/controller/plan/adapter/vsphere/builder_test.go
```diff
@@ -17,30 +17,35 @@ import (
...
 				Origin: ManualOrigin,
 			},
 		},
 		}, "00:50:56:83:25:47:ip:172.29.3.193_00:50:56:83:25:47:ip:fe80::5da:b7a5:e0a2:a097"),
-		Entry("non-static ip", &model.VM{GuestID: "windows9Guest", GuestNetworks: []vsphere.GuestNetwork{{MAC: "00:50:56:83:25:47", IP: "172.29.3.193", Origin: string(types.NetIpConfigInfoIpAddressOriginDhcp)}}}, ""),
-		Entry("non windows vm", &model.VM{GuestID: "other", GuestNetworks: []vsphere.GuestNetwork{{MAC: "00:50:56:83:25:47", IP: "172.29.3.193", Origin: ManualOrigin}}}, ""),
-		Entry("no OS vm", &model.VM{GuestNetworks: []vsphere.GuestNetwork{{MAC: "00:50:56:83:25:47", IP: "172.29.3.193", Origin: ManualOrigin}}}, ""),
+		Entry("non-static ip", &model.VM{GuestID: "windows9Guest", GuestNetworks: []vsphere.GuestNetwork{{MAC: TestGuestNetworkMac, IP: TestGuestNetworkIp, Origin: string(types.NetIpConfigInfoIpAddressOriginDhcp)}}}, ""),
+		Entry("non windows vm", &model.VM{GuestID: "other", GuestNetworks: []vsphere.GuestNetwork{TestGuestNetwork}}, ExpectedMacStaticIps),
+		Entry("no OS vm", &model.VM{GuestNetworks: []vsphere.GuestNetwork{TestGuestNetwork}}, ExpectedMacStaticIps),
```
liranr23, line 48: well, if we have no OS, then we won't have any details of the vNIC GuestNetworks

liranr23 (commented):
Did you test virt-v2v with a map on RHEL to see it's really ignored?

I would put into the commit that on a following patch you are going to add virt-customize script to support RHEL. That will add some value of this PR as a preparation step.

pkg/controller/plan/adapter/vsphere/builder_test.go
yaacov: do you suggest to remove the "no OS vm" test ?

yaacov: > I would put into the commit that on a following patch you are going to add virt-customize script to support RHEL. That will add some value of this PR as a preparation step.

Added to the PR description

yaacov: and the commit message

yaacov:
> Did you test virt-v2v with a map on RHEL to see it's really ignored?

@liranr23 , well, when setting the machine IP to "manual" using  Network manager the agent does not peek it up ... so no mac to ip map is sent to the v2v pod , because in the inventory the origin is null instead of "manual" :-(

looking into that ... we may need to send the explicit request to "PreserveStaticIPs" 

cc:// @ahadas

pkg/controller/plan/adapter/vsphere/builder.go
ahadas: > I can add an explicit `PreserveStaticIPs` but I think @liranr23 way is safer because he also check that the VM actually set the IPs manually and not just ask for static IPs

yeah but on the other hand we don't always know about that information - that's why we added that validation



> i'm ok with adding the `isWindows` and the check in the validator. it's true we check if that's not an empty string in virt-v2v. it makes a bit nonsense to send it but if that help and covers all scenarios that's ok.

I guess you meant `PreserveStaticIPs`, right?

ahadas (approved):
I don't mind adding this new environment variable but I think it is worth checking with @rwmjones how he'd like this to work in virt-v2v for RHEL guests because if we aim to use the same arguments we provide for Windows guests (i.e., mapping MAC address to IP, subnet, etc..) then it would also make sense in my opinion to implement it the same way as for Windows guests in MTV - but that's something you can easily change later on..

yaacov:
> On the subject of this pull request, I'm not sure what this environment variable does.

We use environment variable to pass arguments to the pod we start, then we use this env vars to build the virt-vtv command line args, so when we see `V2V_staticIPs` env var we add a `--mac` arg, 
in this case when I will see `V2V_preserveStaticIPs` i will run `virt-customize` to do the network-manager POC code

yaacov:
Ref: https://github.com/kubev2v/forklift/pull/983 
Update on the pod running virt-v2v to use this env var.

PR #983, opened 2024-08-14
[MTV-595] Add customize disk image step to virt-v2v pod

yaacov:
Ref: https://issues.redhat.com/browse/MTV-595

**Issue:**
When migrating a simple RHEL8 VM from VMware 7 to OCPv 4.13 the name of the network interfaces changes and the static IP configuration for the VM no longer works

**What this pull request do:**

a. If the `os` is RHEL and user opted to `preserveStaticIPs` and the source is `vcenter` :

we look for network manager configuration files:
`/etc/sysconfig/network-scripts/`   - for sysconfig scripts
`/etc/NetworkManager/system-connections/` - for network connection files

b. If we find configuration files that can connect `ip address` to `interface name` we map  `interface name` to mac and address.

c. we create a `/etc/udev/rules.d/70-persistent-net.rules` file to pin the `interface names` to `mac addresses`

**Limitations:**

Only work for network manager
Require the source Virtual machine to run while MTV is collecting inventory to create the mac to ip mapping
 
**Screenshot**

screenshot-console-openshift-console apps mtv82 rhos-psi cnv-qe rhood us-2024 08 27-13_47_24

PR #984, opened 2024-08-16
[MTV-1299] Add windows customize disk image step to virt-v2v pod

mnecas:
### TLDR: 
This PR adds new steps after `virt-v2v` which edit the Windows images to enable the disks after boot.

### Issue: 
When Windows VM boots after the migration it has a new disk interface and does not recognize the VirtIO volumes. By default, they are offline this PR uploads scripts and configures the Windows registries so they are enabled after boot. (Ref: https://issues.redhat.com/browse/MTV-1299)

### Fix: 
To fix the offline disks we need to automatically enable them after boot, but we don't want to enable all disks as some of them (iSCSI etc.) could cause issues if they were brought up before everything would be ready. To this, we need to upload scripts into Windows using `virt-customize --upload`  with fixed firstboot script.

So the current workflow now is:
1. Migrate the VM from VMware using `virt-v2v`
2. Upload our scripts to the VM using `virt-customize` and overwrite the firstboot script

**Additional changes:**
- Added a new `EmbedTool` to create files and their structure from the embedded filesystem 
- Removed the RHEL-specific configs and replaced them with windows from the @yaacov fork
- Used Golang embedded files to more readable load file 
- Update Bazel with the embedded sources

### Notes:
Part of this PR is forked from https://github.com/kubev2v/forklift/pull/983 done by @yaacov
And the scripts to enable the disks are inspired from https://github.com/fabiendupont/ims.premigration-windows/blob/master/files/pre-migrate.ps1 . This script was created with mind that it will be executed on the guest before migraiton, that's why it generates the scripts to the firstboot section.


virt-v2v/cold/customize-image.go
```diff
@@ -0,0 +1,145 @@
...
+	// Get all disk files matching the pattern
+	disks, err := filepath.Glob(filepath.Join(dir, "*-sd*"))
+	if err != nil {
+		return "", fmt.Errorf("error getting disks: %w", err)
+	}
+
+	// Default disk name
+	diskName := "sda"
+
+	// Check if V2V_RootDisk environment variable is set
```
ahadas, line 117: that was added to cope with a scenario of having multi-boot disks but what if there's a single disk that is bootable and it's not sda? it would be better to get the information about the bootable disk from the output of virt-v2v

virt-v2v/cold/entrypoint.go
```diff
@@ -74,6 +82,38 @@ func main() {
...
+		return err
+	}
+
+	operatingSystem, err := GetOperationSystemFromConfig(xmlData)
+	if err != nil {
+		fmt.Printf("Error getting OS ID: %v\n", err)
+	} else {
+		fmt.Printf("Operating System ID: %s\n", operatingSystem)
+	}
+	if source == vSphere {
```
ahadas, line 98: don't we need this also for OVA?

virt-v2v/cold/BUILD.bazel
```diff
@@ -92,7 +92,13 @@ container_image(
 
 go_library(
     name = "cold_lib",
-    srcs = ["entrypoint.go"],
+    srcs = [
+        "customize-image.go",
```
ahadas, line 96: I suppose it will be needed also for warm migrations, right?

virt-v2v/cold/customize-image.go
mnecas: Which virt-v2v output do you mean?

bkhizgiy: It can be scanned from the virt-v2v logs, but it will cause a different issue since the output available only after starting the conversion command, so we will have to run it twice. We had a discussion about this in the multi boot [ticket](https://issues.redhat.com/browse/MTV-1079) and Richard suggested to use `virt-v2v-inspector` prior to executing the command, its probably will be the best solution.

https://libguestfs.org/virt-v2v-inspector.1.html

ahadas: I meant the configuration of the converted VM that virt-v2v generates

yaacov: @bkhizgiy worked on:
https://issues.redhat.com/browse/MTV-1079

I asked her and she sent me this link:
https://issues.redhat.com/browse/MTV-1079?focusedId=25052893&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-25052893

we use:
`info: virt-v2v: virt-v2v 2.5.6rhel=9,release=3.el9 (x86_64)`
and this fix it on 2.5.7 :-( 

in our version we have debug output with the information:
`libguestfs: trace: v2v: inspect_get_mountpoints = ["/boot", "/dev/sda1", "/", "/dev/vg_rhev-orange-03/lv_root"]`
( IMHO it's not a good idea to pars the debug output ... _

@rwmjones also suggested using  virt-v2v-inspector.
https://issues.redhat.com/browse/MTV-1079?focusedId=25052873&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-25052873

IMHO opinion we should wait for the fix in 2.5.7, @mnecas WDYT ?

bkhizgiy: @ahadas It can be taken from the virt-v2v final output, but again, that means we need to run a full conversion before being able to get this info for the actual run....

virt-v2v/cold/BUILD.bazel
yaacov: @ahadas @mnecas can we do the fix for the warm migration in a separate PR ? does MTV-1299 require a fix for the warm case too ?

virt-v2v/cold/customize-image.go
yaacov: > It can be taken from the virt-v2v final output,

@bkhizgiy , that will be great for this use case, because we run it after virt-v2v finish, but on my build we don't have this output ... what am i missing ?

yaacov: maybe i have old virt-v2v, v2.5.6

mnecas: @yaacov  Why can't we bump to 2.5.7 now? Are we missing the build on the u/s? I know we are using 2.6.4 in d/s so we should synchronise them.

@bkhizgiy I would consider the scraping and trying to find some data in logs our last option and would rather go other direction.

What is the difference between using the virt-v2v-inspector and checking the instance XML file? Does it show more details? I'm not that familiar with it but will definitely check it tmr.

I can't access my computer today so can't do much investigations. If possible I would try to find some solutions for which we don't need to way even if it is bit hack, we can refactor later after we will get the fix to unblock users.

virt-v2v/cold/BUILD.bazel
mnecas: Yeah we can split it but Arik had good comment that we need to think about other migration types so we need to move out these scripts out of cold and make some utils which can be shared in both of them.

virt-v2v/cold/customize-image.go
rwmjones: > @ahadas It can be taken from the virt-v2v final output, but again, that means we need to run a full conversion before being able to get this info for the actual run....

Note this is what virt-v2v-inspector does (a full conversion). It's not as expensive as you think because it doesn't do any copying, although it also isn't completely free either (especially for Linux guests because dracut is slow).

rwmjones: As there's a bit of confusion about virt-v2v version numbers, it's a good idea to refer to these upstream branches:

master branch: https://github.com/libguestfs/virt-v2v/commits/master/

- This is the active development branch

rhel-9.5 branch: https://github.com/libguestfs/virt-v2v/commits/rhel-9.5

- Features backported to RHEL 9.5 + RHEL-only patches.  You'll have to scroll down a bit through the RHEL patches to see what we've backported.
- Since we backport aggressively, you'll find almost all upstream patches are already present in RHEL version 2.5.6.

rhel-9.4 (z-stream) branch: https://github.com/libguestfs/virt-v2v/commits/rhel-9.4

- For completeness, what's been backported to 9.4.z

yaacov: the PR we are looking for is in  9.5
https://github.com/libguestfs/virt-v2v/commit/4df3dcf64da91645d20aa9f74cd04fa8985c4610

yaacov: how do we use the rhel-9.5 rpms ?

yaacov: @mnecas hi, so if we go with `virt-v2v-inspector` 
a. we have it already installed
b. running it is "fast" compering to running `virt-v2v`

running it will look like:
`virt-v2v-inspector -i libvirtxml  /var/tmp/v2v/<vm name>.xml -o /var/tmp/v2v/<vm name>.inspect.xml`

and the output will include the root disk:
```xml
<?xml version='1.0' encoding='utf-8'?>
<v2v-inspection>
  <!-- generated by virt-v2v-inspector 2.5.6fedora=40,release=1.fc40 -->
  <program>virt-v2v-inspector</program>
  <package>virt-v2v</package>
  <version>2.5.6</version>
  <disks>
    <disk index='0'>
      <virtual-size>12884901888</virtual-size>
      <allocated estimated='true'>1530933248</allocated>
    </disk>
  </disks>
  <operatingsystem>
    <name>linux</name>
    <distro>rhel</distro>
    <osinfo>rhel8.2</osinfo>
    <arch>x86_64</arch>
    <major_version>8</major_version>
    <minor_version>2</minor_version>
    <package_format>rpm</package_format>
    <package_management>dnf</package_management>
    <product_name>Red Hat Enterprise Linux 8.2 (Ootpa)</product_name>
    <product_variant>unknown</product_variant>
    <root>/dev/vg_rhev-orange-03/lv_root</root>
    <mountpoints>
      <mountpoint dev='/dev/vg_rhev-orange-03/lv_root'>/</mountpoint>
      <mountpoint dev='/dev/sda1'>/boot</mountpoint>
    </mountpoints>
  </operatingsystem>
</v2v-inspection>
```

this is strange because in this vm we have only 1 disk `/sda`

@rwmjones what am i missing ? why i don't see `/` mounted using sda ?

rwmjones: > the PR we are looking for is in 9.5 [libguestfs/virt-v2v@4df3dcf](https://github.com/libguestfs/virt-v2v/commit/4df3dcf64da91645d20aa9f74cd04fa8985c4610)

That commit is present in RHEL 9.5, but not in 9.4-z.

> how do we use the rhel-9.5 rpms ?

You'll have to ask Arik about this, but I believe there was some problem with consuming RHEL 9.5 RPMs in MTV right now.

virt-v2v is shipped using the ordinary RHEL process and schedule.  RHEL 9.5 will ship in November [technically RH-only information, but an open secret].

If you cannot use the 9.5 RPMs early then you'll need to request this to be backported to 9.4-z, which starts with a Jira ticket again RHEL product, virt-v2v component.  Nothing happens without a Jira ticket.

Since 9.4-z is a heavyweight operation involving coordinating many people I'd suggest batching requests for backports together so we don't need to go through it more than once (more).

yaacov: @rwmjones i am trying to use `virt-v2v-inspector`

i get
```xml
 <mountpoints> <mountpoint dev='/dev/vg_rhev-orange-03/lv_root'>/</mountpoint>
 <mountpoint dev='/dev/sda1'>/boot</mountpoint> </mountpoints>
```
i expected one disk `sda` ounted at `/`

what am i missing ?

rwmjones: > @mnecas hi, so if we go with `virt-v2v-inspector` a. we have it already installed b. running it is "fast" compering to running `virt-v2v`
> 
> running it will look like: `virt-v2v-inspector -i libvirtxml /var/tmp/v2v/<vm name>.xml -o /var/tmp/v2v/<vm name>.inspect.xml`
> 
> and the output will include the root disk:
> 
> ```
>  <disks>
>    <disk index='0'>
>      <virtual-size>12884901888</virtual-size>
>      <allocated estimated='true'>1530933248</allocated>
>    </disk>
>  </disks>
>     <mountpoints>
>       <mountpoint dev='/dev/vg_rhev-orange-03/lv_root'>/</mountpoint>
>       <mountpoint dev='/dev/sda1'>/boot</mountpoint>
>     </mountpoints>
>   </operatingsystem>
> </v2v-inspection>
> ```
> 
> this is strange because in this vm we have only 1 disk `/sda`
> 
> @rwmjones what am i missing ? why i don't see `/` mounted using sda ?

This is correct isn't it?  It's showing there is one disk with virtual size 12GB, with about 1.5GB of data on it. With root `/` mounted on a logical volume, and `/boot` on the first partition of the disk.

`/dev/sda1` etc are actually libguestfs standard device names (https://libguestfs.org/guestfs.3.html#block-device-naming).

Inside the guest this might appear as `C:` or `/dev/vda1`.  (We actually have Linux guest device and Windows drive mapping information, but I can't remember if virt-v2v-inspector shows it).

yaacov: > Inside the guest this might appear as C: or /dev/vda1

we need the root disk to know on what disk to run `virt-customize` 
we don't care in this step what this disk is called on the guest, only how it's currently called in our pod, e.g. the file name that `virt-v2c` created in the `-os` dir

our question is how we know what `.img` to use from the files we get in the `-os` after `virt-v2v` finishes, i am missing something here, how do I get what disk include the `/etc` dir, so I can run `virt-customize` on ?

rwmjones: The disks are listed in order they were specified in the source XML.

However I think it'd be possible to put the source filename into the XML too if that helps.  (Virt-v2v's internal metadata model is here: https://github.com/libguestfs/virt-v2v/blob/master/lib/types.mli).  Starts with a Jira ticket (or a pull request if you fancy implementing it).

rwmjones: > how do I get what disk include the /etc dir, so I can run virt-customize on ?

I think actually there's also a misunderstanding here.  For virt-customize you should specify **all** disks of the guest (in order) using multiple `-a` options.  This is because you could have a guest which has, for example, a separate `/usr` from a second disk.  Virt-customize will sort this out if you give it all the disks, you don't have to worry about it.

yaacov: > For virt-customize you should specify all disks of the guest (in order) using multiple -a options.

WOW: 
ok, so we don't actually need to run `virt-v2v-inspector` , just look on all the `.img` files and add them with `-a`

@rwmjones last question ( i hope ) do i need to tell `virt-customize` somehow how to mount this disks, or it knows by inspecting the files ?

```basg
# is this enough ? 
# or i need something more ?  ( is the order important ?)
virt-customize -a disk-sda.img -a disk-sdb.img --run ...
```

rwmjones: virt-customize uses a thing called libguestfs inspection to examine the disks and determine how they would be arranged in the original guest (using clues from `/etc/fstab`, the Windows registry, and a big pile of heuristics), it then mounts up the filesystems automatically.  More here: https://libguestfs.org/guestfs-faq.1.html#why-does-virt-cat-only-work-on-a-real-vm-image-but-virt-df-works-on-any-disk-image

However you might still want to use virt-v2v-inspector (for other reasons, maybe not here).  If you want to know in advance how much space to allocate for the target disks after virt-v2v, or you need to know other information that virt-v2v normally only provides after conversion (like whether virtio-win drivers can be installed), then virt-v2v-inspector does all that without needing to do the copy.  Whether this is useful or not, I don't know, but that's what it's for.

mnecas: Thank you so much for clarification, helps a lot! I don't think that we need it in our use case right now but might need in future.

virt-v2v/cold/customize-image.go
```diff
@@ -0,0 +1,176 @@
...
+// This function looks for disk files in the `dir` directory that match the pattern `*-sd*`. It uses the `V2V_RootDisk`
+// environment variable to determine the disk name, falling back to the default disk name "sda" if the environment variable is not set or invalid.
+//
+// Arguments:
+//   - dir (string): The directory containing the disk files.
+//
+// Returns:
+//   - string: The full path to the root disk image.
+//   - error: An error if the root disk image is not found or if there is an issue reading the disk directory.
+func FindRootDiskImage(dir string) (string, error) {
```
yaacov, line 136: @mnecas we can remove it now ?

virt-v2v/cold/entrypoint.go
```diff
@@ -74,6 +81,52 @@ func main() {
...
+		if strings.Contains(operatingSystem, "win") {
+			rootDiskPath, err := FindRootDiskImage(DIR)
+			if err != nil {
+				fmt.Println("Error looking for root disk path:", err)
+				os.Exit(1)
+			} else {
+				fmt.Printf("Root disk path: %s\n", rootDiskPath)
+			}
```
yaacov, line 119: @mnecas we can remove this code now ?

mnecas: Yes! I rewrote so much because of and forgot to remove it :D

virt-v2v/cold/entrypoint.go
```diff
@@ -62,6 +63,12 @@ func main() {
 		fmt.Println("Error getting XML file:", err)
 		os.Exit(1)
 	}
+	// If needed, customize the VM
+	err = customizeVM(source, xmlFilePath)
```
yaacov, line 67: @mnecas @bkhizgiy heads up, `xmlFilePath` will be `yamlFilePath` see:
https://github.com/kubev2v/forklift/pull/778/files#diff-2bd31ec649de5f659c2c9a2f062e677dd9f4bde75f165d281f32ba98c3d9cc68

rwmjones:
`RunOnce` isn't a reliable way to run scripts at first boot I'm afraid :-(

I would stick to what virt-v2v / virt-customize already does for running firstboot, and if there's some issue with it, let me know what the issue actually is.

mnecas:
> `RunOnce` isn't a reliable way to run scripts at first boot I'm afraid :-(
> 
> I would stick to what virt-v2v / virt-customize already does for running firstboot, and if there's some issue with it, let me know what the issue actually is.

I had a much better experience with it than with the firstboot. The issues which I had with the firstboot were also inconsistent. Sometimes the `Get-Volume` would fail with:
Screenshot from 2024-08-19 10-59-33
Sometimes when I run `Get-Disks` I get no disks. I tried some delay thinking that some services/servers were not ready but every time after 1s the process got terminated after playing with it yesterday I gave up and tried the registries and worked quite well from your blog which helped a lot https://rwmj.wordpress.com/2010/04/29/tip-get-a-windows-vm-to-run-a-batch-file-at-boot/.

rwmjones:
If you're using virt-customize in RHEL 9 it's known to have these issues.  They are fixed upstream & in Fedora already.  No one has yet filed any Jira ticket to backport the fix to RHEL.

mnecas:
> If you're using virt-customize in RHEL 9 it's known to have these issues. They are fixed upstream & in Fedora already. No one has yet filed any Jira ticket to backport the fix to RHEL.

yeah, I had RHEL 9.4 same as with which we distribute the images d/s, which issues? More than happy to file the ticket to backport, we will need to make a list and do it all at once. But this might still take some time, do you think that the `RunOnce` could be at least used as a temporary WA, before we get these fixes d/s?

virt-v2v/cold/customize-image.go
```diff
@@ -0,0 +1,121 @@
...
+	windowsScriptsPath := filepath.Join(DIR, "scripts", "windows")
+	initPath := filepath.Join(windowsScriptsPath, "restore_config_init.bat")
+	restoreScriptPath := filepath.Join(windowsScriptsPath, "restore_config.ps1")
+
+	// Upload scripts to the windows
+	uploadScriptPath := fmt.Sprintf("%s:%s", restoreScriptPath, WIN_FIRSTBOOT_PATH)
+	uploadInitPath := fmt.Sprintf("%s:%s", initPath, WIN_FIRSTBOOT_PATH)
+
+	var extraArgs []string
+	extraArgs = append(extraArgs, getScriptArgs("upload", uploadScriptPath, uploadInitPath)...)
```
yaacov, line 48:
```suggestion
	// Get the list of .ps1 and .bat files in the directory
	files, err := os.ReadDir(windowsScriptsPath)
	if err != nil {
		fmt.Printf("Failed to read directory: %v\n", err)
		return
	}

	var extraArgs []string

	for _, file := range files {
		if !file.IsDir() && (strings.HasSuffix(file.Name(), ".ps1") || strings.HasSuffix(file.Name(), ".bat")) {
			// Construct the local file path and the upload path
			localFilePath := filepath.Join(windowsScriptsPath, file.Name())
			uploadPath := fmt.Sprintf("%s:%s", localFilePath, WIN_FIRSTBOOT_PATH)

			// Append the arguments
			extraArgs = append(extraArgs, getScriptArgs("upload", uploadPath)...)
		}
	}
```

This will upload all .ps1 files, no need to update when adding more scripts ( for example mounting script config map at the scripts path ) WDYT ?

rwmjones:
> > If you're using virt-customize in RHEL 9 it's known to have these issues. They are fixed upstream & in Fedora already. No one has yet filed any Jira ticket to backport the fix to RHEL.
> 
> yeah, I had RHEL 9.4 same as with which we distribute the images d/s, which issues? More than happy to file the ticket to backport, we will need to make a list and do it all at once. 

This upstream commit: https://github.com/libguestfs/guestfs-tools/commit/d7719d6954f48c3a395da7486bcf55fbac66ea07

I really dislike that we're doing development on a released version of RHEL. This risks our customers.  Fedora is the place to do development and integration.

> But this might still take some time, do you think that the `RunOnce` could be at least used as a temporary WA, before we get these fixes d/s?

`RunOnce` only runs when the user logs in (which could be never, or multiple times if multiple users log in).  It's for per-user customizations.

The larger question is whether we're trying to do a hack or whether we're trying to fix this properly.

rwmjones:
> IIUIC then @mnecas used firstboot, and did get logs, speak, the script got run. However, within the script Get-Disks did not return any results.
> 
> Thus I am not sure why the reboot fix would work.

We really ought to be sure that the reboot fix will work, as that's what we're proposing for fixing at least 2 other problems.  If it doesn't, that's serious.  Ming has tested it against about a dozen different versions of Windows.

So I would suggest testing virt-customize with the reboot fix before adding another uncertain [and - actually - known wrong] hack on top here.

virt-v2v/cold/customize-image.go
mnecas: I \was thinking about this but I wanted to keep it modular as we might want some scripts in different paths or overwrite some files. But definitely could visit this in future

virt-v2v/cold/entrypoint.go
mnecas: I'll make it in separate MR

virt-v2v/cold/BUILD.bazel
mnecas: I'll make it in separate MR

yaacov (approved):
I don't understand how the windows first boot work, otherwise looks good to me

virt-v2v/cold/customize-image.go
```diff
@@ -0,0 +1,94 @@
...
+	WIN_FIRSTBOOT_SCRIPTS_PATH = "/Program Files/Guestfs/Firstboot/scripts"
+)
+
+// CustomizeWindows customizes a windows disk image by uploading scripts.
+//
+// The function writes two bash scripts to the specified local tmp directory,
+// uploads them to the disk image using `virt-customize`.
+//
+// Arguments:
+//   - disk ([]string): The list of disk paths which should be customized
```
yaacov, line 25:
```suggestion
//   - disks ([]string): The list of disk paths which should be customized
```

virt-v2v/cold/entrypoint.go
```diff
@@ -73,6 +80,40 @@ func main() {
 		os.Exit(1)
 	}
 }
+func getVmDiskPaths(domain *OvaVmconfig) []string {
```
yaacov, line 83:
```suggestion

func getVmDiskPaths(domain *OvaVmconfig) []string {
```

yaacov (approved):
lgtm, nitpick typos

fabiand:
:rocket: :rainbow:

virt-v2v/cold/entrypoint.go
```diff
@@ -74,6 +81,41 @@ func main() {
...
+			resp = append(resp, disk.Source.File)
+		}
+	}
+	return resp
+}
+
+func customizeVM(source string, xmlFilePath string) error {
+	domain, err := GetDomainFromXml(xmlFilePath)
+	disks := getVmDiskPaths(domain)
+	if err != nil {
```
ahadas, line 97: the error should be checked before calling `getVmDiskPaths`, otherwise we may not get here..

virt-v2v/cold/entrypoint.go
```diff
@@ -62,6 +63,12 @@ func main() {
 		fmt.Println("Error getting XML file:", err)
 		os.Exit(1)
 	}
+	// If needed, customize the VM
```
ahadas, line 66: since only non-boot disks are set as offline, there's no point in going there in case the VM has a single disk, which is supposed to be a very common case...

PR #985, opened 2024-08-20
[MTV-595] Send mac to ip mappings for linux

yaacov:

**Issue:**
When migrating a simple RHEL8 VM from VMware 7 to OCPv 4.13 the name of the network interfaces changes and the static IP configuration for the VM no longer works

**What this PR do:**
This PR check if the OS is not Windows, in this case it collects the mac to static IP map, and send it to the `virt-v2v` pod.

Notes:
  - On linux we don't check for `origin=manual` because this information is missing on linux. 
  - We only send this information when user choose to force static IPs
  - The user must run the VM while MTV inventory server is running to create the mapping, IP information is added to the inventory MTV is collecting only when the VM is running, once the IP data is collected, user can power down the VM and MTV will persist this information.
  - If user choose to force static IP and mapping is missing, validation will fail.

**Why this change is safe:**
Our virt-v2v pod check if we have `V2V_staticIPs` and adds it to the `virt-v2v` CLI tool as `--mac` argument, on non Windows OS, `virt-v2v` tool ignore this CLI argument, see: https://libguestfs.org/virt-v2v.1.html

_... It is currently ignored for Linux guests since they do not have this problem. ..._ 

**Run example:**

```
Preparing virt-v2v
exec: /usr/bin/virt-v2v -v -x --root first -i libvirt -ic vpx://administrator%40vsphere.local@rhev-node-05.rdu2.scalelab.redhat.com/Datacenter/host/auto-test/MTV/f01-h06-000-r640.rdu2.scalelab.redhat.com?no_verify=1 -o local -os /var/tmp/v2v -ip /etc/secret/secretKey --mac 00:50:56:83:f9:7f:ip:172.29.97.122,,32 -it vddk -io vddk-libdir=/opt/vmware-vix-disklib-distrib -io vddk-thumbprint=31:14:EB:9E:F1:78:68:10:A5:78:D1:A7:DF:BB:54:B7:1B:91:9F:30 -- yzamir-03
virt-v2v monitoring: Setting up prometheus endpoint :2112/metrics
virt-v2v monitoring: Prometheus progress counter registered.
info: virt-v2v: virt-v2v 2.5.6rhel=9,release=3.el9 (x86_64)
info: libvirt version: 10.5.0
```

### DIFF
--- a/pkg/controller/plan/adapter/vsphere/builder.go
+++ b/pkg/controller/plan/adapter/vsphere/builder.go
@@ -247,13 +247,14 @@ func (r *Builder) PodEnvironment(vmRef ref.Ref, sourceSecret *core.Secret) (env 
 }
 
 func (r *Builder) mapMacStaticIps(vm *model.VM) (ipMap string, err error) {
-	if !isWindows(vm) {
-		return "", nil
-	}
+	// on windows machines we check if the interface origin is manual
+	// on linux we collect all networks.
+	isWindowsFlag := isWindows(vm)
+
 	configurations := []string{}
 	gatewaySet := false
 	for _, guestNetwork := range vm.GuestNetworks {
-		if guestNetwork.Origin == string(types.NetIpConfigInfoIpAddressOriginManual) {
+		if !isWindowsFlag || guestNetwork.Origin == string(types.NetIpConfigInfoIpAddressOriginManual) {
 			gateway := ""
 			if !gatewaySet {
 				isIpv4 := net.IP.To4(net.ParseIP(guestNetwork.IP)) != nil
@@ -279,7 +280,10 @@ func (r *Builder) mapMacStaticIps(vm *model.VM) (ipMap string, err error) {
 				}
 			}
 			dnsString := strings.Join(guestNetwork.DNS, ",")
-			configurations = append(configurations, fmt.Sprintf("%s:ip:%s,%s,%d,%s", guestNetwork.MAC, guestNetwork.IP, gateway, guestNetwork.PrefixLength, dnsString))
+			configurationString := fmt.Sprintf("%s:ip:%s,%s,%d,%s", guestNetwork.MAC, guestNetwork.IP, gateway, guestNetwork.PrefixLength, dnsString)
+
+			// if DNS is "", we get configurationString with trailing comma, use TrimSuffix to remove it.
+			configurations = append(configurations, strings.TrimSuffix(configurationString, ","))
 		}
 	}
 	return strings.Join(configurations, "_"), nil
--- a/pkg/controller/plan/adapter/vsphere/builder_test.go
+++ b/pkg/controller/plan/adapter/vsphere/builder_test.go
@@ -67,8 +67,8 @@ var _ = Describe("vSphere builder", func() {
 			},
 		}, "00:50:56:83:25:47:ip:172.29.3.193,172.29.3.1,16,8.8.8.8_00:50:56:83:25:47:ip:fe80::5da:b7a5:e0a2:a097,,64,fec0:0:0:ffff::1,fec0:0:0:ffff::2,fec0:0:0:ffff::3"),
 		Entry("non-static ip", &model.VM{GuestID: "windows9Guest", GuestNetworks: []vsphere.GuestNetwork{{MAC: "00:50:56:83:25:47", IP: "172.29.3.193", Origin: string(types.NetIpConfigInfoIpAddressOriginDhcp)}}}, ""),
-		Entry("non windows vm", &model.VM{GuestID: "other", GuestNetworks: []vsphere.GuestNetwork{{MAC: "00:50:56:83:25:47", IP: "172.29.3.193", Origin: ManualOrigin}}}, ""),
-		Entry("no OS vm", &model.VM{GuestNetworks: []vsphere.GuestNetwork{{MAC: "00:50:56:83:25:47", IP: "172.29.3.193", Origin: ManualOrigin}}}, ""),
+		Entry("non windows vm", &model.VM{GuestID: "other", GuestNetworks: []vsphere.GuestNetwork{{MAC: "00:50:56:83:25:47", IP: "172.29.3.193", Origin: ManualOrigin}}}, "00:50:56:83:25:47:ip:172.29.3.193,,0"),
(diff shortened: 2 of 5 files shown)


yaacov:
@liranr23 , @mnecas please review

pkg/controller/plan/adapter/vsphere/builder_test.go
```diff
@@ -67,8 +67,8 @@ var _ = Describe("vSphere builder", func() {
 			},
 		}, "00:50:56:83:25:47:ip:172.29.3.193,172.29.3.1,16,8.8.8.8_00:50:56:83:25:47:ip:fe80::5da:b7a5:e0a2:a097,,64,fec0:0:0:ffff::1,fec0:0:0:ffff::2,fec0:0:0:ffff::3"),
 		Entry("non-static ip", &model.VM{GuestID: "windows9Guest", GuestNetworks: []vsphere.GuestNetwork{{MAC: "00:50:56:83:25:47", IP: "172.29.3.193", Origin: string(types.NetIpConfigInfoIpAddressOriginDhcp)}}}, ""),
-		Entry("non windows vm", &model.VM{GuestID: "other", GuestNetworks: []vsphere.GuestNetwork{{MAC: "00:50:56:83:25:47", IP: "172.29.3.193", Origin: ManualOrigin}}}, ""),
-		Entry("no OS vm", &model.VM{GuestNetworks: []vsphere.GuestNetwork{{MAC: "00:50:56:83:25:47", IP: "172.29.3.193", Origin: ManualOrigin}}}, ""),
+		Entry("non windows vm", &model.VM{GuestID: "other", GuestNetworks: []vsphere.GuestNetwork{{MAC: "00:50:56:83:25:47", IP: "172.29.3.193", Origin: ManualOrigin}}}, "00:50:56:83:25:47:ip:172.29.3.193,,0"),
+		Entry("no OS vm", &model.VM{GuestNetworks: []vsphere.GuestNetwork{{MAC: "00:50:56:83:25:47", IP: "172.29.3.193", Origin: ManualOrigin}}}, "00:50:56:83:25:47:ip:172.29.3.193,,0"),
```
liranr23, line 71: can you add a case for missing DNS?

liranr23 (commented):
Can we prevent the vNIC renaming on the first place?

pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -247,13 +247,14 @@ func (r *Builder) PodEnvironment(vmRef ref.Ref, sourceSecret *core.Secret) (env
...
-	}
+	// on windows machines we check if the interface origin is manual
+	// on linux we collect all networks.
+	isWindowsFlag := isWindows(vm)
+
 	configurations := []string{}
 	gatewaySet := false
 	for _, guestNetwork := range vm.GuestNetworks {
-		if guestNetwork.Origin == string(types.NetIpConfigInfoIpAddressOriginManual) {
+		if !isWindowsFlag || guestNetwork.Origin == string(types.NetIpConfigInfoIpAddressOriginManual) {
```
bkhizgiy, line 257: why we need to save it to var and not simply run `!isWindows(vm) || ....`?

yaacov: hmm, fixing :+1:

pkg/controller/plan/adapter/vsphere/builder_test.go
yaacov: sure

pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -279,7 +280,13 @@ func (r *Builder) mapMacStaticIps(vm *model.VM) (ipMap string, err error) {
 				}
 			}
 			dnsString := strings.Join(guestNetwork.DNS, ",")
-			configurations = append(configurations, fmt.Sprintf("%s:ip:%s,%s,%d,%s", guestNetwork.MAC, guestNetwork.IP, gateway, guestNetwork.PrefixLength, dnsString))
+
+			// if DNS is "", don't add it
+			if dnsString == "" {
+				configurations = append(configurations, fmt.Sprintf("%s:ip:%s,%s,%d", guestNetwork.MAC, guestNetwork.IP, gateway, guestNetwork.PrefixLength))
```
bkhizgiy, line 286: Do we need it twice? if the DNS string is empty wouldn't it simply omitted in the end? if not how about extracting it to a func to reduce code duplication.

yaacov: the template is:
```go
fmt.Sprintf("%s:ip:%s,%s,%d,%s", guestNetwork.MAC, guestNetwork.IP, gateway, guestNetwork.PrefixLength, dnsString)
```

so if `dnsString == ""` we get a string that looks like `00:50:56:83:25:47:ip:172.29.3.193,172.29.3.1,16,`

virt-v2v does not like `,` at the end of the `--mac` command line arg, we can also clean it up by
`str = strings.TrimSuffix(str, ",")
`
not sure what is nicer ?

pkg/controller/plan/adapter/vsphere/builder_test.go
yaacov: added a test case for single static ip, with no DNS

pkg/controller/plan/adapter/vsphere/builder.go
yaacov: simplified the for loop, by calling `isWindows(vm)` inside the loop

pkg/controller/plan/adapter/vsphere/builder.go
bkhizgiy: thanks for clarifying, I personally find the trim nicer, but its really a meter of style so i'm fine with both.

pkg/controller/plan/adapter/vsphere/builder.go
yaacov: oyy, it's calling a function inside a loop, vs using one variable ... reverting the fix ...

pkg/controller/plan/adapter/vsphere/builder.go
yaacov: changed to use trim :+1:

yaacov:
> Can we prevent the vNIC renaming on the first place?

@liranr23 Yes, it requires login as root to the guest vm before the migration install `nmstate` and create `/etc/nmstate/pin-nic.yml ` as described here https://issues.redhat.com/browse/MTV-595?focusedId=25357142&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-25357142

We did not use this option because the user did not want us to log in as root to the guest vm, and install things :-(

liranr23 (approved):

PR #991, opened 2024-08-21
virt-v2v: Replace the Get-Disk and Set-Disk with diskpart

mnecas:
Issue: The Get-Disk and Set-Disk do not support dynamic disks. https://learn.microsoft.com/en-us/powershell/module/storage/get-disk?view=windowsserver2022-ps

### DIFF
--- a/virt-v2v/cold/scripts/windows/9999-restore_config.ps1
+++ b/virt-v2v/cold/scripts/windows/9999-restore_config.ps1
@@ -6,10 +6,28 @@ Write-Output ('') >> $logFile
 
 # script section to re-enable all offline drives
 # Re-enable all offline drives
-Write-Output ('Re-enabling all offline drives') >> $logFile
-Get-Disk | Where { $_.FriendlyName -like '*VirtIO*' } | % {
-     Write-Output ('  - ' + $_.Number + ': ' + $_.FriendlyName + '(' + [math]::Round($_.Size/1GB,2) + 'GB)') >> $logFile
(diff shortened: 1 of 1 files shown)


yaacov (approved):
I don't speak ps1, lgtm

virt-v2v/cold/scripts/windows/9999-restore_config.ps1
```diff
@@ -6,10 +6,28 @@ Write-Output ('') >> $logFile
...
+    # Iterate through each matching disk
+    foreach ($disk in $disks) {
+        $diskNumber = $disk.Index  # Get the disk index which corresponds to the disk number
+
+        # Create a diskpart script to set the disk online
+        $diskpartScript = @"
+select disk $diskNumber
+online disk
+attributes disk clear readonly
+"@
```
fabiand, line 26: This feels quie risky to me.

Take an index from one tool/api, and use it in a different tool/api.

This is just waiting for indexes to be out of sync for whatever reason.

fabiand (changes requested):
Mostly requetsing change to avoid that we accidentally merge it.

virt-v2v/cold/scripts/windows/9999-restore_config.ps1
mnecas: 100% agree and I would like to not use the diskpart

fabiand:
Can we close this?

PR #992, opened 2024-08-22
[MTV-595] Add checks before running virt-v2v customize step

yaacov:
Ref: 
https://issues.redhat.com/browse/MTV-595

Addressing:
https://github.com/kubev2v/forklift/pull/984#pullrequestreview-2249952852

**Issue:**
We run the virt-v2v customization step in cases we know it is not needed

**Fix:**
Add checks, and exit the customization script when we identify that it's not neededL
  - We can't parse .xml ( or .yaml in future ) virt-v2v output
  - We don't recognize the OS name
  - We did not find any disks to customize

### DIFF
--- a/virt-v2v/cold/entrypoint.go
+++ b/virt-v2v/cold/entrypoint.go
@@ -63,7 +63,7 @@ func main() {
 		fmt.Println("Error getting XML file:", err)
 		os.Exit(1)
 	}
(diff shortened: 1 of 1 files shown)


yaacov:
@mnecas please review

mnecas (approved):
LGTM

PR #996, opened 2024-08-27
Port Virt v2v warm migration image to rhel9

jonner:
This is an initial patch series that switches warm migration to use the same rhel9-based image as cold migration does.

It uses an environment variable `V2V_inPlace` to determine whether to use the 'cold' or 'warm' migration workflow (I called it `V2V_inPlace` rather than something like `V2V_warmMigration` because this 'warm' image is used for more than just warm migrations. For example, cold migration to a remote cluster also uses the same image).

The current patch set does not yet add any new features. The goal is to maintain feature parity with the old image while moving to rhel9. Any new features for 'warm' migration can follow in future patch series.

Note that this PR is a draft at the moment because I need to reinstall my cluster and do some additional testing on it before I'm sure that it's working as expected.


pkg/settings/migration.go
```diff
@@ -94,12 +93,13 @@ func (r *Migration) Load() (err error) {
 		return liberr.Wrap(err)
 	}
 	if virtV2vImage, ok := os.LookupEnv(VirtV2vImage); ok {
-		if cold, warm, found := strings.Cut(virtV2vImage, "|"); found {
-			r.VirtV2vImageCold = cold
-			r.VirtV2vImageWarm = warm
+		if img, _, found := strings.Cut(virtV2vImage, "|"); found {
+			// previously, we used different images for cold and warm migration and these
+			// could be specified separately as "COLD_IMAGE|WARM_IMAGE". Accept this syntax,
+			// but just ignore the second image
```
mnecas, line 99: I see that you also removed the warm from `VIRT_V2V_IMAGE`. Think this is no longer needed.

pkg/apis/forklift/v1beta1/plan.go
```diff
@@ -93,9 +93,11 @@ type Plan struct {
 	Referenced `json:"-"`
 }
 
-// Decide if the plan should use EL9 image for virt-v2v conversion.
-// in combination with specified destination provider.
-func (p *Plan) VSphereUsesEl9VirtV2v() (bool, error) {
+// If the plan calls for the vm to be cold migrated to the local cluster, we can
+// just use virt-v2v directly to convert the vm while copying data over. In other
+// cases, we use CDI to transfer disks to the destination cluster and then use
+// virt-v2v-in-place to convert these disks after cutover.
```
mnecas, line 99: FYI @yaacov once adapting the customize for warm migration we will need to adapt it to this flow. So we would modify the image after the cutover and the `virt-v2v-in-place`.

pkg/forklift-api/webhooks/validating-webhook/admitters/plan-admitter.go
```diff
@@ -79,32 +79,6 @@ func (admitter *PlanAdmitter) validateStorage() error {
 	return nil
 }
 
-func (admitter *PlanAdmitter) validateVDDK() error {
```
mnecas, line 82: Why have you removed the vddk validation?

pkg/apis/forklift/v1beta1/plan.go
mnecas: @jonner Could you please add this to the doc https://github.com/kubev2v/forklift/blob/main/docs/use-of-virt-v2v-in-forklift.md?

mnecas (commented):
wow, amazing job! thank you for the PR!
I went quickly over the PR and looks great, I have some comments but knowing this is a draft might not be relevant. I'll take a look again more deeply once the PR is ready

pkg/apis/forklift/v1beta1/plan.go
mnecas: @jonner I have added a `virt-customize` to the warm migration flow, we should add it on top of these changes. I can take over those changes but we will start using the `virt-v2v -o local`. I don't see any possible output. Can I somehow query the guest using the libguestfs tools to get the domain file?
PTAL https://github.com/kubev2v/forklift/pull/1012

virt-v2v/entrypoint.go
```diff
@@ -150,14 +114,21 @@ func customizeVM(source string, xmlFilePath string) error {
...
-		os.Exit(1)
+func convertVirtV2vInPlace() error {
+	args := []string{"-v", "-x", "-i", "libvirtxml"}
+	args = append(args, "--root")
+	if val, found := os.LookupEnv("V2V_RootDisk"); found {
+		args = append(args, val)
+	} else {
+		args = append(args, "first")
 	}
+	args = append(args, "/mnt/v2v/input.xml")
```
mnecas, line 125: I'm not sure if we have the domain available, you might need to specify all disks `virt-v2v-in-place -i disk IMG1 IMG2 ...`

pkg/apis/forklift/v1beta1/plan.go
mnecas: > @jonner I have added a virt-customize to the warm migration flow, we should add it on top of these changes. I can take over those changes but we will start using the virt-v2v -o local. I don't see any possible output. Can I somehow query the guest using the libguestfs tools to get the domain file?
PTAL https://github.com/kubev2v/forklift/pull/1012

please ignore that comment, did not noticed that you are using the mounted domain instead of the /var/tmp/v2v

mnecas: > hmm, I guess I can try to add something, but it seems that this document is significantly out of date already as it describes changes that are being made in forklift 2.4, and we're now on 2.6.x. So it seems that this document needs significant overhaul to be really useful, and it's not as simple as just adding some new information about the virt-v2v-in-place command...

yeah that was just NP don't need to do that, I'll need to update that doc completely

mnecas:
I was able to test it and worked as expected, still have on NP about the settings

pkg/controller/plan/migration.go
```diff
@@ -1647,12 +1647,14 @@ func (r *Migration) updateConversionProgress(vm *plan.VMStatus, step *plan.Step)
...
-		el9, err := r.Context.Plan.VSphereUsesEl9VirtV2v()
+		coldLocal, err := r.Context.Plan.VSphereColdLocal()
 		switch {
 		case err != nil:
 			return liberr.Wrap(err)
-		case el9:
-			if err := r.updateConversionProgressEl9(pod, step); err != nil {
+		case coldLocal:
+			// FIXME: will this work for in-place conversions now that it also uses
+			// virt-v2v-monitor when executing virt-v2v-in-place?
```
mnecas, line 1656: Nope different flow, as we get the progress from CNV CDI instead of monitoring the v2v output. 
I'll remove the comment

mnecas (approved):
LGTM

PR #998, opened 2024-08-28
[MTV-595] Refactor run virt-customize for linux on any non windows

yaacov:

Issue:
The original issue was reported on RHEL8, after discussions with reporter we found out that this fix may be helpful for other distributions.

Concerns:
The fix will work for machines that use network manager, but will no nothing in other cases.

### DIFF
--- a/virt-v2v/cold/entrypoint.go
+++ b/virt-v2v/cold/entrypoint.go
@@ -135,11 +135,11 @@ func customizeVM(source string, xmlFilePath string) error {
 			}
 		}
 
-		// RHEL
-		if strings.Contains(operatingSystem, "rhel") {
+		// Linux
+		if !strings.Contains(operatingSystem, "win") {
 			t := EmbedTool{filesystem: &scriptFS}
 
-			err = CustomizeRHEL(CustomizeDomainExec, disks, DIR, &t)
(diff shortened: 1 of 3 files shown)


virt-v2v/cold/entrypoint.go
```diff
@@ -135,11 +135,11 @@ func customizeVM(source string, xmlFilePath string) error {
 			}
 		}
 
-		// RHEL
-		if strings.Contains(operatingSystem, "rhel") {
+		// Linux
+		if !strings.Contains(operatingSystem, "win") {
```
fabiand, line 139: Could we actually check for the `family == 'linux'` here.

osinfo-db has `family`, see: https://gitlab.com/libosinfo/osinfo-db/-/blob/main/data/os/fedoraproject.org/fedora-38.xml.in?ref_type=heads#L

fabiand: … with that approach we don't risk to be run on i.e. freebsd

yaacov: we use the xml output of virt-v2v, this information is missing, IMHO it's safe to try and run on any machine that is not identifying as windows, the script test for udev script and will not override if exist, and will fail gracefully if it can't find the network manager configuration scripts.

fabiand (approved):

PR #999, opened 2024-08-28
MTV-1354 | virt-v2v add pod anti affinity to the guest conversion pod

mnecas:
Issue: Right now we leave the placement of the conversion pod up to the kubernetes scheduler. But it can cause issues in large scale enviroments as the scheduler can place all migration pods on one node. This can cause the node to run out of the resources and can slow down the migration.

Fix: Add pod anti affinity which will distribute the migration pods evenly across the nodes.

### DIFF
--- a/pkg/controller/plan/kubevirt.go
+++ b/pkg/controller/plan/kubevirt.go
@@ -24,6 +24,7 @@ import (
 	"github.com/openshift/library-go/pkg/template/templateprocessing"
 	batch "k8s.io/api/batch/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/conversion"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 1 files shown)


yaacov (approved):
LGTM

pkg/controller/plan/kubevirt.go
```diff
@@ -1744,6 +1745,29 @@ func (r *KubeVirt) guestConversionPod(vm *plan.VMStatus, vmVolumes []cnv.Volume,
...
+				PodAntiAffinity: &core.PodAntiAffinity{
+					PreferredDuringSchedulingIgnoredDuringExecution: []core.WeightedPodAffinityTerm{
+						{
+							Weight: 100,
+							PodAffinityTerm: core.PodAffinityTerm{
+								TopologyKey: "kubernetes.io/hostname",
+								LabelSelector: &metav1.LabelSelector{
+									MatchExpressions: []metav1.LabelSelectorRequirement{
+										{
+											Key: kApp,
```
yaacov, line 1758: kApp - sounds starange, is this the curect key ?

yaacov: maybe `app` ?

mnecas: The `kApp` is a variable which is set to the pod in `conversionLabels` as a key of label

yaacov: https://github.com/kubev2v/forklift/blob/3ad243d6caa4fac32f6ea18b1158241794701bb8/pkg/controller/plan/kubevirt.go#L2235

mnecas:
Just got info from the scale team that it helped distribute the migration pods across all nodes.

PR #1002, opened 2024-08-30
Dashify network utils script

yaacov:
**Issue:**
Our network config script is writen for `bash` and will fail for distributions that link `/bin/sh` to `dash`

**What happen:**
When running the scripts on `ubuntu` we get a syntex error:
```bash
dash virt-v2v/cold/scripts/rhel/run/network_config_util_test
virt-v2v/cold/scripts/rhel/run/network_config_util_test: 31: 
/github.com/kubev2v/forklift/virt-v2v/cold/scripts/rhel/run/network_config_util.sh: Syntax error: 
"(" unexpected (expecting "then")
```

**What should happen:**
```bash
dash virt-v2v/cold/scripts/rhel/run/network_config_util_test 

Test output:
SUBSYSTEM=="net",ACTION=="add",ATTR{address}=="aa:bb:cc:dd:ee:ff",NAME="eth0"
SUBSYSTEM=="net",ACTION=="add",ATTR{address}=="aa:bb:cc:dd:ee:fe",NAME="eth3"
All tests passed successfully.
```

**Fix:**
Dashifying the code to run on Posix `sh`, so it will work on both `bash` and `dash`

### DIFF
--- a/virt-v2v/cold/scripts/rhel/run/network_config_util.sh
+++ b/virt-v2v/cold/scripts/rhel/run/network_config_util.sh
@@ -28,16 +28,16 @@ log() {
 extract_mac_ip() {
     S_HW=""
     S_IP=""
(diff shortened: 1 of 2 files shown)


mnecas (approved):
lgtm

PR #1006, opened 2024-08-30
MTV-1354 | Ignore namespaces when choosing the v2v pod on the nodes

mnecas:
Issue: When a user has a plans across many namespaces the v2v anti-affinity does not look for the other pods in different namespaces when using the anti-affinity.

Fix: This PR disables the namespace selector so it looks across whole cluster for the pods and distributes the pods evenly.


mnecas:
This change improves the efficiency of the migrations with many different namespaces. It's done by distributing the migration pods evenly across all nodes in the cluster.

Here is an example before and after the change where you can see the network throughput doubled from 200 MBps to 400 MBps
image

yaacov (approved):

PR #1007, opened 2024-09-01
[MTV-595] Add a flush disk step after virt-customize in virt-v2v pod

yaacov:
Issue:
`virt-customize` uses regular disk calls, `kubevirt` execute `qemu` using `cache.direct=true` to access disks.
In cases where the disk image if not flushed after running `virt-customize`, `kubevirt` may use currpted disk image file.

Fix:
Flush all the disk files after using `virt-customize`


fabiand:
Can be closed, the issue was different

PR #1008, opened 2024-09-02
[MTV-595] Add --format raw to virt-customize call

yaacov:
Issue:
When running `virt-customize` we get a warning:
```
WARNING: Image format was not specified for '/dev/block0' and probing guessed raw.
         Automatically detecting the format is dangerous for raw images, write operations on block 0 will be restricted.
         Specify the 'raw' format explicitly to remove the restrictions.
\x1bc\x1b[?7l\x1b[2J\x1b[0mSeaBIOS (version 1.16.3-2.el9)
```

Fix:
Specifying the format explicitly as raw


mnecas (approved):
LGTM

PR #1009, opened 2024-09-02
DO NOT MERGE - linux: A generic approach to try to keep the ifname if…

fabiand:
… possible

This works by looking in ifcfg, network manager to find ifname/macaddr mappings, and then write udev files for those.


fabiand:
Closing, because we will go with a different approach.

PR #1010, opened 2024-09-02
Remove double quates from network config files

yaacov:
**Issue:**
If network configuration file has quotes the name, this quotes will be transferred to the udev config file.

**Fix:**
Remove quotes. 

**Tests:**
`shellcheck virt-v2v/cold/scripts/rhel/run/network_config_util.sh` - pass

``` sh
bash virt-v2v/cold/scripts/rhel/run/network_config_util_test 

Test output:
SUBSYSTEM=="net",ACTION=="add",ATTR{address}=="aa:bb:cc:dd:ee:ff",NAME="eth0"
SUBSYSTEM=="net",ACTION=="add",ATTR{address}=="aa:bb:cc:dd:ee:fe",NAME="eth3"
All tests passed successfully.
```

### DIFF
--- a/virt-v2v/cold/scripts/rhel/run/network_config_util.sh
+++ b/virt-v2v/cold/scripts/rhel/run/network_config_util.sh
@@ -18,6 +18,11 @@ if [ -f "$UDEV_RULES_FILE" ]; then
     exit 0
 fi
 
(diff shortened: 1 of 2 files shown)


mnecas (approved):
LGMT!

PR #1013, opened 2024-09-03
MTV-1377 | Fix the auth to return 401 instead of 403

mnecas:
### First issue
Issue: When we request the inventory web endpoint the auth functions
allows the user to get the endpoint even with the incorrect token.

Fix: Change the permit function logic so we do not set the auth=true.

### Second issue
Issue: Right now the request with an invalid token returns 403 instead of
the expected 401.

Fix: Change the `permit` function to return the status code instead of
Boolean.


yaacov (approved):
LGTM

PR #1017, opened 2024-09-04
MTV-1388 | Add dynamic way to specify the virt-customize

mnecas:
### Issue:
Right now all the run/firstboot scripts are located inside the conversion pod. To get new scripts to the users we need some alternative way to allow users to specify the scripts themselves. This will speed up the migration process as the users won't need to wait for the build and release of the patch. Additionally, the users themselves can tweak the scripts depending on their needs.

### Design:
We have two options use the config maps, mount it to the container and read the mounted directory depending on their name we do virt-customize with them. An alternative solution would be to create CRD with all configurations such as `action`, `path` etc. I have chosen first as this is needed as soon as possible and configmaps are faster to implement.

### User flow:
1. **Create configmap**: 
The user needs to create a config map inside the namespace where they want to migrate the VM. The config map needs to be named  `forklift-virt-customize`.
2. **Populate configmap key and value**: 
Inside the configmap user can specify data with key/value definition. The value is the script itself which they want to run. The key needs to follow regex `([0-9]+_win_firstboot(([\w\-]*).ps1))$` or `^([0-9]+_linux_(run|firstboot)(([\w\-]*).sh))$` depending on where the customer wants the script to run. 
For example `00_win_firstboot_test.ps1` will specify that the data should be interpreted as PowerShell script which should start at boot. Alternatively, the Linux option has not only the `firstboot` but also the `run` option. This will be applied on the VM after virt-v2v conversion, but before the VM is started. 
_Note:_ The number in the beginning of the key sets the order.

### Conversion pod flow:
1. The `forklift-controller` checks if there is a config map inside the namespace. If the config map is present it is automatically mounted to the conversion pod at `/mnt/dynamic_scripts`.
2. The conversion pod checks if the `/mnt/dynamic_scripts` directory is present and if it is, it checks the files and their regexes. 
3. Depending on the VM operating system it will use different regex. From the filename, the conversion pod determines the action required on the `virt-customize` step.

### Additional changes:
- I have moved the static variables into a single file as we are using more and more and it's hard to keep track.
- New batch file which will run all PowerShell scripts inside the windows first boot scripts dir.
- The users can additionally also edit the virt customize config map name in the controller using the variable in controller  `virt_customize_configmap_name`


mnecas:
Tested config map:
image

Result in the VM:
image

operator/roles/forkliftcontroller/defaults/main.yml
```diff
@@ -22,6 +22,7 @@ controller_configmap_name: "{{ controller_service_name }}-config"
 controller_service_name: "{{ app_name }}-controller"
 ovirt_osmap_configmap_name: "forklift-ovirt-osmap"
 vsphere_osmap_configmap_name: "forklift-vsphere-osmap"
+virt_customize_configmap_name: "{{ app_name }}-virt-customize"
```
yaacov, line 25: maybe calling it `forklift-virt-customize `like the osmaps makes more sense because it will make the name deterministic, we will not need to look for different config file depending on the deployment, it will be the same name everywhere, WDYT ?

pkg/controller/plan/kubevirt.go
```diff
@@ -2058,6 +2083,33 @@ func (r *KubeVirt) libvirtDomain(vmCr *VirtualMachine, pvcs []*core.PersistentVo
...
 }
 
+func (r *KubeVirt) findConfigMapInNamespace(name string, namespace string) (configMap *core.ConfigMap, exists bool, err error) {
+	if err != nil {
+		return
+	}
+	list := &core.ConfigMapList{}
+	nameField := "metadata.name"
+	namespaceField := "metadata.namespace"
+	err = r.Destination.Client.List(
```
yaacov, line 2093: do we expect more then one config map, maybe using `r.Destination.Client.Get` makes more sense here ?

virt-v2v/pkg/customize/rhel.go
```diff
@@ -134,3 +119,19 @@ func addLuksKeysToCustomize(extraArgs *[]string) error {
 
 	return nil
 }
+
+func addRhelDynamicScripts(extraArgs *[]string, dir string) error {
+	dynamicScripts, err := getScriptsWithRegex(dir, global.LINUX_DYNAMIC_REGEX)
```
yaacov, line 124: each time we add scripts we need to sort them, so they will run in order, `10-dothing.sh `before `20-something.sh` i am missing where we do it

mnecas: It is already sorted in the filesystem so when we do the `os.ReadDir` we get it in the order

yaacov (approved):
LGTM

jsakil14:
> Tested config map: image
> 
> Result in the VM: image

Could you give an example of a windows interpretation: I want to add this as a CM and a script to enable all drives online after migration - I was looking for options using hooks , but documentation is not clear - so want to try this approach.

do we have to specify to use powershell?

vmtools is not getting uninstalled - attaching logs 

[vmtools.txt](https://github.com/user-attachments/files/17360795/vmtools.txt)


[conversion_pod.log](https://github.com/user-attachments/files/173
[uninstall-vmtools.log](https://github.com/user-attachments/files/17361476/uninstall-vmtools.log)
61475/conversion_pod.log)

 I created this CM - 

kg cm forklift-virt-customize   -o yaml 
apiVersion: v1
data:
  00_win_firstboot_test.ps1: tasklist > c:/temp/list.txt
kind: ConfigMap
metadata:
  creationTimestamp: "2024-10-14T07:56:56Z"
  name: forklift-virt-customize
  namespace: forklift-migration
  resourceVersion: "42578413"
  uid: b644a962-132a-4e54-9985-74e4086b2fb2

image @mnecas

mnecas:
Hi @jsakil14,
the scripts to enable the drivers are already implemented in the latest forklift
please rebuild the virt-v2v+controller image and apply it to your operator.
https://github.com/kubev2v/forklift/blob/main/virt-v2v/pkg/customize/scripts/windows/9999-restore_config.ps1
I need to do a forklift release, I'll try to make 2.7.2.

The config looks correct which version of forklift are you using?

jsakil14:
> * The users can additionally also edit the virt customize config map name in the controller using the variable in controller  `virt_customize_configmap_name`

Could you elaborate on this ,I delpoyed the operator from olm - should I add this env variable to the froklift-controller CR? or just the csv? my operator cluster is different from my target cluster - so I created a config map in tag

```
apiVersion: v1
data:
  00_win_firstboot_test.ps1:
    "Get-Disk | Where-Object { \n    $_.FriendlyName -like
    '*VirtIO*' -and !$_.IsSystem } | ForEach-Object {\n    Write-Host ('  - ' + $_.Number
    + ': ' + $_.FriendlyName + ' (' + [math]::Round($_.Size/1GB, 2) + 'GB)')\n    $_
    | Set-Disk -IsOffline $false\n    $_ | Set-Disk -IsReadOnly $false }\n\nfunction
    Get-VMwareToolsInstallerID {\n    foreach ($item in Get-ChildItem Registry::HKEY_CLASSES_ROOT\\Installer\\Products)
    {\n        if ($item.GetValue('ProductName') -eq 'VMware Tools') {\n            return
    @{\n                reg_id = $item.PSChildName;\n                msi_id = [Regex]::Match($item.GetValue('ProductIcon'),
    '(?<={)(.*?)(?=})').Value\n            }\n        }\n    }\n}\n\n$vmware_tools_ids
    = Get-VMwareToolsInstallerID\n\n$reg_targets = @(\n    \"Registry::HKEY_CLASSES_ROOT\\Installer\\Features\\\",\n
    \   \"Registry::HKEY_CLASSES_ROOT\\Installer\\Products\\\",\n    \"HKLM:\\SOFTWARE\\Classes\\Installer\\Features\\\",\n
    \   \"HKLM:\\SOFTWARE\\Classes\\Installer\\Products\\\",\n    \"HKLM:\\SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Installer\\UserData\\S-1-5-18\\Products\\\"\n)\n\n$VMware_Tools_Directory
    = \"C:\\Program Files\\VMware\"\n$VMware_Common_Directory = \"C:\\Program Files\\Common
    Files\\VMware\"\n\n$targets = @()\n\nif ($vmware_tools_ids) {\n    foreach ($item
    in $reg_targets) {\n        $targets += $item + $vmware_tools_ids.reg_id\n    }\n
    \   $targets += \"HKLM:\\SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Uninstall\\{$($vmware_tools_ids.msi_id)}\"\n}\n\nif
    ([Environment]::OSVersion.Version.Major -lt 10) {\n    $targets += \"HKCR:\\CLSID\\{D86ADE52-C4D9-4B98-AA0D-9B0C7F1EBBC8}\"\n
    \   $targets += \"HKLM:\\SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Uninstall\\{9709436B-5A41-4946-8BE7-2AA433CAF108}\"\n
    \   $targets += \"HKLM:\\SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Uninstall\\{FE2F6A2C-196E-4210-9C04-2B1BC21F07EF}\"\n}\n\nif
    (Test-Path \"HKLM:\\SOFTWARE\\VMware, Inc.\") {\n    $targets += \"HKLM:\\SOFTWARE\\VMware,
    Inc.\"\n}\n\nif (Test-Path $VMware_Tools_Directory) {\n    $targets += $VMware_Tools_Directory\n}\nif
    (Test-Path $VMware_Common_Directory) {\n    $targets += $VMware_Common_Directory\n}\n\n$services
    = Get-Service -DisplayName \"VMware*\"\n$services += Get-Service -DisplayName
    \"GISvc\"\n\nWrite-Host \"The following registry keys, filesystem folders, and
    services will be deleted:\"\nif (!$targets -and !$services) {\n    Write-Host
    \"Nothing to do!\"\n} else {\n    $targets\n    $services\n\n    $services | Stop-Service
    -Confirm:$false\n\n    if (Get-Command Remove-Service -ErrorAction SilentlyContinue)
    {\n        $services | Remove-Service -Confirm:$false\n    } else {\n        foreach
    ($s in $services) {\n            sc.exe DELETE $($s.Name)\n        }\n    }\n\n
    \   foreach ($item in $targets) {\n        if (Test-Path $item) {\n            Remove-Item
    -Path $item -Recurse\n        }\n    }\n\n    Write-Host \"Done. Rebooting now.\"\n
    \   Restart-Computer\n}\n"
kind: ConfigMap
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"v1","data":{"00_win_firstboot_test.ps1":"Get-Disk | Where-Object { \n    $_.FriendlyName -like '*VirtIO*' -and !$_.IsSystem } | ForEach-Object {\n    Write-Host ('  - ' + $_.Number + ': ' + $_.FriendlyName + ' (' + [math]::Round($_.Size/1GB, 2) + 'GB)')\n    $_ | Set-Disk -IsOffline $false\n    $_ | Set-Disk -IsReadOnly $false }\n\nfunction Get-VMwareToolsInstallerID {\n    foreach ($item in Get-ChildItem Registry::HKEY_CLASSES_ROOT\\Installer\\Products) {\n        if ($item.GetValue('ProductName') -eq 'VMware Tools') {\n            return @{\n                reg_id = $item.PSChildName;\n                msi_id = [Regex]::Match($item.GetValue('ProductIcon'), '(?\u003c={)(.*?)(?=})').Value\n            }\n        }\n    }\n}\n\n$vmware_tools_ids = Get-VMwareToolsInstallerID\n\n$reg_targets = @(\n    \"Registry::HKEY_CLASSES_ROOT\\Installer\\Features\\\",\n    \"Registry::HKEY_CLASSES_ROOT\\Installer\\Products\\\",\n    \"HKLM:\\SOFTWARE\\Classes\\Installer\\Features\\\",\n    \"HKLM:\\SOFTWARE\\Classes\\Installer\\Products\\\",\n    \"HKLM:\\SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Installer\\UserData\\S-1-5-18\\Products\\\"\n)\n\n$VMware_Tools_Directory = \"C:\\Program Files\\VMware\"\n$VMware_Common_Directory = \"C:\\Program Files\\Common Files\\VMware\"\n\n$targets = @()\n\nif ($vmware_tools_ids) {\n    foreach ($item in $reg_targets) {\n        $targets += $item + $vmware_tools_ids.reg_id\n    }\n    $targets += \"HKLM:\\SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Uninstall\\{$($vmware_tools_ids.msi_id)}\"\n}\n\nif ([Environment]::OSVersion.Version.Major -lt 10) {\n    $targets += \"HKCR:\\CLSID\\{D86ADE52-C4D9-4B98-AA0D-9B0C7F1EBBC8}\"\n    $targets += \"HKLM:\\SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Uninstall\\{9709436B-5A41-4946-8BE7-2AA433CAF108}\"\n    $targets += \"HKLM:\\SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Uninstall\\{FE2F6A2C-196E-4210-9C04-2B1BC21F07EF}\"\n}\n\nif (Test-Path \"HKLM:\\SOFTWARE\\VMware, Inc.\") {\n    $targets += \"HKLM:\\SOFTWARE\\VMware, Inc.\"\n}\n\nif (Test-Path $VMware_Tools_Directory) {\n    $targets += $VMware_Tools_Directory\n}\nif (Test-Path $VMware_Common_Directory) {\n    $targets += $VMware_Common_Directory\n}\n\n$services = Get-Service -DisplayName \"VMware*\"\n$services += Get-Service -DisplayName \"GISvc\"\n\nWrite-Host \"The following registry keys, filesystem folders, and services will be deleted:\"\nif (!$targets -and !$services) {\n    Write-Host \"Nothing to do!\"\n} else {\n    $targets\n    $services\n\n    $services | Stop-Service -Confirm:$false\n\n    if (Get-Command Remove-Service -ErrorAction SilentlyContinue) {\n        $services | Remove-Service -Confirm:$false\n    } else {\n        foreach ($s in $services) {\n            sc.exe DELETE $($s.Name)\n        }\n    }\n\n    foreach ($item in $targets) {\n        if (Test-Path $item) {\n            Remove-Item -Path $item -Recurse\n        }\n    }\n\n    Write-Host \"Done. Rebooting now.\"\n    Restart-Computer\n}\n"},"kind":"ConfigMap","metadata":{"annotations":{},"name":"forklift-virt-customize","namespace":"forklift-migration"}}
  creationTimestamp: "2024-10-14T10:16:50Z"
  name: forklift-virt-customize
  namespace: forklift-migration
  resourceVersion: "42632690"
  uid: 4125e2b9-c1b5-4c8b-b6c2-10a11e41e3eb
```

mnecas:
you can replace the image in the CSV and the operator will do the rest
ohh customize to remote cluster, it should be done on that cluster and in namepsace where you are migrating to... that said I have not tested it to remote clusters so might not be working correctly

jsakil14:
> you can replace the image in the CSV and the operator will do the rest ohh customize to remote cluster, it should be done on that cluster and in namepsace where you are migrating to... that said I have not tested it to remote clusters so might not be working correctly

ok , will try putting all images to latest flag and will check. Also , you mean that we need forklift operator installed on all the clusters? I thought , it is just needed on 1 admin cluster and we can add others as providers and do the migration.

mnecas:
you are correct only in the admin cluster, from there we create the pods on the remote clusters, but the configurations and images are specified in the master

jsakil14:
@mnecas question reg-configmap - Can I have both linux & windows scripts in a single CM? will the v2v pick up the right one based on the OS flavour?

jsakil14:
> @mnecas question reg-configmap - Can I have both linux & windows scripts in a single CM? will the v2v pick up the right one based on the OS flavour?

BUMP! @mnecas

PR #1019, opened 2024-09-06
Bump forklift csv to 2.7

mnecas:
- bump the csv description to 2.7.
- bump the csv minkube version to ocp 4.15


mnecas:
I tried to bump the csv min kube version to 1.28 which is the ocp 4.15 but the CI failed. We will need to bump it

yaacov (approved):

PR #1020, opened 2024-09-09
MTV-1041 | Change created virt-v2v conversion output to -o kubeVirt

mnecas:
Issue:
We need to provide the firmware for the OVA and right now the firmware is not exposed from `virt-v2v -o local` but is available in the `-o kubevirt`.

Fix: 
Change the output file of the virt-v2v to the `-o kubevirt` and expose that via the server to the controller. 

Notes:
In future, we could use also `virt-v2v-inspector` to expose this parameter https://issues.redhat.com/browse/RHEL-58047. But right now not available in the rhel 9.4 and as we have this parameter in the `-o kubevirt` I don't think that we need to backport it.
This is continuation of: https://github.com/kubev2v/forklift/pull/778


yaacov:
LGTM, is this part of https://github.com/kubev2v/forklift/pull/778 ?

yaacov (approved):

mnecas:
yeah I just rebased it and I'll add some additional commits in followup MRs

pkg/controller/plan/kubevirt.go
```diff
@@ -980,22 +980,22 @@ func (r *KubeVirt) UpdateVmByConvertedConfig(vm *plan.VMStatus, pod *core.Pod, s
 
 	switch r.Source.Provider.Type() {
 	case api.Ova:
-		vmConfigBytes, err := io.ReadAll(resp.Body)
+		vmConf, err := io.ReadAll(resp.Body)
 		if err != nil {
-			return liberr.Wrap(err)
+			return err
```
yaacov, line 985: In other areas of the file, we’re using liberr.Wrap, which provides additional context and tracing. Given that I’m still getting familiar with the codebase, having the error trace could be really helpful for debugging and understanding the flow of issues. Would it make sense to wrap the error here as well for consistency and better traceability?

mnecas: I was thinking to move the `liberr.warp` out of this function to make apply it on all errors, but would be too much changes and did not return this back.

mnecas: fixed

yaacov (approved):
LGTM

PR #1024, opened 2024-09-10
Redirect RHEL run network config utils sh script errors to stderr

yaacov:
Redirect RHEL run network config utils sh script errors to stderr

Issue:
We redirect all output from the script into the udev script file

Fix:
a. Print warning and errors to stderr
b. Redirect only stdout output into the udev file

### DIFF
--- a/virt-v2v/scripts/rhel/run/network_config_util.sh
+++ b/virt-v2v/scripts/rhel/run/network_config_util.sh
@@ -6,15 +6,21 @@ NETWORK_SCRIPTS_DIR="${NETWORK_SCRIPTS_DIR:-/etc/sysconfig/network-scripts}"
 NETWORK_CONNECTIONS_DIR="${NETWORK_CONNECTIONS_DIR:-/etc/NetworkManager/system-connections}"
 UDEV_RULES_FILE="${UDEV_RULES_FILE:-/etc/udev/rules.d/70-persistent-net.rules}"
 
+# Dump debug strings into a new file descriptor and redirect it to stdout.
+exec 3>&1
+log() {
+    echo $@ >&3
+}
+
 # Check if mapping file does not exist
 if [ ! -f "$V2V_MAP_FILE" ]; then
-    echo "File $V2V_MAP_FILE does not exist. Exiting."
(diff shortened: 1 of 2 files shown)


virt-v2v/scripts/rhel/run/network_config_util.sh
```diff
@@ -8,13 +8,13 @@ UDEV_RULES_FILE="${UDEV_RULES_FILE:-/etc/udev/rules.d/70-persistent-net.rules}"
 
 # Check if mapping file does not exist
 if [ ! -f "$V2V_MAP_FILE" ]; then
-    echo "File $V2V_MAP_FILE does not exist. Exiting."
+    echo "File $V2V_MAP_FILE does not exist. Exiting." >&2
```
fabiand, line 11: can we have a function like

```
err() { echo "$@" >&2 ; }
```

fabiand (changes requested):

virt-v2v/scripts/rhel/run/network_config_util.sh
yaacov: done

yaacov: changed to log file descriptor and redirect to stdout as @mnecas suggested

tests:
dash virt-v2v/scripts/rhel/run/network_config_util_test

virt-v2v/scripts/rhel/run/network_config_util.sh
```diff
@@ -110,7 +116,7 @@ check_dupe_hws() {
 
     # If duplicates are found, print an error and exit
     if [ -n "$dupes" ]; then
-        echo "Warning: Duplicate hw: $dupes"
+        err "Warning: Duplicate hw: $dupes"
```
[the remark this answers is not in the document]
yaacov: thanks !

fabiand (commented):
Can we also make a change to explicitly only redirect `echo "SUBSYSTEM==\"net\",ACTION==\"add\",ATTR{address}==\"$(remove_quotes "$S_HW")\",NAME=\"$(remove_quotes "$DEVICE")\""` to the `70-` file?

PR #1026, opened 2024-09-10
Add netplan config parding sh utility for ubuntu static ip

yaacov:
Add netplan config parding sh utility for ubuntu static ip

### DIFF
--- a/virt-v2v/scripts/rhel/run/network_config_util.sh
+++ b/virt-v2v/scripts/rhel/run/network_config_util.sh
@@ -31,13 +31,6 @@ fi
 # Helper functions
 # ----------------
 
-# Get the v2v map file as list 
-v2v_list_as_envs() {
-    while IFS= read -r line; do
-        echo "$line"
-    done < "$V2V_MAP_FILE"
-}
-
 # Clean strigs in case they have quates
 remove_quotes() {
     echo "$1" | tr -d '"' | tr -d "'" | sed 's/^[[:space:]]*//;s/[[:space:]]*$//'
@@ -65,7 +58,7 @@ udev_from_ifcfg() {
     fi
 
     # Read the mapping file line by line
-    v2v_list_as_envs | while read line;
+    cat "$V2V_MAP_FILE" | while read line;
     do
         # Extract S_HW and S_IP
         extract_mac_ip "$line"
@@ -100,7 +93,7 @@ udev_from_nm() {
     fi
 
     # Read the mapping file line by line
-    v2v_list_as_envs | while read line;
+    cat "$V2V_MAP_FILE" | while read line;
     do
         # Extract S_HW and S_IP
         extract_mac_ip "$line"
@@ -152,7 +145,7 @@ udev_from_netplan() {
     }
 
     # Read the mapping file line by line
-    v2v_list_as_envs | while read line;
+    cat "$V2V_MAP_FILE" | while read line;
     do
         # Extract S_HW and S_IP from the current line in the mapping file
         extract_mac_ip "$line"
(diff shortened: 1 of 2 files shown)


fabiand (changes requested):
GOLD for bash/dash yaml parsing.

But let's eval to use `netplan generate` or `netplan get` to avoid this parsing.

fabiand:
With the following script we loop over all configured interfaces and output the one for the given IP address

```console
IP_TO_FIND=10.0.2.15
netplan get ethernets | grep -Eo "^[^[:space:]]+[^:]" | while read IFNAME  ; do netplan get ethernets.$IFNAME.addresses | grep -q $IP_TO_FIND && echo $IFNAME ; done
```

virt-v2v/scripts/rhel/run/network_config_util.sh
```diff
@@ -24,9 +28,19 @@ if [ -f "$UDEV_RULES_FILE" ]; then
...
 fi
 
+# Helper functions
+# ----------------
+
+# Get the v2v map file as list 
+v2v_list_as_envs() {
+    while IFS= read -r line; do
+        echo "$line"
+    done < "$V2V_MAP_FILE"
```
fabiand, line 38: I think this is equal to `cat $V2V_MAP_FILE` as we just echo every line without modifications.

In the original code I was modifying the line to convert them into ENV_KEY=VALUE format to then export them iwth `export`

virt-v2v/scripts/rhel/run/network_config_util.sh
```diff
@@ -104,9 +123,61 @@ udev_from_nm() {
...
+    }
+
+    # Loop over all interface names and treturn the one with target_ip, or null
+    find_interface_by_ip() {
+        target_ip="$1"
+
+        # Loop through all interfaces and check for the given IP address
+        netplan_get ethernets | grep -Eo "^[^[:space:]]+[^:]" | while read IFNAME; do
+            if netplan_get ethernets."$IFNAME".addresses | grep -q "$target_ip"; then
+                echo "$IFNAME"
```
fabiand, line 149: either we should `return` here
or later we need to check that we only found one ifname

virt-v2v/scripts/rhel/run/network_config_util.sh
```diff
@@ -104,9 +123,61 @@ udev_from_nm() {
...
+
+    # Read the mapping file line by line
+    v2v_list_as_envs | while read line;
+    do
+        # Extract S_HW and S_IP from the current line in the mapping file
+        extract_mac_ip "$line"
+
+        # If S_HW and S_IP were not extracted, skip the line
+        if [ -z "$S_HW" ] || [ -z "$S_IP" ]; then
+            continue
```
fabiand, line 162: worth adding a log message?

virt-v2v/scripts/rhel/run/network_config_util.sh
```diff
@@ -104,9 +123,61 @@ udev_from_nm() {
...
+        fi
+
+        # Search the parsed netplan output for a matching IP address
+        interface_name=$(find_interface_by_ip "$S_IP")
+
+        # If no interface is found, skip this entry
+        if [ -z "$interface_name" ]; then
+            continue
```
fabiand, line 170: worth adding a log message?

virt-v2v/scripts/rhel/run/network_config_util.sh
yaacov: thanks ! replaced ...

virt-v2v/scripts/rhel/run/network_config_util.sh
yaacov: added logs for invalid mac to ip lines

virt-v2v/scripts/rhel/run/network_config_util.sh
yaacov: added logs for device not found

virt-v2v/scripts/rhel/run/network_config_util.sh
yaacov: added return here

mnecas (approved):

PR #1028, opened 2024-09-12
MTV-1366 | Add virt-customize to warm migration and OVA

mnecas:
Issue:
Right now we run the virt-customize only on the cold local migration. But we would wanted to run it on all vmware migrations.

Fix:
This commit adds a new stage to the converison pod which scans the guest to get the operating system. It is using the tool `virt-v2v-inspector` which generates inspection xml file. This file contains lot of information about guest, including the OS. This tool is run after the `virt-v2v` step because for the OVA and cold local migration we dont have the disk available before that step.
Additionally this patch adds a new `inspection` endpoint to the conversion pod. The controller requests the endpoint to get the operating system so it could be set as a preference to the kubevirt.

New known issues:
The OVA won't be able to preserve static IP addresses as we get the IP adress configuration by form the vmware guest agent when the VM is running. There might be some information inside the ova file but right now this is not fixed.


pkg/controller/plan/adapter/ova/ovfparser.go
```diff
@@ -120,27 +120,35 @@ func GetFirmwareFromConfig(vmConfigXML string) (firmware string, err error) {
 	return BIOS, nil
 }
 
-func GetOperationSystemFromConfig(vmConfigXML string) (os string, err error) {
-	xmlConf, err := readConfFromXML(vmConfigXML)
+type InspectionOS struct {
```
yaacov, line 123: The xml output of `virt-v2v-inspector` is not strict ovf format, can we move the inspector xml parsing to a general pkg, instead of (re-)using the ova import package ?

virt-v2v/pkg/server/server.go
```diff
@@ -21,6 +22,7 @@ var (
 // Operating System for the VM creation.
 func Start() error {
 	http.HandleFunc("/ovf", ovfHandler)
```
yaacov, line 24: do we still need the ovf endpoint ?

mnecas: In this PR yes, as we are still exposing the firmware to the ova, but that has also its own issues so this will be replaced in followup PR similar to the Bellas changes https://github.com/kubev2v/forklift/pull/1020

pkg/controller/plan/adapter/ova/ovfparser.go
mnecas: Done, ptal

virt-v2v/pkg/utils/command.go
```diff
@@ -131,6 +113,22 @@ func getDiskLink(kind global.MountPath, disk string) (string, error) {
 	), nil
 }
 
+func GetLinkedDisks() ([]string, error) {
+	disks, err := filepath.Glob(
+		filepath.Join(
+			global.DIR,
+			fmt.Sprintf("%s-sd*", os.Getenv("V2V_vmName")),
```
yaacov, line 120: Note: files are called `NAME-sd*`

https://libguestfs.org/virt-v2v-output-local.1.html

> The files will be called:
>  NAME-sda, NAME-sdb, etc.      Guest disk(s).

pkg/controller/plan/adapter/vsphere/inspectionparser.go
```diff
@@ -0,0 +1,85 @@
...
+	OS InspectionOS `xml:"operatingsystem"`
+}
+
+func ParseInspectionFromString(xmlData string) (*InspectionV2V, error) {
+	var xmlConf InspectionV2V
+	err := xml.Unmarshal([]byte(xmlData), &xmlConf)
+	if err != nil {
+		return nil, fmt.Errorf("Error unmarshalling XML: %v\n", err)
+	}
+	return &xmlConf, nil
```
yaacov, line 71: Do we usually use pointers as return values, is InspectionV2V expected to be a large struct ?

mnecas: I copied it from the ovf parser, I'm okay with returning the struct as value

yaacov: lets keep it as is, until we know if it's a intentional, a coding style or not intentional ...

yaacov: ahh... ok I think it's better without the pointer :+1: , missed that you fixed that

pkg/controller/plan/kubevirt.go
```diff
@@ -935,14 +936,27 @@ func (r *KubeVirt) GetGuestConversionPod(vm *plan.VMStatus) (pod *core.Pod, err
...
+	inspectionUrl := fmt.Sprintf("http://%s:8080/inspection", pod.Status.PodIP)
+	resp, err := http.Get(inspectionUrl)
+	if err != nil {
+		return "", liberr.Wrap(err)
+	}
+	inspectionBytes, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return "", liberr.Wrap(err)
+	}
+	defer resp.Body.Close()
```
yaacov, line 949: I thnk the defer will be nicer before the 
`inspectionBytes, err := io.ReadAll(resp.Body)`
line, imidiatly after the` resp, err := http.Get(inspectionUrl)` error check

pkg/controller/plan/kubevirt.go
```diff
@@ -935,14 +936,27 @@ func (r *KubeVirt) GetGuestConversionPod(vm *plan.VMStatus) (pod *core.Pod, err
...
+	return string(inspectionBytes), nil
+}
+
+func (r *KubeVirt) UpdateVmByConvertedConfig(vm *plan.VMStatus, pod *core.Pod, step *plan.Step) error {
 	if pod.Status.PodIP == "" {
```
yaacov, line 954: I know this is not part of this PR, but
pod is a pointer and can theoretically be nil, we need check for pod is nil, before using it ...

pkg/controller/plan/kubevirt.go
```diff
@@ -935,14 +936,27 @@ func (r *KubeVirt) GetGuestConversionPod(vm *plan.VMStatus) (pod *core.Pod, err
 	return
 }
 
-func (r *KubeVirt) UpdateVmByConvertedConfig(vm *plan.VMStatus, pod *core.Pod, step *plan.Step) (err error) {
+func (r *KubeVirt) getInspectionXml(pod *core.Pod) (string, error) {
+	inspectionUrl := fmt.Sprintf("http://%s:8080/inspection", pod.Status.PodIP)
```
yaacov, line 940: pod can theoretically be nil

yaacov (approved):
LGTM

PR #1030, opened 2024-09-12
MTV-446 | Warm: Add cbt check before migration

mnecas:
Issue:
Right now we have a check to fail the migration when we start to create the VM which does not have the CBT enabled. We should throw error much earlier before the customer even start the migration.

Fix:
Add a CBT check to the plan validation. This fix right now affects only VMware migrations.

### DIFF
--- a/pkg/controller/plan/adapter/base/doc.go
+++ b/pkg/controller/plan/adapter/base/doc.go
@@ -138,6 +138,8 @@ type Validator interface {
 	PodNetwork(vmRef ref.Ref) (bool, error)
 	// Validate that we have information about static IPs for every virtual NIC
 	StaticIPs(vmRef ref.Ref) (bool, error)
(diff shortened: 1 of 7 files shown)


mnecas:
New error from UI:
image

mnecas:
Error example from yaml plan:
```
  conditions:
    - category: Critical
      items:
        - ' id:vm-34 name:''mnecas-infra'' '
      lastTransitionTime: '2024-09-12T10:36:03Z'
      message: Changed Block Tracking (CBT) has not been enabled on some VM. This feature is a prerequisite for VM warm migration.
      reason: MissingGuestInformation
      status: 'True'
      type: VMMissingGuestIPs
```

mnecas:
Once I toggle to turn off the warm migration:
image

yaacov (approved):
LGTM

PR #1031, opened 2024-09-12
Support old ubuntu netplan

yaacov:
Support old ubuntu netplan

Issue:
Ubuntu 18 uses vertion of netplan that does not support the get subcommand

Fix:
a. add a python script that mimics netplan get command
b. check if current version of netplan support get
c. run the python script in cases netplan does not support get

### DIFF
--- a/virt-v2v/pkg/customize/scripts/rhel/run/network_config_util.sh
+++ b/virt-v2v/pkg/customize/scripts/rhel/run/network_config_util.sh
@@ -46,6 +46,42 @@ extract_mac_ip() {
     fi
 }
 
+netplan_get_py() {
+    python -c "
+import os
+import yaml
+import sys
+
+netplan_dir = os.getenv('NETPLAN_DIR', '') + '/etc/netplan'
+args = sys.argv[1].split('.')
+
+def find_yaml_files(directory):
+    yaml_files = []
+    for root, dirs, files in os.walk(directory):
+        for file in files:
+            if file.endswith(('.yaml', '.yml')):
+                yaml_files.append(os.path.join(root, file))
+
+    return yaml_files
+
+def get_yaml_path(yaml_data, keys):
+    for key in keys:
+        if yaml_data is None:
+            return None
+        yaml_data = yaml_data.get(key)
+    return yaml_data
+
+yaml_files = find_yaml_files(netplan_dir)
+
+for yaml_file in yaml_files:
+    with open(yaml_file, 'r') as file:
+        yaml_data = yaml.safe_load(file)        
+        result = get_yaml_path(yaml_data, ['network'] + args)
+        if result is not None:
+           print(yaml.dump(result, default_flow_style=False))
+" "$@"
+}
+
 # Network infrastructure reading functions
 # ----------------------------------------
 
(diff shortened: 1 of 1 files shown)


virt-v2v/pkg/customize/scripts/rhel/run/network_config_util.sh
```diff
@@ -133,9 +169,19 @@ udev_from_netplan() {
...
+    netplan_supports_get() {
+        netplan get 2>/dev/null
+        return $?
+    }
+
     # netplan with root dir
     netplan_get() {
-        netplan get --root-dir "$NETPLAN_DIR" "$@" 2>/dev/null
+        if netplan_supports_get; then
+            netplan get --root-dir "$NETPLAN_DIR" "$@" 2>/dev/null
```
mnecas, line 181: Could we pass the stderr to the `log`?

virt-v2v/pkg/customize/scripts/rhel/run/network_config_util.sh
```diff
@@ -46,6 +46,42 @@ extract_mac_ip() {
...
 }
 
+netplan_get_py() {
+    python -c "
+import os
+import yaml
+import sys
+
+netplan_dir = os.getenv('NETPLAN_DIR', '') + '/etc/netplan'
+args = sys.argv[1].split('.')
```
mnecas, line 56: Please add an example of how the args should look like

virt-v2v/pkg/customize/scripts/rhel/run/network_config_util.sh
```diff
@@ -46,6 +46,42 @@ extract_mac_ip() {
...
+    return yaml_data
+
+yaml_files = find_yaml_files(netplan_dir)
+
+for yaml_file in yaml_files:
+    with open(yaml_file, 'r') as file:
+        yaml_data = yaml.safe_load(file)        
+        result = get_yaml_path(yaml_data, ['network'] + args)
+        if result is not None:
+           print(yaml.dump(result, default_flow_style=False))
```
mnecas, line 81: Do we know which python is used in the ubuntu 18?

yaacov: > Do we know which python is used in the ubuntu 18?

python3, on my ubuntu18 (new installed using iso) the python version was 3.6.9

>we would have problems with python2

yes, this script will not work on python2 

AFAIK (* not ubuntu expert)
18.04 - python3 become default python
17.10 - netplan become default network manager
( note: 17 is not LTS version, first LTS version to use netplan is 18.04 )

@cc @fabiand ^^

virt-v2v/pkg/customize/scripts/rhel/run/network_config_util.sh
yaacov: now redirect to file descriptor 3 ( our log file ), and then to screen

virt-v2v/pkg/customize/scripts/rhel/run/network_config_util.sh
mnecas: Okok thanks for checking, sgtm

virt-v2v/pkg/customize/scripts/rhel/run/network_config_util.sh
mnecas: Yeah sounds good to me

mnecas (approved):
Lgtm

virt-v2v/pkg/customize/scripts/rhel/run/network_config_util.sh
```diff
@@ -133,9 +173,20 @@ udev_from_netplan() {
         return 0
     fi
 
+    # Function to check if netplan supports the 'get' subcommand
+    netplan_supports_get() {
+        netplan get 2>/dev/null
```
mnecas, line 178: Ah add it here as well

yaacov: changed

note, this log line is less interesting, because failing here is a "happy path", this method is for testing if this command will error or not, failing is one of the expected paths.

mnecas: yeah aware of it, just want to log in case we will see a problem with the happy path

mnecas (approved):

PR #1038, opened 2024-09-16
Use file name when device is unknown

yaacov:
**Issue:**
In some setups the `ifcgf-*` files use `HWADDR` to identify the interface instead of `DEVICE`,
currently we ignore this config files.

**Fix:**
In case the config file is missing `DEVICE` line, look for `HWADDR` line, if exist the MAC address match, use the file name for the device line.

**Caveat:**
No checks for the device name we take from the file, e.g. "Does this string looks like a network interface name?, do we have duplication ?"

**P.S:**
also check that we don't send the test script to guest file system


mnecas (approved):
lgtm

PR #1039, opened 2024-09-17
Operator | Fix console-plugin template

mnecas:
Issue:
The latest operator fails with error:
`forklift-console-plugin is invalid: spec.proxy[0].endpoint.type: Required value`


yaacov (approved):
LGTM

PR #1040, opened 2024-09-17
add Dockerfile for forklift-controller

ernesgonzalez33:


yaacov:
@ernesgonzalez33 hi, thanks for the `Dockerfile` !,  I see you're planning to add a Dockerfile in the `pkg/controller/` directory to build the command-line tool from `cmd/forklift-controller`. I have a few concerns:
  - Since cmd/forklift-controller contains the tool, it might be more appropriate to place the Dockerfile there. This would make it clearer what the Dockerfile is responsible for and avoid confusion with library packages in pkg/controller/
  - The `pkg/` directory is generally used for reusable libraries, whereas Dockerfiles and other build tools are often better suited in the project root or build dir. Mixing Dockerfiles with packages might make it harder to navigate and maintain the codebase in the long run.
  - If we will have other Dockerfiles in the project, we might want to follow a consistent approach to where they're placed and how they’re structured.
  
  Would it make sense to relocate the Dockerfile to the cmd or specific buid dir ?

ernesgonzalez33:
@yaacov thank you for the review! Yes, the idea is to create Dockerfile for each of the containers that are going to be built for the migration to Konflux. Having a discussion regarding where sounds like a good idea. My first option was a directory called `build` with subdirectories for each container, but putting them where the code is makes even more sense. In the meantime I'll move it to `cmd`, but I'm open to change it to wherever you prefer.

Also, I'm planning on creating a `.dockerignore` file, since right now this is copying the entire repo and that doesn't make a lot of sense

yaacov:
@ernesgonzalez33 i am fine with putting them in the current `/cmd` ( and also fine with a new `/build` dir :-) )
@mnecas WDYT ?

mnecas:
I would do the `/build` dir, as we don't have all images in the `/cmd`, for example, the virt-v2v container is in the `virt-v2v` directory and I would prefer to have everything at one place

ernesgonzalez33:
@yaacov @mnecas moved the Dockerfile to the correct directory and added a `.dockerignore`. PTAL

build/forklift-controller/Dockerfile
```diff
@@ -0,0 +1,25 @@
+FROM registry.access.redhat.com/ubi9/go-toolset:1.21.11-9 AS builder
```
mnecas, line 1: Could we do `Containerfile.controller`, wdyt @yaacov ?

mnecas: Or will we have multiple container files in the directory? Ie also for the downstream? @ernesgonzalez33

yaacov: my vote is for separate directories for each tool,
but I can also see the benefit in one directory with different names.

ernesgonzalez33: I don't think you'll have different Dockerfiles in the same directory, but I find this way easier to know which one is which

mnecas: Ah last note please use the `Containerfile`

ernesgonzalez33: Sure! I'll rename it!

ernesgonzalez33:
@mnecas @yaacov PTAL

build/forklift-controller/Containerfile
```diff
@@ -0,0 +1,26 @@
...
+ENTRYPOINT ["/usr/local/bin/forklift-controller"]
+
+LABEL \
+        com.redhat.component="forklift-controller-container" \
+        name="forklift/forklift-controller-rhel9" \
+        license="Apache License 2.0" \
+        io.k8s.display-name="Forklift" \
+        io.k8s.description="Forklift - Controller" \
+        io.openshift.tags="migration,mtv,forklift" \
+        summary="Forklift - Controller" \
```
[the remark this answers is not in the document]
yaacov: sorry, missed the description fix

yaacov (approved):
LGTM

mnecas (approved):
lgtm

PR #1041, opened 2024-09-18
run,test: Fix netplan test

fabiand:
A small fix to make the netplan tests run as well.

This works y faking the netplan binary.


yaacov (approved):
LGTM

PR #1044, opened 2024-09-19
MTV-1482 | Fix main pod inventory watch NPE

mnecas:
Issue: When the request from main to the inventory pod fails the main pod crashes due to nil pointer.

Fix: Move the request close to case when we get the response body.


pkg/lib/inventory/web/client.go
```diff
@@ -264,10 +264,8 @@ func (r *Client) Watch(url string, resource interface{}, h EventHandler) (status
 				pErr = nil
 				return
 			}
+			defer response.Body.Close()
```
yaacov, line 267: we can actually move it even higher, just after the check response is not nil (L258)

yaacov (approved):
LGTM

Issue #1029, opened 2024-09-12
Unable to boot and start 'windows' vm post migration to kubevirt. 


cniket:
Hello,

After creating a `plan` CR and then creating a `migration` CR for a 'Windows 2k19' VM, a `DV` gets spin-up and disk migration from VMware starts. After successful completion on DV, ultimately a VM gets created and gets listed in 'Running' state.
After that when I am trying to get console access of the VM using command `kubectl virt vnc <vm-name>` , I am getting windows blue screen with error: 'Stop code: INACCESSIBLE BOOT DEVICE', as shown below;
<img width="1279" alt="Screenshot 2024-09-12 at 12 42 02" src="https://github.com/user-attachments/assets/2b573140-f1f3-48ae-93ee-5cb004088cbb">

`storagemap` is using rook ceph storage and the corresponding `pv`,`pvc` are getting created/bound properly. 
```
❯ kg pvc -n virtualmachines |grep w195
plan-w195-warm-vm-4427-jklkt             Bound    pvc-xxxxxxxx-8012-4f35-xxxxx-7dd56xxxxxx   16Gi          RWX            ceph-block     <unset>                 129m    Block
❯
```

The VM manifest already has secureBoot as 'false'. Below is the full manifest:
```
❯ k get vm w195 -n virtualmachines -oyaml
apiVersion: kubevirt.io/v1
kind: VirtualMachine
metadata:
  annotations:
    kubemacpool.io/transaction-timestamp: "2024-09-12T09:03:38.109524776Z"
    kubevirt.io/latest-observed-api-version: v1
    kubevirt.io/storage-observed-api-version: v1
  creationTimestamp: "2024-09-12T07:07:04Z"
  finalizers:
  - kubevirt.io/virtualMachineControllerFinalize
  generation: 5
  labels:
    migration: xxxxxxxx-9a13-xxxxxx-acdb-xxxxxxxxx
    plan: xxxxxx-xxxxx-4bc1-xxxxx-babdxxxxxx
    vmID: vm-4427
  name: w195
  namespace: virtualmachines
  resourceVersion: "130711651"
  uid: xxxxxx-42b7-xxxxxx-4f01bxxxxxxx
spec:
  running: true
  template:
    metadata:
      creationTimestamp: null
      labels:
        app: w195
    spec:
      architecture: amd64
      domain:
        clock:
          timezone: UTC
        cpu:
          cores: 1
          sockets: 2
        devices:
          disks:
          - disk:
[...]

Issue #982, opened 2024-08-14
Consume a new virtio-win which supports win2k22.


ahadas:
I'm merging and backporting this PR.
A note to remember is to update the virt-v2v deps to consume a new virtio-win which supports win2k22.

_Originally posted by @liranr23 in https://github.com/kubev2v/forklift/issues/975#issuecomment-2286337468_

PR #1048, opened 2024-09-19
MTV-1483 | Fix virtio-win driver version

mnecas:
Issue:
The Centos 9 AppStream repo has an outdated Windows virtio driver which does not support Windows 2022 and has issues with 2019. This change updates the rpm to the fedora package.

This is not a proper fix but WA as the base image is ubi9 and we should not mix the packages.

Fixes: https://github.com/kubev2v/forklift/issues/982
Fixes: https://github.com/kubev2v/forklift/issues/1029
Fixes: https://issues.redhat.com/browse/MTV-1483

### DIFF
--- a/virt-v2v/BUILD.bazel
+++ b/virt-v2v/BUILD.bazel
@@ -451,7 +451,7 @@ rpmtree(
         "@util-linux-core-0__2.37.4-20.el9.x86_64//rpm",
         "@vim-minimal-2__8.2.2637-21.el9.x86_64//rpm",
         "@virt-v2v-1__2.5.6-4.el9.x86_64//rpm",
-        "@virtio-win-0__1.9.15-4.el9.x86_64//rpm",
+        "@virtio-win-1.9.40-1.el9.noarch//rpm",
         "@which-0__2.21-29.el9.x86_64//rpm",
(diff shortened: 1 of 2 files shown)


mnecas:
Moving to draft as I'm not sure if I want to go this route

mnecas:
Got build from koji for el9, moving from draft

jsakil14:
> Got build from koji for el9, moving from draft

Will this be pushed to virt-v2v image in quay.io/kubev2v registry? Since , am using olm - I tried build now with the :latest - still its pointing to virtio-win-1.9.15.iso. Or do we still need to rebuild it only with bazel? @mnecas

mnecas:
@jsakil14 strange, it should be updated, let me check.

mnecas:
I opened https://github.com/kubev2v/forklift/pull/1112 you will need to update the controller image after it, hope it will help

jsakil14:
@mnecas was this pushed?

arturshadnik:
Hi @mnecas just following up on this, the `:latest` image from `quay.io/kubev2v` seems to still have the older 1.9.15 version. 
image. When building locally from main, the build fails trying to fetch the updated virtio-win driver. Can you please take a look

PR #1063, opened 2024-09-20
fix: containerfiles now are building local code

ernesgonzalez33:


mnecas (changes requested):
The `github.com/konveyor/forklift-controller` is a Golang package inside the project.
So it is built locally, and it has historical reasons for this.

PR #1085, opened 2024-10-04
MTV-1569 | Fail plan when the conversion pod fails

mnecas:
Issue:
When the virt-v2v pod fails the plan does not fail.

### DIFF
--- a/pkg/controller/plan/migration.go
+++ b/pkg/controller/plan/migration.go
@@ -1146,7 +1146,7 @@ func (r *Migration) execute(vm *plan.VMStatus) (err error) {
 				return err
 			}
 
-			if pod != nil && pod.Status.Phase != core.PodSucceeded {
+			if pod != nil && pod.Status.Phase != core.PodSucceeded && pod.Status.Phase != core.PodFailed {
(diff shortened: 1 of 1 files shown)


yaacov (approved):

pkg/controller/plan/migration.go
```diff
@@ -1146,7 +1146,7 @@ func (r *Migration) execute(vm *plan.VMStatus) (err error) {
 				return err
 			}
 
-			if pod != nil && pod.Status.Phase != core.PodSucceeded {
+			if pod != nil && pod.Status.Phase != core.PodSucceeded && pod.Status.Phase != core.PodFailed {
```
yaacov, line 1149: Question: where do we set the step to "Fail" in this case ?

mnecas: The problem was that it was skipping the error due to different error.
And we set the fail here
https://github.com/kubev2v/forklift/blob/fd1180eb4543ab4130d0afc61e22c5a566b328ae/pkg/controller/plan/migration.go#L1649-L1651

PR #1091, opened 2024-10-08
MTV-1477 | Add DataVolume deletion when archiving

mnecas:
Issue:
When running the migration archive we do not delete the DataVolumes.

Fix:
Add deletion of the DataVolumes when running the archive and the VM migration failed.
At some cases the migration in the context is missing the migration ID so we can use the plan instead to find the DataVolume.

Fixes: https://issues.redhat.com/browse/MTV-1477

### DIFF
--- a/pkg/controller/plan/kubevirt.go
+++ b/pkg/controller/plan/kubevirt.go
@@ -661,7 +661,7 @@ func (r *KubeVirt) getDVs(vm *plan.VMStatus) (edvs []ExtendedDataVolume, err err
 		context.TODO(),
 		dvsList,
 		&client.ListOptions{
-			LabelSelector: k8slabels.SelectorFromSet(r.vmLabels(vm.Ref)),
+			LabelSelector: k8slabels.SelectorFromSet(r.vmAllButMigrationLabels(vm.Ref)),
 			Namespace:     r.Plan.Spec.TargetNamespace,
 		})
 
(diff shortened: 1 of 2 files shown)


yaacov (approved):

pkg/controller/plan/kubevirt.go
```diff
@@ -661,7 +661,7 @@ func (r *KubeVirt) getDVs(vm *plan.VMStatus) (edvs []ExtendedDataVolume, err err
 		context.TODO(),
 		dvsList,
 		&client.ListOptions{
-			LabelSelector: k8slabels.SelectorFromSet(r.vmLabels(vm.Ref)),
+			LabelSelector: k8slabels.SelectorFromSet(r.vmAllButMigrationLabels(vm.Ref)),
```
yaacov, line 664: question:
do we consider the missing migration id a bug ? 
do we know why in some cases it's missing ?

mnecas: > do we consider the missing migration id a bug ?

yes, I think it could cause a problems in future

> do we know why in some cases it's missing ?

still investigating, hope I'll find the answer by the end of the day so I don't need to do this hack
if not I'll file bug for this

mnecas: So the issue is that we do not initiate the migration variable in the context. I was not able to find an easy way how to populate it, I'm probably missing something. However I noticed that we are already using the `vmAllButMigrationLabels`  in all cleanup steps, so I think we should be good to go.

PR #1092, opened 2024-10-08
Use a finalizer to make sure web hooks are removed when operator is uninstalled

jonner:
When uninstalling the forklift operator, not all resources are removed from the cluster. In particular, there are at least 3 validatingwebhookconfiguration objects and 3 mutatingwebhookconfiguration objects that are left behind. By using a finalizer on the operator, we can ensure that these objects are removed from the cluster.

### DIFF
--- a/operator/roles/forkliftcontroller/tasks/main.yml
+++ b/operator/roles/forkliftcontroller/tasks/main.yml
@@ -142,46 +142,10 @@
       state: present
       definition: "{{ lookup('template', 'api/deployment-forklift-api.yml.j2') }}"
 
-  - name: "Delete aggregated validation webhook configurations"
-    k8s:
-      state: absent
-      definition: "{{ lookup('template', 'api/validatingwebhookconfiguration-forklift-api.yml.j2') }}"
-
-  - name: "Setup secrets validating webhook configuration"
-    k8s:
-      state: present
-      definition: "{{ lookup('template', 'api/validatingwebhookconfiguration-secrets.yml.j2') }}"
-
-  - name: "Setup plans validating webhook configuration"
-    k8s:
-      state: present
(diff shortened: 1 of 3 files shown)


operator/roles/forkliftcontroller/tasks/main.yml
```diff
@@ -147,41 +147,15 @@
...
-
-  - name: "Setup providers validating webhook configuration"
-    k8s:
-      state: present
-      definition: "{{ lookup('template', 'api/validatingwebhookconfiguration-providers.yml.j2') }}"
-
   - name: "Delete aggregated mutating webhook configurations"
     k8s:
       state: absent
       definition: "{{ lookup('template', 'api/mutatingwebhookconfiguration-forklift-api.yml.j2') }}"
```
mnecas, line 153: Why we don't move these to the `webhooks.yml`?

operator/roles/forkliftcontroller/tasks/main.yml
```diff
@@ -276,3 +250,11 @@
       namespace: "{{ app_namespace }}"
       name: forklift-must-gather-api
       state: absent
+  when: finalize is not defined
+
+- block:
+  - name: "Remove webhook configuration"
```
mnecas, line 256: Could you please also add a task for the console plugin removal?

yaacov:
Ref: https://issues.redhat.com/browse/MTV-1076

operator/roles/forkliftcontroller/tasks/main.yml
mnecas: yeah please move them to the one file, think it will be more clear

mnecas (approved):

PR #1094, opened 2024-10-09
Fix VMware scheduler test

mnecas:
Now we allow migrating with more disks than the MaxInFlight in case there is no VM being migrated at that time. This is done for the user's convenience.

Related to: https://github.com/kubev2v/forklift/pull/1087

### DIFF
--- a/pkg/controller/plan/scheduler/vsphere/scheduler_test.go
+++ b/pkg/controller/plan/scheduler/vsphere/scheduler_test.go
@@ -84,6 +84,9 @@ func TestScheduler(t *testing.T) {
 		},
 		hostC: {
 			{
+				cost: 11,
+			},
+			{
 				cost: 1,
 			},
 			{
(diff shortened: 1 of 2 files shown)


yaacov (approved):

pkg/controller/plan/scheduler/vsphere/scheduler_test.go
```diff
@@ -83,6 +83,9 @@ func TestScheduler(t *testing.T) {
 			},
 		},
 		hostC: {
+			{
+				cost: 11,
```
yaacov, line 87: question:
in our git automation we run "make ci"
https://github.com/kubev2v/forklift/blob/main/.github/workflows/pull-request.yml#L138

and "ci" does not run "test" :-(
https://github.com/kubev2v/forklift/blob/main/Makefile#L78

do we want to run "make test", am i missing something ?

yaacov: sorry, we run "all" that run "test", resolving

PR #1096, opened 2024-10-09
MTV-1576 | VMware apply CBT from extra args

mnecas:
Issue:
When customer applies the CBT on running VM by:
`govc vm.change -vm VM_NAME -e "ctkEnabled=TRUE"`
The customer gets a critical conidtion which does not allow them to migrate the VM.

Fix:
The changes are not propagated to the VM config but to the extra arguments which we need to query.

Fixes: https://issues.redhat.com/browse/MTV-1576

### DIFF
--- a/pkg/controller/provider/container/vsphere/model.go
+++ b/pkg/controller/provider/container/vsphere/model.go
@@ -3,6 +3,7 @@ package vsphere
 import (
 	"net/url"
 	"sort"
+	"strconv"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


yaacov (approved):

mnecas:
Did more testing and missed some edge case, moving to draft.

mnecas:
Done, added reque in case some plan will fail with this so we will retry

yaacov (approved):

PR #1106, opened 2024-10-14
Fix crash on unexpected ova files

jonner:
See descriptions on the individual commits. This just makes the ova provider a bit more robust to unexpected inputs.

- **ova: Don't crash on ova files when device is missing 'ElementName'**
- **ova: improve handling of extra devices**

### DIFF
--- a/cmd/ova-provider-server/ova-provider-server.go
+++ b/cmd/ova-provider-server/ova-provider-server.go
@@ -466,9 +466,11 @@ func convertToVmStruct(envelope []Envelope, ovaPath []string) ([]VM, error) {
 					newVM.MemoryUnits = item.AllocationUnits
 
 				} else {
-					newVM.Devices = append(newVM.Devices, Device{
-						Kind: item.ElementName[:len(item.ElementName)-2],
-					})
+					if len(item.ElementName) > 2 {
+						newVM.Devices = append(newVM.Devices, Device{
+							Kind: item.ElementName[:len(item.ElementName)-2],
+						})
+					}
 				}
 
 			}
--- a/cmd/ova-provider-server/ova-provider-server.go
+++ b/cmd/ova-provider-server/ova-provider-server.go
@@ -15,6 +15,7 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"unicode"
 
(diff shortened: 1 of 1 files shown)


cmd/ova-provider-server/ova-provider-server.go
```diff
@@ -466,8 +467,27 @@ func convertToVmStruct(envelope []Envelope, ovaPath []string) ([]VM, error) {
...
+						// if the `ElementName` element has a name such as "Hard Disk 1", strip off the
+						// number suffix
+						runes := []rune(item.ElementName)
+						if len(runes) > 2 &&
+							unicode.IsSpace(runes[len(runes)-2]) &&
+							unicode.IsDigit(runes[len(runes)-1]) {
+							itemKind = item.ElementName[:len(item.ElementName)-2]
+						} else {
+							itemKind = item.ElementName
+						}
```
yaacov, line 487:
```suggestion
itemKind = strings.TrimRightFunc(item.ElementName, func(r rune) bool {
        return unicode.IsDigit(r) || unicode.IsSpace(r)
    })
```

IMHO using trim right is nicer here:
 it removes all trailing spaces and digits, not just a specific pattern.
 it is more concise and uses standard library functions, making it easier to read and understand.
 it does not require custom logic
 
using `strings.TrimRightFunc` might be slower (?) but IMHO we can ignore this penalty in our use case.

yaacov: p.s. in both implementations we need to check that the resulting `itemKind` after removing the number suffix is not empty.

yaacov:
@jonner hi, ova test fails, can you take a look ?

cc:// @mnecas

yaacov:
@jonner close in favor of #1111 ?

mnecas (approved):

mnecas:
afaik the main issue [MTV-1577](https://issues.redhat.com/browse/MTV-1577) should be solved by https://github.com/kubev2v/forklift/pull/1111
but this could solve another issue with different ova information
generally lgtm, do we have issue for this case?

yaacov:
> do we have issue for this case?

I don't think we do, @jonner  ?

yaacov:
@jonner @mnecas do you know why this fails the ova test suit ?

PR #1109, opened 2024-10-16
MTV-1536 | Use CDI for disk transfer in cold migration

mnecas:
## Issues:
- Allow migration of "unknow" guests [1] 
Right now we can't migrate an unknown and unsupported operating system which is unsupported by the virt-v2v [2].

- Unifying the process and potential speedup [3]
Right now we are using two different methods for the disk transfer virt-v2v cold local migrations and CDI warm and remote migrations. This brings additional engineering for maintaining two paths. It's harder to debug two different flows. 
The virt-v2v transfers the disks in the sequence whereas using the CDI we can start multiple disk imports in parallel. This can improve the migration speeds.

## Fix:
Start using the CNV CDI for the disk transfer instead of virt-v2v.
MTV is already using the CNV CDI for the warm and remote migration. We just need to adjust the code to remove the virt-v2v transfer and rely on the CNV CDI to do it for us.

## Drawbacks:
- CNV CDI **requires** the VDDK, which was till now highly recommended.
- CNV CDI is not maintained inside the MTV and there might be problems escalating and backporting the patches as CNV has a different release cycle.
- Because we will be migrating all disks in parallel we need to optimise our migration scheduler as we don't want to take too much of the hosts/network resources. I have already done some optimisations in [4,5,6].

## Notes:
This change removes the usage of virt-v2v and we will only use the virt-v2v-in-place.

## Ref:
[1] https://issues.redhat.com/browse/MTV-1536
[2] https://access.redhat.com/articles/1351473
[3] https://issues.redhat.com/browse/MTV-1581
[4] https://github.com/kubev2v/forklift/pull/1088
[5] https://github.com/kubev2v/forklift/pull/1087
[6] https://github.com/kubev2v/forklift/pull/1086

### DIFF
--- a/pkg/apis/forklift/v1beta1/BUILD.bazel
+++ b/pkg/apis/forklift/v1beta1/BUILD.bazel
@@ -23,7 +23,6 @@ go_library(
         "//pkg/apis/forklift/v1beta1/provider",
         "//pkg/apis/forklift/v1beta1/ref",
         "//pkg/lib/condition",
-        "//pkg/lib/error",
         "//vendor/k8s.io/api/core/v1:core",
         "//vendor/k8s.io/apimachinery/pkg/apis/meta/v1:meta",
         "//vendor/k8s.io/apimachinery/pkg/runtime",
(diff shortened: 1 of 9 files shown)


mnecas:
@rwmjones This change removes the usage of virt-v2v for disk transfer and we will only use the virt-v2v-in-place. Not sure when we will get this in, I'm aiming for the next Y release (2.8.0) as it has some changes for the users. Just wanted to let you know.

rwmjones:
We won't be able to support this.  We only support conversions using `virt-v2v`.  We don't support usage of `virt-v2v-in-place` at all, or recommend you use it.  Basically, if you do this, you're on your own with supporting v2v.

mnecas:
Closing, right now we are not moving this direction we might in future

PR #1115, opened 2024-10-18
Add offload plugin implmentation

mnecas:
## Issue:
when migrating the VMs the VM disks need to be transfered over the network.
This is inefficient, takes additional storage and is slow. We need
an alternative way for the disk transfer.

## Design:
This PR is PoC of adding an offload plugin for the disk transfer.
The offload plugin is a specific tool that needs to be implemented
by the CSI provider. To specify the storage offload plugin the user will
need to specify it in the StorageMap destination. This will allow the
users to migrate the VM even with disks across multiple data store types.
For example, some could be managed by the offloadPlugin and some would
still go over the network if needed.

Example of the storage map with offload plugin:
```yaml
spec:
  map:
    - destination:
        offloadPlugin:
          image: 'quay.io/mnecas0/offload:latest'
          vars:
            test1: test1
            test2: test2
        storageClass: nfs-csi
      source:
        id: datastore-30
...
```
The offload plugin is started right after the CreateDataVolumes, this
the way the Kubernetes CSI will create empty PVs into which the disks can be
transfered.

### Variables
The OffloadPlugin step creates a job on the destination cluster. The job
provided offload plugin image and user-defined variables which are 
passed from the storagemap to the job. The job is started with the
following parameters:
- `HOST` = url to the esxi host
- `PLAN_NAME` = plane name
- `NAMESPACE` = namepsace where the migration is running

In addition to these variables, it also mounts the secrets to the vCenter.
The secrets are in the path `/etc/secret` and the files are:
- `accessKeyId` with a username
- `secretKey` with a password

## Note:
This change additionally requires https://github.com/kubev2v/forklift/pull/1109, because right now the
cold migration transfer is managed by the virt-v2v. The https://github.com/kubev2v/forklift/pull/1109 removes
this dependency and moves it out to the CNV CDI. This allows us to split
the transfer and conversion steps which were in the same step from the
forklift perspective. Once the disks are transferred the Forklift does the
`virt-v2v-in-place` on the disks and starts the VM.
In the same way, this step will be done also on the offload plugin as we
will transfer the disks using the offload plugin and then start
`virt-v2v-in-place` on the disks.

## TODO:
- [ ] Add design doc showing larger details, this is just PoC
- [ ] Add check if OffloadPlugin image exists
- [ ] Add check of the offload plugin disk transfer status
- [ ] Allow storage map with OffloadPlugin and without combination
- [ ] Improve the name of the offload pugin job, right now its the VM ID

### DIFF
--- a/operator/config/crd/bases/forklift.konveyor.io_storagemaps.yaml
+++ b/operator/config/crd/bases/forklift.konveyor.io_storagemaps.yaml
@@ -60,6 +60,21 @@ spec:
                           - ReadWriteMany
                           - ReadOnlyMany
                           type: string
+                        offloadPlugin:
+                          description: Offload plugin.
+                          properties:
+                            image:
+                              description: Offload plugin.
+                              type: string
+                            vars:
+                              additionalProperties:
+                                type: string
+                              description: Offload plugin variables passed to the
+                                job
+                              type: object
+                          required:
+                          - image
+                          type: object
                         storageClass:
                           description: A storage class.
                           type: string
--- a/pkg/apis/forklift/v1beta1/BUILD.bazel
+++ b/pkg/apis/forklift/v1beta1/BUILD.bazel
@@ -23,7 +23,6 @@ go_library(
         "//pkg/apis/forklift/v1beta1/provider",
         "//pkg/apis/forklift/v1beta1/ref",
         "//pkg/lib/condition",
-        "//pkg/lib/error",
         "//vendor/k8s.io/api/core/v1:core",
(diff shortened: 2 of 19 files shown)


fabiand (changes requested):
Please include a design document, either in-tree or linked.

fabiand (changes requested):
There are some open questions, let's put this on hold for a little.

mnecas:
Closing in favour of https://github.com/kubev2v/forklift/pull/1369

PR #1119, opened 2024-10-27
change pipelines to follow release-2.7

ernesgonzalez33:
Change all pipelines to follow release-2.7 and add python script in `hack` folder to make this work easy for next versions needed to be supported


ernesgonzalez33:
@mnecas you can merge

PR #1148, opened 2024-10-30
chore(KFLUXMIG-6): update on-cel expression

ergonzal22:
To avoid redundant builds, we are updating the on-cel expressions of all components


ernesgonzalez33:
@mnecas @yaacov the idea of this PR is to only build the components when needed because there are changes that affect them. So please take a look at the changes and let me know if I'm missing folders or files depending on which components need to be built. I will create another PR similar for the 2.7 release

mnecas (approved):

PR #1151, opened 2024-10-30
chore(KFLUXMIG-57): Fix on-cel-expression release-2.7

ernesgonzalez33:


ernesgonzalez33:
@yaacov @mnecas  Same as #1148 but for 2.7

PR #1153, opened 2024-10-30
chore(KFLUXMIG-13): build mtv downstream bundle

ernesgonzalez33:


build/forklift-operator-bundle/Containerfile
```diff
@@ -1,8 +1,10 @@
 FROM quay.io/konflux-ci/operator-sdk-builder@sha256:73fa40301aebf0c62903a5840918251f444212319f95a4e467eeaf0cd8c4349e as builder
 
+ARG STREAM=downstream
+
 COPY ./. /repo
 WORKDIR /repo/operator
-RUN source ./export-vars.sh && \
+RUN source ./export-vars.sh ${STREAM} && env && \
```
ernesgonzalez33, line 7: remove the env

.tekton/forklift-operator-bundle-pull-request.yaml
```diff
@@ -34,6 +34,8 @@ spec:
     value: build/forklift-operator-bundle/Containerfile
   - name: path-context
     value: .
+  - name: build-args-file
+    value: build/forklift-operator-bundle/downstream-args
```
ernesgonzalez33, line 38: Add this to the on-push as well

operator/export-vars.sh
```diff
@@ -26,13 +35,41 @@ images[VALIDATION_IMAGE]=quay.io/redhat-user-workloads/rh-mtv-1-tenant/forklift-
...
+        image="${images[$k]}"
+        new_image="${image/quay.io\/redhat-user-workloads\/rh-mtv-1-tenant\/forklift-operator/kubev2v}"
+        export "$k=$new_image"
+    done
+elif [ $stream == "downstream" ]; then
+    for k in "${!images[@]}"; do
+        image="${images[$k]}"
+        new_image="${image/forklift-operator\/[^@]*@/forklift-operator\/${replacements[$k]}@}"
+        new_image="${new_image/quay.io\/redhat-user-workloads\/rh-mtv-1-tenant\/forklift-operator/registry.redhat.io\/migration-toolkit-virtualization}"
+        export "$k=$new_image"
```
fabiand, line 68: Wouldn't it be easier to manually create two files, like `image-names-downstream.env` and `image-names-upstream.env` which are manually maintained, and then just `source $FN.env` in the two branches?

ernesgonzalez33: But you will lose the automation that creates the PRs with `quay.io/redhat-user-workloads` but then changes to the release registry when building the image. You will only have the nudges when actually releasing

mnecas: Could that automation work around two separate files? 
It would be easier if we will have 2 separate files directly showing the images which are in the bundle.

ernesgonzalez33: @gbenhaim you started with this script. Any thoughts?

gbenhaim: @ernesgonzalez33 you can have 2 different files. Just make sure to reference the upstream file in the upstream dockerfile and the d/s file in the d/s dockerfile.

ernesgonzalez33: Just discussed with @mnecas. For the downstream, they want the references to point directly to registry.redhat.io and be update with every release. Turning this PR back to draft to change that

ernesgonzalez33:
@mnecas Code splitted, PTAL

mnecas (approved):
Lgtm

mnecas:
Is it ready to get merged?

PR #1161, opened 2024-10-31
MTV-1458 | Make Guest information on vNICs is missing a warning

yaacov:
Ref: https://issues.redhat.com/browse/MTV-1458

When selecting "preserve static IP" and migrating a VM without a static IP configuration, then the backend is reporting that the VM was not run. But thi sis incorrect, the VM was run, but it just does not have a static IP configuration (nt inside, and also not repoetred by the VMw API).

Issue:
Currently if a VM does not have static IP, and we set "preserve static IP" the plan is blocked,

Fix:
Changing the condition from "critical" to "warn" will unblock the plan, while still warning the user we may have an issue.


mnecas (approved):
LGTM

PR #1172, opened 2024-11-05
Add recording rule to collect migrations data to telemetry.

bkhizgiy:
Following the request of adding recording rule in https://github.com/openshift/cluster-monitoring-operator/pull/2461

### DIFF
--- a/operator/roles/forkliftcontroller/templates/monitor/recordingrole-migrations.yml.j2
+++ b/operator/roles/forkliftcontroller/templates/monitor/recordingrole-migrations.yml.j2
@@ -1,0 +1,14 @@
+---
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: {{ metrics_rule_name }}
+  namespace: {{ app_namespace }}
+spec:
+  groups:
+  - name: mtv-migrations
+    rules:
+    - record: cluster:mtv_migrations_status_total:max
+      expr: max by(status, provider, mode, target) (mtv_migrations_status_total)
+      labels:
+        app: {{ app_name }}
--- a/operator/config/rbac/leader_election_role.yaml
+++ b/operator/config/rbac/leader_election_role.yaml
@@ -37,6 +37,7 @@ rules:
   - monitoring.coreos.com
   resources:
   - servicemonitors
+  - prometheusrules
   verbs:
   - get
   - create
--- a/operator/roles/forkliftcontroller/defaults/main.yml
+++ b/operator/roles/forkliftcontroller/defaults/main.yml
@@ -119,4 +119,5 @@ metric_service_name: "{{ app_name }}-metrics"
 metric_servicemonitor_name: "{{ app_name }}-metrics"
 metric_interval: "30s"
 metric_port_name: "metrics"
+metrics_rule_name: "{{app_name}}-migration-rules"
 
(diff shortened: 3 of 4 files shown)


yaacov (approved):

operator/roles/forkliftcontroller/templates/monitor/recordingrole-migrations.yml.j2
```diff
@@ -0,0 +1,14 @@
+---
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
```
yaacov, line 4: Q: do we need labels to identify this rule as belonging to the forklift operator application ?

fabiand:
@sradco do you mind giving this a brief look?

operator/roles/forkliftcontroller/templates/monitor/recordingrole-migrations.yml.j2
sradco: The name of the recording rule " cluster:mtv_migrations_status_total:max" includes the mtv.
I think it should be sufficient.

sradco:
Adding @avlitman to also review this.

bkhizgiy:
@sradco It's not possible to move this code under /pkg/monitoring/ as you suggested, since this is the operator code and is being built using a different target. It should be located under the operator folder. Since we are building the monitoring stack differently from CNV (using Ansible vs. Go), it should be added as a template to the operator during its creation.

We could add another folder under /templates/monitoring, but honestly, I don’t think it’s necessary.

sradco:
That was a suggestion,  and if it doesn't fit your project, please feel free to ignore.
It was valuable for CNV to be able to see all monitoring code in the same place for purposes like automatic metrics and recording rules documentation, github permissions for sig-observability team and clear code separation.
Just raising this since it's important to plan this in early states,instead of having to refactor it later(like we needed to).

operator/roles/forkliftcontroller/templates/monitor/recordingrole-migrations.yml.j2
```diff
@@ -0,0 +1,14 @@
...
+  name: {{ metrics_rule_name }}
+  namespace: {{ app_namespace }}
+spec:
+  groups:
+  - name: mtv-migrations
+    rules:
+    - record: cluster:mtv_migrations_status_total:max
+      expr: max by(status, provider, mode, target) (mtv_migrations_status_total)
+      labels:
+        app: {{ app_name }}
```
avlitman, line 14: I would add also annotations
```
      annotations:
        description: "The maximum number of migration statuses, grouped by status, provider, mode, and target."
```

avlitman: sorry for delayed review was sick.. and nice work!

operator/roles/forkliftcontroller/templates/monitor/recordingrole-migrations.yml.j2
```diff
@@ -0,0 +1,14 @@
...
+kind: PrometheusRule
+metadata:
+  name: {{ metrics_rule_name }}
+  namespace: {{ app_namespace }}
+spec:
+  groups:
+  - name: mtv-migrations
+    rules:
+    - record: cluster:mtv_migrations_status_total:max
+      expr: max by(status, provider, mode, target) (mtv_migrations_status_total)
```
avlitman, line 12: are we sure we need all 4 labels? telemetry team is very strict about the number of records added per query. if we can aggregate by less it will be better.

sradco: These labels are needed. It should be fine.

yaacov:
@sradco @avlitman hi, is this RP lgtm-ed by you ?

PR #1175, opened 2024-11-07
MTV-1645 Change default network annotation.

mrnold:
The importer pod can get stuck creating if there are multiple networks and the wrong default network override annotation is configured. This changes the annotation as requested in [MTV-1645](https://issues.redhat.com/browse/MTV-1645).

I reproduced the bug without the change, and managed a successful transfer with the change applied.

### DIFF
--- a/pkg/controller/plan/kubevirt.go
+++ b/pkg/controller/plan/kubevirt.go
@@ -54,7 +54,7 @@ import (
 // Annotations
 const (
 	// Transfer network annotation (value=network-attachment-definition name)
-	AnnDefaultNetwork = "v1.multus-cni.io/default-network"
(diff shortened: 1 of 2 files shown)


yaacov (approved):

pkg/controller/plan/kubevirt.go
```diff
@@ -54,7 +54,7 @@ import (
 // Annotations
 const (
 	// Transfer network annotation (value=network-attachment-definition name)
-	AnnDefaultNetwork = "v1.multus-cni.io/default-network"
+	AnnDefaultNetwork = "k8s.v1.cni.cncf.io/networks"
```
yaacov, line 57: Q: do we also want to change the variable name ? since it's not the "default" network annotation anymore ?

PR #1184, opened 2024-11-13
Reorder warm migration flow for vSphere VMs

mansam:
When performing a warm migration of a vSphere VM, the current version of Forklift creates the snapshot for each precopy, gets the change ID for the previous snapshot, and then deletes the previous snapshot.  Per the [VMware documentation](https://docs.vmware.com/en/VMware-vSphere/8.0/vsphere-vddk-programming-guide/GUID-C9D1177E-009E-4366-AB11-6CD99EF82B5B.html), the correct flow is to record the change IDs for the old snapshot, delete it, and then create the new snapshot.

This PR implements the correct flow by adding vSphere-specific steps to the warm migration itinerary to ensure that old snapshots are cleaned up and change IDs are recorded before creating new snapshots. Due to the change in flow it's now necessary to store the change IDs somewhere between itinerary phases, so a list of disk/changeid pairs has been added to the `Precopy` struct. This change required the CRs to be regenerated.

Correcting the warm migration flow appears, at least in some cases, to address the filesystem corruption documented in https://issues.redhat.com/browse/MTV-1656.

This PR also removes some dead code branches left over from a very old version of Forklift. The `VSphereIncrementalBackup` feature flag is always assumed to be on as this is necessary for the proper warm migration flow, so the alternate code path no longer needs to exist. See https://issues.redhat.com/browse/MTV-1665


pkg/controller/plan/migration.go
```diff
@@ -501,7 +512,12 @@ func (r *Migration) removeWarmSnapshots(vm *plan.VMStatus) {
 	if vm.Warm == nil {
 		return
 	}
-	if err := r.provider.RemoveSnapshots(vm.Ref, vm.Warm.Precopies, r.kubevirt.loadHosts); err != nil {
+	n := len(vm.Warm.Precopies)
```
mnecas, line 515: np: might be a better name now something like `removeLastWarmSnapshot`

mnecas (approved):
I tested it in my env and the warm migration passed and I could boot the VM without a problem.

LGTM!

PR #1189, opened 2024-11-15
VMware warm migration - Fix snapshot cleanup

mnecas:
Issue:
The `CreateSnapshot` is after `RemovePreviousSnapshot` so with the warm migration, we always skipped the snapshot cleanup step. 

Fix:
Get the next migration phase dynamically. This way the VMware migration will run `RemovePreviousSnapshot` and the other migrations will run `CreateSnapshot``.

Ref:
https://issues.redhat.com/browse/MTV-1656


mansam (approved):
LGTM

PR #1193, opened 2024-11-18
MTV-1656 | Disable snapshot consolidation

mnecas:
Issue: When deleting the snapshot the users with slow storage could get data corruption due to the consolidation process. The consolidation writes the snapshot overlay to the disk and if the user has slow storage it can take a long time. However, the Forklift Controller does not check if the consolidation is finished and creates a new snapshot. This new snapshot could be created in the middle of the writing process and it would not see the changes in the underlying image from the consolidation.

Wrong order
1. Delete snapshot
2. Create snapshot
3. Consolidate

Correct order
1. Delete snapshot
2. Consolidate
3. Create snapshot

Fix: Disable consolidation in the middle of migration and consolidate in RemoveFinalSnapshot phase. This will make the warm migration faster as we don't need to wait for the consolidation but at the same time we will consolidate at the end of migration so we won't cause any issues on the users env.

Ref: https://issues.redhat.com/browse/MTV-1656

### DIFF
--- a/pkg/controller/plan/adapter/base/doc.go
+++ b/pkg/controller/plan/adapter/base/doc.go
@@ -106,7 +106,7 @@ type Client interface {
 	// Create a snapshot of the source VM.
 	CreateSnapshot(vmRef ref.Ref, hostsFunc util.HostsFunc) (string, error)
 	// Remove a snapshot.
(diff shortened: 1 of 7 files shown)


mansam (approved):
LGTM

PR #1195, opened 2024-11-18
Add WaitForResult when removing the vmware snapshot

mnecas:
This PR adds a WaitForResults when removing the VMware snapshots so we will get actual results in case something breaks we will be able to see it in the logs.


mnecas:
Fixed, added state to the log

PR #1196, opened 2024-11-19
fix: fix EC issues with rpm task and update build with current activation keys format

ernesgonzalez33:


ernesgonzalez33:
On my side, this can be merged!

mnecas (approved):

PR #1204, opened 2024-11-20
feat: add ova ProductSection to ova provider server resp

arturshadnik:
This PR extends the Inventory service and OVA provider server to also return the ProductSection as part of the VM payload. ProductSection can contain user configurable fields that need to be set in order for the VM to run properly. Only the relevant subset of the original [ProductSection](https://github.com/vmware/govmomi/blob/main/ovf/envelope.go#L77) is added.

Example response:
```json
"product": {
        "Property": [
            {
                "Key": "va-ssh-public-key",
                "Type": "string",
                "Qualifiers": null,
                "UserConfigurable": true,
                "Default": null,
                "Password": null,
                "Label": "Set the SSH public key allowed to access the appliance",
                "Description": "This will enable the SSHD service and configure the specified public key for the user 'bitnami'",
                "Values": null
            },
            {
                "Key": "user-data",
                "Type": "string",
                "Qualifiers": null,
                "UserConfigurable": true,
                "Default": null,
                "Password": null,
                "Label": "User data to be made available inside the instance",
                "Description": "This allows to pass any text to the appliance. It will be executed if it starts with a shebang (\"#!\"). The value should be encoded in base64",
                "Values": null
            }
        ]
    }
```

### DIFF
--- a/cmd/ova-provider-server/ova-provider-server.go
+++ b/cmd/ova-provider-server/ova-provider-server.go
@@ -109,6 +109,7 @@ type VirtualSystem struct {
 		OsType      string `xml:"osType,attr"`
 	} `xml:"OperatingSystemSection"`
 	HardwareSection VirtualHardwareSection `xml:"VirtualHardwareSection"`
+	Product         ProductSection         `xml:"ProductSection"`
 }
 
 type Envelope struct {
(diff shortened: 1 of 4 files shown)


mnecas (commented):
Thanks for the PR! Generally looks good at first glance. But what is the usage? Or why do we need to start collecting it? I'm not against it just wonder what is the root issue

arturshadnik:
Thanks for the feedback @mnecas, this is one part of a change to allow configuring `ovf-env` during OVA migration. I will open an issue with more detail and to track all required changes.

arturshadnik:
> Thanks for the feedback @mnecas, this is one part of a change to allow configuring `ovf-env` during OVA migration. I will open an issue with more detail and to track all required changes.

Created an issue here https://github.com/kubev2v/forklift/issues/1216, let me know your thoughts.

yaacov:
Closing as the code changes and this PR needs to align with new code, feel free to open a new PR aligned with the new code if needed

PR #1208, opened 2024-11-22
fix: change make install target to use the right directory name

arturshadnik:
`make install` was broken because the directory name had an `s` at the end.


yaacov (approved):
Thanks

mnecas (approved):

mnecas:
Thank you for the PR!
LGTM

PR #1209, opened 2024-11-25
MTV-1695 | Add missing NIC for opaque network

mnecas:
Issue: When user has NSXT managed network it is represented as opaque network and the VM has opaque nic.

Fix: Add opaque network model mapping to inventory

Ref: https://issues.redhat.com/browse/MTV-1695

### DIFF
--- a/pkg/controller/provider/web/vsphere/network.go
+++ b/pkg/controller/provider/web/vsphere/network.go
@@ -191,6 +191,8 @@ func (r *Network) With(m *model.Network) {
 	case model.NetDvPortGroup:
 		r.DVSwitch = &m.DVSwitch
 		r.Key = m.Key
+	case model.OpaqueNetwork:
+		r.Key = m.Key
 	case model.NetDvSwitch:
 		r.Host = m.Host
 	}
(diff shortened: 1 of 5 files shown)


pkg/controller/provider/web/vsphere/network.go
```diff
@@ -188,7 +188,7 @@ func (r *Network) With(m *model.Network) {
 	switch m.Variant {
 	case model.NetStandard:
 		r.Tag = m.Tag
-	case model.NetDvPortGroup:
+	case model.NetDvPortGroup, model.OpaqueNetwork:
 		r.DVSwitch = &m.DVSwitch
```
mansam, line 192: `model.OpaqueNetwork` should probably have its own case since it won't have a value for DVSwitch.

mansam (changes requested):
Implementation looks good other than one small detail, but I don't have an environment with an opaque network to test this with.

pkg/controller/provider/web/vsphere/network.go
mnecas: Updated, thanks!

mansam (approved):
LGTM

PR #1224, opened 2024-11-27
Fetch Vsphere instanceUuid

machacekondra:
This PR add fetching of the instanceUuid to About model.


mnecas:
Please link some issues showing why/where you need it.
It does not need to be issue directly on Forklift but could be from your project.

machacekondra:
As part of ticket: https://issues.redhat.com/browse/ECOPROJECT-2376
We need to have unique identification of the Vcenter.
So far @tupyy solved it inside our code here: https://github.com/kubev2v/migration-planner/pull/81

We need it so we can distinguish where the generated OVA is connected to so we don't gather it twice for same Vcenter, so Vcenter UUID is unique identification for us.

PR #1226, opened 2024-11-27
create enhancement proposal for OVF environment

arturshadnik:


arturshadnik:
Hi @mnecas, just wanted to check and see if there are any comments/feedback on this proposal?

yaacov:
@mansam hi, can you take a look ? is this the way we want to go ?

yaacov:
Hi, the OVA feature is developing in new directions, closing to reduce noise in the OVA dev.

TylerGillson:
Hey @yaacov. Could you possibly share links to any new issues or design docs that outline the proposed direction?

yaacov:
@TylerGillson hi, thank you for asking, no, I do not have design docs, we have  https://github.com/kubev2v/forklift/pull/1962 currently in draft mode.

About closing this design doc, this PR is here for almost half a year abd was not picked up for grooming or implementation, it's stale.

mnecas:
I don't think that the #1962 and this PR are the same topic. The #1962 is focusin on importing the OVAs to the kubernetes.
But this PR is about the customization of the OVAs with the vmtoolsd-shim.
We have https://issues.redhat.com/browse/MTV-2465 around this story and we are still doing investigation. The vmtoolsd is complicated as it's VMware tool.

TylerGillson:
Thanks, @mnecas. Yes, `vmtoolsd` is complicated. The shim we proposed is not 😄 It's very lightweight.

mnecas:
Even that is complicated for us as both of them, the shim and vmtoolsd, are black boxes which configure the guest.
And we would need to be injecting the binaries into the guest.

TylerGillson:
Only one binary would need to be injected into the guest: the vmtoolsd shim. All the shim does is expose the ovfenv XML configuration to any process running on the guest that requests it via `vmtoolsd --cmd "info-get guestinfo.ovfenv"`. It supports no other `vmtoolsd` commands.

So the shim doesn't "configure the guest". It just sits on the PATH in replacement of the standard `vmtoolsd` binary and serves XML content to anything that requests it. The XML itself would be mounted onto the guest via a ConfigMap under a well-known path.

Anyhow, very curious to see what alternative solutions you end up pursuing!

mrnold:
I want to keep this open for now. I started looking into it but got sucked into something else, hopefully I can take a closer look soon. It seems like a pretty good solution from what I know so far.

mnecas:
hi, closing as stale, we have been looking into this area but right now i prefer not creating middleware that would translate vmware vmtoolsd

PR #1243, opened 2024-12-03
MTV-1717 | Wait for DV status to not be paused

mnecas:
Issue: When CDI is managing lot of DVs at once it can take some time before the CDI reconciles the DVs status and start the import. In the function `updateCopyProgress` we check the DV status and if it is paused we do continue and as next phase we have snapshot removal. So if the DV is still in paused state and we start the cutover and MTV skips over the phase thinking it already finished as the DV is in paused.

Fix: Add WaitForDataVolumesStatus phase to wait until the DVs do not have the paused status.

Ref: https://issues.redhat.com/browse/MTV-1717

https://github.com/kubev2v/forklift/blob/ea83264881b3dcd4a88dd627c1ca75da2483f408/pkg/controller/plan/migration.go#L1573-L1575

### DIFF
--- a/operator/roles/forkliftcontroller/defaults/main.yml
+++ b/operator/roles/forkliftcontroller/defaults/main.yml
@@ -34,6 +34,7 @@ controller_precopy_interval: 60
 controller_snapshot_removal_timeout_minuts: 120
 controller_snapshot_status_check_rate_seconds: 10
 controller_cleanup_retries: 10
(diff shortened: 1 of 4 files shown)


mansam (approved):

mansam (commented):
Questions inline.

mnecas:
Merging for 2.7.6 release any notes I'll address in further PRs

PR #1244, opened 2024-12-03
Look for annotation on NAD to set transfer route

mansam:
The Plan controller will now look for an annotation `forklift.konveyor.io/route` on the NetworkAttachmentDefinition that was specified as the transfer network for the Plan.

The annotation is expected to have a single IP address as its value, which will be set as the default route when configuring the secondary network on the importer pods.

The annotation is optional, but if it is present then the Plan validator will ensure that its value is a valid IP address.

Replaces https://github.com/kubev2v/forklift/pull/1241

### DIFF
--- a/pkg/controller/plan/kubevirt.go
+++ b/pkg/controller/plan/kubevirt.go
@@ -56,6 +56,10 @@ import (
 
 // Annotations
 const (
+	// Legacy transfer network annotation (value=network-attachment-definition name)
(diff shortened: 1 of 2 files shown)


mnecas (commented):
Yeah right direction but I'm bit worried about backwards compatibility. Could we also support the old multus annotation when the route is not specified?

mnecas:
Ref: https://issues.redhat.com/browse/MTV-1645

PR #1253, opened 2024-12-09
MTV-1753 | Add wait for snapshot removal timeout

mnecas:
Issue: The migration fails with `Another task is already in progress`. This is caused because of the snapshot creation is started right after snapshot removal and the snapshot removal did not finish due to the consolidation.

Fix: I have added a timeout which users can control using the `controller_snapshot_removal_check_retries` variable. This will wait some time right after snapshot removal. This is just a temporary solution to unblock the users. I'll fix it properly using the VM events and tasks, but that will require a lot of changes in the controller inventory and main.

Ref: https://issues.redhat.com/browse/MTV-1753


mnecas:
Merging for 2.7.6 release any notes I'll address in further PRs

PR #1260, opened 2024-12-11
MTV-1632 | Add secureboot to vsphere

mnecas:
Issue: When creating the VM from vSphere on kubevirt the MTV always
defaulted the secureboot to false.

Fix: Add the secureboot to the inventory and to main controller to pass
it to the KubeVirt.

Ref: https://issues.redhat.com/browse/MTV-1632

### DIFF
--- a/pkg/controller/plan/adapter/vsphere/builder.go
+++ b/pkg/controller/plan/adapter/vsphere/builder.go
@@ -668,15 +668,9 @@ func (r *Builder) mapFirmware(vm *model.VM, object *cnv.VirtualMachineSpec) {
 	}
 	switch vm.Firmware {
 	case Efi:
-		// We don't distinguish between UEFI and UEFI with secure boot, but we anyway would have
-		// disabled secure boot, even if we knew it was enabled on the source, because the guest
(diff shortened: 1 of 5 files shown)


pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -668,15 +668,9 @@ func (r *Builder) mapFirmware(vm *model.VM, object *cnv.VirtualMachineSpec) {
 	}
 	switch vm.Firmware {
 	case Efi:
-		// We don't distinguish between UEFI and UEFI with secure boot, but we anyway would have
-		// disabled secure boot, even if we knew it was enabled on the source, because the guest
-		// OS won't be able to boot without getting the NVRAM data. By starting the VM without
-		// secure boot we ease the procedure users need to do in order to make a guest OS that
-		// was previously configured with secure boot bootable.
```
mnecas, line 675: We need to make difference between measured boot and secureboot, the comment about the NVRAM is true but that contains the keys for the measured boot. Which still won't work, but we can manage the secureboot itself.

PR #1266, opened 2024-12-12
Allow setting storage mappings for OVA disks

mansam:
* Exposes OVA appliance disks as mappable storage in the OVA inventory collector.
* Permits mapping any or all OVA appliance disks to their own destination storage classes.
* Any disks that are not specifically mapped are assigned to a default storage class, which must be given a storage mapping.
* No UI changes should be required. The default mapping will continue to appear in the UI dropdown as the only mapping that must be present to continue with the the plan wizard.

Implements https://github.com/kubev2v/forklift/pull/1213

Note: This replaces the old default mapping using the provider's UID with a new one using `default` as the ID, to make the mapping easier to read and write without the UI. This causes a backwards incompatibility with old plans since the ID for the default mapping is different. I think the change is worth it for clarity but I'm not tied to it and it could be changed back to the provider's UID. 

image
image
image


mnecas (approved):
LGTM, before merging this we need to sync with the UI and docs

mansam:
Storages:
image
Disks:
image
I've pushed a forklift-controller image including these changes to my quay account `quay.io/slucidi/forklift-controller:ova-multidisk`

mnecas (approved):

PR #1268, opened 2024-12-13
MTV-1493 | Add missing resource limits and requests

mnecas:
Issue:
If the user sets ClusterResourceQuota the Forklift will start failing as it does not have the limits or requests on the pods which are created form the Forklift Controller.

Fix:
Add a new parameters to Forklift operator which can be configured depedning on the user env.

Ref: https://issues.redhat.com/browse/MTV-1493

### DIFF
--- a/pkg/controller/plan/kubevirt.go
+++ b/pkg/controller/plan/kubevirt.go
@@ -773,6 +773,16 @@ func (r *KubeVirt) createPodToBindPVCs(vm *plan.VMStatus, pvcNames []string) (er
 					// In that case, we could benefit from pulling the image of the conversion pod, so it will be present on the node.
 					Image:   Settings.Migration.VirtV2vImage,
 					Command: []string{"/bin/sh"},
+					Resources: core.ResourceRequirements{
+						Requests: core.ResourceList{
+							core.ResourceCPU:    resource.MustParse(Settings.Migration.VirtV2vContainerRequestsCpu),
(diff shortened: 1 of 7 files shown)


pkg/controller/plan/kubevirt.go
```diff
@@ -1738,6 +1748,16 @@ func (r *KubeVirt) guestConversionPod(vm *plan.VMStatus, vmVolumes []cnv.Volume,
 					MountPath: "/opt",
 				},
 			},
+			Resources: core.ResourceRequirements{
```
mnecas, line 1751: Init container only for mount, don't think we need it to be configurable.

pkg/controller/plan/validation.go
```diff
@@ -1016,6 +1017,16 @@ func createVddkCheckJob(plan *api.Plan) *batchv1.Job {
 					Drop: []core.Capability{"ALL"},
 				},
 			},
+			Resources: core.ResourceRequirements{
```
mnecas, line 1020: Init container only for mount, don't think we need it to be configurable.

pkg/controller/plan/validation.go
```diff
@@ -1057,7 +1068,17 @@ func createVddkCheckJob(plan *api.Plan) *batchv1.Job {
 					InitContainers:  initContainers,
 					Containers: []core.Container{
 						{
-							Name:  "validator",
+							Name: "validator",
```
mnecas, line 1071: Validation contianer only for mount, don't think we need it to be configurable.

PR #1280, opened 2024-12-18
MTV-1804 | Implement VDDK AIO buffer configuration

mnecas:
Issue:
The scale and perf team found a way how to improve the transfer speeds. Right now the only way to enable this feature is to set the v2v extra vars. The v2v extra vars pass the configuration to the virt-v2v and virt-v2v-in-place. The v2v extra vars configuration is general and not specific for VDDK. This causes the warm migration which uses the virt-v2v-in-place to fail as it does not use any VDDK parameters. Those parameters should be passed to the CNV CDI instead.

Fix:
Add a way to easily enable and configure the AIO.
This feature is VDDK and provider-specific as it requires to have specific vSphere and VDDK versions. So we can't enable this feature globally nor by default. So this PR adds the configuration to the Provider spec settings and create a configmap with the necessary configuration and either mounts the configmap to the guest conversion pod for cold migration or passes the configmap name to the CDI DV annotation.

Example:
```
apiVersion: forklift.konveyor.io/v1beta1
kind: Provider
metadata:
  name: vsphere
  namespace: forklift
spec:
  settings:
    sdkEndpoint: vcenter
    vddkInitImage: 'vddk'
    useVddkAioOptimization: 'true'
    vddkConfig: |
      VixDiskLib.nfcAio.Session.BufSizeIn64K=16
      VixDiskLib.nfcAio.Session.BufCount=4
  type: vsphere
```

Ref:
- https://issues.redhat.com/browse/MTV-1804
- https://github.com/kubevirt/containerized-data-importer/pull/3572
- https://docs.redhat.com/en/documentation/migration_toolkit_for_virtualization/2.7/html-single/installing_and_using_the_migration_toolkit_for_virtualization/index#mtv-aio-buffer_mtv

### DIFF
--- a/pkg/apis/forklift/v1beta1/provider.go
+++ b/pkg/apis/forklift/v1beta1/provider.go
@@ -17,6 +17,8 @@ limitations under the License.
 package v1beta1
 
 import (
+	"strconv"
+
 	libcnd "github.com/konveyor/forklift-controller/pkg/lib/condition"
 	core "k8s.io/api/core/v1"
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -59,10 +61,12 @@ const (
 
 // Provider settings.
 const (
-	VDDK    = "vddkInitImage"
-	SDK     = "sdkEndpoint"
-	VCenter = "vcenter"
-	ESXI    = "esxi"
+	VDDK                   = "vddkInitImage"
+	SDK                    = "sdkEndpoint"
+	VCenter                = "vcenter"
+	ESXI                   = "esxi"
+	UseVddkAioOptimization = "useVddkAioOptimization"
+	VddkConfig             = "vddkConfig"
 )
 
 const OvaProviderFinalizer = "forklift/ova-provider"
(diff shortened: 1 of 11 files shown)


pkg/controller/plan/kubevirt.go
```diff
@@ -641,6 +660,83 @@ func (r *KubeVirt) EnsureDataVolumes(vm *plan.VMStatus, dataVolumes []cdi.DataVo
...
+	if r.Source.Provider.UseVddkAioOptimization() {
+		buffSize := r.Source.Provider.Spec.Settings[api.VddkAioBufSize]
+		if buffSize == "" {
+			buffSize = VddkAioBufSizeDefault
+		}
+		buffCount := r.Source.Provider.Spec.Settings[api.VddkAioBufCount]
+		if buffCount == "" {
+			buffCount = VddkAioBufCountDefault
+		}
+		data["vddk-config-file"] = fmt.Sprintf(
```
mrnold, line 674: Should we make this the official ConfigMap key name? In the CDI pull request, I mount the ConfigMap as a volume and then look for the first file in that volume, but it seems like it would be less trouble to just look for this fixed name.

mnecas: yeah works for me

mnecas:
@mrnold do you have some comments? thinking to get this in to 2.8

mrnold:
> @mrnold do you have some comments? thinking to get this in to 2.8

The only question I have is about the likelihood of needing to add more configuration options at some point. Did you consider something like a list of options under spec.settings.vddkConfig instead of having vddkAioBufSize and vddkAioBufCount fixed under spec.settings? That way we could copy the whole list into the VDDK configuration file, and customers would not have to wait for a new MTV release to use new options.

On the other hand, that would probably make for a worse UI, and it would be very easy to mess up imports with invalid options. So I would be fine merging as-is and waiting to see if this ever becomes a real problem.

pkg/apis/forklift/v1beta1/provider.go
```diff
@@ -147,3 +153,16 @@ func (p *Provider) HasReconciled() bool {
 func (p *Provider) RequiresConversion() bool {
 	return p.Type() == VSphere || p.Type() == Ova
 }
+
+// This provider requires VM guest conversion.
```
mrnold, line 157: Probably this needs a different comment.

PR #1281, opened 2024-12-19
MTV-1744 | Allow dots in the VM names

mnecas:
Issue:
Right now we are not allowing to keep the VM names which have the dots in the name but we replace the dots with dashes.

Fix:
Use the IsDNS1123Subdomain validation instead of IsDNS1123Label and ignore the dots in the replacemnet of the new name.

Ref:
https://issues.redhat.com/browse/MTV-1744

### DIFF
--- a/pkg/controller/plan/vm_name_handler.go
+++ b/pkg/controller/plan/vm_name_handler.go
@@ -33,8 +33,8 @@ func (r *KubeVirt) changeVmNameDNS1123(vmName string, vmNamespace string) (gener
 
 // changes VM name to match DNS1123 RFC convention.
 func changeVmName(currName string) string {
-	var underscoreExcluded = regexp.MustCompile("[_.]")
-	var nameExcludeChars = regexp.MustCompile("[^a-z0-9-]")
+	var underscoreExcluded = regexp.MustCompile("^[.]|[_]|[.]$")
+	var nameExcludeChars = regexp.MustCompile("[^a-z0-9-.]")
 
 	newName := strings.ToLower(currName)
 	if len(newName) > NameMaxLength {
--- a/pkg/controller/plan/migration.go
+++ b/pkg/controller/plan/migration.go
@@ -773,7 +773,7 @@ func (r *Migration) execute(vm *plan.VMStatus) (err error) {
 			err = nil
 			break
 		}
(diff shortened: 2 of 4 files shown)


pkg/controller/plan/vm_name_handler.go
```diff
@@ -33,7 +33,7 @@ func (r *KubeVirt) changeVmNameDNS1123(vmName string, vmNamespace string) (gener
 
 // changes VM name to match DNS1123 RFC convention.
 func changeVmName(currName string) string {
-	var underscoreExcluded = regexp.MustCompile("[_.]")
+	var underscoreExcluded = regexp.MustCompile("[_]")
```
yaacov, line 36: we also need to replace or remove  remove "." if it is in the start or end of the name

https://github.com/kubernetes/apimachinery/blob/master/pkg/util/validation/validation.go#L206

mnecas: hmm that's get tricky

yaacov: remove the "." if they are in the start of end ?

fabiand:
@mnecas is this something we can close before 2.8.0?

fabiand:
Does kube permit dots in object names?

fabiand:
Thanks, so what we are saying: Oject names must not contain dots.

Right?

fabiand:
> The name of a Pod must be a valid [DNS subdomain](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names) value, but this can produce unexpected results for the Pod hostname. For best compatibility, the name should follow the more restrictive rules for a [DNS label](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names).

pkg/controller/plan/vm_name_handler.go
mnecas: I have updated it with `^[.]|[_]|[.]$` so we will replace the dots in start or end by dashes

yaacov (approved):
LGTM , need to update the tests

yaacov:
lgtm with the new test

mnecas:
Ah thanks! updated the tests.

Tested manually e2e, you can see the VM name has a dot now, but the name in the VM list is still with dash. I think it's from the UI, created https://issues.redhat.com/browse/MTV-2071
image

PR #1283, opened 2024-12-20
MTV-1610: Avoid uncommon potential inventory crash.

mrnold:
Assuming I have matched up the right code correctly, MTV-1610 shows an inventory crash pointing here: https://github.com/kubev2v/forklift/blob/v2.7.2/pkg/controller/provider/container/vsphere/model.go#L660

This pull request just adds a nil check on the DnsConfig field that appears to cause the problem. I have not been able to configure an IP stack without DNS in vSphere, but this is low risk for a speculative fix. It is pretty much copied from the previous fGuestNet case, and I don't see this DnsConfig used anywhere.


mansam (approved):
LGTM

mnecas (approved):

PR #1287, opened 2024-12-30
MTV-1856: Protect OVA URL parsing

yaacov:
**Issue:**
we do not handle the case of missing ":" in the NFS URL
this will create an unhanded error:

{"level":"info","ts":"2024-12-29 10:46:32.872","msg":"Observed a panic in reconciler: runtime error: index out of range [1] with length 1","controller":"provider","object":\{"name":"ova-server","namespace":"ova-server"}
,"namespace":"ova-server","name":"ova-server","reconcileID":"6322e786-c0cd-42eb-b05c-32d644704caa"}

**Fix:**
check for the number of splits when splitting the URL string using ":"


pkg/controller/provider/ova-setup.go
```diff
@@ -70,6 +70,13 @@ func (r Reconciler) CreateOVAServerDeployment(provider *api.Provider, ctx contex
 
 func (r *Reconciler) createPvForNfs(provider *api.Provider, ctx context.Context, pvNamePrefix string) (pv *core.PersistentVolume, err error) {
 	splitted := strings.Split(provider.Spec.URL, ":")
+	if len(splitted) < 2 {
+		err = fmt.Errorf("invalid provider URL format: %s", provider.Spec.URL)
```
yaacov, line 74: can the URL be a secret we don't want to expose in the logs ?

mnecas (approved):

pkg/controller/provider/ova-setup.go
mnecas: I think it should be okay

PR #1289, opened 2024-12-30
MTV-1682 | Mutate ESXi secret before testing connection

yaacov:
Issue:
When testing ESXi Host connection, we don't take into account the mutation of the secret via the mutation hook.

Fix:
a. use the ESXi provider user and password when testing connection in the admission hook, like the secret will be once it's mutated by the mutation hook.
b. when checking for nil user field, also check for empty string to allow for tooling that sand "" instead of nil.

Images:
Before:
MTV-1682-before

After:
https://github.com/user-attachments/assets/4aa589af-075e-4ec9-a0db-7b3d2375339a

Ref:
https://issues.redhat.com/browse/MTV-1682

### DIFF
--- a/pkg/forklift-api/webhooks/validating-webhook/admitters/secret-admitter.go
+++ b/pkg/forklift-api/webhooks/validating-webhook/admitters/secret-admitter.go
@@ -126,7 +126,7 @@ func (admitter *SecretAdmitter) validateProviderSecret() *admissionv1.AdmissionR
 func (admitter *SecretAdmitter) validateHostSecret() *admissionv1.AdmissionResponse {
 	if hostName, ok := admitter.secret.GetLabels()["createdForResource"]; ok {
 		if _, ok := admitter.secret.Data["user"]; !ok {
-			err := errors.New("Missing credentials on Host secret")
+			err := errors.New("missing credentials on Host secret")
 			return webhookutils.ToAdmissionResponseError(err)
 		}
 		tested, err := admitter.testConnectionToHost(hostName)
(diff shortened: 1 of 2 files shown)


pkg/forklift-api/webhooks/validating-webhook/admitters/secret-admitter.go
```diff
@@ -126,7 +126,7 @@ func (admitter *SecretAdmitter) validateProviderSecret() *admissionv1.AdmissionR
 func (admitter *SecretAdmitter) validateHostSecret() *admissionv1.AdmissionResponse {
 	if hostName, ok := admitter.secret.GetLabels()["createdForResource"]; ok {
 		if _, ok := admitter.secret.Data["user"]; !ok {
-			err := errors.New("Missing credentials on Host secret")
+			err := errors.New("missing credentials on Host secret")
```
yaacov, line 129: unrelated nitpick, error should not be capitalized

pkg/forklift-api/webhooks/validating-webhook/admitters/secret-admitter.go
```diff
@@ -171,6 +171,27 @@ func (admitter *SecretAdmitter) buildProviderCollector(providerType *api.Provide
 	}
 }
 
+func (admitter *SecretAdmitter) ensureEsxiCredentials(provider *api.Provider) (*core.Secret, error) {
```
yaacov, line 174: mimic the mutation hook that check for data["user"] and then for api.SDK

mnecas:
The patch looks good but I'm a bit confused as to why we trigger the secret admitter.
We are storing the network IP inside the secret?

yaacov:
> We are storing the network IP config inside the secret?

Yes we store the url and ip in the secret, I think the logic is that if we have the IP and the credentials in the secret we can validate the secret without needing a the provider/host allowing to first create a secret and then the provider

mnecas (approved):
lgtm

PR #1291, opened 2025-01-06
MTV-1865 | Remove the q35 machine type

mnecas:
Issue:
VMs imported from vSphere always have the q35 machine type set. but the machine type is "floating", and thus the VM abi can change between reboots (i.e. update from rhel 8 to rhel9 or 10 on the host side).

Fix:
Remove the q35 machine type.

Ref: https://issues.redhat.com/browse/MTV-1865

### DIFF
--- a/pkg/controller/plan/adapter/openstack/builder.go
+++ b/pkg/controller/plan/adapter/openstack/builder.go
@@ -407,8 +407,6 @@ func (r *Builder) mapInput(vm *model.Workload, object *cnv.VirtualMachineSpec) {
 }
 
 func (r *Builder) mapResources(vm *model.Workload, object *cnv.VirtualMachineSpec, usesInstanceType bool) {
-	// KubeVirt supports Q35 or PC-Q35 machine types only.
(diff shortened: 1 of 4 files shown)


pkg/controller/plan/adapter/openstack/builder.go
```diff
@@ -407,8 +407,6 @@ func (r *Builder) mapInput(vm *model.Workload, object *cnv.VirtualMachineSpec) {
 }
 
 func (r *Builder) mapResources(vm *model.Workload, object *cnv.VirtualMachineSpec, usesInstanceType bool) {
-	// KubeVirt supports Q35 or PC-Q35 machine types only.
-	object.Template.Spec.Domain.Machine = &cnv.Machine{Type: Q35}
```
fabiand, line 411: The `Q35` definition. Should we drop it as well?

mnecas: ah nice catch, I have also removed additional OpenStack machinetype settings

PR #1293, opened 2025-01-07
MTV-1866 | Remove the requests.memory from the builder

mnecas:
Issue:
MTV is importing VMs with the requests.memory field and the memory.guest field set. This is problematic as it prevents memor-overcommit and is causing un-necessary memory pressure on VMs in the worst case leading to OOM sitiations.

Fix:
Remove the requests.memory from the builders

Ref: https://issues.redhat.com/browse/MTV-1866

### DIFF
--- a/pkg/controller/plan/adapter/vsphere/builder.go
+++ b/pkg/controller/plan/adapter/vsphere/builder.go
@@ -647,11 +647,6 @@ func (r *Builder) setMachine(object *cnv.VirtualMachineSpec) {
 func (r *Builder) mapMemory(vm *model.VM, object *cnv.VirtualMachineSpec) {
 	memoryBytes := int64(vm.MemoryMB) * 1024 * 1024
 	reservation := resource.NewQuantity(memoryBytes, resource.BinarySI)
-	object.Template.Spec.Domain.Resources = cnv.ResourceRequirements{
-		Requests: map[core.ResourceName]resource.Quantity{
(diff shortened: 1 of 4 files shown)


pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -647,11 +647,6 @@ func (r *Builder) setMachine(object *cnv.VirtualMachineSpec) {
 func (r *Builder) mapMemory(vm *model.VM, object *cnv.VirtualMachineSpec) {
 	memoryBytes := int64(vm.MemoryMB) * 1024 * 1024
 	reservation := resource.NewQuantity(memoryBytes, resource.BinarySI)
-	object.Template.Spec.Domain.Resources = cnv.ResourceRequirements{
-		Requests: map[core.ResourceName]resource.Quantity{
-			core.ResourceMemory: *reservation,
-		},
-	}
 	object.Template.Spec.Domain.Memory = &cnv.Memory{Guest: reservation}
```
fabiand, line 650: `reservation` is this the real memory size of the guest?

Or something like ballooned memory.

mnecas: the real memory size of the guest

fabiand: +1

yaacov (approved):
lgtm

PR #1297, opened 2025-01-13
MTV-1864 | Add luks support to the warm migration

mnecas:
Issue:
When creating the warm migration with luks encrypted disks the migration does not even allow us to create the plan. This is because of the validation hook. But the virt-v2v-in-place supports the `--key` parameter and we are already passing it to it.
image

Fix:
Remove the warm check in the plan validation.

Ref: https://issues.redhat.com/browse/MTV-1864


mnecas:
My test:
image
image

yaacov (approved):
lgtm

PR #1303, opened 2025-01-14
MTV-1858 | Remove vm preference temporarily 

mnecas:
Issue:
With older version of the Kubevirt migrating specific windows version the preference always sets BIOS and if we set the UEFI the creation fails. 

Quick fix:
Remove the preference usage.

ref: [MTV-1858](https://issues.redhat.com/browse/MTV-1858) [CNV-49381](https://issues.redhat.com/browse/CNV-49381)


yaacov:
this fix is until CNV-49381 is fixed on CNV side ?

PR #1305, opened 2025-01-20
Chore: run go mod vendor

mnecas:
Run `go mod vendor` to cleanup after bazel


mnecas:
Unrelated linter issues, we should fix those at some point

yaacov (approved):

PR #1307, opened 2025-01-20
Move virt-v2v under cmd and pkg

mnecas:
Issue:
Right now the virt-v2v wrapper has the golang structure but it's in separate directory, we should unify the packages to the same structure.

Fix:
- [x] Move virt-v2v package under cmd and pkg directory
- [x] Update the guest CI to use the new 
- [x] Update the dockerfiles


yaacov (approved):
nice !

PR #1312, opened 2025-01-22
MTV-1929 Clean up unused active source connections.

mrnold:
Fix for MTV-1929, cleans up most active source connections when they are done with their work.


mrnold:
This isn't as effective as I hoped, I still see a bunch of idle connections. Maybe I missed one somewhere.

mrnold:
> This isn't as effective as I hoped, I still see a bunch of idle connections. Maybe I missed one somewhere.

I moved the logouts into migration.go and it works much better now. I don't see any lingering connections from the same client creation, just one from the collector that isn't quite as obvious to fix.

mnecas (approved):

PR #1318, opened 2025-01-27
MTV-1442 | Allow to skip shared disks

mnecas:
Issue: When migrating a VM with shared disks the virt-v2v transfers all disks that are attached to the VM, that includes the shared disks. But if we will transfer 2 VMs with the same shared disk the disk will be migrated twice. This takes a lot of additional resources.

Fix: There are multiple solutions, which vary in complexity. I have chosen the simplest and fastest solution by adding a new Plan parameter migrateSharedDisks this determines if the shared disks should be migrated. By default, it's set to True as that's the default virt-v2v behaviour. If the toggle is turned off the plan will filter out the shared disks and use KubeVirt CDI for disk transfer and virt-v2v-in-place for the guest conversion.

Usage: The user will create 2 plans, one with only one VM and with the `migrateSharedDisks` enabled to transfer the shared disks, and another plan with the toggle disabled which will migrate the additional VMs. If the plans are in correct order (first transfer shared disk) the shared disks will get automatically attached to the new VMs.

Ref: http://issues.redhat.com/browse/MTV-1442

### DIFF
--- a/pkg/controller/plan/adapter/vsphere/builder.go
+++ b/pkg/controller/plan/adapter/vsphere/builder.go
@@ -194,6 +194,9 @@ func (r *Builder) PodEnvironment(vmRef ref.Ref, sourceSecret *core.Secret) (env 
 		err = liberr.Wrap(err, "vm", vmRef.String())
 		return
 	}
+	if !r.Context.Plan.Spec.MigrateSharedDisks {
+		r.removeSharedDisks(vm)
+	}
 
 	macsToIps := ""
 	if r.Plan.Spec.PreserveStaticIPs {
@@ -399,6 +402,9 @@ func (r *Builder) DataVolumes(vmRef ref.Ref, secret *core.Secret, _ *core.Config
 		err = liberr.Wrap(err, "vm", vmRef.String())
 		return
 	}
+	if !r.Context.Plan.Spec.MigrateSharedDisks {
+		r.removeSharedDisks(vm)
+	}
 
 	url := r.Source.Provider.Spec.URL
 	thumbprint := r.Source.Provider.Status.Fingerprint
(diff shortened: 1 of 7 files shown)


mnecas:
Test with 2 VMs, each had one own disk and one shared disk.
The plan `shared-1` had the `migrateSharedDisks: True` and transferred both disks.
image

And the plan `shared-2` had the  `migrateSharedDisks: False` and transferred only non shared disks.
image

pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -723,6 +828,10 @@ func (r *Builder) mapDisks(vm *model.VM, vmRef ref.Ref, persistentVolumeClaims [
 				},
 			},
 		}
+		if disk.Shared {
+			kubevirtDisk.Shareable = ptr.To(true)
```
fabiand, line 832: This line is not needed. it's only needed one one disk is attached multiple times to the same vm.

pkg/controller/plan/kubevirt.go
```diff
@@ -2081,6 +2084,9 @@ func (r *KubeVirt) libvirtDomain(vmCr *VirtualMachine, pvcs []*core.PersistentVo
 		diskSource := libvirtxml.DomainDiskSource{}
 
 		pvc := pvcsByName[vol.PersistentVolumeClaim.ClaimName]
+		if pvc == nil {
+			continue
```
yaacov, line 2088: should we add a log line for this cases, in case a user/qe will need to look for this case ?

pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -491,6 +505,64 @@ func (r *Builder) DataVolumes(vmRef ref.Ref, secret *core.Secret, _ *core.Config
...
+		},
+	)
+	if err != nil {
+		err = liberr.Wrap(err)
+		return
+	}
+	pvcs = make([]*core.PersistentVolumeClaim, len(pvcsList.Items))
+	for i, pvc := range pvcsList.Items {
+		// loopvar
+		pvc := pvc
```
yaacov, line 527: hmmm, neetpick, `pvc := pvc` is not nice, maybe `allocatedPvc := pvc` or `copyPvc := pvc` ?

yaacov (approved):
lgtm, some nitpicks

PR #1325, opened 2025-01-28
MTV-1939 | Disable vTPM for BIOS

mnecas:
Issue: The win22 preference has vTPM by default, which fails to boot.

Fix: Disable vTPM for BIOS.


mnecas:
I reproduced the issue and the VM was failing with `FailedBackendStorageCreate`.
Manual testing passed:
image
image

yaacov (approved):
lgtm

PR #1326, opened 2025-01-30
chore(containers): build: use a cache mount for building

jonner:
When building with bazel, incremental builds (e.g. after changing just a single source file) were pretty quick. After switching away from bazel, running e.g. `make build-controller-image` would do a full rebuild of the entire source code even if only a single file was touched. To reduce the amount of time that builds take, introduce a cache mount in the
build container. This cache will persist the intermediate build artifacts between container builds so that when a single source file is changed, it will only rebuild that one file and re-use the other object files from previous builds. Set the `$GOCACHE` environment variable to point to this cache mount while building.

### DIFF
--- a/build/forklift-api/Containerfile
+++ b/build/forklift-api/Containerfile
@@ -3,7 +3,8 @@ WORKDIR /app
 COPY --chown=1001:0 ./ ./
 ENV GOFLAGS "-mod=vendor -tags=strictfipsruntime"
 ENV GOEXPERIMENT strictfipsruntime
-RUN GOOS=linux GOARCH=amd64 go build -buildvcs=false -ldflags="-w -s" -o forklift-api github.com/konveyor/forklift-controller/cmd/forklift-api
+ENV GOCACHE /go-build/cache
+RUN --mount=type=cache,target=${GOCACHE},uid=1001 GOOS=linux GOARCH=amd64 go build -buildvcs=false -ldflags="-w -s" -o forklift-api github.com/konveyor/forklift-controller/cmd/forklift-api
 
 FROM registry.access.redhat.com/ubi9-minimal:9.5-1736404155
 # Required to be able to get files from within the pod
--- a/build/forklift-controller/Containerfile
+++ b/build/forklift-controller/Containerfile
@@ -3,7 +3,8 @@ WORKDIR /app
 COPY --chown=1001:0 ./ ./
 ENV GOFLAGS "-mod=vendor -tags=strictfipsruntime"
 ENV GOEXPERIMENT strictfipsruntime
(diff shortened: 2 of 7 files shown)


yaacov:
hi, the U,  `unshared` flag in the build line does not work when building within a container.

```
 Containerfile:7
  --------------------
...
  ERROR: failed to solve: invalid field 'U' must be a key=value pair
  make: *** [Makefile:217: build-populator-controller-image] Error 1
  Error: Process completed with exit code 2.
```

yaacov:
> Hmm, it worked for me. In fact, it was not working without it... Are you
using docker or podman?

It fails on our CI task ( ubuntu + docker ), you can see the error in the failing test
https://github.com/kubev2v/forklift/actions/runs/13060991201/job/36443659815?pr=1326

> it worked for me. In fact, it was not working without it

Works on my fedora using podman, with and without the U flag, this is a successful run on my laptop without the U flag:

```
[1/2] STEP 3/7: COPY --chown=1001:0 ./ ./
--> 291785912a27
[1/2] STEP 4/7: ENV GOFLAGS "-mod=vendor -tags=strictfipsruntime"
--> bca4583da81b
[1/2] STEP 5/7: ENV GOEXPERIMENT strictfipsruntime
--> 12fa0fa6784d
[1/2] STEP 6/7: ENV GOCACHE /go-build/cache
--> 37d93be2c35f
[1/2] STEP 7/7: RUN --mount=type=cache,target=${GOCACHE} GOOS=linux GOARCH=amd64 go build -buildvcs=false -ldflags="-w -s" -o manager github.com/konveyor/forklift-controller/cmd/forklift-controller
--> d02e7c79ed39
[2/2] STEP 1/5: FROM registry.access.redhat.com/ubi9/ubi-minimal:9.5-1736404155
```

build/forklift-api/Containerfile
```diff
@@ -3,7 +3,8 @@ WORKDIR /app
 COPY --chown=1001:0 ./ ./
 ENV GOFLAGS "-mod=vendor -tags=strictfipsruntime"
 ENV GOEXPERIMENT strictfipsruntime
-RUN GOOS=linux GOARCH=amd64 go build -buildvcs=false -ldflags="-w -s" -o forklift-api github.com/konveyor/forklift-controller/cmd/forklift-api
+ENV GOCACHE /go-build/cache
+RUN --mount=type=cache,target=${GOCACHE},U GOOS=linux GOARCH=amd64 go build -buildvcs=false -ldflags="-w -s" -o forklift-api github.com/konveyor/forklift-controller/cmd/forklift-api
```
mnecas, line 7: could you please try removing the `U` on Which the CI is failing?

yaacov:
@jonner now it's 
` mkdir /go-build/cache/00: permission denied `

can you use a the local dir for cache ?
maybe something like (didn't test it myself):
`ENV GOCACHE ./go-build/cache`

PR #1327, opened 2025-02-02
MTV-1596 | Add new NetworkNameTemplate field to plan CR

yaacov:
Ref:
https://issues.redhat.com/browse/MTV-1596

RFE:
Requesting the ability for choosing a custom name for the network and disks during migration for VM's.  Currently migrations end up with 'net-0' or 'vol-1' and cannot be modified. A newly created vm typically allows for you to name these.

Fix:
Add an optional network template:
```
NetworkNameTemplate is a template for generating network interface names in the target virtual machine.
It follows Go template syntax and has access to the following variables:
   - .NetworkName: If target network is multus, name of the Multus network attachment definition, empty otherwise.
   - .NetworkNamespace: If target network is multus, namespace where the network attachment definition is located.  
   - .NetworkType: type of the network ("Multus" or "Pod")
   - .NetworkIndex: sequential index of the network interface (0-based)
The template can be used to customize network interface names based on target network configuration.
 Note:
   - This template can be overridden at the individual VM level
   - If not specified on VM level and on Plan leverl, default naming conventions will be used
 Examples:
   "net-{{.NetworkIndex}}"
   "{{if eq .NetworkType "pod"}}pod{{else}}multus-{{.NetworkIndex}}{{end}}"
```

In the plan YAML:

```
spec:
   ...
   networkNameTemplate:   "net-{{.NetworkIndex}}"

   vms:
    - name: some-vm
      networkNameTemplate:   "{{if eq .NetworkType "pod"}}pod{{else}}multus-{{.NetworkIndex}}{{end}}"
```

Screenshots:
plan-yaml-custom-net

vm-yaml-custom-net


pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -592,6 +594,26 @@ func (r *Builder) mapNetworks(vm *model.VM, object *cnv.VirtualMachineSpec) (err
...
+				templateData := api.NetworkNameTemplateData{
+					NetworkName:      mapped.Destination.Name,
+					NetworkNamespace: mapped.Destination.Namespace,
+					NetworkType:      strings.ToLower(string(mapped.Destination.Type)),
+					NetworkIndex:     numNetworks,
+				}
+
+				networkName, err = r.executeTemplate(networkNameTemplate, &templateData)
+				if err != nil {
+					// fallback to default name and reset error
```
mnecas, line 611: Maybe adding a log that the template execution failed?

yaacov: re based and added a log line

mnecas (approved):
lgtm

PR #1338, opened 2025-02-05
MTV-1745 | Improve validating webhooks

jonner:
- **Allow api service account to create `SubjectAccessReview` objects**
- **Validate target namespace from plan webhook**
- **webhook: Disallow plans when providers can't be retrieved**
- **webhooks: add check for permissions on VMs in migration plans**
- **webhooks: check permissions on providers in migration plans**
- **Add migration webhooks**
- **Change api.GetGroupResource() to return a GroupResource**

ref: https://issues.redhat.com/browse/MTV-1745

### DIFF
--- a/pkg/apis/forklift/v1beta1/register.go
+++ b/pkg/apis/forklift/v1beta1/register.go
@@ -39,21 +39,26 @@ var SchemeBuilder = &scheme.Builder{GroupVersion: SchemeGroupVersion}
 
 // TODO: find a better place for this, it would be nice to use it also to
 // determine the plural configuration of a resource for the operator
-func GetGroupResource(required runtime.Object) (group string, resource string, err error) {
+func GetGroupResource(required runtime.Object) (groupresource schema.GroupResource, err error) {
 
 	switch required.(type) {
 	case *Provider:
-		group = SchemeGroupVersion.Group
-		resource = "providers"
+		groupresource = schema.GroupResource{
+			Group:    SchemeGroupVersion.Group,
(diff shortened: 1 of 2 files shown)


pkg/forklift-api/webhooks/validating-webhook/admitters/plan-admitter.go
```diff
@@ -15,6 +15,7 @@ import (
 	api "github.com/konveyor/forklift-controller/pkg/apis/forklift/v1beta1"
 	"github.com/konveyor/forklift-controller/pkg/forklift-api/webhooks/util"
 	liberr "github.com/konveyor/forklift-controller/pkg/lib/error"
+	cnv "kubevirt.io/api/core/v1"
```
yaacov, line 18: can you remove the duplicate line ?

yaacov:
@jonner can you squash the commits ?

yaacov:
> Do you prefer that to using the "squash and merge" feature in github?

I prefer the PR owner to be responsible for the squashing, that way I know if the separate commits are only for reviewing purposes or are a real separation of concerns, because the PR author has the best context and understanding of the changes they've made and can therefore create a clean, logical, and meaningful commit history in the final squashed commit, making it easier for reviewers and future maintainers to understand the evolution of the code.

yaacov:
> I personally prefer more fine-grained commits

My concern with a large number of commits in a pull request is that it complicates back porting.  A smaller, concern specific commits can makes it easier to track and manage the back porting.

PR #1345, opened 2025-02-12
chore(containers): use proper syntax for ENV in Containerfiles

jonner:
The syntax "ENV VAR VALUE" is supported but is deprecated, and docker complains about it. The proper syntax is "ENV VAR=VALUE". Switch to that syntax in all Containerfiles.


yaacov (approved):
nice

PR #1369, opened 2025-02-20
feat: Introduce vsphere xcopy volume populator

rgolangh:
feat: Introducing vSphere generic copy-offload

This commit introduces a new volume populator, `VsphereXcopyVolumePopulator`,
designed to leverage vSphere's XCOPY functionality for efficient volume cloning
during VM migrations.

This enhancement significantly improves the performance of volume cloning for
vSphere VMs, especially for large volumes, by utilizing the server-side XCOPY
operation.

When the feature flag `feature_copy_offload` is true (off by default), the controller will
consult a config map to decided if VM disk from VMWare could be copied
by the storage backend(offloaded) into the newly created PVC.
When the controller creates the PVC for the v2v pod it will also create
a volume popoulator resource of type VSphereXcopyVolumePopulator and set
the filed `dataSourceRef` in the PVC to reference it.

Added a new populator controller for the resource VSPhereXcopyVolumePopulator

A new populator implementation under cmd/vsphere-xcopy-volume-populator
is a cli program that runs in a container that is responsible to perform
XCOPY to effciently copy data from a VMDK to the target PVC. See the
flow in cmd/vsphere-xcopy-volume-populator/README.md
The populator uses the storage API (configurable) to map the PVC to an ESX
then uses Vsphere API to call functions on the ESX to perform the actual
XCOPY command (provided that VAAI and accelerations is enabled on that
ESX).

If a storage provider wants to their storage to be supported they need
to implement a go package named after their product, and mutate main so
their specific code path is initialized. See
cmd/vsphere-xcopy-volume-populator/internal/populator/storage.go

- currently a VM with single disk is supported


Makefile
```diff
@@ -302,3 +311,184 @@ OPA=${HOME}/.local/bin/opa
...
+	FEATURE_VSPHERE_INCREMENTAL_BACKUP="true" \
+	VSPHERE_OS_MAP="forklift-virt-customize" \
+	OVIRT_OS_MAP="forklift-ovirt-osmap" \
+	VIRT_CUSTOMIZE_MAP="forklift-virt-customize" \
+	VSPHERE_COLD_CDI="true" \
+	SSL_CERT_FILE="/home/rgolan/src/kubev2v/forklift/tls-ca-bundle.pem" \
+	./bin/forklift-controller
+
+	#dlv --listen=:5432 --headless=true --api-version=2 exec ./bin/forklift-controller \
+>>>>>>> 826216fd (A working copy-offload with vsphere-xcopy-volume-populator)
```
mnecas, line 494: Rebase leftover artifacts

Makefile
```diff
@@ -302,3 +311,184 @@ OPA=${HOME}/.local/bin/opa
 else
 OPA=$(shell which opa)
 endif
+
+# The directory where the 'crc' binary will be installed (this path
```
mnecas, line 315: I removed most of the crc,minikube and kind

cmd/vsphere-xcopy-volume-populator/Containerfile
```diff
@@ -0,0 +1,36 @@
...
+
+FROM registry.access.redhat.com/ubi9/ubi-minimal
+
+COPY --from=builder /usr/src/app/bin/xcopy-volume-populator /bin/xcopy-volume-populator
+
+RUN ln -s /bin/xcopy-volume-populator /manager
+
+# install powershell
+RUN microdnf install -y libicu
+RUN rpm -Uvh https://github.com/PowerShell/PowerShell/releases/download/v7.5.0/powershell-7.5.0-1.rh.x86_64.rpm
```
mnecas, line 29: Do we need the powershell and the vmkfs.ps1?

cmd/vsphere-xcopy-volume-populator/api/v1beta1/types.go
```diff
@@ -0,0 +1,63 @@
...
+import (
+	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"sigs.k8s.io/controller-runtime/pkg/scheme"
+)
+
+var SchemeGroupVersion = schema.GroupVersion{
+	Group: "forklift.konveyor.io",
+	// Version: "v1alpha1",
+	// workaround mtv code
```
mnecas, line 12: Please not mention the MTV but use Forklift.
What workaround?

cmd/vsphere-xcopy-volume-populator/vsphere-xcopy-volume-populator.go
```diff
@@ -0,0 +1,219 @@
...
+	flag.StringVar(&targetPVC, "target-pvc", "", "Target PVC for population")
+	flag.StringVar(&targetNamespace, "target-namespace", "", "Contents to populate file with")
+	flag.StringVar(&storageVendor, "storage-vendor", "ontap", "The storage vendor to work with. Current values: [ontap,]")
+	flag.StringVar(&storageHostname, "storage-hostname", "", "The storage vendor api hostname")
+	flag.StringVar(&storageUsername, "storage-username", "", "The storage vendor api username")
+	flag.StringVar(&storagePassword, "storage-password", "", "The storage vendor api password")
+
+	flag.StringVar(&vsphereHostname, "vsphere-hostname", "", "vSphere's API hostname")
+	flag.StringVar(&vsphereUsername, "vsphere-username", "", "vSphere's API username")
+	flag.StringVar(&vspherePassword, "vsphere-password", "", "vSphere's API password")
```
mnecas, line 74: Please pass the passwords via file instead of args for security.

cmd/vsphere-xcopy-volume-populator/api/v1beta1/types.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+type VSphereXcopyVolumePopulatorSpec struct {
+	// VmdkPath is the full path the vmdk disk. A valid path format is
+	// '[$DATASTORE_NAME] $VM_NAME/$DISK_NAME.vmdk'
+	VmdkPath string `json:"vmdkPath"`
+	// TargetPVC is the kubernetes PVC name that is used as the target
+	// The controller will resolve the underlying PV and will copy the data
+	// from the vmdk to that target volume
+	TargetPVC string `json:"targetPVC"`
```
mnecas, line 41: Afaik we should not require this, we can get the PVC in the population controller once the volume populator is referenced on the PVC via `dataSourceRef`

mnecas (commented):
I know it's just draft so I had just a quick look.

cmd/vsphere-xcopy-volume-populator/internal/populator/populate.go
```diff
@@ -0,0 +1,54 @@
...
+	return fmt.Sprintf("/vmfs/volumes/%s/%s/%s", d.Datacenter, d.VMName, d.VmdkFile)
+}
+
+func ParseVmdkPath(vmdkPath string) (VMDisk, error) {
+	parts := strings.SplitN(vmdkPath, "] ", 2)
+	if len(parts) != 2 {
+		return VMDisk{}, fmt.Errorf("Invalid vmdkPath %q, should be '[datastore] vmname/vmname.vmdk'", vmdkPath)
+	}
+	datastore := strings.TrimPrefix(parts[0], "[")
+	pathParts := strings.SplitN(parts[1], "/", 2)
```
mnecas, line 50: please check for the size of `pathParts` so we don't panic

cmd/vsphere-xcopy-volume-populator/internal/populator/populate.go
```diff
@@ -0,0 +1,54 @@
...
+	vmname := pathParts[0]
+	vmdk := pathParts[1]
+	return VMDisk{VMName: vmname, Datacenter: datastore, VmdkFile: vmdk}, nil
```
mnecas, line 53: please check that the values are not empty strings

Makefile
mnecas: still applies @rgolangh

pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -1034,32 +1045,213 @@ func (r *Builder) LunPersistentVolumeClaims(vmRef ref.Ref) (pvcs []core.Persiste
...
+
+					vp := api.VSphereXcopyVolumePopulator{
+						ObjectMeta: metav1.ObjectMeta{
+							Name:      commonName,
+							Namespace: namespace,
+							Labels:    labels,
+						},
+						Spec: api.VSphereXcopyVolumePopulatorSpec{
+							VmdkPath:             disk.File,
+							TargetPVC:            commonName,
```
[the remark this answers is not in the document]
mnecas: As alternative to this parameter you can pass the PVC argument within the controller to the population binary  https://github.com/kubev2v/forklift/blob/f7ff69cb52f89afa5888d028e36dca18f24ac051/pkg/lib-volume-populator/populator-machinery/controller.go#L524

cmd/vsphere-xcopy-volume-populator/internal/populator/remote_esxcli.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+	defer func() {
+		p.StorageApi.UnMap(xcopyInitiatorGroup, lun)
+		for _, group := range originalInitiatorGroups {
+			p.StorageApi.Map(group, lun)
+		}
+
+	}()
+
+	_, err = p.VSphereClient.RunEsxCommand(context.Background(), host, []string{"storage", "core", "adapter", "rescan", "-a", "1"})
```
mnecas, line 94: Does the command wait for the rescan?

cmd/vsphere-xcopy-volume-populator/internal/populator/remote_esxcli.go
```diff
@@ -0,0 +1,125 @@
...
+	if err != nil {
+		return err
+	}
+	naa := fmt.Sprintf("naa.%s%x", lun.ProviderID, lun.SerialNumber)
+	_, err = p.VSphereClient.RunEsxCommand(context.Background(), host, []string{"storage", "core", "device", "list", "-d", naa})
+	if err != nil {
+		return fmt.Errorf("failed to locate the target LUN %s. Check the LUN details and the host mapping response: %s", naa, err)
```
mnecas, line 101: Think we should add some retry to this in case the host fails to find it within the time

pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -1034,32 +1045,213 @@ func (r *Builder) LunPersistentVolumeClaims(vmRef ref.Ref) (pvcs []core.Persiste
...
+				vsphereInstance, storageVendorProduct, storageVendorSecretRef := copyOffloadMapping.GetStorageMappingBy(storageClass, ds.Name)
+				if coldLocal && vsphereInstance != "" {
+					namespace := r.Plan.Spec.TargetNamespace
+					// pvs names needs to be less than 63, this leaves 53 chars
+					// for the plan and vm name (2 dashes and 8 chars uuid)
+					commonName := fmt.Sprintf("%s-%s-%s", r.Plan.Name, vm.Name, uuid.New().String()[:8])
+					labels := map[string]string{
+						"migration": string(r.Migration.UID),
+						// we need uniqness and a value which is less than 64 chars, hence using disk.key
+						"vmdkKey": fmt.Sprint(disk.Key),
```
mnecas, line 1115: note: the disk.key is not that unique by itself as the IDs easily repeat for each VM. but with the `vmRef.ID` this combination works well

pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -1034,32 +1045,213 @@ func (r *Builder) LunPersistentVolumeClaims(vmRef ref.Ref) (pvcs []core.Persiste
...
 	return
 }
 
+func (r *Builder) getVolumePopulator(vmdkKey string) (api.VSphereXcopyVolumePopulator, error) {
+	list := api.VSphereXcopyVolumePopulatorList{}
+	err := r.Destination.Client.List(context.TODO(), &list, &client.ListOptions{
+		Namespace: r.Plan.Spec.TargetNamespace,
+		LabelSelector: labels.SelectorFromSet(map[string]string{
+			"migration": string(r.Migration.UID),
+			"vmdkKey":   vmdkKey,
```
mnecas, line 1226: The vmdkKey by itself is not unique enough, you need to also specify the VM as you can have multiple disks in the same plan with the same disk key.

mnecas (changes requested):
We will need to add:
- new image to hold the VIB
- validations
- change the netap lib to older version
want to have a sync with you on the configmap

cmd/vsphere-xcopy-volume-populator/vsphere-xcopy-volume-populator.go
```diff
@@ -0,0 +1,294 @@
...
+		}
+		storageApi = &sm
+	default:
+		klog.Fatalf("Unsupported storage vendor %s use one of [ontap,]", storageVendor)
+	}
+
+	// validations
+	_, err := populator.ParseVmdkPath(sourceVMDKFile)
+	if err != nil {
+		klog.Fatal(err)
```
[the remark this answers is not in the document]
rgolangh: Fatal exits with error 255

cmd/vsphere-xcopy-volume-populator/internal/ontap/ontap.go
```diff
@@ -0,0 +1,112 @@
...
+	// in RHEL lsblk needs that swap. In fedora it doesn't
+	//serialNumber :=  strings.ReplaceAll(l.SerialNumber, "?", "\\\\x3f")
+	serialNumber := l.SerialNumber
+	lun := populator.LUN{Name: l.Name, VolumeHandle: volumeHandle, SerialNumber: serialNumber, ProviderID: OntapProviderID}
+	return lun, nil
+}
+
+func (c *NetappClonner) Get(lun populator.LUN) (string, error) {
+	// this code is from netapp/trident/storage_drivers/ontap/ontap_common.go
+	//reportingNodes, err := c.api.LunMapGetReportingNodes(context.Background(), XCOPY_CLONNER_GROUP, fmt.Sprintf("/vol/%s/lun0",lun))
```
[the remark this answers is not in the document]
rgolangh: ack

cmd/vsphere-xcopy-volume-populator/internal/populator/remote_esxcli.go
rgolangh: It is syncronous on the server AFAIK

pkg/controller/plan/adapter/vsphere/builder.go
rgolangh: ack

cmd/vsphere-xcopy-volume-populator/api/v1beta1/types.go
rgolangh: removing this whole api dir is not needed anymore

cmd/vsphere-xcopy-volume-populator/vsphere-xcopy-volume-populator.go
rgolangh: i'm building on the populator pod to mount the secret as env params. this is not needed anymore and I'll remove it. I think secret-ref is also not needed anymore

rgolangh: perhaps it is just needed for the populator controller to mount it. I'll look again

cmd/vsphere-xcopy-volume-populator/api/v1beta1/types.go
rgolangh: I need the target pvc volume to fetch the volumeHandle from kubernetes because it will only be availabe once the populator pod gets the prime- pvc attached (till then there is no volume created yet on k8s)

rgolangh:
> We will need to add:
> 
>     * new image to hold the VIB
> 
why not ship the VIB in the same xcopy populator image? along with an ansible playbook to deploy it?

>     * validations
> 
what validations you had in mind?

>     * change the netap lib to older version

>       want to have a sync with you on the configmap

mnecas (commented):
I don't see the service account creation on the destination namespace for the volume populator.
AFAIK most users migrate from vSphere but we do also allow the ESXI hosts directly. Could you please try it out aswell?

cmd/vsphere-xcopy-volume-populator/api/v1beta1/types.go
mnecas: You know the PVC in the volume populator controller, the idea would be to pass it from there

cmd/vsphere-xcopy-volume-populator/vmkfstools-wrapper/create-vib.sh
```diff
@@ -0,0 +1,101 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
mnecas: We have decided to move the VIB to the xcopy volume populator image

pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -1034,32 +1044,218 @@ func (r *Builder) LunPersistentVolumeClaims(vmRef ref.Ref) (pvcs []core.Persiste
...
+			return false
+		}
+
+		/// LEFT IN MIDDLE - NEED TO GET THE LOGIC STRAIGHT && EFFICIENT
+		if m.OffloadPlugin != nil && m.OffloadPlugin.VSphereXcopyPluginConfig != nil {
+            klog.Infof("####### RGOLAN found offload plugin: config %+v on ds map  %+v",m.OffloadPlugin.VSphereXcopyPluginConfig, dsMapIn)
+		    return true	
+
+		}
+	}
```
mnecas, line 1070: @rgolangh

cmd/vsphere-xcopy-volume-populator/internal/populator/remote_esxcli.go
```diff
@@ -0,0 +1,143 @@
...
+		}
+
+	}()
+
+    retries := 3
+    for i := 3; i > 0; i--{
+		_, err = p.VSphereClient.RunEsxCommand(context.Background(), host, []string{"storage", "core", "adapter", "rescan", "-a", "1"})
+		if err != nil {
+            klog.Errorf("failed to rescan adapters, probably in progress. Rerty %d/%d", i, retries)
+	        time.Sleep(time.Duration(rand.Intn(10)))
```
mnecas, line 108: Please add FIXME

cmd/vsphere-xcopy-volume-populator/internal/populator/remote_esxcli.go
```diff
@@ -0,0 +1,143 @@
...
+		p.StorageApi.UnMap(xcopyInitiatorGroup, lun, m)
+		for _, group := range originalInitiatorGroups {
+			p.StorageApi.Map(group, lun, m)
+		}
+
+	}()
+
+    retries := 3
+    for i := 3; i > 0; i--{
+		_, err = p.VSphereClient.RunEsxCommand(context.Background(), host, []string{"storage", "core", "adapter", "rescan", "-a", "1"})
```
mnecas, line 105:
```
esxcli storage core adapter rescan --help
Usage: esxcli storage core adapter rescan [cmd options]

Description: 
  rescan                Rescan SCSI HBAs to search for new Devices, remove DEAD paths and update path state. This operation will also run an claim operation equivalent to the claimrule run command and a filesystem rescan.

Cmd options:
  -A|--adapter=<str>    Select the adapter to use when rescanning SCSI adapters. This must be a SCSI HBA name as shown in the adapter list command. This cannot be used with the --all option
  -a|--all              Indicate the rescan should rescan all adapters instead of a specific one.
  -S|--skip-claim       By default after an add operation a claiming session is run to find new devices and have them be claimed by the appropriate Multipath Plugin. Passing this flag will skip that claiming session.
  -F|--skip-fs-scan     This option is deprecated as no filesystem scan is performed by default
  -t|--type=<str>       Specify the type of rescan to perform. Available types are
                            add: Perform rescan and only add new devices if any.
                            all: Perform rescan and do all opertaions (this is the default action.)
                            delete: Perform rescan and only delete DEAD devices.
                            update: Rescan existing paths only and update path states.
````

File: cmd/vsphere-xcopy-volume-populator/vmkfstools-wrapper/Dockerfile [hunk 20]
```diff
@@ -0,0 +1,12 @@
+# as of January 2025 go 1.22 is still the latest go-toolset for ubi-9
+# but since trident depends on 1.23 we must install it.
+# go-toolset 1.23 on ubi-9 is targeted to ubi 9.6 which is around May 2025
+#FROM registry.access.redhat.com/ubi8/go-toolset:1.23.1 AS builder
+
+From docker.io/golang:1.23 AS builder
```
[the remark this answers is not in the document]
rgolangh: works as expected on 1.22, I think we are good to go there

cmd/vsphere-xcopy-volume-populator/vmkfstools-wrapper/foo.vib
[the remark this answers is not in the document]
rgolangh: ack

cmd/vsphere-xcopy-volume-populator/internal/populator/remote_esxcli.go
rgolangh: thanks a lot. this has the potential of making the rescan better.

cmd/vsphere-xcopy-volume-populator/internal/ontap/ontap.go
```diff
@@ -0,0 +1,112 @@
...
+	"fmt"
+	"strings"
+
+	"github.com/kubev2v/forklift/cmd/vsphere-xcopy-volume-populator/internal/populator"
+	drivers "github.com/netapp/trident/storage_drivers"
+	"github.com/netapp/trident/storage_drivers/ontap/api"
+	"k8s.io/klog/v2"
+)
+
+const OntapProviderID = "600a0980"
```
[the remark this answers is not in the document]
borod108: This is an idea that each vendor has that is is the first characters in the NAA of their devices.

cmd/vsphere-xcopy-volume-populator/internal/populator/remote_esxcli.go
```diff
@@ -0,0 +1,143 @@
...
+	"context"
+	"fmt"
+	"math/rand"
+	"time"
+
+	"github.com/kubev2v/forklift/cmd/vsphere-xcopy-volume-populator/internal/vmware"
+	"k8s.io/klog/v2"
+)
+
+const xcopyInitiatorGroup = "xcopy-esxs"
```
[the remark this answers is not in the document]
borod108: this is a group we use that is mapped or exported so that the esxi has access to it, we than map the specific device to this group.

pkg/controller/plan/adapter/vsphere/builder.go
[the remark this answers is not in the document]
mnecas: we will use the `--owner-name`

cmd/vsphere-xcopy-volume-populator/vmkfstools-wrapper/Dockerfile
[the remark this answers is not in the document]
mnecas: please switch to the 1.22 in the dockerfile

cmd/vsphere-xcopy-volume-populator/changes-for-vantara.txt
```diff
@@ -0,0 +1,6 @@
+
```
rgolangh, line 1: will remove this file

cmd/vsphere-xcopy-volume-populator/TODO.md
```diff
@@ -0,0 +1,5 @@
+- LUN esxi FC or Iscsi luns auto discover - there may be deployments with
```
rgolangh, line 1: will remove this file

cmd/vsphere-xcopy-volume-populator/internal/populator/populate.go
rgolangh: done in line 46, unless you mean something else?

rgolangh: ok, pathParts, not parts, ack

cmd/vsphere-xcopy-volume-populator/role.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  annotations:
+    rbac.authorization.kubernetes.io/autoupdate: "true"
+  labels:
+    kubernetes.io/bootstrapping: rbac-defaults
+  name: volumes-access
```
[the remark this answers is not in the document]
rgolangh: this whole role isn't not needed anymore. the real change is adding rules to the forklift-controller role, to create SA with permissions to read PVCs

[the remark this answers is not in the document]
rgolangh: I'm removing this file

cmd/vsphere-xcopy-volume-populator/vmkfstools-wrapper/Dockerfile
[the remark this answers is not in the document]
rgolangh: This file isn't needed, there's not go binary to build in that directory

cmd/vsphere-xcopy-volume-populator/vmkfstools-wrapper/esxi_hosts.yaml
```diff
@@ -0,0 +1,6 @@
+plugin: vmware.vmware.esxi_hosts
```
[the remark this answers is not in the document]
rgolangh: this is according to the doc https://docs.ansible.com/ansible/11/collections/vmware/vmware/esxi_hosts_inventory.html

cmd/vsphere-xcopy-volume-populator/vmkfstools-wrapper/vib-install-playbook.yaml
```diff
@@ -0,0 +1,43 @@
+---
```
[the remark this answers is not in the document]
rgolangh: I included a section to copy the vib and the ansible files to `build/vsphere-xcopy-volume-populator/Containerfile`

I added a section in cmv/vsphere-xcopy-volume-populator/vmkfstools-wrapper/README.md , and I guess that should be the basis to the doc

generic-storage-offloaded-copy.md
```diff
@@ -0,0 +1,25 @@
+prerequisite:
```
rgolangh, line 1: will remove this file

pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -1176,32 +1184,223 @@ func (r *Builder) LunPersistentVolumeClaims(vmRef ref.Ref) (pvcs []core.Persiste
...
+						},
+					}
+
+					if annotations == nil {
+						pvc.Annotations = make(map[string]string)
+					} else {
+						pvc.Annotations = annotations
+					}
+					pvc.Annotations[planbase.AnnDiskSource] = r.baseVolume(disk.File)
+					pvc.Annotations["copy-offload"] = r.baseVolume(disk.File)
```
mnecas, line 1291:
```suggestion
					pvc.Annotations[planbase.AnnDiskSource] = baseVolume(disk.File, false)
					pvc.Annotations["copy-offload"] = baseVolume(disk.File, false)
```

pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -1176,32 +1184,223 @@ func (r *Builder) LunPersistentVolumeClaims(vmRef ref.Ref) (pvcs []core.Persiste
...
+		if fErr != nil {
+			err = fErr
+			return
+		}
+
+		pvblock := core.PersistentVolumeBlock
+		for _, disk := range vm.Disks {
+			if disk.Datastore.ID == ds.ID {
+				storageClass := mapped.Destination.StorageClass
+				coldLocal, vErr := r.Context.Plan.VSphereColdLocal()
```
mnecas, line 1238: The `VSphereColdLocal` does not exist anymore, it was renamed to `ShouldUseV2vForTransfer`, but I don't think that we need to check it here.

operator/roles/forkliftcontroller/templates/controller/copy-offload-config-map.yml.j2
```diff
@@ -0,0 +1,21 @@
+---
```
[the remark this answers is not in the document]
rgolangh: ack

pkg/controller/plan/adapter/vsphere/builder.go
rgolangh: ack

pkg/controller/plan/adapter/vsphere/builder.go
rgolangh: ack

rgolangh: In migration.go, there is a part which check SupportsVolumePopulator, and this returns `true` if the storagemap matches. I don't think that code is cold/warm aware.

mnecas: The name of the method was never perfect, it determines if the virt-v2v should do the transfer of the disks for us. As we are using volume populators the result from it should be false as we are managing the transfer ourselves. 
You could add some logic to the method to include also check for volume populators

rgolangh: since populator volumes are exclusively needed in cold for vsphere, would that be clearer to just return with early with no pvcs, in case plan.spec.warm?
```
func (r *Builder) PopulatorVolumes(....) {

    if r.Plan.Spec.Warm {
      return fmt.Errorf("warm migration and populator volumes dont mix!")
      }
      ...
 }
 ```

File: operator/roles/forkliftcontroller/templates/controller/deployment-controller.yml.j2 [hunk 32]
```diff
@@ -107,10 +107,15 @@ spec:
 {% endif %}
         - name: BLOCK_OVERHEAD
           value: "{{ controller_block_overhead }}"
-{% if controller_vsphere_incremental_backup|bool %}
+{% if feature_copy_offload|bool %}
```
mnecas, line 110: `controller_vsphere_incremental_backup`

rgolangh: ack

mnecas (commented):
lgtm

PR #1374, opened 2025-02-24
MTV-2093 | Sort the disks by bus

mnecas:
Issue: When running a migration of a VM which has various bus controllers the PVs are not populated by correct disk. This causes that some larger disks are in small PVs.
This issue happens due to the libvirt domain-VMware API missmatch. The virt-v2v is outputing the disks in order as they are in the libvirt domain but Forklift is using the order from API and sorting the disks by key. The libvirt is sorted in order of devices SCSI, SATA and IDE wheras sorting by key from VMware API results in order IDE, SATA and SCSI.

Fix: Gather information about the disk busses and sort the disks by SCSI, SATA and IDE. And within each of these categories we sort by the key from lower to higer.

Ref: https://issues.redhat.com/browse/MTV-2093

### DIFF
--- a/pkg/controller/provider/model/vsphere/model.go
+++ b/pkg/controller/provider/model/vsphere/model.go
@@ -264,6 +264,7 @@ type VM struct {
 	Devices               []Device       `sql:""`
 	NICs                  []NIC          `sql:""`
 	Disks                 []Disk         `sql:""`
+	Controllers           []Controller   `sql:""`
 	Networks              []Ref          `sql:""`
 	Concerns              []Concern      `sql:""`
 	GuestNetworks         []GuestNetwork `sql:""`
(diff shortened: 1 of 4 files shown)


pkg/controller/provider/model/vsphere/model.go
```diff
@@ -264,6 +264,7 @@ type VM struct {
 	Devices               []Device       `sql:""`
 	NICs                  []NIC          `sql:""`
 	Disks                 []Disk         `sql:""`
+	Controllers           []Controller   `sql:""`
```
yaacov, line 267: doe it make sense to do the controllers a map by key, instead of an array ?

mnecas: I wanted to go that way but at the same time, I did not want to break the convention.
I think most of these could be maps

yaacov: i am ok with both ways ...

yaacov (approved):
lgtm, 

p.s.
do you want to add some tests to make sure, future us will not refactor this sorting away because they are not needed

PR #1375, opened 2025-02-25
controller: Skip TLS when missing certs

mnecas:
Issue: When running controller locally for development the certificates are usually missing and the main controller can fail to send requests to some existing inventory.

Fix: Skip TLS check when the certs are missing.

### DIFF
--- a/pkg/controller/provider/web/base/client.go
+++ b/pkg/controller/provider/web/base/client.go
@@ -307,6 +307,11 @@ func (c *RestClient) buildTransport() (err error) {
 		transport.TLSClientConfig = &tls.Config{
 			RootCAs: pool,
 		}
+	} else if Settings.Development {
+		// Disable TLS for development when certs are missing
+		transport.TLSClientConfig = &tls.Config{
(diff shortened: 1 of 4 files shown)


pkg/controller/provider/web/base/client.go
```diff
@@ -307,6 +307,11 @@ func (c *RestClient) buildTransport() (err error) {
 		transport.TLSClientConfig = &tls.Config{
 			RootCAs: pool,
 		}
+	} else {
```
yaacov, line 310: how do we know we are in dev mode here, should we be more explicit to to be more clear we don't want this case triggered on production ?

pkg/controller/validation/policy/client.go
```diff
@@ -199,7 +199,7 @@ func (c *Client) buildTransport() (err error) {
 		transport.TLSClientConfig = &tls.Config{
 			RootCAs: pool,
 		}
-	} else {
+	} else if Settings.Development {
```
mnecas, line 202: .

PR #1379, opened 2025-02-26
MTV-1609 | VM source templates should not be selectable in the VM table

Hazanel:
Issue:
VM source templates should not be selectable in the VM table

Fix:
Filter the resources list with only non templates VM sources

### DIFF
--- a/pkg/controller/plan/adapter/vsphere/builder.go
+++ b/pkg/controller/plan/adapter/vsphere/builder.go
@@ -6,7 +6,6 @@ import (
 	"errors"
 	"fmt"
 	"net"
-	"net/netip"
 	liburl "net/url"
 	"path"
 	"regexp"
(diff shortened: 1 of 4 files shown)


pkg/controller/provider/web/vsphere/vm.go
```diff
@@ -69,6 +69,13 @@ func (h VMHandler) List(ctx *gin.Context) {
 	}
 	pb := PathBuilder{DB: db}
 	for _, m := range list {
+		if m.IsTemplate {
```
mnecas, line 72: This works only if the user requests the `model.MaxDetail` otherwise the database is not returning the template.
Could you please check how to add this to the list options?

pkg/controller/provider/web/vsphere/vm.go
```diff
@@ -69,6 +69,20 @@ func (h VMHandler) List(ctx *gin.Context) {
 	}
 	pb := PathBuilder{DB: db}
 	for _, m := range list {
+		if err = db.Get(&m); err != nil {
```
mnecas, line 72: Nope, this will result of creating new DB query per each VM during the listing of all VMs.
Please checkout the `ListOptions` in the `db.List`

mnecas, line 72: Nope, this will result of creating new DB query per each VM during the listing of all VMs.
Please checkout the ListOptions in the db.List

mnecas (changes requested):

PR #1382, opened 2025-02-27
MTV-1966 | Allow to skip shared disks

mnecas:
Issue:
In the current release of Forklift, it is possible to import shared disks using cold migration. This is possible because the virt-v2v is transferring all disks attached to the VM, but because the shared disks are attached to multiple VMs the shared disks get migrated multiple times and users right now, need to delete the extra volumes and reattach the disks.

Fix:
Add a new parameter `migrateSharedDisks` to the plan which will determine if the shared disks should be migrated. This parameter will allow the users to choose if the plan should migrate shared disks or not. When the `migrateSharedDisks` will be enabled the Forklift will use the "normal" cold migration flow using virt-v2v and labelling the shared PVC.
When the `migrateSharedDisks` is disabled the Forklift will use the KubeVirt CDI for disk transfer, and it will skip the shared disks. After the disk transfer, it will try to locate the already shared PVCs and attach them to the VMs. The Forklift will try to automatically find migrated shared PVCs in the namespace and attach them to the VMs.

User flow:
- Turn off all VMs with the attached shared disk on the VMware side.
- Create a plan with a single VM and the `migrateSharedDisks` enabled in the Forklift.
- Start the migration of the first plan and wait for it to finish.
- Create a secondary plan with all other VMs attached and the `migrateSharedDisks` disabled to the same target namespace as the first plan.
- Start the migration of the second plan and wait for it to finish.
- Check if all shared disks are attached to all VMs.

Known issues:
- The shared disks are not passed to the guest conversion so virt-v2v is not aware of them and does not update the config files, for example fstab. This will be fixed in follow-up PR.
- The shared disks can not be migrated when attached to a running VM  (mitigated by validations).
- The shared disks do not support warm migration (mitigated by validations).
- The shared disks can not be migrated to multiple target namespaces (mitigated by validations).

Ref:
- Original issue: https://issues.redhat.com/browse/MTV-1966
- This is a continuation of: https://github.com/kubev2v/forklift/pull/1318

### DIFF
--- a/pkg/controller/plan/kubevirt.go
+++ b/pkg/controller/plan/kubevirt.go
@@ -1708,12 +1708,12 @@ func (r *KubeVirt) guestConversionPod(vm *plan.VMStatus, vmVolumes []cnv.Volume,
 	nonRoot := true
 	allowPrivilageEscalation := false
 	// virt-v2v image
-	coldLocal, vErr := r.Context.Plan.VSphereColdLocal()
+	useV2vForTransfer, vErr := r.Context.Plan.ShouldUseV2vForTransfer()
 	if vErr != nil {
 		err = vErr
 		return
 	}
-	if coldLocal {
+	if useV2vForTransfer {
 		// mount the secret for the password and CA certificate
 		volumes = append(volumes, core.Volume{
 			Name: "secret-volume",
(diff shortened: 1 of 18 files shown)


pkg/controller/plan/kubevirt.go
```diff
@@ -1708,7 +1708,7 @@ func (r *KubeVirt) guestConversionPod(vm *plan.VMStatus, vmVolumes []cnv.Volume,
 	nonRoot := true
 	allowPrivilageEscalation := false
 	// virt-v2v image
-	coldLocal, vErr := r.Context.Plan.VSphereColdLocal()
+	coldLocal, vErr := r.Context.Plan.ShouldUseV2vForTransfer()
```
mansam, line 1711: nit: maybe worth renaming the "coldLocal" variable to match the new method name.

mansam (approved):
LGTM, other than nit about renaming the `coldLocal` variables to match the new method name.

PR #1383, opened 2025-02-27
MTV-2092 | MTV-2105 | spec.running and spec.runStrategy mutually exclusive and spec.running is deprecated on target VM

Hazanel:
Issue:
1. The field spec.running is deprecated
2. Fix spec.running and spec.runStrategy mutually exclusive scenario during migration


mnecas (approved):
LGTM

PR #1401, opened 2025-03-04
Preserve PVC mount index between PVC creation and PVC mounting as a volume

yaacov:
Issue:
When we create the PVC for VM disks we don't give them order.
Then when getting them from k8s we assume they are ordered by mount order, this is not garentead to be true.

Fix:
Add the mounting order as an annotation on the PVC.
Use this annotation to sort the PVC before matching them to the VM volumes.


mnecas (approved):
LGTM

PR #1404, opened 2025-03-05
MTV- 1609 | VM source templates should not be selected in the VM table

Hazanel:
Issue:
VM source templates should not be selectable in the VM table

Fix:
Filter the resources list with only non templates VM sources


mnecas (approved):
lgtm

pkg/controller/provider/web/vsphere/vm.go
```diff
@@ -56,6 +56,8 @@ func (h VMHandler) List(ctx *gin.Context) {
 			ctx.Status(http.StatusInternalServerError)
 		}
 	}()
+
+	h.Detail = model.MaxDetail
```
mnecas, line 60: Ah one more note, please add note that we need the max detail for the IsTemplate

PR #1411, opened 2025-03-06
MTV-2187 | Allow to start migration of shared disk with warning validation

mnecas:
Issue: The shared disk has a warning validation which should not block the migration, but due to bug in the code the migration is blocked even with just warning.

Fix: Becuase we are checking in the migration execution if the validation has a SharedDisk type but we do not check what kind of category it is we block the migration even with the warning. Easy solution is to add new validation type with does not match the blocking.

Ref: https://issues.redhat.com/browse/MTV-2187


yaacov (approved):
lgtm

PR #1412, opened 2025-03-06
Test PR for konflux, DO NOT MERGE

solenoci:


mnecas:
hmm do not see the konflux running on the PR, maybe it does not run on drafts

solenoci:
> hmm do not see the konflux running on the PR, maybe it does not run on drafts

Yeah, I am troubleshooting that now. Trigger for the konflux is defined in `on-cel-expression` in .tekton/* files no mention about draft but I will try with normal one just in case.

PR #1416, opened 2025-03-06
Fix CI

mnecas:
Fix CI, please see my notes on the PR as each of the fixes is quite specific


.github/workflows/pull-request.yml
```diff
@@ -43,9 +43,6 @@ jobs:
           - os: ubuntu-latest
             source_provider: vsphere
 
-          - os: ubuntu-latest
-            source_provider: openstack
-
```
mnecas, line 48: Note: We don't have enough power to maintain the OpenStack right now.

pkg/controller/provider/container/vsphere/model.go
```diff
@@ -880,7 +880,7 @@ func (v *VmAdapter) updateDisks(devArray *types.ArrayOfVirtualDevice) {
 					Key:      disk.Key,
 					File:     backing.FileName,
 					Capacity: disk.CapacityInBytes,
-					Shared:   backing.Sharing != "sharingNone",
+					Shared:   backing.Sharing != "sharingNone" && backing.Sharing != "",
```
mnecas, line 892: Note: The backing.Sharing in the govmomi sim is empty string unlike prod.

pkg/virt-v2v/customize/scripts/rhel/run/test-network_config_util.sh
```diff
@@ -23,7 +23,7 @@ test_dir() {
     export IFQUERY_CMD="
       podman run
       -v $TEST_DIR/etc/network:/etc/network
-      cmd.cat/ifquery ifquery
+      quay.io/kubev2v/ifquery:latest ifquery
```
mnecas, line 26: Note: the ifquery stopped working so I have uploaded the image to the kubev2v so we can use it in our CI

pkg/virt-v2v/customize/scripts/rhel/run/test-network_config_util.sh
```diff
@@ -76,7 +76,7 @@ expected_to_pass_dirs() {
 
 
 expected_to_fail_dirs() {
-    local FAIL_IS_FATAL=false PASS_IS_FATAL=true
+    local FAIL_IS_FATAL=true PASS_IS_FATAL=true
```
mnecas, line 79: Note: Even if it was failing the CI still passed

tests/suit/framework/ova.go
```diff
@@ -17,7 +17,7 @@ func (r *OvaClient) LoadSourceDetails() (vm *OvaVM, err error) {
 		r.storageClass = "nfs-csi"
 	}
 
-	r.vmData.testVMId = "c5686650854d1e69b4123f4bf2e70fe1ed2a"
+	r.vmData.testVMId = "3920ddbed8bb43f1bd6e00e412635f99727d"
```
mnecas, line 20: Note: Once the OVA VM struct changes it changes the database ID and we need to update it here.

We should fix it in a more dynamic way.

pkg/virt-v2v/customize/scripts/rhel/run/systemd-test.d/expected-udev.rule
```diff
@@ -1 +1,3 @@
-SUBSYSTEM=="net",ACTION=="add",ATTR{address}=="aa:bb:cc:dd:ee:fd",NAME="ens160"
+SUBSYSTEM=="net",ACTION=="add",ATTR{address}=="aa:bb:cc:dd:ee:ff",NAME="ens160"
+SUBSYSTEM=="net",ACTION=="add",ATTR{address}=="aa:bb:cc:dd:ee:fe",NAME="eth1"
+SUBSYSTEM=="net",ACTION=="add",ATTR{address}=="aa:bb:cc:dd:ee:fd",NAME="enp3s0"
```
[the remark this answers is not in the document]
yaacov: yes, this is currect:
the 
https://github.com/kubev2v/forklift/blob/main/pkg/virt-v2v/customize/scripts/rhel/run/systemd-test.d/root/tmp/macToIP

file map the ips to the macs in in the expected file

one question:
we have a 
https://github.com/kubev2v/forklift/tree/main/pkg/virt-v2v/customize/scripts/rhel/run/systemd-test.d/root/etc/netplan

this netplan role mixed with the systemd roles
do we want to remove it from this test ?

if we remove it, the result will be just one line:
`SUBSYSTEM=="net",ACTION=="add",ATTR{address}=="aa:bb:cc:dd:ee:ff",NAME="ens160"
`

yaacov (approved):

PR #1419, opened 2025-03-07
MTV-2199 | Set boot disk to always have Virtio bus

mnecas:
Issue:
The Kubevirt VM can not boot from hotpluggable disk with scsi bus. We need to keep the virtio bus for boot disk.
https://issues.redhat.com/browse/VIRTCE-24

Fix: Use virtio for boot disk.

Ref: https://issues.redhat.com/browse/MTV-2199

### DIFF
--- a/pkg/controller/plan/util/utils.go
+++ b/pkg/controller/plan/util/utils.go
@@ -40,6 +40,15 @@ func CalculateSpaceWithOverhead(requestedSpace int64, volumeMode *core.Persisten
 	return spaceWithOverhead
 }
 
+func GetBootDiskNumber(deviceString string) int {
+	deviceNumber := GetDeviceNumber(deviceString)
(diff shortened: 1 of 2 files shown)


yaacov (approved):

pkg/controller/plan/util/utils.go
```diff
@@ -40,6 +40,15 @@ func CalculateSpaceWithOverhead(requestedSpace int64, volumeMode *core.Persisten
 	return spaceWithOverhead
 }
 
+func GetBootDiskNumber(deviceString string) int {
+	resp := GetDeviceNumber(deviceString)
```
yaacov, line 44: `resp` onse as a var name is not clear for me, maybe we can use a different name ?

yaacov (approved):

PR #1420, opened 2025-03-07
MTV-1966 | Add extra shared disk validations

mnecas:
Issue:
Right now when the user will migrate two VMs with the shared disk and the migrateSharedDisks enabled we do not give warning to the user that the disk will be duplicated. Neither if the PVC is already located in the namepsace.

Fix:
Add extra validation for the VMs to show if there would be some disk duplication.

Ref: https://issues.redhat.com/browse/MTV-1966

### DIFF
--- a/pkg/controller/plan/adapter/vsphere/validator.go
+++ b/pkg/controller/plan/adapter/vsphere/validator.go
@@ -5,6 +5,7 @@ import (
 
 	api "github.com/konveyor/forklift-controller/pkg/apis/forklift/v1beta1"
 	"github.com/konveyor/forklift-controller/pkg/apis/forklift/v1beta1/ref"
+	planbase "github.com/konveyor/forklift-controller/pkg/controller/plan/adapter/base"
 	"github.com/konveyor/forklift-controller/pkg/controller/provider/model/vsphere"
 	"github.com/konveyor/forklift-controller/pkg/controller/provider/web"
 	"github.com/konveyor/forklift-controller/pkg/controller/provider/web/base"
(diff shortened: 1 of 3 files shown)


mnecas:
image

mnecas:
image

mnecas:
image

yaacov (approved):

pkg/controller/plan/adapter/vsphere/validator.go
```diff
@@ -219,6 +220,45 @@ func (r *Validator) SharedDisks(vmRef ref.Ref, client client.Client) (ok bool, m
...
+			}
+		}
+		msg := ""
+		for diskFile, occurrences := range sharedDisksDuplicate {
+			if occurrences > 1 {
+				msg = fmt.Sprintf("the shared disk '%s' will be coppied %dx, %s", diskFile, occurrences, msg)
+			}
+		}
+		if msg != "" {
+			msg = fmt.Sprintf("%s please unatach the disks from the VMs except one or remove the VMs from the plan so the disk won't be duplicated", msg)
```
yaacov, line 259:
```suggestion
if len(disks_with_multiple_occurrences) == 1 {
    diskLabel = "disk"
} else {
    diskLabel = "disks"
}

msg = fmt.Sprintf(
    "%s please detach the %s from all but one of the VMs, or remove the VMs from the plan to avoid duplicating the %s",
    msg,
    diskLabel,
    diskLabel,
)
```

yaacov (approved):

PR #1433, opened 2025-03-12
MTV-2171 | New name in ova flow of virt-v2v pod

yaacov:
Ref: https://issues.redhat.com/browse/MTV-2171

Issue:
MTV can use '-on' option to change guest name automatically when migrate guests from vsphere provider, but MTV can't use '-on' option to change the guest name when migrate guests from ova provider, so MTV should add the same function of changing guest name to ova provider as vsphere provider 

Fix:
add the "-on" flag to the OVA flow in the virt-v2v pod

Screenshot:
Before fix:
ova-without-new-name-arg

After fix:
ova-with-new-name-arg

### DIFF
--- a/cmd/virt-v2v/entrypoint.go
+++ b/cmd/virt-v2v/entrypoint.go
@@ -136,7 +136,12 @@ func virtV2vBuildCommand() (args []string, err error) {
 		}
 		args = append(args, vsphereArgs...)
 	case global.OVA:
-		args = append(args, "-i", "ova", os.Getenv("V2V_diskPath"))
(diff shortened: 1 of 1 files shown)


yaacov:
@mnecas hi, tested on my cluster and ready for review

mnecas (approved):
LGTM

mnecas:
@yaacov do we have a target version for this?

yaacov:
> @yaacov do we have a target version for this?

no, but it looks small and safe to be included any version a user may need it

mnecas:
Okay I'll target it to the 2.8.2

PR #1366, opened 2025-02-19
MTV-383 | Allow customization of target vm name

yaacov:
Ref:
https://issues.redhat.com/browse/MTV-383

Issue:
As a migration owner, when creating a migration plan that has VMs with non-conformant names, I would like the ability to change the name of the target VM on the fly before it is migrated.

Currently:
A vm is getting the target name automatically to be unique within the namesapce and to be valid k8s subdomain.

After:
Users can override the automatic mechanism by providing a TargetName

  - If the target name is not valid k8s subdomain, plan validation will fail
  - If the target name is not unique, plan validation will fail
  
 Example:
 ``` bash
kubectl sql "\
       select \
          name as planName, spec.vms[1].name as origName, spec.vms[1].targetName as targetName, \
          status.migration.vms[1].name as migratedVMName \
       from openshift-mtv/plans \
       where targetName is not null"
KIND: Plan	COUNT: 1
planName	origName	targetName  	migratedVMName	
test2   	yzamir_2	my-best-name	my-best-name  	

kubectl sql "select * from openshift-mtv/vms where name = 'my-best-name'"
KIND: VirtualMachine	COUNT: 1
NAMESPACE    	NAME        	CREATION_TIME(RFC3339)       	
openshift-mtv	my-best-name	2025-02-24T11:57:23Z   
```

Plan YAML:
```yaml
spec:
    ...
    vms:
    - id: vm-16117
      name: yzamir_2
      targetName: my-best-name
    ...
```

Screenshots:
YAML:
targetNameYAML

VM:
targetNameVM

PR #1434, opened 2025-03-12
Fix targetName validation if empty

mnecas:
Issue: When the user creates a plan with 2 VMs without specifying the `targetName` one of the VMs gets error `Duplicate targetName`.

Fix: This issue is caused by the targetName being equal to empty string.

Ref: https://github.com/kubev2v/forklift/pull/1366


yaacov (approved):
Thanks !

PR #1436, opened 2025-03-12
Fix shared disks validations

mnecas:
Issue: The condition is adding by specific type. If we want to have different condition msg to different VMs within the same plan we need to change the type so the condition is not overriden by next condition. This way all VMs in the items list had message of the last VM.

Fix: Add VM specific condition types.

Ref:
- https://issues.redhat.com/browse/MTV-2206
- https://issues.redhat.com/browse/MTV-2207
- https://issues.redhat.com/browse/MTV-2209

### DIFF
--- a/pkg/lib/condition/condition.go
+++ b/pkg/lib/condition/condition.go
@@ -324,8 +324,7 @@ func (r *Conditions) HasBlockerCondition() bool {
 // The collection contains blocker conditions that keep the plan reconciling.
 func (r *Conditions) HasReQCondition() bool {
 	return r.HasCondition(ValidatingVDDK) ||
-		r.HasCondition(VMMissingChangedBlockTracking) ||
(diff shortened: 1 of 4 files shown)


pkg/lib/condition/condition.go
```diff
@@ -325,7 +325,7 @@ func (r *Conditions) HasBlockerCondition() bool {
 func (r *Conditions) HasReQCondition() bool {
 	return r.HasCondition(ValidatingVDDK) ||
 		r.HasCondition(VMMissingChangedBlockTracking) ||
-		r.HasCondition(SharedDisks)
+		r.HasCriticalCondition() || r.HasWarnCondition()
```
yaacov, line 328: do we want to block on conditions called "warn" ?

yaacov (approved):

PR #1438, opened 2025-03-12
MTV-2146 | Ignore TLS Cert msg when skipCertVerify

solenoci:
Issue: When user specified they want to skip ceritificate verification we want to just inform the customer about failed connection and not about certificate. The reason behind this is that the GetTlsCertificate is our first attempt to connect to the provider and when connection fails, the customer will still get the certificate error even when they don't care about certificates.

Fix: 'TLS certificate cannot be retrieved' message is ignored when skipVerify flag is set to true and connection fails.

Ref: https://issues.redhat.com/browse/MTV-2146


mnecas (changes requested):
Please add link to the Jira issue to the commit + PR description as reference

mnecas (approved):
lgtm

PR #1446, opened 2025-03-14
Add Fedora guest conversion build for BTRFS support

mnecas:
Issue: The migration of SUSE fails due to the BTRFS. The virt-v2v fails to mount the BTRFS

Fix use Fedora base image which has support of BTRFS.


yaacov (approved):
lgtm

PR #1455, opened 2025-03-17
MTV-2172 | Set owner reference for Migration

solenoci:
Issue: When the Migration CR is created through CLI it is missing owner reference to it's plan. This in turn causes issue when during plan deletion migration CRs are left on the cluster. UI also does not show any migrations in plan's migrations section even though there are CR present.

Creation of migration through UI is not affected because UI automatically sets the owner reference of created resources.

Fix: Set owner preference for migration CR during reconciliation.

Ref: https://issues.redhat.com/browse/MTV-2172


pkg/controller/migration/controller.go
```diff
@@ -217,3 +227,55 @@ func (r *Reconciler) reflectPlan(plan *api.Plan, migration *api.Migration) {
 	}
 	migration.Status.VMs = plan.Status.Migration.VMs
 }
+
+// Set owner reference for Migration.
+// This is needed when the CR is created with CLI and not through UI which automatically
```
yaacov, line 232:
```suggestion
// This is needed so the migration CR will be auto deleted once the plan CR is deleted
```

hi,
Once this is merged the UI (and any other automation) can remove the code that adds the owner reference. once they do this, this comment will become obsolete, WDYT?

solenoci: Hey, good point, I will change it

PR #1460, opened 2025-03-19
MTV-2244 | Check length of generated volume name when using plan string templates

yaacov:
Ref: https://issues.redhat.com/browse/MTV-2244

Issue:
When using template to generate the volume name, we may generate a name that is longer the 63 chars

Fix:
Truncate names longer then 63 chars

Notes:
Truncating the name may create none unique volume names, other heuristics for ensuring valid volume names, like fall back to default name, may have their own issues. 

Screenshots:
Before:
long-pvc-name-before

After:
long-pvc-name-after

Plan with long templates:
long-pvc-name-plan-yaml

### DIFF
--- a/pkg/controller/plan/adapter/vsphere/builder.go
+++ b/pkg/controller/plan/adapter/vsphere/builder.go
@@ -845,6 +845,15 @@ func (r *Builder) mapDisks(vm *model.VM, vmRef ref.Ref, persistentVolumeClaims [
 				// fallback to default name and reset error
 				volumeName = fmt.Sprintf("vol-%v", i)
 			}
+
+			// check if the generated name is longer then 63 characters
+			if len(volumeName) > 63 {
+				// if the generated name is longer then 63 characters, trancate it
+				newVolumeName := volumeName[:63]
+				r.Log.Info("Generated volume name is longer than 63 characters, sanitizing it", "volumeName", volumeName, "newVolumeName", newVolumeName)
+
+				volumeName = newVolumeName
+			}
 		}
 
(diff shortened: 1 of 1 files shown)


yaacov:
@mnecas hi, this PR truncate the volume name to 63 chars, no heuristics to try and find the "important" parts and keep them.

pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -845,6 +845,15 @@ func (r *Builder) mapDisks(vm *model.VM, vmRef ref.Ref, persistentVolumeClaims [
 				// fallback to default name and reset error
 				volumeName = fmt.Sprintf("vol-%v", i)
 			}
+
+			// check if the generated name is longer then 63 characters
+			if len(volumeName) > 63 {
```
mnecas, line 850: Can we know this information during plan validations? So we would tell it to the user so they change the template

yaacov: hmm, it may depend on the PVC name, that may depend on the vm name ...
a lot of options, for example
is the vm name - the vm name, automatically change, have a vm anem template ?
then
is the PVC name ...
then
what is the volume name, and is it longer then 64 ....

mnecas: hmm okay it's an edge case, lets document this so the users will be aware of the possibility of truncating the PVC name

mnecas (approved):
lgtm, followup docs needed for this

PR #1464, opened 2025-03-21
Downgrade virt-v2v in fedora build to the 2.7.7

mnecas:


mnecas:
I tried the fedora image one more time and could not reproduce the issue

PR #1468, opened 2025-03-23
MTV 2285 - Preserve Static IPs fails to match manually configured MAC address on VMware VMs

Hazanel:
In VMware it is possible to manually configure the MAC address on a virtual network adatper. This results in a MAC address that can contain upper and/or lower case characters.
This results in the IP address of the VM not being preserved during migration

Fix:
Ensures that the Mac Address comparison is case-insensitive


mnecas (approved):
lgtm

PR #1469, opened 2025-03-24
MTV-2249 | Validate template for k8s spec

yaacov:
Ref: https://issues.redhat.com/browse/MTV-2249

Issue:
Currently we validate the pvc, network and volume template using label validation, but we need to use DNS1123Label validation 

FIx:
Replace the label validations with DNS1123Label validations

Tasks:
  - [x] Add tests for non valid cases
  - [x] Use sanitized VM name, so all input argument will be k8s valid names
  - [x] Use `IsDNS1123Label` instead of `IsValidLabelValue` for PVC, volume and network templates
 
Screenshot:
After:
pvc-name-invalid

### DIFF
--- a/pkg/controller/plan/adapter/vsphere/builder.go
+++ b/pkg/controller/plan/adapter/vsphere/builder.go
@@ -508,7 +508,7 @@ func (r *Builder) DataVolumes(vmRef ref.Ref, secret *core.Secret, _ *core.Config
 
 					// Create template data
 					templateData := api.PVCNameTemplateData{
-						VmName:        vm.Name,
+						VmName:        r.getPlenVMNewName(vm),
 						PlanName:      r.Plan.Name,
 						DiskIndex:     diskIndex,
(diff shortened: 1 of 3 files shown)


pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -1168,6 +1179,22 @@ func (r *Builder) getPVCNameTemplate(vm *model.VM) string {
 	return ""
 }
 
+// getPlenVMNewName returns the VM new-name if exist, o/w the VM name
+func (r *Builder) getPlenVMNewName(vm *model.VM) string {
+	// Get plan VM
+	planVMStatus := r.getPlanVMStatus(vm)
+	if planVMStatus == nil {
+		return ""
```
yaacov, line 1187:
```suggestion
		return vm.Name
```

mnecas (approved):
LGTM

PR #1474, opened 2025-03-25
MTV-2285 | Mac Address comparison case-insensitive

Hazanel:
Issue:
    In VMware it is possible to manaully configure the MAC address on a virtual network adatper.
    This results in a MAC address that can contain upper and/or lower case characters.

    It appears the validator matches strings without ignoring case.
    This results in the IP address of the VM not being preserved during migration:

    Fix:
    Ensures that the Mac Address comparison is case-insensitive

    Ref: https://issues.redhat.com/browse/MTV-2285


mnecas (approved):
LGTM, alternative to https://github.com/kubev2v/forklift/pull/1468

PR #1480, opened 2025-03-26
MTV-2303 | Sort disks for pvc template indexing same way as the vm builder

yaacov:
Ref: https://issues.redhat.com/browse/MTV-2303

Sort disks for pvc template indexing same way as the vm builder

Issue:
We don't sort the disks in the data-volume method same as in map-disks method

Fix:
Sort disks using same helper sort method used by the mapDisks method


yaacov:
@mnecas hi,

a. most changes are white spaces, best reviewed with "ignore whitespace" on
b. need help matching to JIRA ticket

yaacov:
tested on 1 vm with 3 disks, "works for me" ( :tm: )

mnecas (approved):
lgtm

PR #1489, opened 2025-03-27
MTV-2308 | VM name should not contain multiple dots

yaacov:
Ref: https://issues.redhat.com/browse/MTV-2308

Issue:
When we auto correct the VM name to conform to DNS1123 domain name, we do not trim multiple dots

Fix:
Trim multiple dots

  - [x] Replace multiple dots with one dot
  - [x] Trim dots at the start or end of a name
  - [x] Add test case for multiple dots with "-"
  - [x] Add test case for multiple leading dots 

Screenshots:
Before
dots-before

After
screenshot-console-openshift-console apps yzamir-mtv apps cnv2 engineering redhat com-2025 03 27-17_22_33


mnecas (approved):
LGTM

PR #1495, opened 2025-03-28
Ignore canceled VMs when scheduling

mansam:
Follow up to https://github.com/kubev2v/forklift/pull/1465 that fixes the issue for the other providers.


mnecas (approved):
lgtm

PR #1501, opened 2025-03-31
MTV-2303 (part) | Refactor data volume build loop

yaacov:
Issue:
Building order of data volume is not align with building order of disks

Fix:
Refactor data volume build loop to loop over disks instead of mapping pairs.

Description:
Replace the nested loop with a lookup table, so it will be easier to align the template disk index to the builder disk indexes.

  - [x] Eliminating nested loops
  - [x] Cleaner separation of concerns, looping over disks, matching storage class
  - [x] Efficient lookups using a map instead of repeated searches


yaacov:
@mnecas - best reviewed with "ignore whitespace" on

mnecas (approved):
lgtm

PR #1503, opened 2025-03-31
MTV- 2234 | Plan status is “unknown” during a non admin plan creation

Hazanel:
Issue:
Plan status is “unknown” during a non admin plan creation due to lack of permissions. Error log is ver vague.

Fix:
This scenario is valid but  adding a descriptive error to reflect that better

Ref:https://issues.redhat.com/browse/MTV-2234


pkg/controller/provider/web/vsphere/workload.go
```diff
@@ -112,6 +113,10 @@ func (r *Workload) Expand(db libmodel.DB) (err error) {
 	}
 	err = db.Get(host)
 	if err != nil {
+		log.Error(
```
mnecas, line 116: Could we instead of logging inside the inventory send a error response status to the main?

mnecas (approved):
lgtm

pkg/controller/validation/policy/client.go
```diff
@@ -372,10 +372,11 @@ func (r *Pool) Start() {
...
 					task.String())
 			} else {
-				log.Info(
+				log.Error(
+					task.Error,
 					"VM validation failed.",
-					"task",
-					task.String())
+					"task", task.String(),
+				)
```
mnecas, line 379: Are these changes related to the workload error?

Hazanel: Yes - this log indicates on validation failure  (which was also part of the workload issue)

pkg/controller/provider/web/vsphere/workload.go
```diff
@@ -112,15 +113,15 @@ func (r *Workload) Expand(db libmodel.DB) (err error) {
 	}
 	err = db.Get(host)
 	if err != nil {
-		return
+		return fmt.Errorf("error from %w: For VM '%s', host id '%s' not found - probably due to permissions issue", err, r.VM.ID, r.Host.ID)
```
itroyano, line 116: Any way to fetch the source error here or infer which permissions are missing?

pkg/controller/provider/web/vsphere/workload.go
```diff
@@ -112,15 +113,15 @@ func (r *Workload) Expand(db libmodel.DB) (err error) {
...
 	}
 	r.Host.Host.With(host)
 	cluster := &model.Cluster{
 		Base: model.Base{ID: host.Cluster},
 	}
 	err = db.Get(cluster)
 	if err != nil {
-		return
+		return fmt.Errorf("error from %w: For VM '%s' cluster id '%s' not found - probably due to permissions issue", err, r.VM.ID, host.Cluster)
```
itroyano, line 124: here too

pkg/controller/provider/web/vsphere/workload.go
Hazanel: That's not something  we can deduce from internal error returning from DB query which is very vage and non descriptive

pkg/controller/provider/web/vsphere/workload.go
Hazanel: Same apply as above

pkg/controller/provider/web/vsphere/workload.go
mnecas: I would prefer not to expose the username in the logs and we are seeing which resource we can't access, in this case `host od '%s'`.

But this got me thinking that we could theoretically do this a validation.
We are sending the VM information to the OPA and we could include also the other resources on which the VM depend.
If they would be missing we would fail, creating a condition showing the users the issues before the migration even starts.

pkg/controller/provider/container/ovirt/watch.go
```diff
@@ -298,10 +300,17 @@ func (r *VMEventHandler) validated(batch []*policy.Task) {
...
 	}()
 	for _, task := range batch {
+		forceConcern := false
+
 		if task.Error != nil {
 			r.log.Error(
 				task.Error, "VM validation failed.")
-			continue
+
+			if task.Concerns[0].Label == ForcedConcern {
```
mnecas, line 309: Shouldn't we continue everytime there is error?

pkg/controller/validation/policy/client.go
```diff
@@ -321,6 +325,18 @@ func (r *Worker) run() {
 				task.completed = time.Now()
 			} else {
 				task.Error = err
+				task.Concerns = []model.Concern{
+					//We add a dummy concern to mark forced concern we need to add upon failure in the workload builder
```
mnecas, line 329: What failure in the workload builder?

Hazanel: From the Expand() in workload.go: 
fmt.Errorf("error from %w: For VM '%s' cluster id '%s' not found - probably due to permissions issue", err, r.VM.ID, host.Cluster)

pkg/controller/provider/container/ovirt/watch.go
Hazanel: But the validation didn't actually failed, we want to arise a critical concern , hence we need to continue like a successful validation in order for the concern to be shown.
If we keep the regular flow it won't work and then we can leave the concerns in the logs only

pkg/controller/validation/policy/client.go
```diff
@@ -321,6 +321,15 @@ func (r *Worker) run() {
...
+					//hence it skips the validation and no concern will arise we need to force it for user readability
```
mnecas, line 326: Comments no longer apply

pkg/controller/provider/container/openstack/watch.go
```diff
@@ -302,7 +304,10 @@ func (r *VMEventHandler) validated(batch []*policy.Task) {
 		if task.Error != nil {
 			r.log.Error(
 				task.Error, "VM validation failed.")
-			continue
+
+			if len(task.Concerns) == 0 {
```
mnecas, line 308: Please add a comment that if there are concerns we need to update and commit the changes

pkg/controller/provider/container/openstack/watch.go
```diff
@@ -40,6 +40,8 @@ const (
 	MaxBatch = 1024
 	// Transaction label.
 	ValidationLabel = "VM-validated"
+	// Forced concern label for the validation task.
+	ForcedConcern = "Forced concern"
```
mnecas, line 44: No longer apply

mnecas (changes requested):

mnecas (approved):
lgtm

PR #1504, opened 2025-04-01
MTV-2338 | Add Shared to the pvc template information

yaacov:
Ref: https://issues.redhat.com/browse/MTV-2338

Issue:
As a migration owner,
I want customize the PVC names so I can identify shared disks using the PVC name,
so that when I create VMs with shared disks, I can identify the shared PVCs.

Fix:
Add Shared to the pvc template information

example:
```yaml
pvcNameTemplate: '{{if .Shared}}shared-{{end}}{{.VmName}}-{{.DiskIndex}}'
```

Screenshots:
Plan yaml:
pvc-template-shared

Plan running:
pvc-template-shared-run


mnecas (approved):
lgtm

PR #1510, opened 2025-04-02
MTV-2345 | Add optional vddk image global settings

yaacov:
Ref: https://issues.redhat.com/browse/MTV-2345

Add optional vddk image global settings

Issue:
As a migration administrator I would like to create the VDDK container image automatically and set it as a global settings for all vmWare providers

What happen now:
As a migration administrator I set the vddk image for each new vmware provider

What happen with the fix:
As a migration administrator I create a vddk image using my automation, and set it to be the default image
When creating new vmware provider I can skip setting the vddk image, and migration will fall back into the image I set in the global settings.

Fix:
  - [x] Add an environment variable `VDDK_IMAGE` that can be used to init the global default VDDK container image
  - [x] Add a controller settings field  `vddk_image` to set the environment variable dynamically
  - [x] Use providers vddk image, if missing use gloable default vddk image, if missing default to no vddk image
  - [x] Update vddk image validations 
  
 Why this is safe:
 when the configuration is empty, fall back to using providers vddk image or empty
 
 Screenshot:
 global-vddk-image


mnecas (approved):
lgtm

PR #1511, opened 2025-04-02
Restrict VM name to 64 chars

yaacov:
Restrict VM name to 64 chars

Issue:
We may use the VM name as source for entities that may require shorted length then K8s subdomain length

Fix:
Restrict the VM name to 63 chars, similar to K8s labels


mnecas:
Fix related to https://github.com/kubev2v/forklift/pull/1505

mnecas (approved):
lgtm

PR #1513, opened 2025-04-02
MTV-1178| Mapping static ip network to nad network get warning message

Hazanel:
Issue:
static ip from vsphere isn't always being assigned, hence when preserving  static ip is requested on a multiple nics vm  a warning is issued by the UI

Fix:
If a vm static ip is not assign - we'll assign it if possible.

Ref:https://issues.redhat.com/browse/MTV-1778

### DIFF
--- a/pkg/controller/plan/adapter/vsphere/builder.go
+++ b/pkg/controller/plan/adapter/vsphere/builder.go
@@ -260,13 +260,10 @@ func (r *Builder) PodEnvironment(vmRef ref.Ref, sourceSecret *core.Secret) (env 
 }
 
 func (r *Builder) mapMacStaticIps(vm *model.VM) (ipMap string, err error) {
-	// on windows machines we check if the interface origin is manual
-	// on linux we collect all networks.
-	isWindowsFlag := isWindows(vm)
 
 	var configurations []string
 	for _, guestNetwork := range vm.GuestNetworks {
-		if !isWindowsFlag || guestNetwork.Origin == string(types.NetIpConfigInfoIpAddressOriginManual) {
+		if guestNetwork.Origin == string(types.NetIpConfigInfoIpAddressOriginManual) {
 			gateway := ""
 			isIpv4 := net.IP.To4(net.ParseIP(guestNetwork.IP)) != nil
 			for _, ipStack := range vm.GuestIpStacks {
@@ -288,10 +285,6 @@ func (r *Builder) mapMacStaticIps(vm *model.VM) (ipMap string, err error) {
 		}
 	}
 	return strings.Join(configurations, "_"), nil
-}
-
-func isWindows(vm *model.VM) bool {
-	return strings.Contains(vm.GuestID, WindowsPrefix) || strings.Contains(vm.GuestName, WindowsPrefix)
 }
 
 // Retrieve the IP address of an ESXI host from its Management Network VNIC or fall back to the hostname.
--- a/pkg/controller/provider/container/vsphere/model.go
+++ b/pkg/controller/provider/container/vsphere/model.go
@@ -533,6 +533,8 @@ func (v *VmAdapter) Model() model.Model {
 // Apply the update to the model.
 func (v *VmAdapter) Apply(u types.ObjectUpdate) {
 	v.Base.Apply(&v.model.Base, u)
+
+	staticIpEncountered := false
 	for _, p := range u.ChangeSet {
 		switch p.Op {
 		case Assign:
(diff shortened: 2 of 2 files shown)


pkg/controller/provider/container/vsphere/model.go
```diff
@@ -790,6 +797,20 @@ func (v *VmAdapter) Apply(u types.ObjectUpdate) {
...
 		}
 	}
+
+	// There might be cases which GuestNetworks.origin (Which indicates the ip address type )is missing
+	// So we need to mark (if possible) the static ip address by setting the GuestNetworks.origin field to manual.
+	if !skipStaticIpAssign {
+		for _, nic := range v.model.NICs {
+			for i := range v.model.GuestNetworks {
+				guestNetwork := &v.model.GuestNetworks[i]
+				if strings.TrimSpace(guestNetwork.Origin) == "" && nic.MAC == guestNetwork.MAC {
```
yaacov, line 807: if we are testing for the existence of a nic with specific MAC, we can replace the nested loop with a dictionary.

``` go
	// Build a map of NICs by lowercase MAC address
	nicMap := make(map[string]struct{})
	for _, nic := range v.model.NICs {
		nicMap[strings.ToLower(nic.MAC)] = struct{}{}
	}

	// Iterate over guest networks and assign origin if MAC is in NICs
	for i := range v.model.GuestNetworks {
		guestNetwork := &v.model.GuestNetworks[i]
		if strings.TrimSpace(guestNetwork.Origin) == "" {
			if _, found := nicMap[strings.ToLower(guestNetwork.MAC)]; found {
				guestNetwork.Origin = string(types.NetIpConfigInfoIpAddressOriginManual)
			}
		}
	}
	```

Comment (yaacov) [reply pkg/controller/provider/container/vsphere/model.go:807]: p.s.
if you check that you don't overwrite the origin if exist, you don't need the `skipStaticIpAssign` check

Comment (yaacov) [reply pkg/controller/provider/container/vsphere/model.go:807]: p.p.s
can be refactored into a method with meaningful name

Comment (yaacov) [reply pkg/controller/provider/container/vsphere/model.go:807]: p.p.p.s
this is a hack we should only use on none windows machine, on windows we can trust the value of the `origin`  we only need to guess the origin using this hack if we have a linux machine.

a. we should add a guard to use this only on none windows machines
b. we should add a comment that this is a hack

File: pkg/controller/plan/adapter/vsphere/builder.go [hunk 2]
```diff
@@ -260,13 +260,10 @@ func (r *Builder) PodEnvironment(vmRef ref.Ref, sourceSecret *core.Secret) (env
 }
 
 func (r *Builder) mapMacStaticIps(vm *model.VM) (ipMap string, err error) {
-	// on windows machines we check if the interface origin is manual
-	// on linux we collect all networks.
-	isWindowsFlag := isWindows(vm)
```
mnecas, line 265: why don't we need this? 
tbh I don't remember why we needed it in the first place, @yaacov do you remember by any chance?

yaacov: on windows we have the `origin` and we don't need to invent it's value buy guessing

yaacov: now that we have code that "invent" the "origin" for none windows vms, we can omit this check, because we have the origin for all machines (in case we find a nic that match the mac of a network)

yaacov:
lgtm

yaacov (approved):

mnecas (commented):
The original issue is that you have one DHCP one static and we show error.
If I understand this issue correctly we just need to expose the information if the interface is static or DHCP to the UI and in the UI determine by the mapping.

pkg/controller/provider/container/vsphere/model.go
```diff
@@ -790,6 +799,34 @@ func (v *VmAdapter) Apply(u types.ObjectUpdate) {
...
 		}
 	}
+
+	if !staticIpEncountered {
+		v.assignStaticIpOrigin()
+	}
+}
+
+// There might be cases where GuestNetworks.origin (which indicates the IP address type) is missing.
+// In such cases, we need to mark the static IP address by setting the GuestNetworks.origin field to "manual".
```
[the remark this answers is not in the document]
Hazanel: manual means static ip
so when GuestNetworks.origin we need to decide if it's manual or not

mnecas:
Closing, we are not getting enough information for this and the information that we are getting we are exposing to the UI

PR #1522, opened 2025-04-04
MTV-2172 | Fix nil ptr dereference

solenoci:
Issue: A nil pointer dereference occurred when error happened inside `setOwnerReference` and migration obj was replaced with nil. Defer function then referenced nil migration obj.

Fix: Don't overwrite migration obj with nil on error and don't execute the `setOwnerReference` function at all when plan is not found.

Ref: https://issues.redhat.com/browse/MTV-2172

### DIFF
--- a/pkg/controller/migration/controller.go
+++ b/pkg/controller/migration/controller.go
@@ -129,24 +129,13 @@ func (r Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (r
 		r.Log.V(2).Info("Conditions.", "all", migration.Status.Conditions)
 	}()
 
-	// Set owner reference (OR) for migration CR if it was created using CLI
-	// Try to set OR before doing anything with the migration CR so we fail fast.
-	// Skip setting of OR if plan is not found
-	plan := &api.Plan{}
-	err = r.Client.Get(
-		context.TODO(),
-		client.ObjectKey{
-			Namespace: migration.Spec.Plan.Namespace,
-			Name:      migration.Spec.Plan.Name,
-		},
-		plan,
-	)
-	if err == nil {
-		err = r.setOwnerReference(migration, plan)
-		if err != nil {
-			r.Log.Error(err, "Could not set migration owner reference.")
-			return
-		}
+	// Set owner reference for migration CR if it was created using CLI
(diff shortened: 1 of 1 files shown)


yaacov:
lgtm

btw/ you can experiment with moving the plan validation into the "set owner ref" method, so the call will stay a one liner
and the method will now be "set owner if owner exist" method

pkg/controller/migration/controller.go
```diff
@@ -129,12 +129,24 @@ func (r Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (r
 		r.Log.V(2).Info("Conditions.", "all", migration.Status.Conditions)
 	}()
 
-	// Set owner ref for migration CR if it was created using CLI
-	// Try to set owner reference before doing anything with the migration CR so we fail fast.
-	migration, err = r.setOwnerReference(migration)
-	if err != nil {
-		r.Log.Error(err, "Could not set migration owner reference.")
-		return
+	// Set owner reference (OR) for migration CR if it was created using CLI
```
yaacov, line 132: nitpick, the (OR) confuses me, looks like a boolean expression :-)

yaacov (approved):
lgtm

nitpicks about a comment and where to put the check plan exist code

pkg/controller/migration/controller.go
yaacov: thanks :green_heart:

pkg/controller/migration/controller.go
```diff
@@ -249,19 +249,21 @@ func (r *Reconciler) setOwnerReference(migration *api.Migration) (*api.Migration
 		},
 		plan,
 	)
-
 	if err != nil {
-		return nil, err
+		if k8serr.IsNotFound(err) {
```
yaacov, line 253:
```suggestion
                // ... some reason we ignore this error ...
		if k8serr.IsNotFound(err) {
```

much nicer fix then adding the check before calling this method ! nice catch, making this method more flexible !

yaacov (approved):
lgtm

yaacov (approved):

mnecas (approved):

PR #1531, opened 2025-04-06
MTV-2354 | Add template helper string and math methods

yaacov:
Ref: https://issues.redhat.com/browse/MTV-2354

Add template helper string and math methods

 - [x] Add template pkg with int math and string methods
 - [x] Apply the pkg to plan templates
 - [x] Add tests for template methods
 - [x] Add readme for the template methods 

Screenshots:
After:
pvc-template-with-funcs-1

pvc-template-with-funcs


mnecas (approved):
lgtm

PR #1532, opened 2025-04-06
MTV-2054, MTV-2027, MTV-1962 |  Delete plan garbage collection

Hazanel:
Issue:
All the bugs describe certain cenarios where additional manual steps are required to fully delete a plan.
A plan must be archived before deletion to ensure its associated resources are properly cleaned up in the correct order.

Solution
The current plan is now set as an OwnerReference for all of its dependent resources.
This enables Kubernetes automatic garbage collection, ensuring that when the plan is deleted,
all associated resources are also cleaned up without requiring extra steps.

Signed-off-by: ehazan <ehazan@redhat.com>

### DIFF
--- a/pkg/controller/plan/kubevirt.go
+++ b/pkg/controller/plan/kubevirt.go
@@ -483,9 +483,11 @@ func (r *KubeVirt) EnsureVM(vm *plan.VMStatus) error {
 	}
 
 	for _, pvc := range pvcs {
-		ownerRefs := []meta.OwnerReference{vmOwnerReference(virtualMachine)}
 		pvcCopy := pvc.DeepCopy()
-		pvc.OwnerReferences = ownerRefs
+		pvc.OwnerReferences = createOwnerReferences(
+			virtualMachine.ObjectMeta,
+			metav1.TypeMeta{APIVersion: "kubevirt.io/v1",
+				Kind: "VirtualMachine"}, true)
 		patch := client.MergeFrom(pvcCopy)
 		err = r.Destination.Client.Patch(context.TODO(), pvc, patch)
 		if err != nil {
@@ -871,9 +873,10 @@ func (r *KubeVirt) createPodToBindPVCs(vm *plan.VMStatus, pvcNames []string) (er
 	allowPrivilageEscalation := false
 	pod := &core.Pod{
 		ObjectMeta: meta.ObjectMeta{
-			Namespace:    r.Plan.Spec.TargetNamespace,
-			Labels:       r.consumerLabels(vm.Ref, false),
-			GenerateName: r.getGeneratedName(vm) + "pvcinit-",
+			Namespace:       r.Plan.Spec.TargetNamespace,
+			OwnerReferences: createOwnerReferences(r.Plan.ObjectMeta, r.Plan.TypeMeta, false),
+			Labels:          r.consumerLabels(vm.Ref, false),
+			GenerateName:    r.getGeneratedName(vm) + "pvcinit-",
 		},
 		Spec: core.PodSpec{
 			RestartPolicy: core.RestartPolicyNever,
@@ -1383,9 +1386,10 @@ func (r *KubeVirt) dataVolumes(vm *plan.VMStatus, secret *core.Secret, configMap
 	annotations[AnnDeleteAfterCompletion] = "false"
 	dvTemplate := cdi.DataVolume{
 		ObjectMeta: meta.ObjectMeta{
-			Namespace:    r.Plan.Spec.TargetNamespace,
-			Annotations:  annotations,
-			GenerateName: r.getGeneratedName(vm),
+			Namespace:       r.Plan.Spec.TargetNamespace,
+			Annotations:     annotations,
+			GenerateName:    r.getGeneratedName(vm),
+			OwnerReferences: createOwnerReferences(r.Plan.ObjectMeta, r.Plan.TypeMeta, false),
 		},
 	}
 	dvTemplate.Labels = r.vmLabels(vm.Ref)
@@ -1923,10 +1927,11 @@ func (r *KubeVirt) guestConversionPod(vm *plan.VMStatus, vmVolumes []cnv.Volume,
 	// pod
 	pod = &core.Pod{
 		ObjectMeta: meta.ObjectMeta{
-			Namespace:    r.Plan.Spec.TargetNamespace,
-			Annotations:  annotations,
-			Labels:       r.conversionLabels(vm.Ref, false),
-			GenerateName: r.getGeneratedName(vm),
+			Namespace:       r.Plan.Spec.TargetNamespace,
+			Annotations:     annotations,
+			Labels:          r.conversionLabels(vm.Ref, false),
+			GenerateName:    r.getGeneratedName(vm),
+			OwnerReferences: createOwnerReferences(r.Plan.ObjectMeta, r.Plan.TypeMeta, false),
 		},
 		Spec: core.PodSpec{
 			SecurityContext: &core.PodSecurityContext{
@@ -2400,8 +2405,9 @@ func (r *KubeVirt) ensureLibvirtConfigMap(vmRef ref.Ref, vmCr *VirtualMachine, p
 func (r *KubeVirt) configMap(vmRef ref.Ref) (object *core.ConfigMap, err error) {
 	object = &core.ConfigMap{
 		ObjectMeta: meta.ObjectMeta{
-			Labels:    r.vmLabels(vmRef),
-			Namespace: r.Plan.Spec.TargetNamespace,
+			Labels:          r.vmLabels(vmRef),
+			Namespace:       r.Plan.Spec.TargetNamespace,
+			OwnerReferences: createOwnerReferences(r.Plan.ObjectMeta, r.Plan.TypeMeta, false),
 			GenerateName: strings.Join(
 				[]string{
 					r.Plan.Name,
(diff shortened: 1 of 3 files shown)


pkg/controller/plan/kubevirt.go
```diff
@@ -2686,21 +2690,38 @@ func (r *VirtualMachine) Conditions() (cnd *libcnd.Conditions) {
...
-		Controller:         &isController,
+
+	switch v := any(obj).(type) {
+	case *api.Plan:
+		apiVersion = v.APIVersion
+		kind = v.Kind
+		name = v.Name
+		uid = v.UID
+	case *cnv.VirtualMachine:
+		apiVersion = "kubevirt.io/v1"
```
solenoci, line 2708: Can't we also use `v.APIVersion` here?

pkg/controller/plan/kubevirt.go
```diff
@@ -2686,21 +2690,38 @@ func (r *VirtualMachine) Conditions() (cnd *libcnd.Conditions) {
...
+		kind = "VirtualMachine"
```
solenoci, line 2709: Same as with comment above.

pkg/controller/plan/kubevirt.go
```diff
@@ -2686,21 +2690,38 @@ func (r *VirtualMachine) Conditions() (cnd *libcnd.Conditions) {
...
 }
 
-// Create an OwnerReference from a VM.
-func vmOwnerReference(vm *cnv.VirtualMachine) (ref meta.OwnerReference) {
+// Generic function to create OwnerReferences based on the input type.
+func setOwnerReferences[T any](obj T) []meta.OwnerReference {
+	var apiVersion, kind string
+	var name string
+	var uid types.UID
 	blockOwnerDeletion := true
```
solenoci, line 2698: In current state it should not cause any issues but in the future it might.

solenoci (commented):
Just a few nitpicks I am unsure about.

pkg/controller/plan/kubevirt.go
Hazanel: If the owner reference is set correctly it shouldn't be problematic

pkg/controller/plan/kubevirt.go
```diff
@@ -2686,21 +2690,38 @@ func (r *VirtualMachine) Conditions() (cnd *libcnd.Conditions) {
 	return
 }
 
-// Create an OwnerReference from a VM.
-func vmOwnerReference(vm *cnv.VirtualMachine) (ref meta.OwnerReference) {
+// Generic function to create OwnerReferences based on the input type.
+func setOwnerReferences[T any](obj T) []meta.OwnerReference {
```
yaacov, line 2694: nitpick, `setOwnerReferences` sounds like something the set the owner ref, e.g. it will set the owner ref of the obj,
but in this case it creates an owner ref obj, can you think of a name that will not have this issue ?

pkg/controller/plan/kubevirt.go
yaacov, line 2698: As discussed in mtg (Apr 10) we do not want to block in case of Plan owner, to unblock deletion of plans stuck on children objects deletion, stuck plans may block finalize call of children creating cyclic block loop.

pkg/controller/plan/kubevirt.go
```diff
@@ -2814,8 +2835,9 @@ func (r *KubeVirt) CreatePvForNfs() (pvName string, err error) {
 	labels := map[string]string{"provider": r.Plan.Provider.Source.Name, "app": "forklift", "migration": r.Migration.Name, "plan": string(r.Plan.UID), "ova": OvaPVLabel}
 	pv := &core.PersistentVolume{
 		ObjectMeta: meta.ObjectMeta{
-			GenerateName: pvcNamePrefix,
-			Labels:       labels,
+			GenerateName:    pvcNamePrefix,
+			OwnerReferences: setOwnerReferences(r.Plan),
```
yaacov, line 2839: pv is owned by pvc, so we don't want to touch it's owner

pkg/controller/plan/kubevirt.go
```diff
@@ -2846,9 +2868,10 @@ func (r *KubeVirt) CreatePvcForNfs(pvcNamePrefix, pvName, vmID string) (pvcName
 	labels := map[string]string{"provider": r.Plan.Provider.Source.Name, "app": "forklift", "migration": string(r.Migration.UID), "plan": string(r.Plan.UID), "ova": OvaPVCLabel, kVM: vmID}
 	pvc := &core.PersistentVolumeClaim{
 		ObjectMeta: meta.ObjectMeta{
-			GenerateName: pvcNamePrefix,
-			Namespace:    r.Plan.Spec.TargetNamespace,
-			Labels:       labels,
+			GenerateName:    pvcNamePrefix,
+			Namespace:       r.Plan.Spec.TargetNamespace,
+			OwnerReferences: setOwnerReferences(r.Plan),
```
yaacov, line 2873: AFAIK pvc is owned by the dv, can you check that we don't break the delete chain here ?

I think we should own the DV it should own the PVC and it should own the PV

pkg/controller/plan/kubevirt.go
```diff
@@ -483,9 +483,8 @@ func (r *KubeVirt) EnsureVM(vm *plan.VMStatus) error {
 	}
 
 	for _, pvc := range pvcs {
-		ownerRefs := []meta.OwnerReference{vmOwnerReference(virtualMachine)}
 		pvcCopy := pvc.DeepCopy()
-		pvc.OwnerReferences = ownerRefs
+		pvc.OwnerReferences = setOwnerReferences(virtualMachine)
```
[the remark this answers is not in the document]
solenoci: DV gets deleted after successful VM creation/migration, do we want to preserve them?

solenoci:
I made a diagram of references using Elad's changes, it works and the refs are correctly set. They are just blocking.
image

pkg/controller/plan/kubevirt.go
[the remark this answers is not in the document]
Hazanel: And if the migration hasn't started ?
you need to point to the plan.
After the migration is done the owner will be changed to the target VM

pkg/controller/plan/kubevirt.go
yaacov: this PVC should own the PV, see comment about PV owner

yaacov:
@solenoci Nice diagram :+1: it helps understand who owned by how, can you add the DV and PV references to the diagram (see my comments above about not setting DV owner and setting of PV owner) ?

pkg/controller/plan/kubevirt.go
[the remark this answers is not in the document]
yaacov: we delete the DV manually someware else, we can't take ownership of the DV, so taking ownership of the PVC is the best thing to do here

solenoci:
I think I don't understand you Kobi. 

The diagram is only for vsphere migration. For OVA, there's a need to separate the PVC/PV for NFS and PVC/PV/DV for VM. 

NFS part:
- The PVC for NFS should be ideally owned by migration resource (but it can stay as Elad made it as plan) as they are created for each migration conversion pod separately, thus when migration is deleted the PVC/PV for NFS are no longer needed. 
- The PV for NFS should ideally be owned by PVC or by nothing and have reclaimPolicy set to delete.
- There is no DV for NFS

VM part (is same for all providers):
- The DV for VM during migration should be ideally owned by the migration as well, but can also be owned by plan, after migration/cancel it will be deleted.
- The PVC for VM during migration should be owned by DV, after migration it will be owned by VM. After migration cancel it will be deleted.
- The PV for VM does not have an owner set, but it has reclaimPolicy set to Delete, so we don't need to worry about referencing PV.

What do you think? If I am mistaken, please correct me.
image

yaacov:
> The DV for VM during migration should be ideally owned by the migration as well, but can also be owned by plan, after migration/cancel it will be deleted.
> The PVC for VM during migration should be owned by DV, after migration it will be owned by VM. After migration cancel it will be deleted.
> The PV for VM does not have an owner set, but it has reclaimPolicy set to Delete, so we don't need to worry about owner of PV.

Usually we create a DV and we own it ( who ever created it, in out case the plan ? )
and then automatically it owns the PVC and PV because it created them.

So in our case, the plan creates the DV and own it, 
Then when the VM is created it is added as owner of the DV.

Do you know why we change the ownership of PVC instead of DV ?

yaacov:
> The PV for NFS should ideally be owned by PVC or by nothing and have reclaimPolicy set to delete.

@solenoci I agree, do you know how to set the `reclaimPolicy` currectly so after the PVC release it , it will be deleted ?
I am ok with just set the owner as the PVC

yaacov (approved):
lgtm

solenoci:
> Do you know why we change the ownership of PVC instead of DV ?
AFAIK, DV is present just during migration, after migration, it gets deleted, why we do it that way, I have no idea.

lgtm

pkg/controller/plan/kubevirt.go
```diff
@@ -2686,21 +2690,40 @@ func (r *VirtualMachine) Conditions() (cnd *libcnd.Conditions) {
 	return
 }
 
-// Create an OwnerReference from a VM.
-func vmOwnerReference(vm *cnv.VirtualMachine) (ref meta.OwnerReference) {
-	blockOwnerDeletion := true
+// Generic function to create OwnerReferences based on the input type.
+func createOwnerReferences[T any](obj T) []meta.OwnerReference {
```
mnecas, line 2694: Would prefer not to use any generics.
To make it general we can use the metadata which all k8s resources have.
```
func createOwnerReferences(objMeta metav1.ObjectMeta, typeMeta metav1.TypeMeta) []meta.OwnerReference {
	blockOwnerDeletion := false
	isController := false
	return []meta.OwnerReference{
		{
			APIVersion:         typeMeta.APIVersion,
			Kind:               typeMeta.Kind,
			Name:               objMeta.Name,
			UID:                objMeta.UID,
			Controller:         &isController,
			BlockOwnerDeletion: &blockOwnerDeletion,
		},
	}
}
```

usage:
```
createOwnerReferences(virtualMachine.ObjectMeta, virtualMachine.TypeMeta)
```

mnecas (changes requested):

pkg/controller/plan/kubevirt.go
```diff
@@ -483,9 +483,9 @@ func (r *KubeVirt) EnsureVM(vm *plan.VMStatus) error {
 	}
 
 	for _, pvc := range pvcs {
-		ownerRefs := []meta.OwnerReference{vmOwnerReference(virtualMachine)}
 		pvcCopy := pvc.DeepCopy()
-		pvc.OwnerReferences = ownerRefs
+		pvc.OwnerReferences = createOwnerReferences(virtualMachine.ObjectMeta, metav1.TypeMeta{APIVersion: "kubevirt.io/v1",
```
mnecas, line 487: This is for the VM creation, thought that we agreed on this being the `BlockOwnerDeletion: true`

mnecas (approved):
lgtm

PR #1541, opened 2025-04-07
Refactor v2v

mnecas:
- [x] Add abstraction on top of CommandBuilder, CommandExecutor, FileSystem, EmbedFileSystem, which will allow us to inject the mocked dependecies for the testing
- [x] Add mockgen to create a mocked interfaces https://github.com/uber-go/mock
- [x] Rewrite the tests using those mocked interfaces
- [x] Add args, so we don't need to use only env vars
- [x] Move all const dirs to configurable AppConfig
- [x] Add settings to contain all env variables, move all settings variables to the settings, add all discovery to the init
- [x] Add disk struct with a type (filesystem or block), add to that struct links to those disks
- [x] Refactor v2v to use those disks
- [x] Create a Conversion, Customize structs containing needed information in argument instead of passing them.
- [x] Unify the v2v args/env var format -> Create CommandBuilder
- [x] Add validation step of the args


mnecas:
Manual testing passed: 
- cold migration - centos 8 - with LUKS
- warm migration - win 16 - with static IPs

cmd/virt-v2v/entrypoint.go
```diff
@@ -2,279 +2,76 @@ package main
...
+		fmt.Println("Failed to load variables", err)
 		os.Exit(1)
 	}
-
-	// virt-v2v or virt-v2v-in-place
-	if _, found := os.LookupEnv("V2V_inPlace"); found {
-		err = runVirtV2vInPlace()
-	} else {
-		err = runVirtV2v()
+	if err = virtV2VPrepEnvironment(env); err != nil {
```
yaacov, line 21: virtV2VPrepEnvironment is a generic name, maybe two methods, one for ca certs and another for dir setup ?

cmd/virt-v2v/entrypoint.go
```diff
@@ -2,279 +2,76 @@ package main
...
-		fmt.Println("Failed to execute virt-v2v command:", err)
+		fmt.Println("Failed prepare conversion", err)
 		os.Exit(1)
 	}
 
-	// virt-v2v-inspector
-	var disks []string
-	disks, err = utils.GetLinkedDisks()
+	// virt-v2v or virt-v2v-in-place
+	err = convert.ConvertGuest()
```
yaacov, line 32: hmm... IDK, doe it make sense to be explicit here, doing the 
`if c.IsInPlace { ...
`
here ? it will be more clear what happen here ... ?

pkg/virt-v2v/conversion/conversion.go
```diff
@@ -0,0 +1,235 @@
...
+	v2vCmdBuilder.AddPositional(c.LibvirtDomainFile)
+
+	v2vCmd := v2vCmdBuilder.Build()
+	v2vCmd.SetStdout(os.Stdout)
+	v2vCmd.SetStderr(os.Stderr)
+	return v2vCmd.Run()
+}
+
+func (c *Conversion) addVirtV2vArgs(cmd utils.CommandBuilder) (err error) {
+	cmd.AddFlag("-v").
```
yaacov, line 130: Q: do we use "-on" always ? if we do, we can add it here and remove it elsewhere ?

mnecas: ah good point!

yaacov (approved):

PR #1542, opened 2025-04-08
MTV-2356 | Safe vm name for templates

yaacov:
Ref: https://issues.redhat.com/browse/MTV-2356

Safe vm name for templates

Issue:
VM name may include dots, we need to sanitize the vm name before using it in the template.

Fix:
Check length and remove "." from vm name, before using it in the teamples.

Screenshots:
After plan spec:
vm-vith-dots-volume-name-template

After plan run:
vm-vith-dots-volume-name-template-1

### DIFF
--- a/pkg/controller/plan/adapter/vsphere/builder.go
+++ b/pkg/controller/plan/adapter/vsphere/builder.go
@@ -556,6 +556,17 @@ func (r *Builder) DataVolumes(vmRef ref.Ref, secret *core.Secret, _ *core.Config
 			}
 
 			generatedName, err := r.executeTemplate(pvcNameTemplate, &templateData)
+
+			// Validate that template output is a valid k8s label
+			errs := k8svalidation.IsDNS1123Label(generatedName)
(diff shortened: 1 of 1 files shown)


pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -1240,17 +1241,38 @@ func (r *Builder) getPVCNameTemplate(vm *model.VM) string {
...
+	// but we need to check if it is a valid k8s label
+
+	// Check if new vm name is valid k8s label
+	if len(newName) > 63 {
+		// if the new name is longer then 63 characters, trancate it
+		newName = newName[:63]
+	}
+
+	// Validate that template output is a valid k8s label
+	errs := k8svalidation.IsDNS1123Label(newName)
```
mnecas, line 1269: Please add this also after the replace to check that the errors were resolved

mnecas (approved):
lgtm

PR #1545, opened 2025-04-08
Prepare for provider-specific migrators

mansam:
This commit prepares the way for breaking up the migration runner into provider-specific implementations. Provider specific migrators will now handle the itineraries and progress through the pipeline steps, and eventually provide implementations of the phases. The current pipelines have been moved into a shared BaseMigrator that the migration runner delegates to during calls to `execute()`, though the phase implementations currently remain in the migration runner.

During each call the `execute()` the migration runner will call `ExecutePhase()` on the provider migrator, which will return a value indicating whether it provides the phase implementation. If that value is false, it will fall through to the implementation in the migration runner.

The KubeVirt live migration work will build on this by using the provider-specific Migrator interface to implement the live migration plan type. Beyond the live migration work, I hope that this can be used over the long term to extract the provider specific implementation details that have leaked into the common parts of the plan controller and encapsulate them again in provider modules.

### DIFF
--- a/pkg/controller/plan/migration.go
+++ b/pkg/controller/plan/migration.go
@@ -23,8 +23,6 @@ import (
 
 	libcnd "github.com/konveyor/forklift-controller/pkg/lib/condition"
 	liberr "github.com/konveyor/forklift-controller/pkg/lib/error"
-	libitr "github.com/konveyor/forklift-controller/pkg/lib/itinerary"
-
 	"github.com/konveyor/forklift-controller/pkg/settings"
 	batchv1 "k8s.io/api/batch/v1"
 	core "k8s.io/api/core/v1"
@@ -41,17 +39,6 @@ import (
 const (
 	NoReQ   = time.Duration(0)
 	PollReQ = time.Second * 3
-)
-
-// Predicates.
-var (
-	HasPreHook              libitr.Flag = 0x01
-	HasPostHook             libitr.Flag = 0x02
-	RequiresConversion      libitr.Flag = 0x04
-	CDIDiskCopy             libitr.Flag = 0x08
-	VirtV2vDiskCopy         libitr.Flag = 0x10
-	OpenstackImageMigration libitr.Flag = 0x20
-	VSphere                 libitr.Flag = 0x40
 )
 
 // Phases.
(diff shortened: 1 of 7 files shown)


pkg/controller/plan/migration.go
```diff
@@ -110,68 +111,6 @@ const (
 	DvStatusCheckRetriesAnnotation = "dvStatusCheckRetries"
 )
 
-var (
```
mnecas, line 113: You got leftover predicates `// Predicates.`  which are not used.

pkg/controller/plan/migrator/base/migrator.go
```diff
@@ -0,0 +1,318 @@
...
+		next = step.Name
+	}
+	r.Log.Info("Itinerary transition", "current phase", status.Phase, "next phase", next)
+	return
+}
+
+func (r *BaseMigrator) ExecutePhase(vm *plan.VMStatus) (ok bool, err error) {
+	// return ok = false to delegate to default itinerary implementation
+	// in plan/migration.go
+	return
```
mnecas, line 246: Is this for future usage?

pkg/controller/plan/migrator/base/migrator.go
```diff
@@ -0,0 +1,318 @@
...
+	case OpenstackImageMigration:
+		allowed = r.context.Plan.IsSourceProviderOpenstack()
+	case VSphere:
+		allowed = r.context.Plan.IsSourceProviderVSphere()
+	}
+
+	return
+}
+
+func (r *BasePredicate) Count() int {
```
mnecas, line 316: What's the usage of this?

pkg/controller/plan/migrator/base/migrator.go
mansam: Yes. This is an extension point for a Migrator to provide its own implementation of a migration phase. Eventually the shared implementation in the migration runner in controller/plan/migration.go:execute() should go away and phase execution should be done by the provider migrators. In order to avoid requiring the entire system to be refactored at once, the BaseMigrator in this module just hands control back to the shared implementation.

(This way refactoring can be done in stages, first with the shared implementation being moved into this BaseMigrator, and then eventually broken down into vsphere, ovirt, ova, etc packages.)

pkg/controller/plan/migrator/base/migrator.go
mansam: In the current system, there is one migration predicate and there is just a constant with a hard coded number of flags in the pkg/lib/itinerary package. The purpose of this is to encapsulate it behind the Predicate interface. It will make slightly more sense once I add a file that got missed when I committed...

pkg/controller/plan/migration.go
mansam: Thanks

mnecas (approved):
lgtm, will wait with the merging to next week

mnecas (approved):

PR #1575, opened 2025-04-15
MTV-2362 | Can't migrate guests from secure esxi proivder on MTV

Hazanel:
Issue:
Can't migrate guests from secure esxi proivder on MTV with self signed certificate.

Fix:
Add the server certificate to the certificate pool and try to establish a TLS connection with the server.

Ref: https://issues.redhat.com/browse/MTV-2362


pkg/controller/provider/validation.go
```diff
@@ -235,8 +238,34 @@ func (r *Reconciler) validateSecret(provider *api.Provider) (secret *core.Secret
...
+
+			tlsConfig := &tls.Config{
+				RootCAs: x509.NewCertPool(),
+			}
+
+			// Add the server certificate to the certificate pool
+			tlsConfig.RootCAs.AddCert(cert)
+
+			// We want to confirm that url.Host does not already include a port to avoid duplication
+			host := url.Host
```
mnecas, line 252: This uses the specified url, but if the user specifies the host, we resolve the hostname to the IP (added in your PR https://github.com/kubev2v/forklift/pull/1490). In these cases the certs can act differently
Could we get the address from `getSourceDetails`  or something similar?

pkg/controller/provider/validation.go
```diff
@@ -235,8 +238,34 @@ func (r *Reconciler) validateSecret(provider *api.Provider) (secret *core.Secret
 		}
 
 		url, _ := url.Parse(provider.Spec.URL)
-		if crt, err := util.GetTlsCertificate(url, secret); err == nil {
-			provider.Status.Fingerprint = util.Fingerprint(crt)
+		if cert, err := util.GetTlsCertificate(url, secret); err == nil {
```
mnecas, line 241: This is for the VSphere case, what about ESXi?

pkg/controller/host/validation.go
```diff
@@ -236,6 +237,20 @@ func (r *Reconciler) validateSecret(host *api.Host) (err error) {
...
+			err := r.VerifyTLSConnection(host.Spec.IpAddress, secret)
+			if err != nil {
+				switch {
+				case errors.Is(err, base.ErrProviderConnect):
+					cnd.Message = "Failed to connect to provider."
+				case errors.Is(err, base.ErrTLSCert):
+					cnd.Message = "TLS certificate cannot be retrieved."
+				case errors.Is(err, base.ErrTLSConnection):
+					cnd.Message = "Failed to create a secure connection to server."
+				}
```
mnecas, line 250: Can we set the error to the cnd.Message? So we don't need to do this case?

pkg/controller/provider/validation.go
```diff
@@ -234,32 +235,33 @@ func (r *Reconciler) validateSecret(provider *api.Provider) (secret *core.Secret
...
+		err := r.VerifyTLSConnection(provider.Spec.URL, secret)
+		if err != nil {
+			switch {
+			case errors.Is(err, base.ErrProviderConnect):
 				provider.Status.SetCondition(libcnd.Condition{
 					Type:     ConnectionTestFailed,
 					Status:   True,
 					Reason:   Tested,
 					Category: Critical,
+					Message:  "Failed to connect to provider",
```
mnecas, line 247: Same here that we would just pass the error because that's the only difference between these conditions

pkg/controller/base/controller.go
```diff
@@ -20,6 +27,12 @@ const (
 	LongReQ = time.Second * 30
 )
 
+var (
+	ErrTLSCert         = errors.New("failed to get TLS certificate")
+	ErrTLSConnection   = errors.New("failed to create a secure TLS connection")
+	ErrProviderConnect = errors.New("failed to connect to provider")
+)
+
```
mnecas, line 35: Please move these to the return statment as these are used just once

pkg/controller/base/controller.go
```diff
@@ -103,3 +110,43 @@ func (r *Reconciler) Record(object runtime.Object, cnd libcnd.Conditions) {
...
 }
+
+func (r *Reconciler) VerifyTLSConnection(rawURL string, secret *core.Secret) error {
+	parsedURL, err := url.Parse(rawURL)
+	if err != nil {
+		return fmt.Errorf("invalid URL: %w", err)
+	}
+
+	// Attempt to get certificate
+	cert, err := util.GetTlsCertificate(parsedURL, secret)
```
mnecas, line 121: Should we check if the cert is not nil?

mnecas (commented):
One small NP but generally looks good

mnecas (approved):

PR #1577, opened 2025-04-15
Offload | Add vendor to the volume populator

mnecas:
Issue: For d/s build we need to have vendored deps

Fix: Run `go mod vendor`


mnecas:
fyi @rgolangh

PR #1579, opened 2025-04-16
OCP Live Migration

mansam:


vendor/sigs.k8s.io/mcs-api/pkg/apis/v1alpha1/serviceexport.go
```diff
@@ -0,0 +1,98 @@
...
+	// +optional
+	// +patchStrategy=merge
+	// +patchMergeKey=type
+	// +listType=map
+	// +listMapKey=type
+	Conditions []ServiceExportCondition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type"`
+}
+
+// ServiceExportConditionType identifies a specific condition.
+type ServiceExportConditionType string
```
yaacov, line 52: Q: in other parts of our API the condition types are also strings, over time we added more and more conditions as strings, and now it's hard to keep track of all our possible conditions :-(

does it make sense to make this type more limiting, so we will have to update it each time we add new condition type, making a source of truth for all our possible condition for this CR ?

cc:// @solenoci hi, if we made the Plan condition types more strict, would it help with your task of creating a unified "state" field ?

yaacov: > Forklift is not the only possible writer,

Ahh, i missed that, why would some other operator change the status of a resource Forklift is managing ?

mansam: No reason some other operator would change it, but because CRs _can_ be changed by an API user (even though it's unlikely) I think it's important to consider that when considering how to restrict field values (to make sure we either completely prevent values we don't expect via something like CRD validation or are tolerant of unexpected values).

yaacov: Sorry for taking over this PR, this discussion is a bit off topic :-)

> Having less conditions 

My concern is about havin "known" conditions, for example if you want to build a method that will take the conditions and return one "status" code, it is helpful that you covered all the checked all the used conditions types. it the condition type is a string, you never know if you tested for all available conditions.

pkg/apis/forklift/v1beta1/plan.go
```diff
@@ -134,6 +140,9 @@ type PlanSpec struct {
 	// Determines if the plan should skip the guest conversion.
 	// +kubebuilder:default:=false
 	SkipGuestConversion bool `json:"skipGuestConversion,omitempty"`
+	// Migration type. e.g. "cold", "warm", "live". Supersedes the `warm` boolean if set.
+	// +optional
```
mnecas, line 144:
```
	// +kubebuilder:validation:Enum=cold;warm;live
```

pkg/controller/plan/migrator/ocp/live.go
```diff
@@ -0,0 +1,1601 @@
...
+		step, found := vm.FindStep(r.Step(vm))
+		if !found {
+			vm.AddError(fmt.Sprintf("Step '%s' not found", r.Step(vm)))
+			return
+		}
+		vm.MarkedStarted()
+		step.Phase = api.StepCompleted
+		vm.Phase = r.Next(vm)
+	case PreHook, PostHook:
+		// delegate to common pipeline
```
mnecas: Note: need more info about this

pkg/controller/plan/migrator/ocp/live.go
```diff
@@ -0,0 +1,1601 @@
...
+		})
+	if err != nil {
+		err = liberr.Wrap(err)
+	}
+	return
+}
+
+// Ensurer has the limited responsibility of ensuring resources
+// are present in the destination cluster and namespace.
+type Ensurer struct {
```
mnecas, line 957: +1

pkg/controller/plan/migrator/ocp/live.go
```diff
@@ -0,0 +1,1601 @@
...
+
+// EnsureConfigMaps exist in the destination cluster's target namespace. We attempt to create ConfigMaps
+// with the same name that they have on the source cluster because they are likely to be shared between
+// multiple VMs. If one with a matching name already exists, we assume it's the intended ConfigMap for
+// the VM to mount.
+// TODO: consider raising a concern at the VM or plan level if a configmap with the desired
+// name already exists but does not have the annotation indicating that Forklift created it.
+func (r *Ensurer) EnsureConfigMaps(vm *planapi.VMStatus, configMaps []core.ConfigMap) (err error) {
+	for _, configMap := range configMaps {
+		err = r.Destination.Client.Create(context.Background(), &configMap)
```
mnecas: I see in all of them we just try to create it not update it. Could there be a problems with the long running migrations or changing plans over longer time? Know it most likely does not apply here just wanted double check.

pkg/controller/plan/migrator/ocp/live.go
```diff
@@ -0,0 +1,1601 @@
...
+		}
+		vm.Phase = r.Next(vm)
+	case EnsureInstanceType:
+		step, found := vm.FindStep(r.Step(vm))
+		if !found {
+			vm.AddError(fmt.Sprintf("Step '%s' not found", r.Step(vm)))
+			return
+		}
+		var required bool
+		required, err = r.RequiresLocalInstanceType(vm)
```
mnecas, line 415: So this should not affect the VMs from templates right?

pkg/controller/plan/migrator/ocp/live.go
mnecas: think we should apply this to the original migration

pkg/controller/plan/migrator/ocp/live.go
```diff
@@ -0,0 +1,1601 @@
...
+			// or preference with the same name is the same and use it, and if there
+			// isn't one with the same name then create it. This situation exists with
+			// secrets and configmaps as well, and like that situations it could result
+			// in failed migrations if the CRs are actually different.
+			// We don't want to create our own with different names, because that will
+			// result in the VM spec changing every time the VM is migrated between clusters.
+			//
+			// If the VM refers to cluster scope preferences and instance types rather than
+			// local scope, we should probably not attempt to create missing ones.
+			Instancetype: source.Object.Spec.Instancetype,
```
mnecas, line 1146: Right now MTV has some what support of instance types but it does not apply well so lot of VMs might be created from templates.

mnecas (commented):
Do we have some validations in place to check that the feature is enabled on the Kubevirt?

pkg/controller/plan/migrator/ocp/live.go
```diff
@@ -0,0 +1,1601 @@
...
+			Preference:   source.Object.Spec.Preference,
+			Template:     source.Object.Spec.Template.DeepCopy(),
```
mnecas, line 1148: This might break if the source.Object.Spec.Template is nil

pkg/controller/plan/migrator/ocp/live.go
mnecas: Please add ok=false or switch to ok=false by default

pkg/controller/plan/adapter/ova/adapter.go
```diff
@@ -17,12 +16,12 @@ func (r *Adapter) Builder(ctx *plancontext.Context) (builder base.Builder, err e
...
 // Constructs a OVA validator.
-func (r *Adapter) Validator(plan *api.Plan) (validator base.Validator, err error) {
-	v := &Validator{plan: plan}
-	err = v.Load()
-	if err != nil {
-		return
-	}
+func (r *Adapter) Validator(ctx *plancontext.Context) (validator base.Validator, err error) {
+	v := &Validator{Context: ctx}
+	//err = v.Load()
```
mansam: todo: delete

pkg/controller/plan/migrator/ocp/live.go
```diff
@@ -0,0 +1,1779 @@
...
+	CreateSecrets                          = "CreateSecrets"
+	CreateConfigMaps                       = "CreateConfigMaps"
+	EnsurePreference                       = "EnsurePreference"
+	EnsureInstanceType                     = "EnsureInstanceType"
+	EnsureDataVolumes                      = "EnsureDataVolumes"
+	CreateTarget                           = "CreateTarget"
+	CreateVirtualMachineInstanceMigrations = "CreateVirtualMachineInstanceMigrations"
+	WaitForStateTransfer                   = "WaitForStateTransfer"
+	PostHook                               = "PostHook"
+	Completed                              = "Completed"
```
mnecas: NP: Some of these are duplicated from v1beta1 api

pkg/controller/plan/migrator/ocp/live.go
```diff
@@ -0,0 +1,1779 @@
...
+		next = Completed
+		if err != nil {
+			r.Log.Error(err, "Next phase failed.")
+		}
+	} else {
+		next = step.Name
+	}
+	r.Log.Info("Itinerary transition", "current phase", vm.Phase, "next phase", next)
+	return
+}
```
mnecas, line 219: This looks quite generic to all migratiros could this be moved to out to base?

pkg/controller/plan/migrator/ocp/live.go
```diff
@@ -0,0 +1,1779 @@
...
+
+// Ensurer has the limited responsibility of ensuring resources
+// are present in the destination cluster and namespace.
+type Ensurer struct {
+	*plancontext.Context
+	SourceClient client.Client
+}
+
+// EnsureTargetVMIM
+func (r *Ensurer) EnsureTargetVMIM(vm *planapi.VMStatus, target *cnv.VirtualMachineInstanceMigration) (out *cnv.VirtualMachineInstanceMigration, err error) {
```
mnecas, line 963: Please move the ensurere to separate file, I love the idea and want to start refactorying other methods to it

mansam:
@mnecas Does this need anything else?

I'm not sure what the .kustomized_manifests issue is, attempting to regenerate stuff doesn't result in any changes.

mnecas:
@mansam please try `make generate-manifests kustomized-manifests`

PR #1580, opened 2025-04-16
MTV-2025, MTV-1673 | NICs mapping is unordered

Hazanel:
Issue:
When migrating VMs with multiple NICs the mapping order must match source VM order

Fix:
Iterate over vm.NICs directly, preserving the original order of network interfaces from the source VM during migration. Previously, networks were generated based on the order of network mappings (netMapIn), which could result in a different NIC order in the target VM.

Ref: https://issues.redhat.com/browse/MTV-2025
     https://issues.redhat.com/browse/MTV-1673

### DIFF
--- a/pkg/controller/plan/adapter/vsphere/builder.go
+++ b/pkg/controller/plan/adapter/vsphere/builder.go
@@ -700,89 +700,75 @@ func (r *Builder) mapNetworks(vm *model.VM, object *cnv.VirtualMachineSpec) (err
 
 	numNetworks := 0
 	netMapIn := r.Context.Map.Network.Spec.Map
-	for i := range netMapIn {
-		mapped := &netMapIn[i]
 
-		// Skip network mappings with destination type 'Ignored'
-		if mapped.Destination.Type == Ignored {
+	for _, nic := range vm.NICs {
+		mapped := r.findNetworkMapping(nic, netMapIn)
+
+		// Skip if no valid mapping found or the destination type is Ignored
+		if mapped == nil || mapped.Destination.Type == Ignored {
 			continue
 		}
 
-		ref := mapped.Source
-		network := &model.Network{}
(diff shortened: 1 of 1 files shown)


pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -700,86 +700,72 @@ func (r *Builder) mapNetworks(vm *model.VM, object *cnv.VirtualMachineSpec) (err
...
-		if fErr != nil {
-			err = fErr
-			return
-		}
+	for _, nic := range vm.NICs {
+		var mapped *api.NetworkPair
+		var network *model.Network
+
+		// Find the mapping for this NIC
+		for i := range netMapIn {
```
mnecas, line 709: please move this finind of the mapping to separate function

mnecas (commented):
generally lgtm, small nit pick
were you able to verify that this solved the issue?

Hazanel:
Yes I did
Source VM:
image

Dest VM:
image

pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -700,91 +702,84 @@ func (r *Builder) mapNetworks(vm *model.VM, object *cnv.VirtualMachineSpec) (err
...
+		if err := r.Source.Inventory.Find(network, candidate.Source); err != nil {
+			continue
+		}
+
+		if (network.Variant == vsphere.NetDvPortGroup || network.Variant == vsphere.OpaqueNetwork) &&
+			nic.Network.ID == network.Key || nic.Network.ID == network.ID {
+			return candidate, nil
+		}
+	}
+	return nil, ErrNoNetworkMapping
```
mnecas, line 780: Don't think that we need to return the error, we already check if the map is nil and then continue. Plust there are no other errors from this function

mnecas (approved):

PR #1584, opened 2025-04-17
MTV-2157 | Enable nested virt on VMs from VMware

mnecas:
Issue:
When a VM from VMware has a Hardware virtualization enabled we are not propagating this to the VM.

Fix:
If VM is using the Hardware virtualization use the CPU host-model and set the CPU feature flags for the nested virt as option, this way the VM can run on both architectures depending on the host.

Ref:
- https://issues.redhat.com/browse/MTV-2157
- https://access.redhat.com/solutions/6692341
- https://issues.redhat.com/browse/CNV-60150


yaacov (approved):
lgtm

PR #1599, opened 2025-04-22
Add snapshot creation and deletion to copy offload

mnecas:
Issue: To make sure that we are not touching the VMDKs we will create a snapshot on top of the VM to protect the disks.

Fix: To combine the warm migration snapshot creation for the offload usage I needed to create a new status for the snapshot informaiton.


mnecas:
@rgolangh @mansam could you please take a look?

mansam (approved):
This approach seems pretty reasonable to me. I think that over the long term we'll want to think about how to report on provider- or migration-type-specific details in the migration status without having to add new top level structs (like `warm` and `offloadSnapshot`), but that's a design issue to be worked out later.

PR #1604, opened 2025-04-24
Offload | Certificate tool - Sanity test

amitosw15:


cmd/vsphere-xcopy-volume-populator/certificate-tool/go.mod
```diff
@@ -0,0 +1,53 @@
+module certificate-tool
+
+go 1.23.0
```
mnecas, line 3: @rgolangh we started distributing the 1.23 build image so this is no longer a blocker on our side, and if needed,d we can also bump the other parts of the offload plugin.
https://catalog.redhat.com/software/containers/ubi9/go-toolset/61e5c00b4ec9945c18787690

rgolangh: it is not that important for that tool. Lets follow the cmd/vsphere-xcopy-volume-populator/go.mod version, which is 1.22 for now.

cmd/vsphere-xcopy-volume-populator/certificate-tool/assets/manifests/xcopy-setup/controller.yaml
```diff
@@ -0,0 +1,49 @@
...
+        - args:
+            - -v
+            - "5"
+          env:
+            - name: ${.POD_NAMESPACE}
+              valueFrom:
+                fieldRef:
+                  apiVersion: v1
+                  fieldPath: metadata.namespace
+            - name: OVIRT_POPULATOR_IMAGE
```
rgolangh, line 27: this is nice to have, though not a must,  to have the other 2 populators working. @mnecas in case you are interested to take this further.

cmd/vsphere-xcopy-volume-populator/vmkfstools-wrapper/vmkfstools-wrapper.sh
```diff
@@ -1,63 +1,113 @@
...
-# Ensure that both flags are provided
+
+
 if [ -z "$source_vmdk" ] || [ -z "$target_lun" ]; then
     usage
 fi
 
+if [ "$verify_md5_value" = "true" ]; then
+    do_md5_check="true"
+fi
```
rgolangh, line 44: duplicate vars?

cmd/vsphere-xcopy-volume-populator/certificate-tool/internal/k8s/rbac.go
```diff
@@ -0,0 +1,101 @@
...
+				APIGroups: []string{""},
+				Resources: []string{"persistentvolumeclaims"},
+				Verbs:     []string{"get", "list", "watch", "patch", "create", "update", "delete"},
+			},
+		},
+	}
+}
+
+// NewClusterRole returns a ClusterRole for managing various K8s and Forklift resources.
+func NewClusterRole(roleName string) *rbacv1.ClusterRole {
```
rgolangh, line 28: in fact this is equivalent to `operator/config/rbac/populator-controller/*.yaml`
check if you can use that will always be consistent with the current code

cmd/vsphere-xcopy-volume-populator/certificate-tool/Makefile
```diff
@@ -0,0 +1,98 @@
...
+check-tools:
+	@command -v yq >/dev/null || (echo "yq not installed" && exit 1)
+	@command -v kubectl >/dev/null || (echo "kubectl not installed" && exit 1)
+	@command -v go >/dev/null || (echo "go not installed" && exit 1)
+
+build: ## Build the certificate-tool CLI binary.
+	go build -o $(BINARY) .
+
+prepare: check-tools build ## Run certificate-tool create-env subcommand.
+	@kubectl apply -f $(OPENSTACK_POPULATOR)
```
rgolangh, line 52: apply is mutating the file you are working on - it sets an annotation on the resource in question with the former state. I think in most cases we need `create` or `replace`

rgolangh: I think we can get rid of kubectl completely, and use the client-go library. It supports KUBECONFIG anyway, and with that we loose the dependency on kubectl as a tool, and we also just read `config/static_values.yaml` into the code and then replace the values in the go code itself, and remove the stage where we read yamls to env vars and so on... so we won't need `yq` tool either, because we can just read yaml using go. So we would be left only with go code. which means that people can run the binary , with no deps at all

rgolangh:
ping

PR #1605, opened 2025-04-24
Add advancedOption parameter to /hosts/{id} API

machacekondra:
This PR add a collection of `advanedOption` field from `HostSystem` object. This option is only a reference so we store a reference in a database. User can fetch value of `advancedOption` setting from the `/hosts/{id}` endpoint by calling:
```
/providers/vsphere/1/hosts/{id}?advancedOption={optionName}
```

Issue: https://issues.redhat.com/browse/ECOPROJECT-2736

### DIFF
--- a/pkg/controller/provider/container/openstack/collector.go
+++ b/pkg/controller/provider/container/openstack/collector.go
@@ -2,6 +2,7 @@ package openstack
 
 import (
 	"context"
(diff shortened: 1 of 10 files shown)


mnecas (approved):
lgtm

mnecas (approved):

PR #1610, opened 2025-04-24
xcopy-populator: Make the populator PVC RWX

rgolangh:
This is mainly to allow live migrations in OCP

### DIFF
--- a/pkg/controller/plan/adapter/vsphere/builder.go
+++ b/pkg/controller/plan/adapter/vsphere/builder.go
@@ -1256,7 +1256,6 @@ func (r *Builder) PopulatorVolumes(vmRef ref.Ref, annotations map[string]string,
 						Annotations: annotations,
 					},
 					Spec: core.PersistentVolumeClaimSpec{
-						AccessModes:      []core.PersistentVolumeAccessMode{core.ReadWriteOnce},
 						StorageClassName: &storageClass,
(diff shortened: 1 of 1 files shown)


pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -1256,7 +1256,7 @@ func (r *Builder) PopulatorVolumes(vmRef ref.Ref, annotations map[string]string,
 						Annotations: annotations,
 					},
 					Spec: core.PersistentVolumeClaimSpec{
-						AccessModes:      []core.PersistentVolumeAccessMode{core.ReadWriteOnce},
+						AccessModes:      []core.PersistentVolumeAccessMode{core.ReadWriteMany},
```
mnecas, line 1259: Won't this be problematic with the StorageClass which does not support the RWX?

rgolangh: I suppose this PVC becomes the VM pvc, don't we want to make that RWX by default to make live migration available? 
Looking further into the code I see the better thing to do is to take that access mode from the storage mapping of the destination - https://github.com/kubev2v/forklift/blob/9ee5931531f937cb4dd03d435aee1108ec9ad64f/pkg/apis/forklift/v1beta1/mapping.go#L95

mnecas (approved):

PR #1611, opened 2025-04-24
Konflux downstream

solenoci:
Enable hermetic build for individual forklift components.

Issue: If we want to release using Konflux we have to satisfy release policies. We have to have hermetic builds from officially supported images.

Fix:
- Enable hermetic build for validation
- Enable hermetic build for populator-controller
- Enable hermetic build for ova-provider-server
- Enable hermetic build for openstack-populator
- Enable hermetic build for forklift-operator
- Enable hermetic build for forklift-controller
- Enable hermetic build for forklift-api
- Enable hermetic build for forklift-operator-bundle
- Add .kustomized_manifests for forklift-operator-bundle
- Add validation of .kustomized_manifests

Ref: https://issues.redhat.com/browse/MTV-2280


solenoci:
Still WIP, so I expect some failures in Konflux

build/vsphere-xcopy-volume-populator/Containerfile-downstream
```diff
@@ -0,0 +1,45 @@
...
+# them in subsequent builds if they change
+COPY go.mod go.sum ./
+RUN go mod download && go mod verify
+
+COPY . .
+RUN --mount=type=cache,target=/go/pkg/mod \
+    --mount=type=cache,target=/.cache/go-build \
+    go build -o bin/vsphere-xcopy-volume-populator
+
+RUN make vmkfstools-wrapper
```
mnecas, line 15: Is this running under the forklift project root or under the offload dir?
If it's under root you need to use the midstream example of the dockerfile

solenoci: It's running in project root, however the downstream containerfile is not used anywhere yet as it is missing a component in Konflux. For that, we have to have merged https://gitlab.cee.redhat.com/releng/konflux-release-data/-/merge_requests/5196 and after that we can modify the xcopy component to reflect reality. So the reason why it does not make sense yet is that it will need to be changed later either way.

mnecas (approved):
lgtm

PR #1617, opened 2025-04-25
MTV-2368 | Migration plan hangs and other migration

Hazanel:
Issue:
The running migration plan hangs and other migration plans cannot start when the guest migration fails in the ImageConversion phase and the PVC status remains pending

Fix:
The loop that builds kDisks was skipped entirely because no matching PVC was found for any disk. This line assumes at least one disk exists in kDisks, which isn't guaranteed. That causes an exception that blocks the entire plan and subsequent ones. Before accessing kDisks[bootDisk], add a guard clause

Ref: https://issues.redhat.com/browse/MTV-2368


pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -980,6 +981,21 @@ func (r *Builder) mapDisks(vm *model.VM, vmRef ref.Ref, persistentVolumeClaims [
...
+		return fmt.Errorf("no disks were successfully mapped for VM %s", vm.Name)
+	}
+
+	if bootDisk < 0 || bootDisk >= len(kDisks) {
+		r.Log.Info("Boot disk index out of range", "bootDisk", bootDisk, "diskCount", len(kDisks), "vm", vm.Name)
+		return fmt.Errorf("boot disk index %d is out of range for VM %s (have %d disks)", bootDisk, vm.Name, len(kDisks))
+	}
 	// For multiboot VMs, if the selected boot device is the current disk,
 	// set it as the first in the boot order.
 	kDisks[bootDisk].BootOrder = ptr.To(uint(1))
```
mnecas, line 1001: please move this under the `if len(kDisks) == 0 {` to the `else if bootDisk < len(kDisks)` and remove the error above.

Hazanel: Like this ?
if len(kDisks) == 0 {
		r.Log.Info("No disks were successfully mapped", "vm", vm.Name, "vmID", vmRef.ID)
		for _, d := range disks {
			r.Log.Info("Unmapped disk", "diskFile", d.File)
		}
		for key, pvc := range pvcMap {
			r.Log.Info("Available PVC mapping", "diskKey", key, "pvcName", pvc.Name)
		}
		return fmt.Errorf("no disks were successfully mapped for VM %s", vm.Name)
	} else if bootDisk <  len(kDisks) {
		r.Log.Info("Boot disk index out of range", "bootDisk", bootDisk, "diskCount", len(kDisks), "vm", vm.Name)
		return fmt.Errorf("boot disk index %d is out of range for VM %s (have %d disks)", bootDisk, vm.Name, len(kDisks))
	}

Hazanel: Like this?
} else if bootDisk < len(kDisks) {
		r.Log.Info("Boot disk index out of range", "bootDisk", bootDisk, "diskCount", len(kDisks), "vm", vm.Name)
		return fmt.Errorf("boot disk index %d is out of range for VM %s (have %d disks)", bootDisk, vm.Name, len(kDisks))
	}

PR #1618, opened 2025-04-25
xcopy-populaotr: allow the populator SA fetch PVs

rgolangh:
The populator needs to use the PV spec.CSI.volumeHandle because some
storage vendors need that to identify the LUN and fetch its details. The
PVC name or volumeName is not acceptable.
The volumeHadle is by the CSI spec the identifier in the storage system
so it is expected every provider can use that to uniquely identify
that LUN

### DIFF
--- a/pkg/controller/plan/adapter/vsphere/builder.go
+++ b/pkg/controller/plan/adapter/vsphere/builder.go
@@ -11,6 +11,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"time"
 
 	"k8s.io/klog/v2"
 
@@ -1530,10 +1531,10 @@ func (r *Builder) getNetworkNameTemplate(vm *model.VM) string {
 
 // MergeSecrets merges the storage vendor secret into the migration secret
 func (r *Builder) mergeSecrets(migrationSecret, migrationSecretNS, storageVendorSecret, storageVendorSecretNS string) error {
-	dst := &core.Secret{}
+	dst := core.Secret{}
 	if err := r.Destination.Get(context.Background(), client.ObjectKey{
 		Name:      migrationSecret,
-		Namespace: migrationSecretNS}, dst); err != nil {
+		Namespace: migrationSecretNS}, &dst); err != nil {
 		return fmt.Errorf("failed to get migration secret: %w", err)
 	}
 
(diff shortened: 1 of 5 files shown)


mnecas (commented):
Think this needs the extra permissions in the operator

pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -1574,9 +1576,22 @@ func (r *Builder) mergeSecrets(migrationSecret, migrationSecretNS, storageVendor
 			dst.Data["GOVMOMI_INSECURE"] = value
 		}
 	}
+
 	// Update secret1 with the merged data.
-	if err := r.Destination.Update(context.Background(), dst); err != nil {
-		return fmt.Errorf("failed to update secret1: %w", err)
+	maxRetries := 10
+	for i := 0; i <= maxRetries; i++ {
```
mnecas, line 1582: What's the problem that we need to add the retries? collision?

pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -1574,9 +1576,22 @@ func (r *Builder) mergeSecrets(migrationSecret, migrationSecretNS, storageVendor
...
+		current := core.Secret{}
+		if err := r.Destination.Get(context.Background(), client.ObjectKey{
+			Name:      migrationSecret,
+			Namespace: migrationSecretNS}, &current); err != nil {
+			return fmt.Errorf("failed to get migration secret: %w", err)
+		}
+		current.Data = dst.Data
+		if err := r.Destination.Update(context.Background(), &current); err != nil {
+			r.Log.Info(fmt.Sprintf("Try numer %d - failed to update secret1: %v",i, err))
+			time.Sleep(2 * time.Second)
```
mnecas, line 1592: The controller is running in a single process without go routines this will halt all migrations happening at that time. Trying to avoid the sleep as much as possible

pkg/controller/plan/adapter/vsphere/builder.go
rgolangh: so maybe we should fail in case the update collides?
Perhaps a patch is a better approach here

rgolangh:
Dropping this PR, it seems uneeded now

PR #1619, opened 2025-04-25
Move externally-facing constants to API package

mansam:
Goals:
- Reduce duplication by defining constants in a single place that can be imported elsewhere without import cycles.
- Expose constants that make up part of Forklift's API envelope (plan statuses, conditions) in the `api` package.

This could possibly be taken further, but this is a starting point for consideration. Importing from the `api` package and the new names do make the constants slightly harder to read while skimming code, but it avoids error-prone redefinition in multiple packages, and it prevents some confusing situations that currently exist where a vague constant like `Running` is defined in one package for a particular use and imported elsewhere for a conceptually unrelated purpose.

### DIFF
--- a/pkg/apis/forklift/v1beta1/doc.go
+++ b/pkg/apis/forklift/v1beta1/doc.go
@@ -22,5 +22,81 @@ limitations under the License.
 // +groupName=forklift.konveyor.io
 package v1beta1
 
+import libcnd "github.com/konveyor/forklift-controller/pkg/lib/condition"
+
 // GlanceSource represents an image of which the source is Glance, to be used in storage mapping
 const GlanceSource = "glance"
+
+// Conditions
+const (
+	ConditionExecuting = "Executing"
+	ConditionRunning   = "Running"
+	ConditionPending   = "Pending"
+	ConditionCanceled  = "Canceled"
+	ConditionSucceeded = "Succeeded"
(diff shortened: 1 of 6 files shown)


pkg/apis/forklift/v1beta1/doc.go
```diff
@@ -22,5 +22,81 @@ limitations under the License.
...
 package v1beta1
 
+import libcnd "github.com/konveyor/forklift-controller/pkg/lib/condition"
+
 // GlanceSource represents an image of which the source is Glance, to be used in storage mapping
 const GlanceSource = "glance"
+
+// Conditions
+const (
+	ConditionExecuting = "Executing"
```
mansam, line 32: Another approach could be to put all condition related consts in their own subpackage, so they can be imported as `condition.Executing` rather than `api.ConditionExecuting`, if that's more appealing or easier to read.

yaacov: I like to have one place where i can see all the possible "statuses" and "conditions" , the use case I see, is a user trying to understand the "good" flow and possible "bad" flows, having a "one source of truth" for all possible flows.

Happy with current implementation, as it is a "one place" that cover all possible flows.

mnecas (approved):
lgtm

PR #1648, opened 2025-04-30
MTV-2451 | Add virt-v2v-inspector step to warm migratioin

mnecas:
Issue:
When runnning a warm migration the guest conversion, is happening after we transfer all disks. So if there is problem with the guest conversion we fail late and users need to retry the disk transferes.

Fix:
To work around this issue we can add virt-v2v-inspector which would run on already running VMs. The requiremet for that step is that we would need to create a snapshot so the underlaying disk is not locked by the runnign VM. Additionally the virt-v2v-inspector needs to allow us to run on running VMs https://issues.redhat.com/browse/RHEL-88543.

Ref: https://issues.redhat.com/browse/MTV-2451


pkg/controller/plan/migrator/base/doc.go
```diff
@@ -58,6 +58,7 @@ var (
 			{Name: api.PhaseCreateInitialSnapshot},
 			{Name: api.PhaseWaitForInitialSnapshot},
 			{Name: api.PhaseStoreInitialSnapshotDeltas, All: VSphere},
+			{Name: api.PhaseVirtV2vInspection, All: RequiresConversion},
```
mnecas, line 61: @mansam I was trying to move this to a separate step, but the `PhaseCreateDataVolumes`, which is after this step, is still in the initilize step. So we would need to move it out, which didn't seem doable.
Got any ideas how we could have this as separate step. Didn't want the initialize step to take such long time.

PR #1656, opened 2025-05-04
Add VDDK BuildConfig, ImageStream and endpoints

AvielSegev:
https://github.com/user-attachments/assets/df839748-0010-46f3-b01d-6b9cd62b8eac

- Introduce OpenShift BuildConfig (`vddk`) and ImageStream for building/pushing the VDDK image
- Implement `VddkHandler` with endpoints:
- `POST /vddk/build-image` to upload tar and trigger build
- `GET  /vddk/download-tar` to retrieve the uploaded archive
- `GET  /vddk/image-url` to verify image availability and return its Docker reference

Please view the attached video to see the flow. After the merge, we should connect the UI to use those endpoints.


yaacov:
NICE !

you added the API endpoint in the forklift inventory server, adding vddk sounds more an API thing then an inventory thing
can you move the endpoint to the api server ?
https://github.com/ahadas/forklift/tree/main/pkg/forklift-api

operator/config/rbac/vddk-builder-role.yaml
```diff
@@ -0,0 +1,43 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: vddk-builder
+  namespace: openshift-mtv
```
yaacov, line 5: I think we don't put the final namespace here, only a placeholder that will be changed later by the controller who deploy the different resources, can you look into it ?

yaacov: cc:// @mnecas @solenoci  hi, do you know how we handle namespaces in the operator configs to the operator will deploy in the user defined namespace?

mnecas: yup 
```
 namespace: {{ app_namespace }}
```

pkg/controller/provider/web/vsphere/vddk.go
```diff
@@ -0,0 +1,217 @@
...
+const (
+	defaultNamespaceFile = "/var/run/secrets/kubernetes.io/serviceaccount/namespace"
+	uploadDirPerm        = 0700
+	VddkRoot             = ProvidersRoot + "/" + "vddk" // Route
+)
+
+var (
+	buildConfigName  = "vddk"
+	registryImageTag = "vddk:latest"
+	vddkTarFileName  = "vddk.tar.gz"
```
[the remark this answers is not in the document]
AvielSegev: It’s not included in the current implementation, but it’s definitely something to consider as a follow-up task. Right now, each upload overrides the tar file and pushes a new image to the registry. If we want to support VDDK versioning, we’ll also need to discuss the design and how we want to presented it in the UI.

AvielSegev:
> NICE !
> 
> you added the API endpoint in the forklift inventory server, adding vddk sounds more an API thing then an inventory thing can you move the endpoint to the api server ? https://github.com/ahadas/forklift/tree/main/pkg/forklift-api

Thanks @yaacov !

@mnecas, I think we wanted it to be part of the inventory (inside forklift-controller), right?

mnecas:
> > NICE !
> > you added the API endpoint in the forklift inventory server, adding vddk sounds more an API thing then an inventory thing can you move the endpoint to the api server ? https://github.com/ahadas/forklift/tree/main/pkg/forklift-api
> 
> Thanks @yaacov !
> 
> @mnecas, I think we wanted it to be part of the inventory (inside forklift-controller), right?

The forklift-api is now used as kubernetes admission webhooks, think we should rename it as it causes confusion.
Yeah I was suggesting the inventory which is part of the forklift-controller.

pkg/controller/provider/web/vsphere/vddk.go
```diff
@@ -0,0 +1,217 @@
...
+// VddkHandler provides endpoints for VDDK image management.
+type VddkHandler struct {
+	Handler
+}
+
+// AddRoutes registers the VDDK-specific HTTP routes on the given Gin engine.
+func (h *VddkHandler) AddRoutes(e *gin.Engine) {
+	e.POST(VddkRoot+"/build-image", h.BuildImage)
+	e.GET(VddkRoot+"/image-url", h.ImageUrl)
+	e.GET(VddkRoot+"/download-tar", h.DownloadVddkTar)
```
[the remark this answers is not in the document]
AvielSegev: Thanks!
I’ve done this for the `/build-image` and `/image-url` endpoints. Is it okay to leave `/download-tar` authentication free? It only exposes the uploaded VDDK tar file, which the build pod then pulls. If we require authentication, we’ll need to supply a token to the build (for example, by mounting a ServiceAccount token or creating a Kubernetes Secret tied to an existing or dedicated ServiceAccount). it's more simple to keep it that way.

operator/config/rbac/vddk-builder-role.yaml
AvielSegev: The YAML files are now in Ansible format and use the app_namespace variable to retrieve the namespace from the default file.

pkg/controller/provider/web/vsphere/vddk.go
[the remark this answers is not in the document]
AvielSegev: OK, Now all the endpoints are using Auth:)

pkg/controller/provider/web/vsphere/vddk.go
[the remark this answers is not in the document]
AvielSegev: I added a mechanism to prevent race conditions. Once a user triggers the upload endpoint, the server waits until the build completes, is canceled, or fails (with 1 min timeout) before allowing another `/build-image` request. This ensures that multiple users can’t overwrite each other’s vddk.tar.gz file and the image will be built from the correct one.

AvielSegev:
Hi @mnecas and @yaacov,
thank you for your comments! I’ve made the requested changes—please review them and let me know if you spot any other issues. I’ve also marked the above conversation as resolved and mentioned the things I added to resolve  😊

operator/config/rbac/role.yaml
AvielSegev, line 1: Is this the correct file to extend the forklift-operator’s permissions?
Is it needed for the creation of the BuildConfig and the ImageStream during the installation process.

I found it difficult to determine exactly where the forklift-operator’s permissions come from.

AvielSegev: Never mind — I found the correct one: `operator/config/rbac/leader_election_role.yaml`

pkg/controller/provider/web/vsphere/vddk.go
```diff
@@ -0,0 +1,306 @@
...
+func (h *VddkHandler) BuildImage(ctx *gin.Context) {
+	status, err := h.Prepare(ctx)
+	if status != http.StatusOK {
+		ctx.Status(status)
+		base.SetForkliftError(ctx, err)
+		return
+	}
+
+	buildLock.Lock()
+	if isBusy {
```
mnecas, line 62: Why do we need the `isBusy`?
If it's to just show that the build process is running, couldn't we replace it with `TryLock`?

mnecas: Bith having problems orienting myself with the lock and isBusy variable.

pkg/controller/provider/web/vsphere/vddk.go
```diff
@@ -0,0 +1,306 @@
...
+func (h *VddkHandler) ImageUrl(ctx *gin.Context) {
+	status, err := h.Prepare(ctx)
+	if status != http.StatusOK {
+		ctx.Status(status)
+		base.SetForkliftError(ctx, err)
+		return
+	}
+
+	// return error in case build is running
+	if isBusy {
```
mnecas, line 128: TryLock?

operator/roles/forkliftcontroller/templates/controller/buildconfig-vddk.yml.j2
```diff
@@ -0,0 +1,30 @@
...
+    secrets:
+      - secret:
+          name: "{{ controller_service_name }}-token"
+        destinationDir: secrets
+  strategy:
+    dockerStrategy: {}
+  output:
+    to:
+      kind: ImageStreamTag
+      name: "{{ vddk_image_stream_name }}:latest"
```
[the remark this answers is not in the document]
AvielSegev: Each time, the latest image will be overridden, but the URL returned from` image-url ` endpoint provides vddk@<sha256>, which should be sufficient.

pkg/controller/provider/web/vsphere/vddk.go
```diff
@@ -0,0 +1,306 @@
...
+func resetBusy() {
+	buildLock.Lock()
+	isBusy = false
+	buildLock.Unlock()
+}
+
+// waitForBuildAndResetBusy will poll your Build every second, stop on a terminal phase
+// or after 1 minute, then invoke resetBusy().
+func waitForBuildAndResetBusy(buildClient buildclientset.Interface, namespace, buildName string) {
+	ctx, cancel := context.WithTimeout(context.TODO(), time.Minute)
```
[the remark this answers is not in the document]
AvielSegev: Okay, how much time do you think is sufficient?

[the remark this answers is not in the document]
mnecas: hmm 5min?
but lets make it also configurable

pkg/controller/provider/web/vsphere/vddk.go
AvielSegev: Removed this flag

pkg/controller/provider/web/vsphere/vddk.go
AvielSegev: It's an idea I got from another project, but it seems to be redundant. I converted it to use TryLock().

pkg/controller/provider/web/vsphere/vddk.go
```diff
@@ -0,0 +1,306 @@
...
+	}
+
+	buildLock.Lock()
+	if isBusy {
+		buildLock.Unlock()
+		JSONError(ctx, http.StatusServiceUnavailable, "Server is busy processing another build. Please try again later.")
+		return
+	}
+	isBusy = true
+	buildLock.Unlock()
```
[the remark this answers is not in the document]
AvielSegev: Because I want the go waitForBuildAndUnlock() goroutine to unlock the mutex when the build is finished or the timeout is reached.

pkg/controller/provider/web/vsphere/vddk.go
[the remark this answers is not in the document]
AvielSegev: What do you mean? Like an environment variable that holds the timeout?

pkg/controller/provider/web/vsphere/vddk.go
```diff
@@ -0,0 +1,296 @@
...
+
+	if _, err := io.Copy(dst, src); err != nil {
+		JSONError(ctx, http.StatusInternalServerError, fmt.Sprintf("error copy to the local file: %v. ", err))
+		buildLock.Unlock()
+		return
+	}
+
+	if err := BuildAndPushImage(); err != nil {
+		JSONError(ctx, http.StatusInternalServerError, err.Error())
+		buildLock.Unlock()
```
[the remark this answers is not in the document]
AvielSegev: thanks! 
Used without `buildLock.TryLock() `- we discussed that TryLock() actually locks the mutex.

pkg/controller/provider/web/vsphere/vddk.go
```diff
@@ -0,0 +1,299 @@
...
+	"path/filepath"
+	"strings"
+	"sync"
+	"time"
+)
+
+const (
+	defaultNamespaceFile = "/var/run/secrets/kubernetes.io/serviceaccount/namespace"
+	uploadDirPerm        = 0600
+	VddkRoot             = ProvidersRoot + "/" + "vddk" // Route
```
mansam, line 28: This is trampling on the path to the vsphere provider collection, which would be `/providers/vsphere/:id`.

pkg/controller/provider/web/vsphere/vddk.go
```diff
@@ -0,0 +1,299 @@
...
+// AddRoutes registers the VDDK-specific HTTP routes on the given Gin engine.
+func (h *VddkHandler) AddRoutes(e *gin.Engine) {
+	e.POST(VddkRoot+"/build-image", h.BuildImage)
+	e.GET(VddkRoot+"/image-url", h.ImageUrl)
+	e.GET(VddkRoot+"/download-tar", h.DownloadVddkTar)
+}
+
+// BuildImage receives a VDDK tar file, writes it to disk,
+// and triggers an OpenShift BuildConfig to build and push the image.
+func (h *VddkHandler) BuildImage(ctx *gin.Context) {
```
mansam, line 54: Using a mutex in the handler to attempt to prevent more than one image build at a time feels risky. Can we look at the status of the build to decide what to do?

Rather than trying to make upload exclusive, I wonder if we could let them happen simultaneously (dump them into a temp dir with random names) and have a worker thread that kicks off builds one at a time. I'm assuming it's possible to parameterize the build request with the name of a specific file to download based on what I saw in the docs about passing args to the build, but I'm not sure if I understood properly.  Then we wouldn't need to worry about multiple uploads tramping on each other, and the last uploaded file would wind up in the latest image build.

pkg/controller/provider/web/vsphere/vddk.go
```diff
@@ -0,0 +1,299 @@
...
+	}
+
+	go waitForBuildAndUnlock(buildClient, namespace, buildObj.Name, waitForBuildTimeout)
+
+	return nil
+}
+
+// currentNamespace will return the namespace this pod is running in,
+// or an error if it can’t be determined.
+func currentNamespace() (string, error) {
```
mansam, line 230: This would be a good thing to collect and surface in the Settings package, where multiple things could access it.

pkg/controller/provider/web/vsphere/vddk.go
```diff
@@ -0,0 +1,299 @@
...
+		JSONError(ctx, http.StatusInternalServerError, fmt.Sprintf("error copy to the local file: %v. ", err))
+		return
+	}
+
+	if err = BuildAndPushImage(); err != nil {
+		JSONError(ctx, http.StatusInternalServerError, err.Error())
+		return
+	}
+
+	JSONSuccess(ctx, "VDDK build started; check your registry in OpenShift", nil)
```
mansam, line 112: I wonder if this might be more appropriate as a 202: Accepted.

pkg/controller/provider/web/vsphere/vddk.go
```diff
@@ -0,0 +1,299 @@
...
+	}
+	if err != nil {
+		return "", false, fmt.Errorf("error: %w. could not get ImageStreamTag: %s", err, registryImageTag)
+	}
+	return ist.Image.DockerImageReference, true, nil
+}
+
+// JSONError sends a standardized error response.
+func JSONError(ctx *gin.Context, code int, msg string) {
+	ctx.JSON(code, gin.H{"status": "error", "message": msg})
```
mansam, line 259: This pattern isn't used elsewhere and it's not saving any lines of code, so I would prefer to just use the call to `ctx.JSON` directly.

pkg/controller/provider/web/vsphere/vddk.go
```diff
@@ -0,0 +1,299 @@
...
+}
+
+// JSONSuccess sends a standardized success response with optional data.
+func JSONSuccess(ctx *gin.Context, msg string, data gin.H) {
```
mansam, line 263: I would prefer to just use the call to ctx.JSON directly. Especially since it's called only twice, once for each branch of the conditional here.

mansam (changes requested):
I have a few concerns about this, mostly pertaining to the use of a mutex and goroutine to make uploads exclusive. I think it might be less fragile to set up a worker to observe uploads and trigger builds. I've elaborated on this in my inline comments.

This appears to assume an OpenShift cluster, so it probably needs to be gated behind the `OpenShift` flag (in the settings package, this is derived from the environment). @mnecas what do you think?

I also want to mention a concern that this presumes that Forklift is installed on the same cluster that the VMs will be migrated to. In the majority of vSphere to Kubevirt migrations that will probably be true, but this does break the assumption that Forklift may be managing the migration from a cluster other than the destination. I understand that this might already be factored in to the design of this feature.

pkg/controller/provider/web/vsphere/vddk.go
```diff
@@ -0,0 +1,299 @@
...
+	VddkRoot             = ProvidersRoot + "/" + "vddk" // Route
+)
+
+var (
+	buildConfigName     = "vddk"
+	registryImageTag    = "vddk:latest"
+	vddkTarFileName     = "vddk.tar.gz"
+	uploadDir           = "/tmp/uploads"
+	buildLock           sync.Mutex
+	waitForBuildTimeout = time.Minute * 2
```
mansam, line 37: Most of these settings could probably be exposed in an appropriate section in the Settings package.

pkg/controller/provider/web/vsphere/vddk.go
AvielSegev: Can you please elaborate on why it feels risky?

Actually, we discussed the scenario where more than one request arrives at the same time, but we treated it as an edge case rather than something we intend to support. I agree that supporting multiple requests would be preferable, but is it truly necessary? It introduces more complexity:

1. Saving the files in a temporary directory with random names.

2. Modifying the Dockerfile’s download link to use the unique filename, ensuring the correct VDDK tar is downloaded.

3. OpenShift doesn’t support—at least to my knowledge—dynamically outputting to different ImageStreamTags. This leaves us with two options:

Retain the mutex to prevent race conditions and ensure each user receives the correct image URL (which is exactly what we’re trying to avoid), or Implement a mechanism to tag the latest ImageStreamTag with the random name using oc tag, which feels risky without the mutex.

There may also be other challenges with that approach.

pkg/controller/provider/web/vsphere/vddk.go
AvielSegev: Could you suggest an alternative route? 
Does `/providers/vsphere-vddk/... ` sound good?

pkg/controller/provider/web/vsphere/vddk.go
AvielSegev: OK, Done:)

pkg/controller/provider/web/vsphere/vddk.go
AvielSegev: OK, Done:)

pkg/controller/provider/web/vsphere/vddk.go
AvielSegev: I agree, Done:)

pkg/controller/provider/web/vsphere/vddk.go
AvielSegev: I discovered that an environment variable already holds this value, so the function is redundant. I’m now using the value from the settings package. PTAL.

pkg/controller/provider/web/vsphere/vddk.go
AvielSegev: I discovered that an environment variable already holds this value, so the function is redundant. I’m now using the value from the settings package. PTAL.

pkg/controller/provider/web/vsphere/vddk.go
mansam: It can't be under `/providers` either, since `providers` is a collection and the segment containing `vsphere` is also variable. This section of the API contains inventory details for specific Provider resources, and since this feature affects Forklift itself rather than a specific instance of a Provider it should be outside of that hierarchy.

pkg/controller/provider/web/vsphere/vddk.go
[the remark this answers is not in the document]
AvielSegev: Now that I’ve removed the use of mutex, this isn’t needed anymore

pkg/controller/provider/web/vsphere/vddk.go
```diff
@@ -0,0 +1,299 @@
...
+	uploadDirPerm        = 0600
+	VddkRoot             = ProvidersRoot + "/" + "vddk" // Route
+)
+
+var (
+	buildConfigName     = "vddk"
+	registryImageTag    = "vddk:latest"
+	vddkTarFileName     = "vddk.tar.gz"
+	uploadDir           = "/tmp/uploads"
+	buildLock           sync.Mutex
```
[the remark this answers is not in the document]
AvielSegev: Done:)

pkg/controller/provider/web/vsphere/vddk.go
AvielSegev: Changed to `/vddk` under the root. I will update the PR description as well...

pkg/controller/provider/web/vsphere/vddk.go
AvielSegev: We discussed this during the meeting and agreed that it’s fine now.

pkg/controller/provider/web/vsphere/vddk.go
AvielSegev: These variables are no longer needed:

```
buildLock           sync.Mutex
waitForBuildTimeout = time.Minute * 2
```
And these, as we discussed, don’t need customization:
```
vddkTarFileName     = "vddk.tar.gz"
uploadDir           = "/tmp/uploads"
```

pkg/controller/provider/web/vsphere/vddk.go
AvielSegev: I’ve made the requested changes: removed the mutex from the handler, now each upload is written to its own  random filename, then passing the specific filename as a build-arg so we can avoid race conditions without locking.

AvielSegev:
> I have a few concerns about this, mostly pertaining to the use of a mutex and goroutine to make uploads exclusive. I think it might be less fragile to set up a worker to observe uploads and trigger builds. I've elaborated on this in my inline comments.
> 
> This appears to assume an OpenShift cluster, so it probably needs to be gated behind the `OpenShift` flag (in the settings package, this is derived from the environment). @mnecas what do you think?
> 
> I also want to mention a concern that this presumes that Forklift is installed on the same cluster that the VMs will be migrated to. In the majority of vSphere to Kubevirt migrations that will probably be true, but this does break the assumption that Forklift may be managing the migration from a cluster other than the destination. I understand that this might already be factored in to the design of this feature.

@mansam Thanks for sharing!

I’ve implemented your approach and removed the mutex. The current flow is:

1. User uploads the file.
2. The file is saved with a random name.
3. The BuildConfig is triggered (OpenShift queues builds by default, so no separate worker is needed) and also passing build-arg for downloading the correct file.
4. The user receives a 202 response with the build name. 

Then, The user can then watch progress via the `/image-url` endpoint—passing the build name as a parameter—and eventually will receive the image URL.

I also added a simple check that returns “busy” if there is any build not in a terminal state (e.g., running, pending, initializing). I’m not sure this check is still necessary; I’d appreciate your thoughts.

Additionally, I added the watch option for the UI. PTAL.

mansam (approved):
LGTM, thanks for revising. Left one note about optionally checking whether this is an OpenShift cluster, but it's good as it is. :+1:

pkg/controller/provider/web/vsphere/doc.go
```diff
@@ -65,5 +65,10 @@ func Handlers(container *container.Container) []libweb.RequestHandler {
 				base.Handler{Container: container},
 			},
 		},
+		&VddkHandler{
```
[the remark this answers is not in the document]
mnecas: yup sounds like good idea

mnecas (commented):
The build, kustomized manigests and golint is failing please take a look

pkg/controller/provider/web/vsphere/doc.go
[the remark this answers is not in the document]
AvielSegev: I’ve also added the OPENSHIFT flag to the inventory, so now it adds the Handler whenever it recognizes an OpenShift cluster.

AvielSegev:
> The build, kustomized manigests and golint is failing please take a look

I think it's OK now. [code/snyk (Hybrid Platforms - Migration Toolkit for Virtualization)](https://app.snyk.io/org/hybrid-platforms-migration-toolkit-for-virtualization/project/2d256196-f48d-42e6-91f9-02037f267aa3/pr-check/cb66feff-5ecf-4f98-923c-71510d1a9df6) is failing but I don't have permissions to see.

mnecas (approved):
lgtm

moadqassem:
Just a little note here,  if I am not running forklift in an openshift cluster and naturally the  BuildConfig crd doesn't exist, how would the ansible operator pass? Also  there are no condition available to set it to false other than simply comment the `Create VDDK BuildConfig` and the `Create VDDK ImageStream` tasks in the main.yml file. Am I overlooking something here?

AvielSegev:
> Just a little note here, if I am not running forklift in an openshift cluster and naturally the BuildConfig crd doesn't exist, how would the ansible operator pass? Also there are no condition available to set it to false other than simply comment the `Create VDDK BuildConfig` and the `Create VDDK ImageStream` tasks in the main.yml file. Am I overlooking something here?

Hi @moadqassem, thank you for bringing this up.
Let me know if I’m wrong, but I think the Ansible operator will still pass while reporting the failed task (Let me know if I'm wrong).
Anyway, it would be great to perform those tasks only if it’s running under an OpenShift cluster, similar to what’s done here:


```
	if settings.Settings.OpenShift {
		handlers = append(
			handlers,
			&VddkHandler{
				Handler: base.Handler{
					Container: container,
				},
			},
		)
	} 
```

PR #1662, opened 2025-05-05
MTV-2075 | Dont override vm original name

yaacov:
Ref:
https://issues.redhat.com/browse/MTV-2075 - use correct vm name in migration status
https://issues.redhat.com/browse/MTV-2076 - add display name

Issue:
When changing the vm name to comply with DNS1123 we unintentionally remove the original name from the migration.

Fix:
Don't change the original name in the migration CR

Screenshots:
Before and After:
migration-vm-name

Before:
migration-vm-name-before

After:
migration-vm-name-after

VM annotations:
vm-name-after

### DIFF
--- a/pkg/controller/plan/kubevirt.go
+++ b/pkg/controller/plan/kubevirt.go
@@ -73,6 +73,8 @@ const (
 	AnnImporterPodName = "cdi.kubevirt.io/storage.import.importPodName"
 	//  Original VM name on source (value=vmOriginalName)
 	AnnOriginalName = "original-name"
+	// Openshift display name annotation (value=vmName)
+	AnnDisplayName = "openshift.io/display-name"
 	//  Original VM name on source (value=vmOriginalID)
 	AnnOriginalID = "original-ID"
 	// DV deletion on completion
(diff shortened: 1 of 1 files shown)


pkg/controller/plan/kubevirt.go
```diff
@@ -73,6 +73,8 @@ const (
 	AnnImporterPodName = "cdi.kubevirt.io/storage.import.importPodName"
 	//  Original VM name on source (value=vmOriginalName)
 	AnnOriginalName = "original-name"
+	// Openshift display name annotation (value=vmName)
+	AnnDisplayName = "openshift.io/display-name"
```
mnecas, line 77: Out of curiosity: Is this a generic annotation for any resource?

yaacov: This annotation is used in openshift (didn't find it in vanilla k8s),
It is currently an RFE for OCP-V [ CNV-59643 ]

For openshift Projects:
https://access.redhat.com/solutions/1982113
For openshift Templates:
https://docs.redhat.com/en/documentation/openshift_container_platform/3.11/html/developer_guide/dev-guide-templates#using-the-instantapp-templates

It is suggested for kubevirt VMs (on openshift clusters):
https://issues.redhat.com/browse/CNV-59643

yaacov: > Thinking if we should do something like that also for MTV resources.

:shrug: 

I added it for: https://issues.redhat.com/browse/CNV-59643

yaacov: @mnecas should we remove the "original-name" annotation ?

mnecas: hmm yeah, we are not displaying the `original-name` anywhere, right?

yaacov: > we are not displaying the original-name anywhere, right?

no, we are not using it, before this PR it showed the wrong value anyways ...

yaacov: removed the unused  "original-name" annotation

PR #1667, opened 2025-05-06
MTV-2314 | Warn about potentially unsupported OVAs.

mrnold:
Issue:
Errors in OVA import due to differences in expected OVF formatting, in this case "MegaBytes" as a unit instead of "bytes * 2^20".

Fix:
Address this one unit instance by allowing "MegaBytes" as a resource unit, and add a warning for OVFs that look like they might have been exported from anything other than VMware. 

Ref: https://issues.redhat.com/browse/MTV-2314

### DIFF
--- a/cmd/ova-provider-server/ova-provider-server.go
+++ b/cmd/ova-provider-server/ova-provider-server.go
@@ -124,6 +124,7 @@ type Envelope struct {
 type VM struct {
 	Name                  string
 	OvaPath               string
+	OvaSource             string
 	OsType                string
 	RevisionValidated     int64
 	PolicyVersion         int
--- a/cmd/ova-provider-server/ova-provider-server.go
+++ b/cmd/ova-provider-server/ova-provider-server.go
@@ -114,6 +114,7 @@ type VirtualSystem struct {
 
 type Envelope struct {
 	XMLName        xml.Name        `xml:"Envelope"`
+	Attributes     []xml.Attr      `xml:",any,attr"`
 	VirtualSystem  []VirtualSystem `xml:"VirtualSystem"`
 	DiskSection    DiskSection     `xml:"DiskSection"`
 	NetworkSection NetworkSection  `xml:"NetworkSection"`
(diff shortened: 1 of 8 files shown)


cmd/ova-provider-server/ova-provider-server.go
```diff
@@ -429,6 +431,45 @@ func readOVF(ovfFile string) (*Envelope, error) {
 	return &envelope, nil
 }
 
+const (
+	Unknown    = "Unknown"
+	VMware     = "VMware"
+	VirtualBox = "VirtualBox"
+	RHV        = "RHV"
```
mnecas, line 438: Please switch to upstream oVirt to keep it consistent with other parts of Forklift.

cmd/ova-provider-server/ova-provider-server.go
```diff
@@ -429,6 +431,45 @@ func readOVF(ovfFile string) (*Envelope, error) {
...
+)
+
+// Check the OVF XML for any markers that might cause import problems later on.
+// Not guaranteed to correctly guess the OVA source, but should be good enough
+// to filter out some obvious problem cases.
+func guessOvaSource(envelope Envelope) string {
+	namespaceMap := map[string]string{
+		"http://schemas.citrix.com/ovf/envelope/1": Xen,
+		"http://www.citrix.com/xenclient/ovf/1":    Xen,
+		"http://www.virtualbox.org/ovf/machine":    VirtualBox,
```
mnecas, line 449: You are missing the ovirt, here is ovf example 
[vm.ovf.txt](https://github.com/user-attachments/files/20069993/vm.ovf.txt)

cmd/ova-provider-server/ova-provider-server.go
mrnold: Done.

cmd/ova-provider-server/ova-provider-server.go
mrnold: Thanks, added.

mnecas (approved):
lgtm

Issue #1354, opened 2025-02-15
Network interface not UP after migration


cloudcafetech:
I migrate Ubuntu VM from VMware to Kubevirt using forklift, migration went well but in kubevirt migrated VM network interface did not UP.

Source (Vmware) VM:

![Image](https://github.com/user-attachments/assets/dafd6e2b-02b0-4b93-8ca4-3adb6856e446)

Destination (Kubevirt) VM:

![Image](https://github.com/user-attachments/assets/694edad0-684f-4eb6-bc58-e49276900d9f)

Plan:

```
apiVersion: forklift.konveyor.io/v1beta1
kind: Plan
metadata:
  name: vm-mig-plan
  namespace: konveyor-forklift
spec:
  preserveStaticIPs: true
  archived: false
  description: ''
  map:
    network:
      name: vmware-network-map
      namespace: konveyor-forklift
    storage:
      name: vmware-storage-map
      namespace: konveyor-forklift
  provider:
    destination:
      name: host
      namespace: konveyor-forklift
    source:
      name: vsphere-provider
      namespace: konveyor-forklift
  targetNamespace: virtualmachines
  vms:
    - hooks: []
      name: vw-jumphost
  warm: false
```

mnecas:
Sorry for the late response, https://github.com/kubev2v/forklift/blob/main/pkg/virt-v2v/customize/scripts/rhel/run/network_config_util.sh

mnecas:
Ah after reading the code, we do it also for the DHCP but that script is missing the cloud.cfg so all we need to do is to update the script to look at the cloud init configs and create the udev rules from it.

cloudcafetech:
@mnecas Thanks.

Can you help me one more thing though its not V2V migration but its live migration inside cluster.

I am using Kubevirt along with Multus & whereabouts.

My network attachment definition as follows 

```
apiVersion: "k8s.cni.cncf.io/v1"
kind: NetworkAttachmentDefinition
metadata:
  name: bridge-whereabouts
  namespace: virtualmachines
spec:
  config: '{
      "cniVersion": "0.3.1",
      "name": "bridge-whereabouts",
      "type": "bridge",
      "bridge": "br0",
      "promiscMode": true,       
      "ipam": {
        "type": "whereabouts",
        "range": "192.168.0.0/24",
        "range_start": "192.168.0.20",
        "range_end": "192.168.0.50",
        "gateway": "192.168.0.1",
        "routes": [
[...]

PR #1706, opened 2025-05-07
MTV-2312 | Maintain network during VM migration with DHCP

jonner:
When migrating a VM with a network interface that uses DHCP, make a best effort to maintain network connectivity after migration. We use the same scripts that we use for static IP interface mapping, but attempt to use dhcp lease files on the vm to determine the interface name associated with a particular IP and create a udev rule for that device. Note that this only works for vms that are using NetworkManager or dhclient to handle DHCP. If using systemd-networkd, the dhcp lease files are not available in the filesystem while the vm is not running so we don't handle that scenario.


pkg/virt-v2v/customize/scripts/rhel/run/network_config_util.sh
```diff
@@ -13,7 +15,7 @@ NETPLAN_DIR="${NETPLAN_DIR:-/}"
 # Dump debug strings into a new file descriptor and redirect it to stdout.
 exec 3>&1
 log() {
-    echo $@ >&3
+    echo "$@" >&3
```
yaacov, line 18: :+1:

pkg/virt-v2v/customize/scripts/rhel/run/network_config_util.sh
```diff
@@ -84,7 +86,7 @@ udev_from_ifcfg() {
     fi
 
     # Read the mapping file line by line
-    cat "$V2V_MAP_FILE" | while read line;
+    cat "$V2V_MAP_FILE" | while read -r line;
```
yaacov, line 89: :+1:

pkg/virt-v2v/customize/scripts/rhel/run/network_config_util.sh
```diff
@@ -151,6 +153,182 @@ udev_from_nm() {
     done
 }
 
+# Attempt to parse the `timestamps` file and find the most recent timestamp
```
yaacov, line 156:
```suggestion
# Attempt to parse the `timestamps` file and find the matching timestamp
```

this method finds the most recent or just the first one ?

pkg/virt-v2v/customize/scripts/rhel/run/network_config_util.sh
```diff
@@ -151,6 +153,182 @@ udev_from_nm() {
...
+        # parse the filenames of the matching lease files and grab the device name of
+        # the most recent one
+        DEVICE=$(for FILENAME in $LEASE_FILES;
+        do
+            log "Checking $FILENAME"
+            # Filenames are of the form 'prefix-$(UUID)-$(INTERFACE_NAME).lease'
+            FILENAME_PARTS=$(echo "$FILENAME" | sed -n 's|^.*-\([0-9a-f]\{8\}-[0-9a-f]\{4\}-[0-9a-f]\{4\}-[0-9a-f]\{4\}-[0-9a-f]\{12\}\)-\(.*\)\.lease$|\1 \2|p')
+            if [ -n "$FILENAME_PARTS" ]; then
+                UUID=$(echo "$FILENAME_PARTS" | cut -d' ' -f1)
+                INTERFACE=$(echo "$FILENAME_PARTS" | cut -d' ' -f2-)
```
yaacov, line 222:
```suggestion
                INTERFACE=$(echo "$FILENAME_PARTS" | cut -d' ' -f2)
```
no need for -

yaacov:
LGTM

@jonner @mnecas hi, a VM using DHCP will start using static IP , what am i missing ?

mnecas:
FYI @cloudcafetech
Fixes: https://github.com/kubev2v/forklift/issues/1354

mnecas:
@jonner which operating systems could this include/which operating systems do use these lease configurations?

mnecas:
@cloudcafetech please see explanation from @jonner around the possible OS versions.
I need to sync with QE before merging. I see that the RFE wasn't targeted to the 2.9 so need to check with them about capacity.

mnecas (approved):
lgtm, but need to wait for a bit before merging

PR #1745, opened 2025-05-09
Collect information about the storage protocols used by the hosts

nirarg:
This info is concluded from the HostBusAdapter type Create Const string for each protocol
This is done as part of exposing the information in Migration-planner
Jira ticket: ECOPROJECT-2858


nirarg:
This PR a way to collect the storage protocol
/cc @machacekondra

pkg/controller/provider/model/vsphere/model.go
```diff
@@ -141,6 +153,7 @@ type Host struct {
 	Datastores         []Ref          `sql:""`
 	HostScsiDisks      []HostScsiDisk `sql:""`
 	AdvancedOptions    Ref            `sql:""`
+	StorageProtocols   []ProtocolType `sql:""`
```
machacekondra, line 156: I am unsure if this could be just array of protocol types, because we need to be able to map it later to specific storage ( I think via naa)

nirarg: Added more information, please take a look

pkg/controller/provider/container/vsphere/model.go
```diff
@@ -376,14 +376,59 @@ func (v *HostAdapter) Apply(u types.ObjectUpdate) {
...
 					for _, iScsiLun := range array.ScsiLun {
 						hostScsiDisk := model.HostScsiDisk{}
-						scsiLun := iScsiLun.GetScsiLun()
-						hostScsiDisk.CanonicalName = scsiLun.CanonicalName
-						hostScsiDisk.Vendor = strings.TrimSpace(scsiLun.Vendor)
-						v.model.HostScsiDisks = append(v.model.HostScsiDisks, hostScsiDisk)
+						if disk, ok := iScsiLun.(*types.HostScsiDisk); ok {
+							hostScsiDisk.CanonicalName = disk.CanonicalName
+							hostScsiDisk.Vendor = strings.TrimSpace(disk.Vendor)
+							hostScsiDisk.Model = strings.TrimSpace(disk.Vendor)
```
machacekondra, line 382: Unsure we really need to collect a model here.

pkg/controller/provider/container/vsphere/model.go
```diff
@@ -376,14 +376,59 @@ func (v *HostAdapter) Apply(u types.ObjectUpdate) {
...
+						v.model.HbaDiskInfo = append(v.model.HbaDiskInfo, hbaDiskInfo)
+					}
+				}
+			case fScsiTopology:
+				if array, cast := p.Val.(types.ArrayOfHostScsiTopologyInterface); cast {
+					for _, scsiTopologyInterface := range array.HostScsiTopologyInterface {
+						hostScsiTopology := model.HostScsiTopology{}
+						hostScsiTopology.HbaKey = scsiTopologyInterface.Key
+						for _, scsiTopologyTarget := range scsiTopologyInterface.Target {
+							hostScsiTopology.ScsiDiskKeys = append(hostScsiTopology.ScsiDiskKeys, scsiTopologyTarget.Key)
```
machacekondra, line 427: IMHO you need here `lun.key` not `key`. But unsure how you plan to do the mappings.

pkg/controller/provider/container/vsphere/model.go
nirarg: This was leftover, I removed it, thank you for the catch

pkg/controller/provider/container/vsphere/model.go
machacekondra: Yeah, I think you are comparing two non-comparable items. I think you need to fetch `lun.key`

nirarg: You right, I. used the wrong value here
But I think it sgould be lun. scsiLun (not key)
I used the first lun only (lun[0]) as its seems there is only one in this list and anyway it should be enough

nirarg: The work on the migration-planner and testing it still in progress

pkg/controller/provider/container/vsphere/collector.go
```diff
@@ -728,6 +730,7 @@ func (r *Collector) propertySpec() []types.PropertySpec {
 				fVNIC,
 				fScsiLun,
 				fAdvancedOption,
+				fHostBusAdapter,
```
machacekondra, line 733: fScsiTopology

pkg/controller/provider/container/vsphere/model.go
nirarg: The change was validated with migartion-planner cahnges:
https://github.com/kubev2v/migration-planner/pull/292
The data was collected as expected
@machacekondra @mnecas 
Can you please have the final review and merge?

machacekondra (approved):

mnecas (approved):

PR #1763, opened 2025-05-12
MTV-2025 | MTV-1673 | NICs mapping is unordered

Hazanel:
MTV-2025 | MTV-1673 | NICs mapping is unordered
MTV-2489 | Update warning message for VM static ip setting Issue:
1. When migrating VMs with multiple NICs the mapping order must match source VM order
2. Display warning when one or more interfaces lack static IP configuration

Fix:
1. Iterate over vm.NICs directly, preserving the original order of network interfaces from the source VM during migration. Previously, networks were generated based on the order of network mappings (netMapIn), which could result in a different NIC order in the target VM.
2. Get warning message when no interface in VM with static ip

Ref: https://issues.redhat.com/browse/MTV-2025
https://issues.redhat.com/browse/MTV-1673
https://issues.redhat.com/browse/MTV-2489


mnecas (approved):
lgtm

PR #1764, opened 2025-05-12
fix: close template string

playworker:


mnecas (approved):

mnecas:
Nice catch, thank you!

PR #1821, opened 2025-05-13
MTV-2364 | Inconsistent hostname

Hazanel:
Issue:
Inconsistency in the hostname between the source and destination VMs Due to lack of host name in the incventory.

Fix:
 Fetching the hostname for each VM and apply that to the destination VM.

 Ref: https://issues.redhat.com/browse/MTV-2364

### DIFF
--- a/pkg/virt-v2v/conversion/conversion.go
+++ b/pkg/virt-v2v/conversion/conversion.go
@@ -149,7 +149,9 @@ func (c *Conversion) addVirtV2vArgs(cmd utils.CommandBuilder) (err error) {
 func (c *Conversion) addVirtV2vVsphereArgs(cmd utils.CommandBuilder) (err error) {
 	cmd.AddArg("-i", "libvirt").
 		AddArg("-ic", c.LibvirtUrl).
-		AddArg("-ip", c.SecretKey)
+		AddArg("-ip", c.SecretKey).
+		AddArg("--hostname", c.HostName)
+
 	err = c.addCommonArgs(cmd)
 	if err != nil {
 		return err
--- a/validation/policies/io/konveyor/forklift/vmware/hostname.rego
+++ b/validation/policies/io/konveyor/forklift/vmware/hostname.rego
@@ -1,0 +1,31 @@
+package io.konveyor.forklift.vmware
+
+import future.keywords.in
+
+is_empty_hostname {
+    input.hostName == ""
+}
+
+
+is_localhost_hostname {
+    input.hostName == "localhost.localdomain"
+}
+
+concerns[flag] {
+    is_empty_hostname
+    flag := {
+        "category": "Warning",
+        "label": "Empty Host Name",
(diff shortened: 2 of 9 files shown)


pkg/virt-v2v/conversion/conversion.go
```diff
@@ -150,6 +150,9 @@ func (c *Conversion) addVirtV2vVsphereArgs(cmd utils.CommandBuilder) (err error)
 	cmd.AddArg("-i", "libvirt").
 		AddArg("-ic", c.LibvirtUrl).
 		AddArg("-ip", c.SecretKey)
+	if c.HostName != "" {
+		cmd.AddArg("--hostname", c.HostName)
+	}
```
mnecas, line 155: NP: The `AddArg` is already checking if the variable is empty, otherwise it would break the key arg format.

validation/policies/io/konveyor/forklift/vmware/hostname.rego
```diff
@@ -0,0 +1,45 @@
...
+        "assessment": "The 'hostname' field is missing or empty. The hostname might be renamed during migration."
+    }
+}
+
+concerns[flag] {
+    is_invalid_hostname_format
+    flag := {
+        "category": "Warning",
+        "label": "Invalid Host Name",
+        "assessment": "The 'hostname' does not comply with the DNS subdomain format defined in RFC 1123. A valid hostname must consist of lowercase letters (a–z), numbers (0–9), periods (.), and hyphens (-). It must be 1–63 characters per label, start and end with an alphanumeric character, and must not include uppercase letters, spaces, or special characters. The hostname will be renamed automatically during migration to meet the RFC convention."
```
mnecas, line 34: Does this apply? Don't think this should be anywhere in the kubernetes resources

Hazanel: Eventually we're speaking about internal vm hostname  - so you're right.
Dropped that.

validation/policies/io/konveyor/forklift/vmware/hostname.rego
```diff
@@ -0,0 +1,31 @@
...
+    }
+}
+
+
+concerns[flag] {
+    is_localhost_hostname
+    flag := {
+        "category": "Warning",
+        "label": "Default Host Name",
+        "assessment": "The 'hostname' is set to 'localhost.localdomain', which is a default value and may cause conflicts during migration. Consider changing it to a unique hostname."
```
mnecas, line 29: One more thing, what kind of conflits could happen? From my understanding it could be renamed after migration

Hazanel: A conflict between hostname to /etc/hostname

mnecas: In that case i would mention again that the `The hostname might be renamed during migration.`

mnecas (approved):
lgtm

PR #1855, opened 2025-05-15
Bump go toolset version

yaacov:
Bump go toolset version

Issue:
Some tools and libraries require new version of the go toolset


solenoci:
The operator bundle test will fail every time for now, please ignore it

solenoci:
lgtm

PR #1857, opened 2025-05-15
Disable ova and vsphere CI as people accidentally overlook failing CI

mnecas:
Disable ova and vsphere CI as people accidentally overlook failing CI


yaacov (approved):
LGTM

PR #1859, opened 2025-05-15
MTV-2501 | Dangling plan

Hazanel:
Issue:
Deleting a provider with existing related plan causes error when archiving, leading to leftovers if plan is force deleted

Fix:
Allowing archiving dangling plan in the provider hook.

Ref: https://issues.redhat.com/browse/MTV-2501

### DIFF
--- a/pkg/controller/plan/controller.go
+++ b/pkg/controller/plan/controller.go
@@ -210,6 +210,11 @@ func (r Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (r
 	// Validations.
 	err = r.validate(plan)
 	if err != nil {
+		if r.isDanglingArchivedPlan(plan) {
+			r.Log.Info("Dangling Plan - Aborting reconcile of plan without source provider.")
+			r.archive(plan)
+			r.updatePlanStatus(plan)
+		}
 		return
 	}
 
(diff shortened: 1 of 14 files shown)


pkg/controller/plan/controller.go
```diff
@@ -189,7 +189,7 @@ func (r Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (r
 	}()
 
 	// Don't reconcile if the plan is archived.
-	if plan.Spec.Archived && plan.Status.HasCondition(Archived) {
+	if r.shouldAbortArchived(plan) {
```
mansam, line 192: Please don't obscure these simple conditions behind method calls. It makes it harder to read when you have to jump to a method to find out what the condition is.

pkg/controller/plan/controller.go
```diff
@@ -210,9 +210,12 @@ func (r Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (r
 	// Validations.
 	err = r.validate(plan)
 	if err != nil {
+		if r.isDanglingArchivedPlan(plan) {
```
mansam, line 213: It would probably be worthwhile to check what the specific error is before archiving, and the condition probably does not need to be hidden behind a method call.

pkg/controller/plan/controller.go
```diff
@@ -263,6 +254,33 @@ func (r Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (r
 	return
 }
 
+func (r *Reconciler) archiveAndUpdate(plan *api.Plan) {
```
mansam, line 257: I feel that hiding two method calls behind one makes the code significantly harder to follow.

mansam (changes requested):
I have a few maintainability nitpicks about this, but I also don't think this will work as written. Archiving requires a plancontext to be constructed, which will fail (for good reason) with no source provider. Very little in the archiving flow will work, as it assumes a source provider is present since it reuses the migration runner for a number of things. Archiving probably needs to be refactored substantially to enable it to work with a missing provider.

pkg/controller/plan/controller.go
```diff
@@ -237,19 +240,7 @@ func (r Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (r
...
-	plan.Status.ObservedGeneration = plan.Generation
-	// At this point, the plan contains data that is not persisted by design, like the Referenced data
-	// and the staged flags in the status, and more data that has been loaded in the validate function,
-	// like the name of the VMs in the spec section, therefore we don't want the plan to be overridden
-	// by data from the server (even the spec section is overridden) and so we pass a copy of the plan
-	err = r.Status().Update(context.TODO(), plan.DeepCopy())
-	if err != nil {
-		return
-	}
+	r.updatePlanStatus(plan)
```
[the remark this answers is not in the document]
mansam: You've changed the error handling behavior here, so an error while updating the Plan status no longer aborts the reconcile before `execute()`. I am not certain what the consequences will be if the reconcile continues without the status updating succeeding, but the status update failing probably indicates a big enough problem that reconciliation should stop.

PR #1860, opened 2025-05-15
Automatically handle Pipeline step transitions when moving to the next phase

mansam:
This PR adds a method to the migration runner called `NextPhase()` that will move the VM to the next phase and automatically complete the current pipeline step if the next phase belongs to a different one. It's no longer necessary for the last phase in a step to manually complete it which will allow phases to be reordered and moved between steps more easily.

(Extracted from https://github.com/kubev2v/forklift/pull/1648)


mnecas (approved):
Thank you!

PR #1864, opened 2025-05-16
MTV-2271 | Add release config file for konflux

solenoci:
Add a file that holds all the info for configuring a release


mnecas (approved):
lgtm

PR #1886, opened 2025-05-20
MTV-2524 | Round VMware disk capacity up to next megabyte.

mrnold:
Issue:
A VMware disk image was set to a number of bytes that did not align to 1MiB, and the imported VM would not start.

Fix:
I have not yet been able to set a disk to such a size in VMware, but a speculative fix is to change the conversion from bytes to MiB in the vSphere builder, which just divides by 0x100000 without rounding up first. With non-aligned size values, this division results in a smaller number, for example: 19000005632 / 0x100000 = 18999148544, vs. rounding up to 19000197120.

Reference:
https://issues.redhat.com/browse/MTV-2524

### DIFF
--- a/pkg/controller/plan/adapter/vsphere/builder.go
+++ b/pkg/controller/plan/adapter/vsphere/builder.go
@@ -1087,7 +1087,7 @@ func (r *Builder) Tasks(vmRef ref.Ref) (list []*plan.Task, err error) {
 		vm.RemoveSharedDisks()
 	}
 	for _, disk := range vm.Disks {
-		mB := disk.Capacity / 0x100000
+		mB := utils.RoundUp(disk.Capacity, 0x100000) / 0x100000
 		list = append(
 			list,
 			&plan.Task{
--- a/pkg/controller/plan/util/utils.go
+++ b/pkg/controller/plan/util/utils.go
@@ -21,7 +21,7 @@ const (
 	diskPrefix = "/dev/sd"
 )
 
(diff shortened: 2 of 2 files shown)


rwmjones:
I am of course completely unfamiliar with MTV code.

The old code  herewas certainly wrong - truncating the disk size by blindly dividing and rounding down is not correct.

However if virt-v2v was being presented with a target block device which was too small, we would have failed the copy.  As that didn't happen (we saw a target block device which was the same size as the source, even if the size was not a multiple of 4K) I can't help thinking that it must be some other code that needs to be changed.

mansam:
The method that is doing the division (`Builder.Tasks()`) is only used to report approximate transfer progress in the Plan CR. The actual resource request is made in `Builder.DataVolumes()` by passing the source disk's capacity in bytes (called `Capacity` here, but taken directly from the vsphere object's `CapacityInBytes` field) as a resource request in the DataVolume spec:

```go
dvSpec := cdi.DataVolumeSpec{
	Source: &dvSource,
	Storage: &cdi.StorageSpec{
		Resources: core.ResourceRequirements{
			Requests: core.ResourceList{
				core.ResourceStorage: *resource.NewQuantity(disk.Capacity, resource.BinarySI),
			},
		},
		StorageClassName: &storageClass,
	},
}
```

mrnold:
Thanks, I will change that too.

mnecas (approved):
lgtm

PR #1893, opened 2025-05-21
MTV-2524 | Round VMware disk capacity up to next megabyte, part 2.

mrnold:
Issue: need one more change for previously merged pull request #1886.

Fix: also align disk size in DataVolume creation request. I still haven't managed to reproduce the original bug, but this probably won't hurt anything.

Ref: https://github.com/kubev2v/forklift/pull/1886


rwmjones:
This looks better!  Rounds up the size to 1M which ought to be enough for anybody.

mnecas (approved):

PR #1896, opened 2025-05-22
MTV-2367 | Keep the disk order between Kubernetes clusters

mnecas:
Issue:
When migrating the VM between two Kubernetes clusters, the disks are not sorted in the original order.
This causes the devices to be named differently.

Fix:
Right now, the disks are created from a map which does not have order.
We need to switch to the original disk order. 

Ref: https://issues.redhat.com/browse/MTV-2367


solenoci:
lgtm

PR #1898, opened 2025-05-22
VIB ensure install

rgolangh:
https://issues.redhat.com/browse/ECOPROJECT-3066

Motivation
The VIB is a critical piece of the flow, and it will probably go under
changes as we go. We do have an ansible script to install it, but that
actions is manual. Instead we can try to always look for the vib on the
ESX in context, and check the version, and remediate if necessary. That
way we are always updated and aligned to the populator version.

Modification
With the former fix making the wrapper VIB not needing a restart of
hostd, the installation process becomes quicker and safer.
The populator will fetch the VIB version and compare to the desired
version. If not there it will upload the vib, from the container
filesystem, to the same datastore the VM is on, and then install it,
all through remote ESX commands through the API.

Result
The populator pod automatically checks and installs the VIB on each
attempt to run clone the disk. We pay few seconds in the worst case,
just once per ESX.

### DIFF
--- a/cmd/vsphere-xcopy-volume-populator/internal/populator/remote_esxcli.go
+++ b/cmd/vsphere-xcopy-volume-populator/internal/populator/remote_esxcli.go
@@ -77,6 +77,10 @@ func (p *RemoteEsxcliPopulator) Populate(sourceVMDKFile string, volumeHandle str
 	}
 	klog.Infof("Got ESXI host: %s", host)
 
+	err = ensureVib(p.VSphereClient, host, vmDisk.Datastore, VibVersion)
+	if err != nil {
+		return fmt.Errorf("failed to ensure VIB is installed: %w", err)
+	}
 	// for iSCSI add the host to the group using IQN. Is there something else for FC?
 	r, err := p.VSphereClient.RunEsxCommand(context.Background(), host, []string{"storage", "core", "adapter", "list"})
 	if err != nil {
--- a/cmd/vsphere-xcopy-volume-populator/internal/populator/vib.go
+++ b/cmd/vsphere-xcopy-volume-populator/internal/populator/vib.go
@@ -1,0 +1,92 @@
+package populator
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/kubev2v/forklift/cmd/vsphere-xcopy-volume-populator/internal/vmware"
+	"github.com/vmware/govmomi/object"
+	"k8s.io/klog/v2"
+)
+
+const (
+	vibName     = "vmkfstools-wrapper"
+	vibLocation = "/bin/vmkfstools-wrapper.vib"
+)
+
+// VibVersion is set by ldflags
+var VibVersion = "x.x.x"
+
+// ensure vib will fetch the vib version and in case needed will install it
+// on the target ESX
+func ensureVib(client vmware.Client, esx *object.HostSystem, datastore string, desiredVibVersion string) error {
+	klog.Infof("ensuring vib version on ESXi %s: %s", esx.Name(), VibVersion)
+
+	version, err := getViBVersion(client, esx)
+	if err != nil {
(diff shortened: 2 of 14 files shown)


rgolangh:
@borod108 @mnecas PTAL - I wrote exactly in the commit messages the nature of the change.

cmd/vsphere-xcopy-volume-populator/internal/populator/vib.go
```diff
@@ -0,0 +1,75 @@
...
+	"fmt"
+
+	"github.com/kubev2v/forklift/cmd/vsphere-xcopy-volume-populator/internal/vmware"
+	"github.com/vmware/govmomi/object"
+	"k8s.io/klog/v2"
+)
+
+const (
+	vibName     = "vmkfstools-wrapper"
+	vibVersion  = "0.1.0"
```
mnecas, line 14: Could we reuse the `CUSTOM_VIB_VERSION` so we won't run out of sync in future?

cmd/vsphere-xcopy-volume-populator/internal/populator/vib.go
```diff
@@ -0,0 +1,75 @@
...
+	return r[0].Value("Version"), err
+}
+
+func uploadVib(client vmware.Client, esx *object.HostSystem, datastore string) (string, error) {
+	ds, err := client.GetDatastore(context.Background(), datastore)
+	if err != nil {
+		return "", fmt.Errorf("failed to upload file: %w", err)
+	}
+
+	if err = ds.UploadFile(context.Background(), vibLocation, vibName+".vib", nil); err != nil {
```
mnecas, line 61: What if the upload fails due to certs/permissions?

mnecas: Note: We will need to document the needed permissions for this.

cmd/vsphere-xcopy-volume-populator/internal/populator/remote_esxcli.go
```diff
@@ -79,8 +79,12 @@ func (p *RemoteEsxcliPopulator) Populate(sourceVMDKFile string, volumeHandle str
 	}
 	klog.Infof("Got ESXI host: %s", host)
 
+	err = ensureVib(p.VSphereClient, host, vmDisk.Datastore, vibVersion)
+	if err != nil {
+		return fmt.Errorf("failed to ensure VIB is installed: %w", err)
```
mnecas, line 84: Do we want to fail fail when we fail this step?
What if the VIB is present but we can't upload to it because of the certs/permissions/networking?

cmd/vsphere-xcopy-volume-populator/internal/populator/vib.go
rgolangh: is the permission to upload to a datastore considered less common?

cmd/vsphere-xcopy-volume-populator/internal/populator/vib.go
rgolangh: I thought about a build flag that I'll push using LDFLAG

cmd/vsphere-xcopy-volume-populator/internal/populator/remote_esxcli.go
rgolangh: if the vib is already present we don't upload it again , unless the version changed. 
So this is once per ESX, per vib version

cmd/vsphere-xcopy-volume-populator/internal/populator/vib.go
rgolangh: done, 
see vmkfstools-wrapper/version.mk usage in the vsphere-xcopy-volume-populator/Makefile and vsphere-xcopy-volume-popualtor/vmkfstools-wrapper/Makefile

PR #1903, opened 2025-05-26
MTV-2612 |  Duplicate default persistent routes exist in VM after migration

Hazanel:
Issue:
When a VM configured with a static IP and gateway undergoes a cold migration (with "preserve static IP" enabled), Windows sometimes creates a new persistent route for the gateway without removing the old one. This results in duplicate persistent routes in the VM's routing table, where the newly created route becomes the active one.

Fix:
Remove persistent routes duplications during boot time.

Ref: https://issues.redhat.com/browse/MTV-2612

### DIFF
--- a/pkg/virt-v2v/customize/scripts/windows/9999-remove_duplicate_persistent_routes.ps1
+++ b/pkg/virt-v2v/customize/scripts/windows/9999-remove_duplicate_persistent_routes.ps1
@@ -1,0 +1,94 @@
+# Requires Administrator privileges to run
+
+Write-Host "Starting persistent route cleanup script..." -ForegroundColor Green
+
+# Function to convert CIDR prefix length to subnet mask
+function Convert-PrefixToMask($prefix) {
+    $bin = ("1" * $prefix).PadRight(32, "0")
+    $octets = @(
+        $bin.Substring(0, 8)
+        $bin.Substring(8, 8)
+        $bin.Substring(16, 8)
+        $bin.Substring(24, 8)
+    ) | ForEach-Object { [Convert]::ToInt32($_, 2) }
+
+    return ($octets -join ".")
+}
+
+# Get persistent routes
+try {
+    $routes = Get-NetRoute -PolicyStore PersistentStore -ErrorAction Stop
+} catch {
+    Write-Host "Error retrieving persistent routes: $_" -ForegroundColor Red
+    Exit 1
+}
+
+$routes = $routes | Where-Object { $_.AddressFamily -eq "IPv4" }
+
+$groupedRoutes = $routes | Group-Object {
+    "$($_.DestinationPrefix)-$($_.NextHop)-$($_.RouteMetric)"
+} | Where-Object { $_.Count -gt 1 }
+
+if (-not $groupedRoutes) {
+    Write-Host "No duplicate persistent routes found." -ForegroundColor Green
+} else {
+    Write-Host "Cleaning up duplicate persistent routes..." -ForegroundColor Yellow
+
+    foreach ($group in $groupedRoutes) {
+        $toKeep = $group.Group[0]
+        $dest = $toKeep.DestinationPrefix
+        $gateway = $toKeep.NextHop
+        $metric = $toKeep.RouteMetric
+        $interfaceIndex = $toKeep.InterfaceIndex
+
+        $parts = $dest.Split("/")
+        if ($parts.Count -ne 2) {
+            Write-Host "  Invalid destination prefix format: $dest" -ForegroundColor Red
+            continue
+        }
+
+        $network = $parts[0]
+        $prefix = [int]$parts[1]
+        $netmask = Convert-PrefixToMask $prefix
+        $metricStr = if ($metric -is [int]) { "METRIC $metric" } else { "" }
+
+        # Delete all matching routes
+        foreach ($route in $group.Group) {
+            try {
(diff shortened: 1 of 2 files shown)


rwmjones:
How will this be tested before it reaches a customer?

Hazanel:
@rwmjones  - It was reproduced and verified against the fix

rwmjones:
> It was reproduced and verified against the fix

Can you outline the method that you used to reproduce and verify the fix?

Hazanel:
@rwmjones  - In the bug description which attached in the commit message there is a description  of reproducing - it's part of our list that we check on win vm's after migration.
The way to verify is to see that the duplicated routing table per default getaway is gone

rwmjones:
I don't want to know how it could be reproduced, I want to know how **you** did it.  I want to see the detailed logs that you must have collected before and after conversion when you reproduced and then later verified that you had fixed the bug, from the firstboot scripts that run inside the guest, and the virt-v2v log generated by MTV.

Hazanel:
Before_fix
bootLogPostFix
[cold-win2019-vm-post-fix-virt-v2v.txt](https://github.com/user-attachments/files/20504959/cold-win2019-vm-post-fix-virt-v2v.txt)
[cold-win2019-vm-prior-fix-virt-v2v.log](https://github.com/user-attachments/files/20504961/cold-win2019-vm-prior-fix-virt-v2v.log)
Post_fix

rwmjones:
I'm going to ask Yan and Vadim if there's a way to do this programmatically without parsing over the output of a command.  In particular, I think this only works for IPv4.

rwmjones:
I guess we could just do nothing if the `Get-NetRoute` API is not around, as this is a corner case anyway.  Can you help Elad to adjust his patch to work with the API please.

pkg/virt-v2v/customize/scripts/windows/9999-remove_duplicate_persistent_routes.ps1
```diff
@@ -0,0 +1,87 @@
...
+        $routes += [PSCustomObject]@{
+            Network = $match[0].Groups[1].Value
+            Netmask = $match[0].Groups[2].Value
+            Gateway = $match[0].Groups[3].Value
+            Metric = $match[0].Groups[4].Value
+            OriginalLine = $line.Trim()
+            Key = "$($match[0].Groups[1].Value)-$($match[0].Groups[2].Value)-$($match[0].Groups[3].Value)-$($match[0].Groups[4].Value)"
+        }
+    }
+}
```
mnecas: Is there some alternative to the `route print`?
I don't like string parsing out of cli as the formatting can change between versions.

pkg/virt-v2v/customize/scripts/windows/9999-remove_duplicate_persistent_routes.ps1
```diff
@@ -0,0 +1,87 @@
...
+        $routeToPreserve = $group.Group[0]
+        $network = $routeToPreserve.Network
+        $netmask = $routeToPreserve.Netmask
+        $gateway = $routeToPreserve.Gateway
+        $metric = $routeToPreserve.Metric
+
+        $deleteCmd = "route DELETE $network MASK $netmask $gateway"
+        Write-Host "  Deleting ALL duplicates: $deleteCmd" -ForegroundColor Red
+        try {
+            Invoke-Expression $deleteCmd | Out-Null
```
mnecas: Why to use the `Invoke-Expression`?

mnecas: And ignoring the out of it?
shouldn't we pass it to the host logs in case something happens?

pkg/virt-v2v/customize/scripts/windows/9999-remove_duplicate_persistent_routes.ps1
```diff
@@ -0,0 +1,87 @@
...
+            Write-Host "    Deleted successfully." -ForegroundColor Green
+        } catch {
+            Write-Host "    Failed to delete route: $_" -ForegroundColor Red
```
mnecas: How do we know what caused the fail?

pkg/virt-v2v/customize/scripts/windows/9999-remove_duplicate_persistent_routes.ps1
```diff
@@ -0,0 +1,80 @@
...
+                Remove-NetRoute -DestinationPrefix $route.DestinationPrefix -NextHop $route.NextHop `
+                    -InterfaceIndex $route.InterfaceIndex -PolicyStore PersistentStore -Confirm:$false -ErrorAction Stop
+                Write-Host "  Deleted: $($route.DestinationPrefix) via $($route.NextHop)" -ForegroundColor Red
+            } catch {
+                Write-Host "    Failed to delete route: $($_.Exception.Message)" -ForegroundColor Red
+            }
+        }
+
+        # Re-add one preserved route using legacy route.exe
+        $command = "route -p ADD $network MASK $netmask $gateway"
```
mnecas: Could we use the New-NetRoute instead? This way we won't need to deal with the cmd and mask calculation

Hazanel: "New-NetRoute" requires -PolicyStore PersistentStore which has stricter policy although being used in other commands successfully like  "Remove-NetRoute"
I can add New-NetRoute and a fallback to cmd - that's our best option right now

PR #1906, opened 2025-05-26
xcopy-populator: Fix the VMDisk parsing from path + tests

rgolangh:

### DIFF
--- a/cmd/vsphere-xcopy-volume-populator/internal/populator/populate_test.go
+++ b/cmd/vsphere-xcopy-volume-populator/internal/populator/populate_test.go
@@ -1,0 +1,110 @@
+package populator_test
+
+import (
+	"testing"
+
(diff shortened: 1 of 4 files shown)


cmd/vsphere-xcopy-volume-populator/internal/populator/populate_test.go
```diff
@@ -0,0 +1,110 @@
+package populator_test
+
+import (
+	"testing"
+
+	"github.com/kubev2v/forklift/cmd/vsphere-xcopy-volume-populator/internal/populator"
+)
+
+func TestVMDisk_Path(t *testing.T) {
```
borod108, line 9: We already fixed the parsing in the main branch, and took a bit different approach, would be cool if we could use the tests, though they need to be adapted a little.

rgolangh: done.

PR #1907, opened 2025-05-27
Fix downstream v2v image to contain the virt-v2v-in-place

mnecas:
Isuse:
The warm migraiton fails with
```
Building command:virt-v2v-in-place[-v -x -i libvirtxml --root first /mnt/v2v/input.xml]Failed to execute virt-v2v command exec: "virt-v2v-in-place": executable file not found in $PATH
```

Fix:
Add `ENV PATH="$PATH:/usr/libexec"`


solenoci (approved):
lgtm, if you want, go ahead and merge as I want to include it in the new build

PR #1911, opened 2025-05-28
fix os assessment description

AvielSegev:
Each concern is part of a VM, so, like any other Rego rule, it makes sense not to return the formatted` input.guestId`.


yaacov:
> I have a concern about a case where an admin that is not the VM owner create a migration plan and one vm fails because of unsupported OS, can the admin know what is the offending unsupported OS ?

After discussion with @AvielSegev , he explained that in migration-planner, they create a report grouping and summarizing warnings and errors, for their use case they take the first "assessment" from each  "label" as the "assessment" for the whole group.

For this use case we need to keep the "assessment" clean of individual VM information, o/w we will break the  migration-planner reports.

PR #1928, opened 2025-05-29
MTV-2263 | Plan does not detect changes

Hazanel:
Issue:
Require source VMs with shared disks to be powered off before migration

When migrating VMs with shared disks, all related VMs must be in a powered-off state. If any are powered on, the migration plan enters an error state. Changes to VM power state after plan creation are not automatically detected; users must create a new plan or copy the existing one to refresh VM states.

Fix:
Reqeue plans with blocker conditions to reconsile queue as well.

Ref:
https://issues.redhat.com/browse/MTV-2263


pkg/controller/plan/controller.go
```diff
@@ -447,7 +447,7 @@ func (r *Reconciler) execute(plan *api.Plan) (reQ time.Duration, err error) {
 	if migration == nil {
 		r.Log.Info("No pending migrations found.")
 		plan.Status.DeleteCondition(Executing)
-		reQ = NoReQ
+		reQ = base.SlowReQ
```
mnecas, line 450: @mansam  do you have any alternative idea how to check outside changes?

mnecas: I was thinking that we could send some request from inventory to the main once the VM gets updated to trigger the reconcile async. But afaik we would need to update the CR do this trigger and I'm not aware of having a trigger API to restart reconcile.

Hazanel: @mnecas  - Any decision/thoughts about this one?

PR #1930, opened 2025-05-29
MTV-2809 | xcopy-populator: Add Pure FlashArray provider

rgolangh:
This work adds support for Pure FlashArray.

Main difference of this provider to others is that it doesn't allow one
host to belong to more than one hostGroups. Hence, map and unmap will be
performed on the host itself, and not use xcopy-esxs groups. For this
reason the EnsureClonnerIgroup for this provider returns a mapping
context with a list of hosts that match the ESX IQN or WWN. This list is
used later by the map and unmap.

Tested with iSCSI.
Untested with FC.

https://issues.redhat.com/browse/MTV-2809


rgolangh:
Still some rough edges I need to work on, but functional

cmd/vsphere-xcopy-volume-populator/internal/pure/flashArray.go
```diff
@@ -0,0 +1,192 @@
...
+		klog.Fatalf("Error getting array status: %v", err)
+	}
+	klog.Infof("Array Name: %s, ID: %s", array.ArrayName, array.ID)
+	return FlashArrayClonner{client: client}, nil
+}
+
+// EnsureClonnerIgroup creates or updates an initiator group with the clonnerIqn
+func (f *FlashArrayClonner) EnsureClonnerIgroup(initiatorGroup string, clonnerIqn []string) (populator.MappingContext, error) {
+
+	if true {
```
xinzheyang: if true?

cmd/vsphere-xcopy-volume-populator/internal/pure/flashArray.go
```diff
@@ -0,0 +1,192 @@
...
+			}
+			for _, wwn := range h.Wwn {
+				if slices.Contains(clonnerIqn, wwn) {
+					klog.Infof("adding host to group %v", h.Name)
+					hostNames = append(hostNames, h.Name)
+				}
+			}
+		}
+		return populator.MappingContext{"hosts": hostNames}, nil
+	}
```
xinzheyang: lines below this return won't get executed? I think you didn't mean to keep this if block

cmd/vsphere-xcopy-volume-populator/internal/pure/flashArray.go
```diff
@@ -0,0 +1,192 @@
...
+
+	klog.Infof("starting ensure clonner group with %v %v", initiatorGroup, clonnerIqn)
+	g, err := f.client.Hostgroups.GetHostgroup(initiatorGroup, nil)
+	if err != nil {
```
xinzheyang: I see initiatorGroup is a hardcoded name. How do we handle cleanup from previous workloads, either successful or failed in the middle.
and if we have concurrent executions of this populator workload, wouldn't this run into problems when everybody is dealing with the same hgroup?

cmd/vsphere-xcopy-volume-populator/internal/pure/flashArray.go
```diff
@@ -0,0 +1,192 @@
...
+func (f *FlashArrayClonner) CurrentMappedGroups(targetLUN populator.LUN, context populator.MappingContext) ([]string, error) {
+	connectedHostGroups := []string{}
+	hgs, err := f.client.Hostgroups.ListHostgroups(nil)
+	if err != nil {
+		return nil, err
+	}
+	for _, hg := range hgs {
+		connections, err := f.client.Hostgroups.ListHostgroupConnections(hg.Name)
+		if err != nil {
+			return nil, err
```
xinzheyang: it's more straightforward to use ListVolumeSharedConnections

cmd/vsphere-xcopy-volume-populator/internal/pure/flashArray.go
```diff
@@ -0,0 +1,192 @@
...
+				}
+
+			}
+		}
+	}
+
+	if true {
+		return targetLUN, nil
+	}
+
```
xinzheyang: didn't mean to keep this?

cmd/vsphere-xcopy-volume-populator/internal/pure/flashArray.go
```diff
@@ -0,0 +1,192 @@
...
+				klog.Infof("disconnecting host %s from volume %s", host, targetLUN.Name)
+				_, err := f.client.Hosts.DisconnectHost(host, targetLUN.Name)
+				if err != nil {
+					return err
+				}
+
+			}
+		}
+	}
+	return nil
```
xinzheyang, line 111: if you connected lun to hgroup directly, then just disconnect disconnect hgroup. no need to iterate over hosts and manipulate connections, which defeats the purpose of hgroups.

cmd/vsphere-xcopy-volume-populator/internal/pure/flashArray.go
rgolangh: strictly for testing, I removed it

cmd/vsphere-xcopy-volume-populator/internal/pure/flashArray.go
rgolangh: I am not using hgroups at all in the end, as I saw that pure rejects an attempt to add a host to a group if it already belongs to one. 
Hence I call the connection on an individual host and then revert at the end.

cmd/vsphere-xcopy-volume-populator/internal/pure/flashArray.go
rgolangh: correct, I'm cleaning this up, this was temporary

cmd/vsphere-xcopy-volume-populator/internal/pure/flashArray.go
```diff
@@ -0,0 +1,146 @@
...
+func (f *FlashArrayClonner) CurrentMappedGroups(targetLUN populator.LUN, context populator.MappingContext) ([]string, error) {
+	connectedHostGroups := []string{}
+	hgs, err := f.client.Hostgroups.ListHostgroups(nil)
+	if err != nil {
+		return nil, err
+	}
+	for _, hg := range hgs {
+		connections, err := f.client.Hostgroups.ListHostgroupConnections(hg.Name)
+		if err != nil {
+			return nil, err
```
xinzheyang: this doesn't make any sense if we are not using hgroup at all?
does the cleanup code (whoever is calling CurrentMappedGroups) do the right thing if we are not using hgroups?

cmd/vsphere-xcopy-volume-populator/internal/pure/flashArray.go
```diff
@@ -0,0 +1,146 @@
...
+				}
+
+			}
+		}
+	}
+
+	if true {
+		return targetLUN, nil
+	}
+
```
xinzheyang: remove if true.
and we need to remove the code below this?

xinzheyang (changes requested):

cmd/vsphere-xcopy-volume-populator/internal/populator/vib.go
```diff
@@ -24,6 +24,9 @@ var VibVersion = "x.x.x"
 // ensure vib will fetch the vib version and in case needed will install it
 // on the target ESX
 func ensureVib(client vmware.Client, esx *object.HostSystem, datastore string, desiredVibVersion string) error {
+	if true {
```
mnecas: :eyes: @rgolangh

operator/.downstream_manifests
```diff
@@ -54,7 +54,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.17.0
+    controller-gen.kubebuilder.io/version: v0.18.0
   name: hooks.forklift.konveyor.io
```
rgolangh: I think the makefile  generate-manifest is using the local developer kubectl. we should instead do that in a container. @mnecas what version of kubectl you run?

cmd/vsphere-xcopy-volume-populator/internal/pure/flashArray.go
xinzheyang: @rgolangh I think this comment is not addressed. shouldn't this function be a no-op if none of our connections are through hgroup?

xinzheyang (changes requested):

cmd/vsphere-xcopy-volume-populator/internal/pure/flashArray.go
rgolangh: yes, I'm removing this, as we don't use host groups. thanks

mnecas (approved):

PR #1934, opened 2025-05-30
MTV-2471 | Windows-Preserve multiple ip’s per nic

Hazanel:
Issue:
Nics with multiple Ip's are not preserved.
Only the primary ip is preserved  from virt-v2v but the rest are droped due to re-assigning of default-gatway .

Fix:
Preserve the complementry IP's seperatly to avoid re-assigning of default-gatway .

Ref:
https://issues.redhat.com/browse/MTV-2471

### DIFF
--- a/pkg/virt-v2v/customize/customize.go
+++ b/pkg/virt-v2v/customize/customize.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"reflect"
 	"regexp"
 	"strings"
 	"text/template"
@@ -33,6 +34,46 @@ type Customize struct {
 	commandBuilder     utils.CommandBuilder
 	fileSystem         utils.FileSystem
 	embeddedFileSystem EmbedTool
+}
+
+type IPConfig struct {
+	MAC string
+	IPs []IPEntry
+}
+
+type IPEntry struct {
+	IP           string
+	Gateway      string
+	PrefixLength string
+	DNS          []string
+}
+
+func formatIPs(ips []IPEntry) string {
+	var b strings.Builder
+	b.WriteString("(\n")
+	for i, ip := range ips {
+		b.WriteString("'" + ip.IP + "'")
+		if i < len(ips)-1 {
+			b.WriteString(",")
+		}
+		b.WriteString("\n")
+	}
+	b.WriteString(")")
+	return b.String()
+}
+
+func formatDNS(dns []string) string {
+	var b strings.Builder
+	b.WriteString("(\n")
+	for i, ip := range dns {
+		b.WriteString("'" + ip + "'")
+		if i < len(dns)-1 {
+			b.WriteString(",")
+		}
+		b.WriteString("\n")
+	}
+	b.WriteString(")")
+	return b.String()
 }
 
 func NewCustomize(cfg *config.AppConfig, disks []string, operatingSystem utils.InspectionOS) *Customize {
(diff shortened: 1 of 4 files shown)


Hazanel:
This PR is pending on #1903  just in order to avoid merge conflicts, since #1903  requires static_ip enabled and this one requires another step within static_ip enabled part

pkg/virt-v2v/customize/scripts/windows/9999-preserve_complementry_ips_per_nic.ps1.tmpl
```diff
@@ -0,0 +1,50 @@
...
+$adapters = @()
+While (-Not $adapters -and $startdate.AddMinutes(5) -gt (Get-Date)) {
+    Start-Sleep -Seconds 5
+    $adapters = Get-NetAdapter -Physical | Where DriverFileName -eq "netkvm.sys"
+    Write-Host "Waiting for netkvm.sys adapters..."
+}
+
+# Define MACs and IPs for each adapter
+$networkConfigs = @(
+{{- range $i, $cfg := . }}
```
mnecas, line 12: Why is there dash `-`?

mnecas: The dot `.` is the input? having problems finding where and what you populate with the `configs []IPConfig`

pkg/virt-v2v/customize/scripts/windows/9999-preserve_complementry_ips_per_nic.ps1.tmpl
```diff
@@ -0,0 +1,50 @@
...
+    @{
+        MAC = '{{ lower $cfg.MAC }}'
+        IPs = @(
+        {{- range $j, $ip := $cfg.IPs }}
+            '{{ $ip.IP }}'{{ if ne (add $j 1) (len $cfg.IPs) }},{{ end }}
```
mnecas: Wouldn't it be more clear to do these evaluation inside the go and use template to just put the strings?

pkg/virt-v2v/customize/scripts/windows/9999-preserve_complementry_ips_per_nic.ps1.tmpl
Hazanel: In Go templates:  
 1. The "-"  is a whitespace trimming modifier
 2. The "." represents the current context or data - must be a slice or array of configurations

This line tells Go to:

Iterate over the input (.),

Trim any white-space before the range.

That's what it looks like after generating the script:
$networkConfigs = @(
    @{
        MAC = '00-50-56-be-56-a1'
        IPs = @(
            '192.168.20.37',
            '192.168.20.38',
            '192.168.20.39',
            '192.168.20.40'
        )
        PrefixLength = 24
        DNS = @(
            '192.168.20.200',
            '192.168.20.201'
        )
    },
    @{
        MAC = '00-50-56-be-74-f8'
        IPs = @(
            '10.2.122.78'
        )
        PrefixLength = 23
        DNS = @(
            '10.2.114.201',
            '10.2.114.202'
        )
    }
)

pkg/virt-v2v/customize/scripts/windows/9999-preserve_complementry_ips_per_nic.ps1.tmpl
Hazanel: Hard to tell - I'm biased :), but  I think it's more elegant that way cause we keep the go code itself clean

mnecas: I do still find it really hard to read, and think we should input the, DNS, MAC and IP listis and just placing them to in the template without this manipulation @yaacov wdyt?

yaacov: hmm ...
my 2 cents, 

it is a loop within a loop in a template that will get interpreted into a shell script,
no way to make it nice :-)

it can be nicer is we extract the ip and dns loops into dedicated functions, so the template will look more like:
```go
MAC = '{{ lower $cfg.MAC }}'
IPs = {{ formatIPs $cfg.IPs }}
PrefixLength = {{ (index $cfg.IPs 0).PrefixLength }}
DNS = {{ formatDNS (index $cfg.IPs 0).DNS }}
```

a `formatIPs` method can then move the comlexity to go, a language we feel better maintaining:
```go
// formatIPs creates a PowerShell array string from IP addresses
func formatIPs(ips []NetworkIP) string {
	if len(ips) == 0 {
		return "@()"
	}
	
	var ipStrings []string
	for _, ip := range ips {
		ipStrings = append(ipStrings, "'"+ip.IP+"'")
	}
	
	return "@(" + strings.Join(ipStrings, ", ") + ")"
}
```

the pros of this approach is moving the complexity to a language we feel better maintaining and can more easily read and validate

mnecas: Yeah that looks like something I had in mind

pkg/virt-v2v/customize/customize.go
```diff
@@ -115,6 +156,86 @@ func (c *Customize) addDisksToCustomize(cmdBuilder utils.CommandBuilder) {
...
+	funcMap := template.FuncMap{
+		"lower": strings.ToLower,
+		"add": func(a, b int) int {
+			return a + b
+		},
+		"len": func(v interface{}) int {
+			return reflect.ValueOf(v).Len()
+		},
+		"formatIPs": func(ips []IPEntry) string {
+			return formatIPs(ips, "            ")
```
mnecas: What are these spaces?

yaacov:
lgtm

PR #1947, opened 2025-06-04
MTV-1369 | Add UDN support

mnecas:
# Add UDN support to the Forklift

User Defined Network (UDN) is a new feature in OVN-K that allows users to isolate the namespace networking.
Because of this isolation, we need to expose some of the ports via `k8s.ovn.org/open-default-ports` so we can access the information from the UDN.

This new network type is an alternative to the pod default network. In the YAML, the network mapping will look identical to the `type: pod` mapping, but the controller will decide according to the destination namespace behaviour if we need to create a VM with UDN support.

Kubevirt supports only L3 networks, so I have decided to filter out the network from the inventory and add validations to block the migrations when an L3 network is specified.

This feature works for the following providers:
- [x] oVirt
- [x] OSP
- [x] VMware
- [x] OVA
- [ ] OCP (missing [MTV-3262](https://issues.redhat.com/browse/MTV-3262))

Notes:
- This PR is missing OCP to OCP migration, which was split to separate the epic https://issues.redhat.com/browse/MTV-3262
- The OCP CNV below 4.20 can not specify the MAC address on the YAML; otherwise, it will not get an IP address from the UDN DHCP. https://issues.redhat.com/browse/CNV-66820

## Summary by CodeRabbit

- New Features
  - UDN networking: Pod networking binds to an L2 bridge when a User Defined Network is detected.
  - Conversions on UDNs are annotated to open default ports (TCP 2112, 8080).
  - Conditional MAC assignment governed by a new feature flag that enables MACs only on supported cluster versions.

- Bug Fixes
  - Validation now detects and rejects unsupported UDN configurations (e.g., Layer3 OVN overlays).

- Chores
  - Removed obsolete Go toolchain pinning.

- Style
  - Minor whitespace cleanup in task definitions.

### DIFF
--- a/pkg/apis/forklift/v1beta1/plan.go
+++ b/pkg/apis/forklift/v1beta1/plan.go
@@ -42,6 +42,11 @@ const (
 	MigrationOnlyConversion MigrationType = "conversion"
 )
 
+const (
+	namespaceLabelPrimaryUDN = "k8s.ovn.org/primary-user-defined-network"
+	nadLabelUDN              = "k8s.ovn.org/user-defined-network"
+)
+
 // PlanSpec defines the desired state of Plan.
 type PlanSpec struct {
 	// Description
@@ -277,7 +282,7 @@ func (r *Plan) DestinationHasUdnNetwork(client k8sclient.Client) bool {
 	if err != nil {
 		return false
 	}
-	_, hasUdnLabel := namespace.ObjectMeta.Labels["k8s.ovn.org/primary-user-defined-network"]
+	_, hasUdnLabel := namespace.ObjectMeta.Labels[namespaceLabelPrimaryUDN]
 	if !hasUdnLabel {
 		return false
 	}
@@ -285,7 +290,7 @@ func (r *Plan) DestinationHasUdnNetwork(client k8sclient.Client) bool {
 	nadList := &k8snet.NetworkAttachmentDefinitionList{}
 	listOpts := []k8sclient.ListOption{
 		k8sclient.InNamespace(r.Spec.TargetNamespace),
-		k8sclient.MatchingLabels{"k8s.ovn.org/user-defined-network": ""},
+		k8sclient.MatchingLabels{nadLabelUDN: ""},
 	}
 
 	err = client.List(context.TODO(), nadList, listOpts...)
--- a/pkg/apis/forklift/v1beta1/plan.go
+++ b/pkg/apis/forklift/v1beta1/plan.go
@@ -18,6 +18,7 @@ package v1beta1
 
 import (
 	"context"
+
 	k8snet "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
 	"github.com/kubev2v/forklift/pkg/apis/forklift/v1beta1/plan"
 	"github.com/kubev2v/forklift/pkg/apis/forklift/v1beta1/provider"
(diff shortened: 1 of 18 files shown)


pkg/apis/forklift/v1beta1/plan.go
```diff
@@ -236,6 +239,32 @@ func (p *Plan) ShouldUseV2vForTransfer() (bool, error) {
...
+	}
+	_, hasUdnLabel := namespace.ObjectMeta.Labels["k8s.ovn.org/primary-user-defined-network"]
+	if !hasUdnLabel {
+		return false
+	}
+	// TODO: mnecas not sure if we should check also the NAD or only namespace
+	key = k8sclient.ObjectKey{
+		Namespace: r.Spec.TargetNamespace,
+		Name:      "primary-udn",
+	}
```
mnecas: Does not apply to cudn 
maybe label `k8s.ovn.org/user-defined-network: ''`

pkg/controller/plan/adapter/openstack/builder.go
```diff
@@ -605,8 +608,15 @@ func (r *Builder) mapNetworks(vm *model.Workload, object *cnv.VirtualMachineSpec
 
 				switch networkPair.Destination.Type {
 				case Pod:
-					kNetwork.Pod = &cnv.PodNetwork{}
-					kInterface.Masquerade = &cnv.InterfaceMasquerade{}
+					if r.Plan.DestinationHasUdnNetwork(r.Destination) {
+						kNetwork.Pod = &cnv.PodNetwork{}
+						kInterface.Binding = &cnv.PluginBinding{
+							Name: "l2bridge",
```
mnecas: should be const

pkg/controller/plan/kubevirt.go
```diff
@@ -1970,6 +1981,17 @@ func (r *KubeVirt) guestConversionPod(vm *plan.VMStatus, vmVolumes []cnv.Volume,
...
+	if r.Plan.DestinationHasUdnNetwork(r.Destination) {
+		metricsPort := OpenPort{Protocol: "tcp", Port: 2112}
+		dataServerPort := OpenPort{Protocol: "tcp", Port: 8080}
+		ports := []OpenPort{metricsPort, dataServerPort}
+		var yamlPorts []byte
+		yamlPorts, err = yaml.Marshal(ports)
+		if err != nil {
+			return
+		}
+		annotations["k8s.ovn.org/open-default-ports"] = string(yamlPorts)
```
mnecas: should be const

pkg/apis/forklift/v1beta1/plan.go
```diff
@@ -264,6 +268,37 @@ func (p *Plan) ShouldUseV2vForTransfer() (bool, error) {
 	}
 }
 
+func (r *Plan) DestinationHasUdnNetwork(client k8sclient.Client) bool {
```
yaacov, line 276: we are abusing the api definitions file, this method is doing api calls using a k8s client, maybe we can put this method some ware else ?

mnecas: Hmm, good point, but not sure where else to put this. It is used across all providers

pkg/apis/forklift/v1beta1/plan.go
```diff
@@ -264,6 +268,37 @@
...
+
+	err = client.List(context.TODO(), nadList, listOpts...)
+	if err != nil {
+		return false
+	}
+
+	if len(nadList.Items) > 0 {
+		return true
+	}
+	return false
```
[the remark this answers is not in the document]
mnecas: The CUDNs have the same labels

pkg/apis/forklift/v1beta1/plan.go
yaacov: will it look better in adapter base ?

PR #1949, opened 2025-06-05
MTV-2669 | An incorrect IP address keeps the VMware provider stuck in a non-ready state

Hazanel:
Issue:
Creating a VMware provider with wrong ip for example https://10.10.10.10/sdk ending with creating the provider with "undefined" state and the forkflift-controller gets stuck.

Fix:
1. Add a timeout to the tls-certificate method.
2. Change provider condition upon timeout
3. Update provider status after validation failure Ref: https://issues.redhat.com/browse/MTV-2669

Ref: https://issues.redhat.com/browse/MTV-2669

### DIFF
--- a/pkg/lib/util/util.go
+++ b/pkg/lib/util/util.go
@@ -6,13 +6,33 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"fmt"
+	"net"
 	liburl "net/url"
 	"strconv"
+	"time"
 
 	api "github.com/konveyor/forklift-controller/pkg/apis/forklift/v1beta1"
 	liberr "github.com/konveyor/forklift-controller/pkg/lib/error"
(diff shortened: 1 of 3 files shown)


pkg/lib/util/util.go
```diff
@@ -35,7 +55,7 @@ func GetTlsCertificate(url *liburl.URL, secret *core.Secret) (crt *x509.Certific
 	}
 	// disable verification since we don't trust it yet
 	cfg.InsecureSkipVerify = true
-	conn, err := tls.Dial("tcp", host, cfg)
+	conn, err := dialTLSWithTimeout(host, cfg, 5*time.Second)
```
mnecas: Please make this timeout configurable from the controller settings and forklift-controller CR

PR #1958, opened 2025-06-05
Collect vendor and model of the host

AvielSegev:
Adding the Collection of each host’s hardware vendor (hardware.systemInfo.vendor) and model (hardware.systemInfo.model) in the vSphere collector, and add corresponding Vendor and Model fields to the Host struct.

Signed-off-by: Aviel Segev <asegev@redhat.com>


AvielSegev:
> code lgtm what do you need the info for? what are the possible values of it? I expect it to be identical as we have there only the esxi hosts.
> 
> Please add some jira ticket to the PR

Thanks @mnecas and sorry for omitting the Jira ticket. This PR relates to [ECOPROJECT-3072](https://issues.redhat.com/browse/ECOPROJECT-3072).
Example for output: 
Vendor | Model
------ | --------------------
HPE    | ProLiant DL380 Gen10

PR #1970, opened 2025-06-09
MTV-2687: Block warm migraiton without vddk

mnecas:
Issue:
Failed to migrate VM from vsphere provider skipped VDDK acceleration

Fix:
Block the warm plans migration without the VDDK as it's mandatory.

Ref: https://issues.redhat.com/browse/MTV-2687


yaacov (approved):
:+1:

PR #1989, opened 2025-06-10
MTV-2671 |  Migration from RHV to OpenShift Virtualization is failing with error "Running and RunStrategy are mutually exclusive"

Hazanel:
Issue:
Creating a virtual machine with "Operating System" type "Linux" in RHV, the inventory cache of forklift have other_linux as OS type. This will not match any cluster preference and fallback to template selection, For template selection, it will search for templates with label rhel8.1 . That will be rhel8 templates and in OpenShift Virtualization < 4.18, these templates have "running" in spec which shuld be deprecated but we still need to backword compatible.

Fix:
Adding running to template and identify which methos is being used in the template and update accordingly .

ref: https://issues.redhat.com/browse/MTV-2671


pkg/controller/plan/kubevirt.go
```diff
@@ -1472,16 +1472,24 @@ func (r *KubeVirt) virtualMachine(vm *plan.VMStatus, sortVolumesByLibvirt bool)
...
+		running = true
+	} else {
+		runStrategy = cnv.RunStrategyHalted
+		running = false
 	}
 
-	// Assign the determined run strategy to the object
-	object.Spec.RunStrategy = &runStrategy
+	// Apply the run configuration based on what's already defined in the object spec
+	if object.Spec.RunStrategy != nil {
```
yaacov: another option is to clear the object.Spec.Running

e.g.
```go
object.Spec.RunStrategy = &runStrategy
object.Spec.Running = nil // Ensure running is not set
```

ensureing running is null is more inline with deprecating the usage of running

yaacov (approved):

PR #2002, opened 2025-06-12
MTV-2707 | xcopy-popualtor: Pass the vmId to the populator

rgolangh:
Motivation
Up until now the populator was deducing the name of the VM from the path
of the vmdk file. That approach breaks when 1. a VM is renamed 2. the
disk is a shared disk 3. there are more vm's with same name on another
DC. Probably more cases exists.

Modification
Forkflift will add the vmId to the VsphereXcopyVolumePopulator resource,
which now have a new field, VmId. From there on that is passed all the
way to the command line argument --source-vm-id and there its used to
find the ESX by VM ID.

Result
The migration runs exactly the same, there are no visible changes, or UX
changes, this is all transparent to the user.
Renamed or imported VMs or any of the cases mentioned above will migrate
without any problems.

https://issues.redhat.com/browse/MTV-2707


cmd/vsphere-xcopy-volume-populator/internal/vmware/client.go
```diff
@@ -72,7 +72,7 @@ func (c *VSphereClient) RunEsxCommand(ctx context.Context, host *object.HostSyst
 	return res.Values, nil
 }
 
-func (c *VSphereClient) GetEsxByVm(ctx context.Context, vmName string) (*object.HostSystem, error) {
+func (c *VSphereClient) GetEsxByVm(ctx context.Context, vmId string) (*object.HostSystem, error) {
 	finder := find.NewFinder(c.Client.Client, true)
 
 	//FIXME - need to trace the VM by the datastore, which we should have because
```
rgolangh: fix that already - maybe we donh't need to set a datacenter if we get the vm by id. anyhow, get rid of the datacednter part because there may be a case where there is no datacenger

rgolangh: /hold need to change a small thing with the default DC (remove it)

rgolangh: /hold cancel
This is ready and working

mnecas (approved):

PR #2029, opened 2025-06-18
chore | Sent an error on inventory client watch fail

yaacov:
Issue:
In inventory web client, we may return misleading error when  status != http.StatusOK
in this case the err value can be `nil` and the error is lost

Fix:
Add an error the case of   status != http.StatusOK

Before:
`{"level":"error","ts":"2025-06-17 11:07:17.634","logger":"storageMap","msg":"","error":"Not Implemented","stacktrace":"github ... `

After:
`{"level":"error","ts":"2025-06-18 07:28:06.685","logger":"storageMap","msg":"","error":"inventory client watch return with none StatusOK","stacktrace":"github.com/kubev2v/forklift/pkg/controller/ ... `

### DIFF
--- a/pkg/lib/inventory/web/client.go
+++ b/pkg/lib/inventory/web/client.go
@@ -285,6 +285,12 @@ func (r *Client) Watch(url string, resource interface{}, h EventHandler) (status
 	}
 	status, err = post(reader)
 	if err != nil || status != http.StatusOK {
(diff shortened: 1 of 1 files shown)


yaacov:
@solenoci please revirew

Hazanel (approved):

PR #2076, opened 2025-06-23
Fix "Not implemented" error for ensureWatch in ocp providers

jonner:
Since OCP providers no longer provide 'watch' functionality via the inventory web interface, we provide a fake watch implementation that simply drips events into the channel and causes the objects to re-reconcile periodically.


solenoci:
The konflux error is weird... everything works as it should and I can't find any evidence of this... so I think it is safe to ignore

yaacov (approved):
lgtm

mnecas (approved):

PR #2077, opened 2025-06-24
ECOPROJECT-2558, MTV-1616 | Streamline the VDDK image creation

AvielSegev:
During the UI implementation, we encountered some CORS-related issues and would like to wait until the [PR](https://github.com/kubev2v/forklift-console-plugin/pull/1651) is merged to determine whether it’s a local problem.

I tried using the new upload feature to create a VDDK image with the forklift dev operator, but it seems to have the same CORS issue.

The problem: For POST requests, the web server enforces CORS. Currently, allowedOrigins is initialized as an empty array (and can be configured using the CORS_ALLOWED_ORIGINS environment variable); therefore, the w.allow method, which is responsible for authorizing requests, drops them (for POST) and returns a 403.

While debugging, I noticed that the UI requests originate from the console URL. This PR suggests fixing the problem by adding the console URL to the allowed origins. 

I’m not sure this is the best approach—please feel free to comment or suggest an alternative.

[ECOPROJECT-2558](https://issues.redhat.com/browse/ECOPROJECT-2558)
[MTV-1616](https://issues.redhat.com/browse/MTV-1616)
Cors Issue

### DIFF
--- a/pkg/settings/settings.go
+++ b/pkg/settings/settings.go
@@ -1,12 +1,18 @@
 package settings
 
 import (
+	"context"
 	"fmt"
 	"os"
 	"strconv"
 
 	api "github.com/kubev2v/forklift/pkg/apis/forklift/v1beta1"
 	liberr "github.com/kubev2v/forklift/pkg/lib/error"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/rest"
 )
 
 // Global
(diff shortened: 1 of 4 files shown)


pkg/settings/settings.go
```diff
@@ -132,3 +138,40 @@ func GetVDDKImage(providerSpecSettings map[string]string) string {
...
+		return "", fmt.Errorf("failed to get console Route: %w", err)
+	}
+
+	host, found, err := unstructured.NestedString(obj.UnstructuredContent(), "spec", "host")
+	if err != nil {
+		return "", err
+	}
+
+	if !found {
+		return "", fmt.Errorf("console Route is missing spec.host")
```
mnecas, line 173: Not sure if we should fail the whole inventory when the route is missing.

pkg/settings/settings.go
```diff
@@ -132,3 +138,40 @@ func GetVDDKImage(providerSpecSettings map[string]string) string {
...
+
+	routeGVR := schema.GroupVersionResource{
+		Group:    "route.openshift.io",
+		Version:  "v1",
+		Resource: "routes",
+	}
+
+	obj, err := dynClient.
+		Resource(routeGVR).
+		Namespace("openshift-console").
```
mnecas, line 161: Is it always in the `openshift-console`?

pkg/settings/settings.go
```diff
@@ -132,3 +138,40 @@ func GetVDDKImage(providerSpecSettings map[string]string) string {
...
+		Get(context.Background(), "console", metav1.GetOptions{})
+	if err != nil {
+		return "", fmt.Errorf("failed to get console Route: %w", err)
+	}
+
+	host, found, err := unstructured.NestedString(obj.UnstructuredContent(), "spec", "host")
```
mnecas, line 167: Why unstructured?

yaacov:
> but it seems to have the same CORS issue.

the UI proxy the call, ALL calls should use the UI proxy and they will have no CORS issues 

did you check that ALL requests use the console URL ?
maybe they try to connect directly to our server ?

UI proxy all requests:
https://github.com/openshift/console/blob/main/pkg/proxy/proxy.go#L56

yaacov:
p.s.
they also delete all the headers you try to add here:
https://github.com/openshift/console/blob/main/pkg/proxy/proxy.go#L37

yaacov:
@AvielSegev what is the URL you are using ?

pkg/settings/settings.go
```diff
@@ -132,3 +138,40 @@ func GetVDDKImage(providerSpecSettings map[string]string) string {
 
 	return vddkImage
 }
+
+func getOpenshiftConsoleOrigin() (string, error) {
```
yaacov, line 142: this should be done on the console side:
https://github.com/openshift/console

pkg/settings/settings.go
yaacov: routes are OCP specific, we don't have them on K8s clusters, meaning our KIND baded CI will fail

AvielSegev:
We're probably going to solve this without any changes to Forklift.
Thanks to @yaacov for helping with the debugging and being involved in the discussions with the OpenShift Console team.
I'll close this PR once [this PR](https://github.com/openshift/console/pull/15227) is merged and we've verified that it resolves the issue.

PR #2078, opened 2025-06-24
PAC rules prevented the build of API image

solenoci:
forklift-api should be rebuild even when there are changes to the pkg/controller directory and vice versa.


yaacov:
don't we need lib as well ?

yaacov (approved):
lgtm,
maybe we need to consider more dependencies, e.g. pkg/lib ?

solenoci:
We will need to go through all of the PAC rules and ensure so this does not happen in the future

PR #2083, opened 2025-06-24
MTV-2793 | Add plan warning for unsupported OVA sources.

mrnold:
Reference: https://issues.redhat.com/browse/MTV-2793

Issue: the existing warning from MTV-2314 was only shown on the OVA provider inventory page, and should really also be shown on the plan page.

Fix: during plan validation, check for any OVAs with the "Unsupported OVA source" concern label and add a new warning to the plan if it is found.

### DIFF
--- a/pkg/controller/plan/validation.go
+++ b/pkg/controller/plan/validation.go
@@ -16,6 +16,7 @@ import (
 	"github.com/kubev2v/forklift/pkg/controller/plan/adapter"
 	plancontext "github.com/kubev2v/forklift/pkg/controller/plan/context"
 	"github.com/kubev2v/forklift/pkg/controller/provider/web"
+	"github.com/kubev2v/forklift/pkg/controller/provider/web/ova"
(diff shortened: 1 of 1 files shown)


pkg/controller/plan/validation.go
```diff
@@ -629,6 +638,15 @@ func (r *Reconciler) validateVM(plan *api.Plan) error {
 				setOfTargetName[vm.TargetName] = true
 			}
 		}
+		// check for supported OVA source
+		if ova, ok := v.(*ova.VM); ok {
+			for _, concern := range ova.Concerns {
+				// match label from ova/export_source.rego
+				if concern.Label == "Unsupported OVA source" {
```
yaacov: we now have ids :tada: , we can test by id witch is nicer then by label:

"id": "ova.source.unsupported",

mrnold: Thanks, this is nicer.

PR #2087, opened 2025-06-25
MTV-2802 | Mutate host secret add skip verify flag

yaacov:
Ref: https://issues.redhat.com/browse/MTV-2802

**Issue:**
In https://github.com/kubev2v/forklift/pull/2073 we "hot fixed" the missing host secret field by forcing insecure skip verify to true.

**Fix:**
  - Revert #2073
  - Add a secret mutation hook to set the missing `insecureSkipVerify` field in the host secret based on provider settings.

**Concerns:**
  - skip verify settings may be different for the vCenter and the ESXi.
  - a ca cert that works for vCenter may not work for the ESXi , causing TLS error.

**Notes:**
  - users can set skip verify flag in host secret manually, overriding copy from vCenter secret
  - users can set `cacert` field in host secret to use custom cacert for the ESXi host
 
**How to tests:**
_Scenario I: host secret without skip flag, copy from provider_
a. create a vCenter provider + secret with skip verify set to `true`
b. create a Host + secret without setting skip verify.

Check the new Host secret, a new field with skip verify set to `true` should be set.

_Scenario I: host secret without skip flag, copy from provider_
a. create a vCenter provider + secret with skip verify set to `fasle`
b. create a Host + secret without setting skip verify.

Check the new Host secret, a new field with skip verify set to `false` should be set.

_Scenario III: host secret with skip flag, ignore provider settings_
a. create a vCenter provider + secret with skip verify set to `true`
b. create a Host + secret with setting skip verify set to `false` and `cacert` data field to valid certificate.

Check the new Host secret, a field with skip verify set to `false` should not be changed.

### DIFF
--- a/pkg/controller/plan/adapter/vsphere/host.go
+++ b/pkg/controller/plan/adapter/vsphere/host.go
@@ -5,8 +5,10 @@ import (
 	liburl "net/url"
 	"time"
 
+	"github.com/kubev2v/forklift/pkg/controller/base"
 	model "github.com/kubev2v/forklift/pkg/controller/provider/web/vsphere"
 	liberr "github.com/kubev2v/forklift/pkg/lib/error"
+	"github.com/kubev2v/forklift/pkg/lib/util"
 	"github.com/vmware/govmomi"
 	"github.com/vmware/govmomi/find"
 	"github.com/vmware/govmomi/session"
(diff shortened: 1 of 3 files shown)


yaacov:
@Hazanel @mnecas please review

pkg/controller/plan/adapter/vsphere/host.go
```diff
@@ -73,9 +75,20 @@ func (r *EsxHost) connect(ctx context.Context) (err error) {
 		r.user(),
 		r.password())
 
-	// Always use insecure skip verify for ESX hosts.
-	soapClient := soap.NewClient(url, true)
-	soapClient.SetThumbprint(url.Host, r.thumbprint())
+	thumbprint := r.thumbprint()
+	skipVerifying := base.GetInsecureSkipVerifyFlag(r.Secret)
```
mnecas, line 79: What will hapen to already existing hosts?

yaacov: missing flag is interpreted as "false", so they will start requiring tls verification

a. the issue is that missing value and explicit "false" look the same ...

do you think we should implement a new `GetInsecureSkipVerifyFlagForHost` that will default to `true` in case of missing value ?

yaacov: hmm, a work around is simple, 
existing hosts that will become invalid due to this fix, can be revived easily by re-creating them.
the re-create w/a can work both manually and using CI.

PR #2092, opened 2025-06-26
MTV-2808 | fix the automatic clusterrolebinding creation for the populator SA

rgolangh:
The binding was duplicating the subjects. The patch implementation now
don't include those.
Also the creation of the binding was change to not run over other fields
as we saw that sometimes it fails

https://issues.redhat.com/browse/MTV-2808


rgolangh:
@borod108 PTAL

PR #2096, opened 2025-06-26
VM-2774 | Deleted VM causes entire plans reconciliations to fail

Hazanel:
Issue:
Reconciliation of every plan randomly fails if there is a plan that's referring to an inexistent VM

Fix:
1. Set deleted VM status as canceled and propagate it up.
2. Ignore canceled VM's during plans schedule builder.

Ref: https://issues.redhat.com/browse/MTV-2774

### DIFF
--- a/pkg/controller/plan/controller.go
+++ b/pkg/controller/plan/controller.go
@@ -257,6 +257,9 @@ func (r Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (r
 	// The plan is updated as needed to reflect status.
 	result.RequeueAfter, err = r.execute(plan)
 	if err != nil {
+		if updateErr := r.updatePlanStatus(plan); updateErr != nil {
+			r.Log.Error(err, "failed to update plan status")
+		}
 		return
 	}
 
--- a/pkg/controller/plan/scheduler/vsphere/scheduler.go
+++ b/pkg/controller/plan/scheduler/vsphere/scheduler.go
@@ -5,12 +5,12 @@ import (
 	"errors"
 	"sync"
 
-	"github.com/kubev2v/forklift/pkg/controller/provider/web"
-
 	api "github.com/kubev2v/forklift/pkg/apis/forklift/v1beta1"
 	"github.com/kubev2v/forklift/pkg/apis/forklift/v1beta1/plan"
 	plancontext "github.com/kubev2v/forklift/pkg/controller/plan/context"
+	"github.com/kubev2v/forklift/pkg/controller/provider/web"
 	model "github.com/kubev2v/forklift/pkg/controller/provider/web/vsphere"
+	libcnd "github.com/kubev2v/forklift/pkg/lib/condition"
 	liberr "github.com/kubev2v/forklift/pkg/lib/error"
 )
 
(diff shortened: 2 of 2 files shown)


pkg/controller/plan/scheduler/vsphere/scheduler.go
```diff
@@ -117,9 +118,22 @@ func (r *Scheduler) buildInFlight() (err error) {
...
 		vm := &model.VM{}
 		err = r.Source.Inventory.Find(vm, vmStatus.Ref)
 		if err != nil {
+			if errors.As(err, &web.NotFoundError{}) {
+				vmStatus.SetCondition(libcnd.Condition{
+					Type:     api.ConditionCanceled,
+					Status:   libcnd.True,
+					Category: api.CategoryAdvisory,
+					Reason:   NotFound,
+					Message:  "VM was deleted from inventory.",
```
yaacov:
```suggestion
					Message:  "VM was not found in inventory.",
```

we know it's not found and we assume it was deleted, IMHO, best to state what we know and let the user assume it's not found because it was deleted

pkg/controller/plan/controller.go
```diff
@@ -257,6 +257,9 @@ func (r Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (r
 	// The plan is updated as needed to reflect status.
 	result.RequeueAfter, err = r.execute(plan)
 	if err != nil {
+		if err = r.updatePlanStatus(plan); err != nil {
```
yaacov:
```suggestion
		if updateErr = r.updatePlanStatus(plan); updateErr != nil {
```

in this case we don't actuall change the original err, this is good, we do want to return the `execute` error
but reading this code is confusing, a user may think we change the err, and will return the `updatePlanStatus` error

to make it easier to read i will chose a different var name for the `updatePlanStatus` error

a. It's immediately obvious that you're capturing a different error for the update operation
b. No risk of accidentally overwriting the original err in some future code update

pkg/controller/plan/scheduler/vsphere/scheduler.go
```diff
@@ -161,10 +175,7 @@ func (r *Scheduler) buildInFlight() (err error) {
 			vm := &model.VM{}
 			err = r.Source.Inventory.Find(vm, vmStatus.Ref)
 			if err != nil {
-				if errors.As(err, &web.NotFoundError{}) {
-					continue
-				}
-				if errors.As(err, &web.RefNotUniqueError{}) {
+				if errors.As(err, &web.NotFoundError{}) || errors.As(err, &web.RefNotUniqueError{}) {
```
yaacov: nitpick: i like the two ifs better, for me it's easier to read two ifs then an if with or ...

yaacov (approved):
3 nitpicks:
a. var naming
b. condition naming

pkg/controller/plan/controller.go
yaacov: Thank you for noticing, fixing suggestion, fiding `=` to` :=`

`if updateErr := r.updatePlanStatus(plan); updateErr != nil {`

PR #2101, opened 2025-06-29
Keep plus and spaces in vm name and convert to dash

yaacov:
Ref: https://issues.redhat.com/browse/MTV-2805

Plan with "+" within VM name replaced with " " as space in the Inventory

Issue:
We use URL decode on vSphere names, but we should use Path decode that preserve + and other URL signs

Before:
```
$ oc mtv inventory vms vsphere -q "where name ~= 'ameen'"
NAME                       ID          POWER       CPU         MEMORY      DISK USAGE  GUEST OS    CONCERNS (C/W/I)
mtv-func-rhel8-ameen//  a  vm-7415     Off         2           4.0 GB      24.0 GB                 0/2/0
```

After:
```
$ oc mtv inventory vms vsphere -q "where name ~= 'ameen'"
NAME                       ID          POWER       CPU         MEMORY      DISK USAGE  GUEST OS    CONCERNS (C/W/I)
mtv-func-rhel8-ameen//++a  vm-7415     Off         2           4.0 GB      24.0 GB                 0/2/0
```


Hazanel (approved):
lgtm

pkg/controller/plan/vm_name_handler.go
AmenB, line 1: Please also add "*" as it act the same as "+"

name: mtv-func-rhel8-****A
newName: mtv-func-rhel8-a

yaacov: Thanks :+1:  , fixing

yaacov: added a test for "*"

PR #2134, opened 2025-07-01
MTV-2839 | Add Dell PowerFlex copy-offload provider

rgolangh:
- xcopy-populator: use PersistentVolume type instead of volumeHandle
- **MTV-2839 | Add Dell PowerFlex copy-offload provider**

This work add Dell PowerFlex support and was tested with PowerFlex version Version 4.6.1.0 

https://issues.redhat.com/browse/MTV-2839


rgolangh:
- need to add 'naa.' prefix for the rest of the providers as the NAA string is being moved to be entirely controlled by the provider code. Main reason is that for powerflex the naa has a differnt prefix (because of the special device they create)
- need to clean up log messages 
- populator works well, but end to end with MTV is not yet tested
-

rgolangh:
> * need to add 'naa.' prefix for the rest of the providers as the NAA string is being moved to be entirely controlled by the provider code. Main reason is that for powerflex the naa has a differnt prefix (because of the special device they create)
> 
done
>     * need to clean up log messages
> 
done
>     * populator works well, but end to end with MTV is not yet tested

done

cmd/vsphere-xcopy-volume-populator/internal/populator/vib.go
```diff
@@ -24,6 +24,9 @@ var VibVersion = "x.x.x"
 // ensure vib will fetch the vib version and in case needed will install it
 // on the target ESX
 func ensureVib(client vmware.Client, esx *object.HostSystem, datastore string, desiredVibVersion string) error {
+	if true {
+		return nil
+	}
```
mnecas: @rgolangh :eyes:

cmd/vsphere-xcopy-volume-populator/internal/populator/remote_esxcli.go
```diff
@@ -126,16 +149,21 @@ func (p *RemoteEsxcliPopulator) Populate(vmId string, sourceVMDKFile string, vol
 
 	mappingContext, err := p.StorageApi.EnsureClonnerIgroup(xcopyInitiatorGroup, hbaUIDs)
 
+	mapTarget, ok := mappingContext[MapTargetContextKey]
+	if ok {
+		xcopyInitiatorGroup = mapTarget.(string)
```
mnecas: What if the mapTarget won't be found?

cmd/vsphere-xcopy-volume-populator/internal/populator/remote_esxcli.go
```diff
@@ -89,16 +91,37 @@ func (p *RemoteEsxcliPopulator) Populate(vmId string, sourceVMDKFile string, vol
...
+			continue
+		}
+		// powerflex handling - scini is the powerflex kernel module and is not
+		// using any iqn/wwn to identity the host. Instead extract the SdcGuid
+		// as the possible clonner identifier
+		if slices.Contains(driver, "scini") {
+			sciModule, err := p.VSphereClient.RunEsxCommand(context.Background(), host, []string{"system", "module", "parameters", "list", "-m", "scini"})
+			if err != nil {
+				// TODO skip, but print the error. Perhaps this handling is better suited per-vendor?
+				klog.Infof("failed to fetch the scini module parameters %s: ", err)
```
mnecas, line 110: hmm maybe add `continue`

cmd/vsphere-xcopy-volume-populator/internal/populator/remote_esxcli.go
```diff
@@ -89,16 +91,37 @@ func (p *RemoteEsxcliPopulator) Populate(vmId string, sourceVMDKFile string, vol
...
+			}
+			for _, v := range sciModule {
```
mnecas: NP: could you please be more verbose with the vars? not sure what is `v`

cmd/vsphere-xcopy-volume-populator/internal/powerflex/powerflex.go
```diff
@@ -0,0 +1,185 @@
...
+	}
+	return currentMappedSdcs, nil
+}
+
+// EnsureClonnerIgroup implements populator.StorageApi.
+func (p *PowerflexClonner) EnsureClonnerIgroup(initiatorGroup string, clonnerIqn []string) (populator.MappingContext, error) {
+	klog.Infof("ensuring initiator group %s for clonners %v", initiatorGroup, clonnerIqn)
+
+	mappingContext := make(map[string]any)
+	system, err := p.Client.FindSystem("", "", "")
```
mnecas, line 44: Looks for default?

cmd/vsphere-xcopy-volume-populator/internal/powerflex/powerflex.go
```diff
@@ -0,0 +1,185 @@
...
+		if sdc.OSType != "Esx" {
+			continue
+		}
+		klog.Infof("@@@@@@@@@@@ sdc %+v", sdc)
+		klog.Infof("@@@@@@@@@@@ clonnerIqn %+v", clonnerIqn)
+		klog.Infof("@@@@@@@@@@@ sdc.SdcGUID %+v", sdc.SdcGUID)
+		if slices.Contains(clonnerIqn, sdc.SdcGUID) {
+			klog.Infof("found SDC ID %+v", sdc)
+			mappingContext[populator.MapTargetContextKey] = sdc.ID
+			return mappingContext, nil
```
mnecas: Why do we return map if it will contain only one key?

mnecas:
generally lgtm just some questions and NP

cmd/vsphere-xcopy-volume-populator/internal/powerflex/powerflex.go
rgolangh: mapping context is a generic map construct for all the providers. Hitachi for example uses few keys internally.

cmd/vsphere-xcopy-volume-populator/internal/powerflex/powerflex.go
rgolangh: yes, I'm bit unsure about this one. I think that API call with empty strings is the way to get the default one

cmd/vsphere-xcopy-volume-populator/internal/populator/remote_esxcli.go
rgolangh: if not found then the xcoyInitiator group stays in tact. But now as I look on it again I realize I'll just push this logic into the powerflex Map call and handle it there. There is no reason to use it here, this is a provider specific usage.
The general rule I guess is that mapping context should be set and get in the provider code, and only be passed in the remote_esxcli.

cmd/vsphere-xcopy-volume-populator/internal/populator/vib.go
rgolangh: thanks! that is my usual kill switch for vib installation....

cmd/vsphere-xcopy-volume-populator/internal/powerflex/powerflex.go
mnecas: ah understand

cmd/vsphere-xcopy-volume-populator/internal/populator/remote_esxcli.go
mnecas: I won't lie, not exactly following, but I was mainly checking code smell
If you tell me there is good reason for it, I'm okay with it

cmd/vsphere-xcopy-volume-populator/internal/populator/remote_esxcli.go
rgolangh: ack

cmd/vsphere-xcopy-volume-populator/internal/populator/remote_esxcli.go
rgolangh: ack

PR #2170, opened 2025-07-03
MTV-2849 | Need to keep the default gateway setting after migration

Hazanel:
Issue:
Regression - default gateway settings are missing after a cold migration with preserve static ip.

The root cause is  the duplication routes being deleted. The remove duplications flow is:
1. Deletes all duplicates . 2.Tries to re-add the route using New-NetRoute.
3. If that fails, it falls back to the "route" command

If re-adding a route fails silently, the gateway route is simply gone, leading to:
1. No default gateway showing. 2.No connectivity despite a configured IP.

Fix:
Add InterfaceIndex to "route" command .
This removes all ambiguity and ensures:
1. The correct interface gets the route.
2. You avoid silent misrouting or ignored routes. 3.The default gateway is reliably assigned, even during early boot or interface flapping.

Ref: https://issues.redhat.com/browse/MTV-2849


Hazanel:
Destination  VM status:
image

image

yaacov (approved):

PR #2174, opened 2025-07-03
max add powermax

rgolangh:
- **xcopy-populator: use PersistentVolume type instead of volumeHandle**
- **MTV-2855 | Add Dell PowerMax**


cmd/vsphere-xcopy-volume-populator/internal/powermax/powermax.go
```diff
@@ -0,0 +1,288 @@
...
+		klog.Infof("group %s doesn't exist - create it", initiatorGroup)
+		_, err := p.client.CreateStorageGroup(ctx, p.symmetrixID, initiatorGroup, "none", "", true, nil)
+		if err != nil {
+			klog.Errorf("failed to create group %v ", err)
+			return nil, err
+		}
+	}
+
+	//change info message
+	klog.Infof("storage group %+v", sg)
```
shrihn-dell: Entire SG object dump not necessary. SG name would suffice I guess

cmd/vsphere-xcopy-volume-populator/internal/powermax/powermax.go
```diff
@@ -0,0 +1,288 @@
...
+		pg, err := p.client.GetPortGroupByID(ctx, p.symmetrixID, pgId)
+		if err != nil {
+			return nil, err
+		}
+		klog.Infof("port group protocol %+v", pg.PortGroupProtocol)
+		portGroupID = pg.PortGroupID
+	}
+
+	//change info message
+	klog.Infof("host group %+v", hg)
```
shrihn-dell: Logging host group name would suffice I guess

cmd/vsphere-xcopy-volume-populator/internal/powermax/powermax.go
```diff
@@ -0,0 +1,288 @@
...
+	klog.Infof("port groups %+v", pgs.PortGroupIDs)
+
+	portGroupID := ""
+	for _, pgId := range pgs.PortGroupIDs {
+		pg, err := p.client.GetPortGroupByID(ctx, p.symmetrixID, pgId)
+		if err != nil {
+			return nil, err
+		}
+		klog.Infof("port group protocol %+v", pg.PortGroupProtocol)
+		portGroupID = pg.PortGroupID
```
shrihn-dell: Filters port groups by protocol and then takes the last port group ID from the filtered list.  You need to consider port status otherwise you will end up selecting an offline port group. Add a check `if port.PortStatus != ONLINE` for all the ports in pg.SymmetrixPorts. A PG is viable if all the ports in the group are online.

cmd/vsphere-xcopy-volume-populator/internal/powermax/powermax.go
```diff
@@ -0,0 +1,288 @@
...
+				_, err := p.client.UpdateHostGroupHosts(ctx, p.symmetrixID, initiatorGroup, []string{host.HostID})
+				if err != nil {
+					return nil, err
+				}
+			}
+		}
+	}
+
+	protocolType := ""
+	for _, iqn := range clonnerIqn {
```
shrihn-dell, line 155: There is a possibility that you could have mixed initiator types which needs to be handled.

cmd/vsphere-xcopy-volume-populator/internal/powermax/powermax.go
```diff
@@ -0,0 +1,288 @@
...
+				if strings.HasSuffix(iqn, initiator) {
+					if !slices.Contains(hostIdsToAdd, host.HostID) {
+						hostIdsToAdd = append(hostIdsToAdd, host.HostID)
+					}
+				}
+			}
+		}
+	}
+
+	hg, err := p.client.GetHostGroupByID(ctx, p.symmetrixID, initiatorGroup)
```
shrihn-dell, line 127: A general comment. Add retries to all PowerMax REST API calls. Use 3-5 retry attempts with increasing delays.

rgolangh: is that an expected part of the API availability? is the CSI driver expected to do the same, and if so, is there already a wrapper for that behaviour?

cmd/vsphere-xcopy-volume-populator/internal/powermax/powermax.go
rgolangh: ack

cmd/vsphere-xcopy-volume-populator/internal/powermax/powermax.go
rgolangh: ack

rgolangh: there is no direct field named 'Name' on storage group. would that be the BaseSLOName?

cmd/vsphere-xcopy-volume-populator/internal/powermax/powermax.go
rgolangh: I guess only work with SCSI_FC and iSCSI protocols right?

rgolangh: also, I'll exit that loop if I have the porgroup with all the ports online, no need to proceed to the next iteration. Please confirm

rgolangh: regarding filtering, I already get the port group by protocol few lines above:

```
	protocolType := ""
	for _, iqn := range clonnerIqn {
		if strings.HasPrefix(iqn, "iqn") {
			protocolType = "iscsi"
		} else if strings.HasPrefix(iqn, "wwn") {
			protocolType = "fibre"
		}
	}

	pgs, err := p.client.GetPortGroupList(ctx, p.symmetrixID, protocolType)
```

it should be enough no?

cmd/vsphere-xcopy-volume-populator/internal/powermax/powermax.go
shrihn-dell: TBH, I don't know if CSI has a retry layer for U4P APIs. I presume three are, but I am not sure. May be Utkarsh who worked on CSI driver can comment. I feel it is a best practice to have retries for transient errors instead of failing the migration straight away.

cmd/vsphere-xcopy-volume-populator/internal/powermax/powermax.go
shrihn-dell: You can use storage group id. Same as name.

cmd/vsphere-xcopy-volume-populator/internal/powermax/powermax.go
shrihn-dell: > also, I'll exit that loop if I have the porgroup with all the ports online, no need to proceed to the next iteration. Please confirm

Yes, if all the ports are online then you have the port group and exit the loop.

PR #2184, opened 2025-07-08
MTV-6466 | Add IPV6 preserve static IP

Hazanel:
Add support for preserving IPv6 static addresses during Windows VM migration,
including IPv6-only and dual-stack (IPv4 + IPv6) configurations.

Controller (builder.go):
- Refactored mapMacStaticIps with device-based gateway matching for multi-NIC
- Added IPv6 gateway selection for global addresses
- Filter SLAAC and link-local addresses
- Sort networks IPv4-first for backward compatibility with virt-v2v

Windows firstboot scripts:
- Updated both PowerShell templates (standard + registry) to handle per-IP
  gateway and prefix instead of a single per-adapter PrefixLength
- Added default route configuration (New-NetRoute) for IPv4 and IPv6
- Added idempotency checks, error handling, and MAC normalization
- Fixed registry template to use hashtable fields ($ip.IPAddress, $ip.PrefixLength)

Customize (customize.go):
- formatIPs now emits PowerShell hashtables with IPAddress/Gateway/PrefixLength
- formatDNS filters blank entries
- Relaxed input parsing: DNS is optional (min 3 fields instead of 4)

Ref: https://issues.redhat.com/browse/MTV-3507


Hazanel:
@coderabbitai review

pkg/virt-v2v/customize/customize.go
```diff
@@ -53,7 +53,7 @@ func formatIPs(ips []IPEntry) string {
 	var b strings.Builder
 	b.WriteString("(\n")
 	for i, ip := range ips {
-		b.WriteString("'" + ip.IP + "'")
+		b.WriteString(fmt.Sprintf("    @{ IPAddress = '%s'; Gateway = '%s'; PrefixLength = %s }", ip.IP, ip.Gateway, ip.PrefixLength))
```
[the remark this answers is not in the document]
Hazanel: @coderabbitai  - we validate and control the input so it's not critical

Hazanel:
@coderabbitai review

Kyl-Bempah:
When https://github.com/kubev2v/forklift/pull/3359 is merged, pulling it into your branch may help with the virt-v2v failure

Hazanel:
@Kyl-Bempah, Thanks, I rebase anytime before pushing so I guess it will be reflected once it's merged

Hazanel:
@coderabbitai review

mnecas:
closing as stale, it has lot of potential and once we will have request around this we should take a look at this PR again. Right now i don't want to maintian additional code which we can't test

Hazanel:
Hey @coderabbitai - ptal

Hazanel:
Hey @coderabbitai 

P2 — ULA pushback :

ULA addresses (fc00::/7) are the IPv6 equivalent of private IPv4 ranges like 192.168.x.x or 10.x.x.x. We never filter those during migration, and ULA should be treated identically. The isNonGlobalIPv6() function exists for gateway selection only — selectGateway() uses it to avoid assigning default routes for ULA on Linux (where they'd be meaningless). The address itself is always preserved. Filtering ULA from shouldIncludeNetwork() would break VMs in private IPv6 networks.

please follow up and close resolved issues

Hazanel:
@coderabbitai - I meant check also the other issues you opened and if they were addressed to your satisfaction  please resolve them

Hazanel:
Thanks @coderabbitai for your review!
I attached the testing visual results in the ticket itself  - [IPv6_Results](https://redhat.atlassian.net/browse/MTV-3507?focusedCommentId=18054993)

mnecas (changes requested):
please remove the registry script, trying to reduce the scope of places where something can break. If needed we can expand it later 

can we test allso the complementry ips per nic? in these cases?

Hazanel:
@mnecas  - Just removed the registry
About the complementary - I actually tested that - our vm has multiple IP's per NI so the script was triggered

pkg/virt-v2v/customize/customize.go
mnecas, line 76: @coderabbitai is there some alternative i don't like formatting of the string in the controller and then replacing string inside the script could we use some go templates?

mnecas: @Hazanel wdyt?

Hazanel: @mnecas  -  That's quite a change but if it's important I'll do it

Hazanel: @coderabbitai - please check my changes according to you suggestion

Hazanel: @coderabbitai - V2V_staticIPs is an internal transport format ,the controller builds it from hypervisor inventory data that is already validated

mnecas (approved):
yeah looks much better thanks!
lgtm

PR #2223, opened 2025-07-17
MTV-2799: Fetch vsphere domain xml from libvirt

jonner:
Rather than generating the libvirt xml ourselves, we should use libvirt to connect to vsphere and generate a domain xml document for the vm. This will provide the most complete domain description possible and give virt-v2v the best chance to convert the disks successfully.

The generated libvirt xml will obviously have the disk paths to the vsphere data store locations which are not accesible to the virt-v2v pod, so we still need to generate our own disks xml which point to the imported PVCs, but the rest of the xml will be what was generated by the libvirt vmware driver.

## Summary by CodeRabbit

* **New Features**
  * In-place conversions now fetch and adjust a VM's libvirt domain XML so disk paths are correct before conversion.
  * New CLI option to provide an access key path for vSphere integrations; default libvirt domain file moved to /tmp/input.xml.
  * Added bundled libvirt support enabling richer libvirt operations and event handling.

* **Bug Fixes**
  * Replaced fragile config-map based domain XML handling with a writable temporary volume for more robust in-place flows.
  * Improved error reporting around libvirt domain retrieval.

* **Chores**
  * CI/build pipelines and builder images updated to install libvirt dependencies and toolchain components.

### DIFF
--- a/build/virt-v2v/Containerfile-upstream
+++ b/build/virt-v2v/Containerfile-upstream
@@ -1,6 +1,7 @@
 # Build virt-v2v binary
-FROM registry.access.redhat.com/ubi9/go-toolset:1.24.4-1753221510 AS builder
+FROM quay.io/centos/centos:stream9 AS builder
 USER 0
+RUN --mount=type=cache,target=/var/cache/dnf dnf -y install --enablerepo=crb libvirt-devel golang
 WORKDIR /app
 COPY --chown=1001:0 ./ ./
 ENV GOFLAGS "-mod=vendor -tags=strictfipsruntime"
--- a/cmd/virt-v2v/entrypoint.go
+++ b/cmd/virt-v2v/entrypoint.go
@@ -42,7 +42,22 @@ func main() {
 	} else {
 		// virt-v2v or virt-v2v-in-place
 		if convert.IsInPlace {
-			err = convert.RunVirtV2vInPlace()
+			// fetch xml description of the guest from libvirt to help virt-v2v make the conversion
+			err = func() error {
+				domainXML, err := convert.GetDomainXML()
+				if err != nil {
+					return fmt.Errorf("failed to get domain XML: %v", err)
+				}
+
+				err = os.WriteFile(convert.LibvirtDomainFile, []byte(domainXML), 0644)
+				if err != nil {
+					return fmt.Errorf("failed to write domain XML file: %v", err)
+				}
+				return nil
+			}()
+			if err == nil {
+				err = convert.RunVirtV2vInPlace()
+			}
 		} else {
 			err = convert.RunVirtV2v()
 		}
--- a/pkg/controller/plan/kubevirt.go
+++ b/pkg/controller/plan/kubevirt.go
@@ -3,7 +3,6 @@ package plan
 import (
 	"context"
 	"encoding/json"
-	"encoding/xml"
 	"errors"
 	"fmt"
 	"io"
@@ -50,7 +49,6 @@ import (
 	instancetypeapi "kubevirt.io/api/instancetype"
 	instancetype "kubevirt.io/api/instancetype/v1beta1"
 	cdi "kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1"
-	libvirtxml "libvirt.org/libvirt-go-xml"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	k8sutil "sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
@@ -995,14 +993,6 @@ func (r *KubeVirt) EnsureVirtV2vPod(vm *plan.VMStatus, vmCr *VirtualMachine, pvc
 		return
 	}
 
-	var libvirtConfigMap = &core.ConfigMap{}
-	if podType == VirtV2vConversionPod {
-		libvirtConfigMap, err = r.ensureLibvirtConfigMap(vm.Ref, vmCr, pvcs)
-		if err != nil {
-			return
-		}
-	}
-
 	var vddkConfigMap *core.ConfigMap
 	if r.Source.Provider.UseVddkAioOptimization() {
 		vddkConfigMap, err = r.ensureVddkConfigMap()
@@ -1016,7 +1006,7 @@ func (r *KubeVirt) EnsureVirtV2vPod(vm *plan.VMStatus, vmCr *VirtualMachine, pvc
 	if podType == VirtV2vConversionPod {
 		vmVolumes = vmCr.Spec.Template.Spec.Volumes
 	}
-	newPod, err := r.getVirtV2vPod(vm, vmVolumes, libvirtConfigMap, vddkConfigMap, pvcs, v2vSecret, podType)
+	newPod, err := r.getVirtV2vPod(vm, vmVolumes, vddkConfigMap, pvcs, v2vSecret, podType)
 	if err != nil {
 		return
 	}
@@ -1955,8 +1945,8 @@ func (r *KubeVirt) getConvertorAffinity() *core.Affinity {
 	}
 }
 
-func (r *KubeVirt) getVirtV2vPod(vm *plan.VMStatus, vmVolumes []cnv.Volume, libvirtConfigMap *core.ConfigMap, vddkConfigmap *core.ConfigMap, pvcs []*core.PersistentVolumeClaim, v2vSecret *core.Secret, podType int) (pod *core.Pod, err error) {
-	volumes, volumeMounts, volumeDevices, err := r.podVolumeMounts(vmVolumes, libvirtConfigMap, vddkConfigmap, pvcs, vm, podType)
+func (r *KubeVirt) getVirtV2vPod(vm *plan.VMStatus, vmVolumes []cnv.Volume, vddkConfigmap *core.ConfigMap, pvcs []*core.PersistentVolumeClaim, v2vSecret *core.Secret, podType int) (pod *core.Pod, err error) {
+	volumes, volumeMounts, volumeDevices, err := r.podVolumeMounts(vmVolumes, vddkConfigmap, pvcs, vm, podType)
 	if err != nil {
 		return
 	}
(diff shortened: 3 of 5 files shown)


build/forklift-controller/Containerfile-downstream
```diff
@@ -1,3 +1,4 @@
+# FIXME: use a base image that includes libvirt-devel
 FROM registry.redhat.io/ubi9/go-toolset:1.23.6-1745588370 AS builder
```
jonner: @solenoci here is the downstream containerfile we were discussing where we will need libvirt packages installed into the ubi9 base image

solenoci: @jonner Hey, the packages were added to the controller and the changes are merged to main.

jonner:
> Would it be possible to move this from controller to v2v conversion pod?

Sorry, I probably should have mentioned this in the PR. I was intending to move it to the v2v conversion pod, but the domain xml that we get from libvirt contains the disk sources with paths that reference the vmware storage location, and we need to convert those to the imported PVC paths. We don't have the disk PVC map in the v2v pod, so it seemed better to keep it in the controller pod.

jonner:
@coderabbitai full review

jonner:
I've pushed an updated branch. As I mentioned in my comment above, the libvirt usage couldn't really be moved to the conversion pod because of the due to the disk-to-pvc mapping issue. This new branch does contain a bit more handling for cases where libvirt is not available, and where the libvirt version is too old to support the 'cacert' url parameter. That functionality was introduced in libvirt 11.5.0, which is not available in rhel9. So currently, fetching the domain xml from libvirt will only happen if no ca cert is specified for the provider (i.e. 'skip certification validation' is checked). In the future as the base ubi image is updated to versions with newer libvirt, it will be used for all providers.

yaacov:
Sorry for the layman question, i am probably missing somthing, don't we want kubevirt output instead of ovirt output ?
( e.g. `virt-v2v -o kubevirt `instead of `virt-v2v -o ovirt` )

jonner:
> Sorry for the layman question, i am probably missing somthing, don't we want kubevirt output instead of ovirt output ? ( e.g. `virt-v2v -o kubevirt `instead of `virt-v2v -o ovirt` )

I'm not sure I understand. I don't see anything like that in the repository or in this PR.

jonner:
@yaacov see above

yaacov:
> @yaacov see above

Since we are migrating into kubevirt and not ovirt [1], does it make sense to parse the libvirt xml, when we can get the kubevirt yaml [2] and parse it using go struct ? what am i missing ?

[1] https://libguestfs.org/virt-v2v.1.html // look for -o kubevirt
[2] https://github.com/kubev2v/forklift/blob/main/pkg/virt-v2v/conversion/conversion.go#L133

mrnold:
> Since we are migrating into kubevirt and not ovirt [1], does it make sense to parse the libvirt xml, when we can get the kubevirt yaml [2] and parse it using go struct ? what am i missing ?
> 
> [1] https://libguestfs.org/virt-v2v.1.html // look for -o kubevirt [2] https://github.com/kubev2v/forklift/blob/main/pkg/virt-v2v/conversion/conversion.go#L133

I don't think there's any KubeVirt YAML to parse at this stage, the VM hasn't been created yet. This is using libvirt independent of any oVirt stuff to connect to VMware (https://libvirt.org/drvesx.html) and turn VMware's VM definition into libvirt's XML format to give to virt-v2v as an input. That way virt-v2v can make better decisions about how to convert the disk contents, and then forklift can create the destination VM.

mrnold:
I have no idea about any of the build or vendoring stuff, but the actual XML part looks okay to me.

jonner:
@yaacov maybe I got confused by the reference to `-o ovirt`, which afaict is not used anywhere in the codebase. Were you actually talking about `-o libvirt`?

The code that you are pointing to (with the `-o kubevirt` option set) is when we're converting a guest with the `virt-v2v` binary. `virt-v2v` will connect to the source provider and copy the disk, convert it, and output it to a new directory. But in the code path that is used in this patch, we're converting a guest with the `virt-v2v-in-place` binary. The disks have already been copied from the source provider to the destination provider. `virt-v2v-in-place` does not have a `-o kubevirt` option because we're modifying the local guest disks in place rather than copying them and generating new disks in an output directory.

mnecas:
> > Would it be possible to move this from controller to v2v conversion pod?
> 
> Sorry, I probably should have mentioned this in the PR. I was intending to move it to the v2v conversion pod, but the domain xml that we get from libvirt contains the disk sources with paths that reference the vmware storage location, and we need to convert those to the imported PVC paths. We don't have the disk PVC map in the v2v pod, so it seemed better to keep it in the controller pod.

we have the disks within the v2v pod, I have created abstraction around it aswell where we can point to the disk itself if it's block device or file or we create symlinks for the virt-v2v output
https://github.com/kubev2v/forklift/blob/29b60e21c388420e6e0e79a55802c2c0a201ab45/pkg/virt-v2v/conversion/conversion.go#L19
https://github.com/kubev2v/forklift/blob/29b60e21c388420e6e0e79a55802c2c0a201ab45/pkg/virt-v2v/conversion/disk.go#L22

build/virt-v2v/Containerfile-upstream
```diff
@@ -1,6 +1,7 @@
 # Build virt-v2v binary
-FROM registry.access.redhat.com/ubi9/go-toolset:1.24.4-1753221510 AS builder
+FROM quay.io/centos/centos:stream9 AS builder
 USER 0
+RUN --mount=type=cache,target=/var/cache/dnf dnf -y install --enablerepo=crb libvirt-devel golang
 WORKDIR /app
```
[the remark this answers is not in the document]
jonner: The version of go in centos stream 9 is actually 1.25.x, not 1.22.x

cmd/virt-v2v/entrypoint.go
```diff
@@ -42,7 +42,22 @@ func main() {
...
+				if err != nil {
+					return fmt.Errorf("failed to get domain XML: %v", err)
+				}
+
+				err = os.WriteFile(convert.LibvirtDomainFile, []byte(domainXML), 0644)
+				if err != nil {
+					return fmt.Errorf("failed to write domain XML file: %v", err)
+				}
+				return nil
+			}()
```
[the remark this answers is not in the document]
jonner: personally, I find the anonymous function makes things slightly more clear because it allows me to treat getting the xml and writing it to file as a single fallible operation. If that operation succeeds, I can run the virt-v2v command without nesting the virt-v2v command inside multiple if statements.

jonner:
> > > Would it be possible to move this from controller to v2v conversion pod?
> > 
> > 
> > Sorry, I probably should have mentioned this in the PR. I was intending to move it to the v2v conversion pod, but the domain xml that we get from libvirt contains the disk sources with paths that reference the vmware storage location, and we need to convert those to the imported PVC paths. We don't have the disk PVC map in the v2v pod, so it seemed better to keep it in the controller pod.
> 
> we have the disks within the v2v pod, I have created abstraction around it aswell where we can point to the disk itself if it's block device or file or we create symlinks for the virt-v2v output
> 
> https://github.com/kubev2v/forklift/blob/29b60e21c388420e6e0e79a55802c2c0a201ab45/pkg/virt-v2v/conversion/conversion.go#L19
> 
> https://github.com/kubev2v/forklift/blob/29b60e21c388420e6e0e79a55802c2c0a201ab45/pkg/virt-v2v/conversion/disk.go#L22

My latest patchset moves the libvirt conversion down to the virt-v2v pod and uses the disk array as you suggested.

pkg/virt-v2v/conversion/conversion.go
```diff
@@ -255,3 +257,110 @@ func (c *Conversion) addVirtV2vRemoteInspectionArgs(cmd utils.CommandBuilder) (e
...
+		return "", fmt.Errorf("failed to parse libvirt URL: %w", err)
+	}
+
+	usernameData, err := os.ReadFile("/etc/secret/accessKeyId")
+	if err != nil {
+		return "", fmt.Errorf("failed to read username from secret: %w", err)
+	}
+	username := string(usernameData)
+
+	passwordData, err := os.ReadFile("/etc/secret/secretKey")
```
mnecas: Please use `SecretKey`

pkg/virt-v2v/conversion/conversion.go
```diff
@@ -255,3 +257,110 @@ func (c *Conversion) addVirtV2vRemoteInspectionArgs(cmd utils.CommandBuilder) (e
...
+
+// retrieve and modify the domain XML from libvirt
+func (c *Conversion) GetDomainXML() (string, error) {
+	libvirtURL, err := url.Parse(c.LibvirtUrl)
+
+	if err != nil {
+		return "", fmt.Errorf("failed to parse libvirt URL: %w", err)
+	}
+
+	usernameData, err := os.ReadFile("/etc/secret/accessKeyId")
```
mnecas: Please create var for this

pkg/controller/plan/kubevirt.go
```diff
@@ -2253,21 +2242,16 @@ func (r *KubeVirt) podVolumeMounts(vmVolumes []cnv.Volume, libvirtConfigMap *cor
 	}
 
 	if podType == VirtV2vConversionPod {
-		// add volume and mount for the libvirt domain xml config map.
-		// the virt-v2v pod expects to see the libvirt xml at /mnt/v2v/input.xml
+		// Create a writable directory for virt-v2v to generate domain XML files for vSphere migrations
```
mnecas: Think you could write to /tmp/

jonner: changed

pkg/virt-v2v/conversion/conversion.go
jonner: done

pkg/virt-v2v/conversion/conversion.go
jonner: done

jonner:
> lgtm, @jonner could you please sync with @solenoci I'm bit worried about the d/s v2v build. If I remember correctly this will require specific libvirt so we have all correct headers.

Indeed. I don't know much about how the downstream builds are done, so I'm not sure exactly how to make the necessary changes. As referenced in the very first comment on this thread, I believe @solenoci already added the libvirt dependency to the forklift-controller base image. But with the changes that @mnecas requested and which I implemented in the latest version of this patch, the forklift-controller image no longer needs the libvirt dependency and the virt-v2v container needs it instead.

@solenoci please let me know if you need to me to do anything to make things happen here.

rwmjones:
Isn't it done here? https://github.com/kubev2v/forklift/tree/main/build/virt-v2v

jonner:
Yes, but as far as I can recall, the libvirt development packages are not available to install directly in the ubi image, so will need to do something like https://github.com/kubev2v/forklift/commit/cde1933f37c0eec07ddbeab80f3afde18bdb9214 to enable libvirt in the base image.

solenoci:
@jonner Hi, do you have the resources/capacity to try the migration functionality with these images please?

controller_image_fqin: 'quay.io/redhat-user-workloads/rh-mtv-1-tenant/forklift-operator-dev-preview/forklift-controller-dev-preview:on-pr-e3228053924736792884ca0fadddd035026cbd84'

virt_v2v_image_fqin: 'quay.io/redhat-user-workloads/rh-mtv-1-tenant/forklift-operator-dev-preview/virt-v2v-dev-preview:on-pr-e3228053924736792884ca0fadddd035026cbd84'

jonner:
> @jonner Hi, do you have the resources/capacity to try the migration functionality with these images please?
> 
> controller_image_fqin: 'quay.io/redhat-user-workloads/rh-mtv-1-tenant/forklift-operator-dev-preview/forklift-controller-dev-preview:on-pr-e3228053924736792884ca0fadddd035026cbd84'
> 
> virt_v2v_image_fqin: 'quay.io/redhat-user-workloads/rh-mtv-1-tenant/forklift-operator-dev-preview/virt-v2v-dev-preview:on-pr-e3228053924736792884ca0fadddd035026cbd84'

Basic smoke test looks fine with those images.

PR #2237, opened 2025-07-21
build: Remove LIBGUESTFS_BACKEND=direct everywhere

rwmjones:
This should *never* be used.  It causes libguestfs to use an untested code path.  It also stops libvirt from containerizing the qemu subprocess, reducing security.

Wherever this is seen, it should be removed.  If doing so causes a problem (which is unlikely as it puts us back on the tested code path) then file a detailed bug about the problem, but do not reintroduce the environment variable.

### DIFF
--- a/build/virt-v2v/Containerfile
+++ b/build/virt-v2v/Containerfile
@@ -1,7 +1,5 @@
 FROM registry.access.redhat.com/ubi9:9.5-1747219013 AS appliance
 
-ENV LIBGUESTFS_BACKEND=direct
-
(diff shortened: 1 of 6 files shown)


rwmjones:
Martin pointed out that unfortunately this fails to build with:
```
libvirt: XML-RPC error : Failed to connect socket to '/var/run/libvirt/virtqemud-sock': No such file or directory
```
We're currently looking at how to fix this.

rwmjones:
See https://github.com/kubev2v/forklift/pull/2242

PR #2238, opened 2025-07-21
build: virt-v2v: Simplify package download

rwmjones:
See individual commit messages for description.

I was only partially able to test this. Download worked (which is the main point of the change), but installation failed because of some broken dependencies.


rwmjones:
The update is minor, just rebasing on top of current `main` branch and a bit of tidy-up.

PR #2245, opened 2025-07-22
MTV-2703 | Migrate VM with nvme direct attached disk, the plan hang

Hazanel:
Issue:
NVME disk type isn't supported , hence a migration plan will hang during disk allocation.

Fix:
Add validation to fail plan when unsupported disks (e.g., NVMe) are used

This adds a validation step that checks for unsupported disk types during plan creation. If such disks (currently NVMe) are detected and the user proceeds despite existing warnings, the plan will now explicitly fail to prevent unsupported migrations.

Ref: https://issues.redhat.com/browse/MTV-2703


mrnold:
The validation messages look good, but if you proceed anyway does the plan still hang? I just wanted to make sure I'm not missing anything.

Hazanel:
@mrnold  - It won't even start...
image

yaacov (approved):

PR #2259, opened 2025-07-23
MTV-2894 | OCP->OCP migrate VM with Pre/Post migration hooks, plan failed in Initialize phase.

Hazanel:
Issue:
    OCP->OCP migrate VM with Pre/Post migration hooks, plan failed in Initialize phase, with the error below:
    
    "invalid resource name "namespace/vm-name": [may not contain '/'], meaning that vm name was concatenated from    namespace and vm-name with / which is illegal.
    
    Fix:
    keep the vm-name as is.
    
    Ref: https://issues.redhat.com/browse/MTV-2894


mnecas (approved):
lgtm

PR #2260, opened 2025-07-23
MTV-2810 | Migrate VM from OCP to OCP hang

Hazanel:
Issue:
OCP vm name with dots causes the migration to hang.

Fix:
Add additional validations to inform users during plan creation that OCP VMs names with dots are not supported.

Ref: https://issues.redhat.com/browse/MTV-2810

### DIFF
--- a/pkg/controller/plan/adapter/base/doc.go
+++ b/pkg/controller/plan/adapter/base/doc.go
@@ -161,6 +161,8 @@ type Validator interface {
 	VMMigrationType(vmRef ref.Ref) (bool, error)
 	// Validate that the VM disks are supported.
 	UnSupportedDisks(vmRef ref.Ref) ([]string, error)
+	// Validate that the VM name is valid.
+	ValidVmName(vmRef ref.Ref) (bool, error)
 }
 
 // DestinationClient API.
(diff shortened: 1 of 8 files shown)


mnecas (commented):
lgtm, but @yaacov could you please take a look aswell?

yaacov:
https://issues.redhat.com/browse/MTV-2810 is about ocp to ocp hangs on valid names ( dots are allowed in ocp+cnv )

my concern is that after this fix, we will not be able to migrate vms with valid names.

@mnecas @Hazanel what is blocking us from fixing this issue by allowing vms with valid names to successfully migrate ?
is the problem on our side ? should we open a bug on a tool we relay on for successful migration ?

Hazanel:
@yaacov There is an open issue on CNV side [CNV-60965](https://issues.redhat.com/browse/CNV-60965) which is the root cause.

yaacov:
> @yaacov There is an open issue on CNV side [CNV-60965](https://issues.redhat.com/browse/CNV-60965) which is the root cause.

lets be verbose about the root cause in our JIRA ticket and in our code, that this is temporary fix and we will remove it once https://issues.redhat.com/browse/CNV-60965 is resolved, maybe even point to the CNV issue in the message we show the user, @mnecas @Hazanel  WDYT ?

Hazanel:
CNV team response : 
	
It's merged and backported to 4.18.z. Next 4.18 should have it, so I'd say it's not worth it (and it's not MTV's problem).
I think we can close this PR

Hazanel:
This is a CNV issue which is fixed and soon to be ready.

PR #2300, opened 2025-07-24
MTV-3025 | Add TargetPowerState to migration plan CR

yaacov:
Ref:
https://issues.redhat.com/browse/MTV-3025

Issue:
Currently there is no user controlled method of ensuring the desired power state of a VM Post Migration.

Per [MTV-811](https://issues.redhat.com/browse/MTV-811), it is currently determined by the power state of the VM on the source environment.  There are circumstances where a user would like to migrate a VM that is running on the source provider, but ensure the VM is powered off post migration in order to preserve start up dependencies between VMs.   Adding this as an option to the plan would be useful for these circumstances

Fix:
  - On plan level:
  Add `TargetPowerState` , `TargetPowerState` specifies the desired power state of all target VM after migration.
      - "on": Target VM will be powered on after migration
      - "off": Target VM will be powered off after migration
      - "auto" or nil (default): Target VM will match the source VM's power state
  - On sngle VM level:
  Add `TargetPowerState` , `TargetPowerState` specifies the desired power state of the target VM after migration.
      - "on": Target VM will be powered on after migration
      - "off": Target VM will be powered off after migration
      - "auto" or nil (default): Target VM will match the source VM's power state

## Summary by CodeRabbit

* **New Features**
  * Added an option to specify the desired power state ("on", "off", or "auto") for target virtual machines after migration. If set to "auto" or left unset, the target VM will match the source VM's power state.
  * Enhanced migration plans and VM specifications to support this new power state setting.

* **Documentation**
  * Updated descriptions to explain the new power state options and their behavior.

### DIFF
--- a/operator/.downstream_manifests
+++ b/operator/.downstream_manifests
@@ -943,6 +943,17 @@ spec:
                         If not provided, the original VM name will be used and automatically adjusted to meet k8s DNS1123 requirements.
                         If provided, this exact name will be used instead. The migration will fail if the name is not unique or already in use.
                       type: string
+                    targetPowerState:
+                      description: |-
+                        TargetPowerState specifies the desired power state of the target VM after migration.
+                        - "on": Target VM will be powered on after migration
+                        - "off": Target VM will be powered off after migration
+                        - "auto" or nil (default): Target VM will match the source VM's power state
+                      enum:
+                      - "on"
+                      - "off"
+                      - auto
+                      type: string
                     type:
                       description: Type used to qualify the name.
                       type: string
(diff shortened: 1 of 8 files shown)


mnecas:
please run `make manifests generate-manifests kustomized-manifests` but LGTM!

yaacov:
@mnecas addressed review by rabbit , run `make manifests generate-manifests kustomized-manifests`

operator/.downstream_manifests
```diff
@@ -2840,12 +2856,17 @@ spec:
                 type: object
                 x-kubernetes-map-type: atomic
               type:
+                allOf:
```
mnecas: First time seeing the `allOf` think it's from the `type MigrationType string`
does it have some side affect? @mansam wdyt?

yaacov: That file is auto generated by `make manifests`, yes AFAIU it's from the type, Rabbit thought it is nicer to use a type instead of string and I agreed with it .
AFAIU it's just syntactic sugar to help the compiler, IDEs and linters keep the enum safe.

mansam: It seems like the tool is generating the CRD wrong, since it's duplicating the enum (and then having to add `allOf`). There should only be one schema to validate against.

I'm not a huge fan of aliasing strings like that because it doesn't buy very much (any string can be cast to one, so it doesn't really protect the field) and tends to get in the way, but it's ultimately harmless.

mansam: I suspect that if you remove the kubebuilder enum annotation then it will figure it out from the type and get rid of the duplication.

yaacov: @mansam thank you!, removed the annotation and it worked as expected and the duplication removed :+1:

yaacov:
@mnecas I implemented @mansam suggested fix, and the duplication was removed, please re review

PR #2311, opened 2025-07-25
MTV-3035 | Switch MTV LIBGUESTFS_BACKEND to libvirt

mnecas:
Issue:
The LIBGUESTFS_BACKEND=direct is not tested and not supported path, we
need to switch to LIBGUESTFS_BACKEND=libvirt:qemu:///session.

This switch caused multiple issues:
- https://issues.redhat.com/browse/RHEL-105490
- https://issues.redhat.com/browse/RHEL-105529

Both of them are caused by running the libguestfs inside container.

For RHEL-105490 we need to set home directly, becuase QEMU users as
which this container is running does not have permissions to write to
the root directory. We will create home directory for that user.

For RHEL-105529 the virt-v2v had problem cleaning up the libguestfs VMs
in the contianer. This is cause as there is not init process PID 1 which
would cleanup the zombie processes. For this we need to use some init
process such as `catatonit` and run it before the virt-v2v.

## Summary by CodeRabbit

* **Chores**
  * Improved container environment by setting a dedicated home directory to prevent permission errors.
  * Introduced a process management tool to ensure proper cleanup and prevent hangs during operations.
  * Updated backend connection method to use the officially supported libvirt session for enhanced stability.

### DIFF
--- a/build/virt-v2v/Containerfile
+++ b/build/virt-v2v/Containerfile
@@ -12,9 +12,17 @@ RUN subscription-manager refresh && \
 # LIBGUESTFS_BACKEND=direct is required to work around the following bug:
 # https://issues.redhat.com/browse/RHEL-104684
 # and it can be removed when that bug gets fixed.
+#
+# It is also required during contianer runtime to work around the following bug:
+# https://issues.redhat.com/browse/RHEL-105490 workaround is to set ENV HOME=/tmp
+#
+# Additional issue is that the virt-v2v will get stuck when destroying domain
+# https://issues.redhat.com/browse/RHEL-105529
+ENV LIBGUESTFS_BACKEND=direct
+
 RUN mkdir -p /usr/local/lib/guestfs/appliance && \
     cd /usr/local/lib/guestfs/appliance && \
-    LIBGUESTFS_BACKEND=direct libguestfs-make-fixed-appliance . && \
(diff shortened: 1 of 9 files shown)


rwmjones:
Let's NOT do this!  Let's fix the problem with libvirt.  I believe there's a simple workaround. https://issues.redhat.com/browse/RHEL-105490?focusedId=27664749&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-27664749

mnecas:
@rwmjones updated the PR with the `ENV HOME` and with the `catatonit`. Could you please take a look?

build/virt-v2v/Containerfile
```diff
@@ -45,6 +45,23 @@ RUN subscription-manager refresh && \
         virtio-win && \
     dnf clean all
 
+# This prevents libvirt from attempting to create files in the root directory,
+# which would lead to permission errors.
+# Refer to: https://issues.redhat.com/browse/RHEL-105490
+RUN mkdir /home/qemu && chown qemu:qemu /home/qemu
+ENV HOME=/home/qemu
+
+# The 'virt-v2v' tool can hang when destroying libguestfs domains due to the
```
rwmjones: libvirt, not virt-v2v ...

rwmjones:
Yup, looks fine, thanks.

PR #2348, opened 2025-07-28
MTV-2384 | Migration failed with unexpected error "Another task is already in progress"

Hazanel:
Issue:
- The Plan controller performed multiple status updates in same reconcile loop
- Race conditions between validation updates and execution updates
- "Operation cannot be fulfilled: object has been modified" errors
- Failed phase transitions led to duplicate snapshot creation attempts
- DNS1123 invalid VM names particularly affected due to additional validation updates

**Root Cause:**
Multiple controller instances reading stale Plan ResourceVersions:
1. Controller A: validation update (Plan v100 → v101)
2. Controller B: reads stale Plan v100, tries update → CONFLICT
3. Controller A: execution update → potential conflict
4. Failed phase transitions caused re-execution of snapshot creation phases

Fix:
1. Proactive conflict prevention: Always fetch latest Plan before status update
2. Resilient fallback: If fetch fails, attempt update with current version
3. Controller-level duplicate prevention: Check existing precopy tasks before snapshot creation
4. vSphere-level duplicate detection: Detect running CreateSnapshot_Task and reuse task ID

Ref: https://issues.redhat.com/browse/MTV-2384

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved snapshot creation reliability by preventing duplicate snapshot tasks on virtual machines.  
  * Enhanced plan status updates to avoid conflicts by ensuring updates use the latest resource version.
* **Style**
  * Removed unnecessary logging statements for cleaner output.

### DIFF
--- a/pkg/controller/plan/adapter/vsphere/client.go
+++ b/pkg/controller/plan/adapter/vsphere/client.go
@@ -28,9 +28,10 @@ import (
 )
 
 const (
-	snapshotName = "forklift-migration-precopy"
-	snapshotDesc = "Forklift Operator warm migration precopy"
-	taskType     = "Task"
+	snapshotName           = "forklift-migration-precopy"
+	snapshotDesc           = "Forklift Operator warm migration precopy"
+	taskType               = "Task"
+	createSnapshotTaskName = "CreateSnapshot_Task"
 )
 
 // vSphere VM Client
@@ -47,12 +48,62 @@ func (r *Client) CreateSnapshot(vmRef ref.Ref, hostsFunc util.HostsFunc) (snapsh
 	if err != nil {
 		return
 	}
+
+	// Check if there's already a running CreateSnapshot task - prevents duplicates
+	if existingTaskId := r.findRunningSnapshotTask(vmRef, vm, hostsFunc); existingTaskId != "" {
+		return "", existingTaskId, nil
+	}
+
 	task, err := vm.CreateSnapshot(context.TODO(), snapshotName, snapshotDesc, false, true)
 	if err != nil {
 		err = liberr.Wrap(err)
 		return
 	}
-	return "", task.Reference().Value, nil
+
+	creationTaskId = task.Reference().Value
+	return "", creationTaskId, nil
+}
+
+// Check if there's already a running CreateSnapshot task on the VM to prevent duplicates
+func (r *Client) findRunningSnapshotTask(vmRef ref.Ref, vm *object.VirtualMachine, hosts util.HostsFunc) string {
+	// Get the ESXi client
+	client, err := r.getClientFromVmRef(vmRef, hosts)
+	if err != nil {
+		return ""
+	}
+
+	// Create property collector
+	pc := property.DefaultCollector(client)
+	pc, err = pc.Create(context.TODO())
+	if err != nil {
+		return ""
+	}
+	//nolint:errcheck
+	defer pc.Destroy(context.TODO())
+
+	// Get VM recent tasks
+	var vmObj mo.VirtualMachine
+	err = pc.RetrieveOne(context.TODO(), vm.Reference(), []string{"recentTask"}, &vmObj)
+	if err != nil {
+		return ""
+	}
+
+	// Check for running CreateSnapshot tasks
+	for _, taskRef := range vmObj.RecentTask {
+		var task mo.Task
+		err = pc.RetrieveOne(context.TODO(), taskRef, []string{"info"}, &task)
+		if err != nil {
+			continue
+		}
+
+		// Check if this is a running CreateSnapshot task
+		if task.Info.Name == createSnapshotTaskName &&
+			(task.Info.State == types.TaskInfoStateRunning || task.Info.State == types.TaskInfoStateQueued) {
+			return taskRef.Value
+		}
+	}
+
+	return ""
 }
 
 // Remove a VM snapshot.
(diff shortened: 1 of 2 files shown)


Hazanel:
Why We Must Wait for Task Completion When Creating a Snapshot of a VM with an Invalid DNS-1123 Name
When creating a snapshot of a virtual machine in vSphere, the platform returns a Task object that must be monitored to determine when (and whether) the operation completes. Under normal conditions, polling the task until it's marked as success or error is a best practice. However, this becomes critical in edge cases — particularly when the VM name does not conform to DNS-1123 standards.

🧱 What’s the Problem?
VM names that contain invalid characters (e.g., capital letters, underscores, or special symbols) break DNS-1123 compliance. While vSphere allows such names, they can interfere with internal assumptions in management layers or third-party tools.

During snapshot creation for such a VM:

The CreateSnapshot_Task might complete successfully.

But the resulting snapshot ID is not immediately accessible via the task’s .Result field.

Worse: if you proceed too early, vSphere will reject property queries (like listing snapshot.rootSnapshotList) with obscure internal SOAP errors related to malformed ManagedObjectReference parsing.

🧪 What We Observed
In these cases, if we:

Immediately query the VM (e.g., for its snapshot tree) after checking taskInfo.Result,

We sometimes get an error like: "Another task is already in progress"


while parsing MoRef for ManagedObject of type vmodl.ManagedObject
This happens because:

vSphere may still be internally registering the snapshot,

Even though the task appears Success, the snapshot isn’t fully committed to inventory.

Why WaitForTaskCompletion Fixes It
By explicitly waiting for the task ID (using getTaskById) to:

Reach TaskInfoStateSuccess, and

Remain stable for a grace period (e.g., 15 seconds),
we give vSphere enough time to complete all side effects — including registering the new snapshot reference in the VM's managed object tree.

This eliminates the race condition and avoids triggering a SOAP fault during the next snapshot operation or query.

🧰 TL;DR
We wait for snapshot task completion explicitly, even if the task looks "done", because:

VM names with invalid DNS-1123 characters break internal assumptions.

Snapshot creation appears successful but isn't fully finalized in inventory.

Without waiting, follow-up API calls may crash due to malformed object references.

pkg/controller/plan/migration.go
```diff
@@ -960,6 +960,13 @@ func (r *Migration) execute(vm *plan.VMStatus) (err error) {
 			precopy := plan.Precopy{Snapshot: snapshot, CreateTaskId: taskId, Start: &now}
 			vm.Warm.Precopies = append(vm.Warm.Precopies, precopy)
 			r.resetPrecopyTasks(vm, step)
+			// When original VM name is DNS-1123 invalid , we need to wait for the CreateSnapshot task to complete
+			if vm.NewName != "" {
+				err := r.provider.WaitForTaskCompletion(vm.Ref, taskId, r.kubevirt.loadHosts, 10*time.Second)
```
mnecas: For waiting, we have a  separate phase like `PhaseWaitForPreviousSnapshotRemoval`, and we are waiting for the task to finish in those. 
Reason for this i that we will check the task every reconcile (3s) and we will not halt the reconcile process.

pkg/controller/plan/adapter/vsphere/client.go
```diff
@@ -64,6 +65,42 @@ func (r *Client) RemoveSnapshot(vmRef ref.Ref, snapshot string, hosts util.Hosts
 	return
 }
 
+// This function polls a vSphere task (identified by taskId) until it finishes
+// either successfully or with an error — or until a specified timeout is reached.
+// Key Roles:
+// 1. Ensures the task has truly finished, not just marked as “in progress.
+// 2. Helps avoid race conditions where the task is marked success but the result (e.g., snapshot or VM state) is not yet fully available.
+// 3. Particularly important for VMs with invalid DNS-1123 names, where vSphere may take extra time to finalize task side effects.
+func (r *Client) WaitForTaskCompletion(vmRef ref.Ref, taskId string, hosts util.HostsFunc, timeout time.Duration) error {
```
mnecas: We have already function to check the tasks please take a look at `CheckSnapshotReady` and `CheckSnapshotRemove`

pkg/controller/plan/adapter/vsphere/client.go
```diff
@@ -248,7 +299,6 @@ func (r *Client) CheckSnapshotRemove(vmRef ref.Ref, precopy planapi.Precopy, hos
 
 // Check if a snapshot is ready to transfer.
 func (r *Client) CheckSnapshotReady(vmRef ref.Ref, precopy planapi.Precopy, hosts util.HostsFunc) (ready bool, snapshotId string, err error) {
-	r.Log.Info("Check Snapshot Ready", "vmRef", vmRef, "precopy", precopy)
```
mnecas: Why removing the logging?

mnecas (commented):
Ahhh! Makes sense if there is collision between the objects.
This could be a good workaround, but could we prevent this collision?
Know @yaacov was investigating these problems. @yaacov do you remember similar issue?

Hazanel:
That's the issue:

image

Hazanel:
Untitled diagram _ Mermaid Chart-2025-08-07-113150

pkg/controller/plan/controller.go
```diff
@@ -275,14 +275,32 @@ func (r *Reconciler) updatePlanStatus(plan *api.Plan) error {
...
+
+		// Fallback: try update with current plan
+		plan.Status.ObservedGeneration = plan.Generation
+		if updateErr := r.Status().Update(context.TODO(), plan.DeepCopy()); updateErr != nil {
+			r.Log.Error(updateErr, "Failed to update plan status with current version")
+			return updateErr
+		}
+		return nil
+	}
+
```
mnecas, line 293: @yaacov ptal this could be interesting for you as you were looking in the conflits

mnecas (approved):
lgtm!

yaacov:
lgtm

PR #2354, opened 2025-07-29
MTV-2404 | Can't migrate a guest that contains a disk with 0 size

Hazanel:
Issue:
Can't migrate a guest if the guest contains a disk with 0 size on MTV.

Fix:
1. Opa validation - Add a criticl concern
2. VM validation - Fail the migration plan if the vm has a disk with a 0 size.

Ref: https://issues.redhat.com/browse/MTV-2404


pkg/controller/plan/validation.go
```diff
@@ -636,6 +637,14 @@ func (r *Reconciler) validateVM(plan *api.Plan) error {
 		Message:  "%s disks are not supported for migration.",
 		Items:    []string{},
 	}
+	invalidDiskSizes := libcnd.Condition{
+		Type:     InvalidDiskSizes,
+		Status:   True,
+		Reason:   NotValid,
+		Category: api.CategoryCritical,
+		Message:  "VM has disks with invalid sizes (≤ 0).",
```
mnecas: NP: Please remove the `(≤ 0)`

mnecas (commented):
One small nitpick but looks good!

PR #2372, opened 2025-07-30
MTV-2702 | Migration plan failed and conversion pod is not created.

Hazanel:
Issue:
When a MAC address conflict occurs between the source VM being migrated and an existing VM on the destination (CNV/OpenShift) cluster, the failure is not clearly surfaced at the Plan or Migration level. Instead, the error is buried deep in the Plan YAML under the VM status section, specifically within the ImageConversion step under status.migration.vms[].pipeline[].error.reasons.

As a result, users do not see the conflict reflected in the overall failure reason of the plan, leading to difficulty in diagnosing the root cause without manually inspecting the full Plan YAML.

main error from the plan :

Source VM has a mac address conflict with one or more destination VMs.

Fix:
Perform the validation during plan creation to block migrations with invalid configurations from starting.

Ref:
https://issues.redhat.com/browse/MTV-2702

## Summary by CodeRabbit

* **New Features**
  * Added validation to detect and report MAC address conflicts between source and destination virtual machines during migration planning.
  * Users will now receive detailed messages highlighting any MAC address conflicts found, including which destination VMs are affected.

* **Bug Fixes**
  * Improved reliability of migration plans by preventing the creation of plans with conflicting MAC addresses.

* **Tests**
  * Removed outdated unit tests related to previous MAC conflict detection logic.
  * Added comprehensive new tests for MAC address conflict detection utilities and related data structures.
  * Added new tests to ensure MAC conflict checks run only for live migrations where applicable.


Hazanel:
An example of the new message :
image

pkg/controller/plan/adapter/ocp/validator.go
```diff
@@ -178,6 +179,11 @@ func (r *Validator) InvalidDiskSizes(vmRef ref.Ref) ([]string, error) {
 	return []string{}, nil
 }
 
+func (r *Validator) MacConflicts(vmRef ref.Ref) ([]planbase.MacConflict, error) {
```
mansam, line 182: OCP to OCP live migrations can have MAC conflicts.

pkg/controller/plan/adapter/base/mac_conflicts.go
```diff
@@ -0,0 +1,109 @@
...
+	if err != nil {
+		return nil, liberr.Wrap(err, "vm", vmRef.String())
+	}
+	return vm, nil
+}
+
+// GetDestinationVMsFromInventory retrieves and extracts destination VMs using the common pattern
+// This works for all providers (vsphere, ovirt, openstack, ova, ocp)
+func GetDestinationVMsFromInventory(client interface {
+	List(interface{}, ...webbase.Param) error
```
mnecas: Don't like having generic `interface`, the `client` should be type `web.Client`

mnecas: If you are having problems with cyclic dependencies, please create new interface in top of the file and reference which interface you are using in commnet. Or try to move the interface to separate module.

pkg/controller/plan/validation.go
```diff
@@ -814,6 +823,33 @@ func (r *Reconciler) validateVM(plan *api.Plan) error {
...
+			}
+
+			// Build detailed message with grouped conflicts
+			var conflictDetails []string
+			for destinationVM, macs := range vmConflictsByVM {
+				if len(macs) == 1 {
+					conflictDetails = append(conflictDetails, fmt.Sprintf("MAC %s conflicts with destination VM %s", macs[0], destinationVM))
+				} else {
+					conflictDetails = append(conflictDetails, fmt.Sprintf("MACs %s conflict with destination VM %s", strings.Join(macs, ", "), destinationVM))
+				}
```
mnecas: Don't think you need this if case, as the `strings.Join` wiht just one item joins without appliing the `", "`

mnecas (commented):
Like moving it to uniformed interface!
Please add unit tests

PR #2377, opened 2025-07-31
MTV-3075 | Migrating 2 NICs to the same NAD succeeds

Hazanel:
Issue:
Created a plan with a VM that has multiple NICs assigned to the same target NAD. Migration succeeded when it should have failed. Post-migration, the VM failed to boot properly.

Fix:
Prevent migration when multiple NICs target the same network (not just POD as we had)

Ref: https://issues.redhat.com/browse/MTV-3075

## Summary by CodeRabbit

* **New Features**
  * Enhanced network mapping validation for virtual machines, ensuring only one Pod network is mapped per VM and Multus network destination names are unique.

* **Refactor**
  * Unified network mapping validation logic across all providers for improved consistency and maintainability.
  * Renamed validation methods and updated related messages for broader and clearer network mapping checks.

### DIFF
--- a/pkg/controller/plan/adapter/base/network_validation.go
+++ b/pkg/controller/plan/adapter/base/network_validation.go
@@ -1,0 +1,90 @@
+package base
+
+import (
+	api "github.com/kubev2v/forklift/pkg/apis/forklift/v1beta1"
+	"github.com/kubev2v/forklift/pkg/apis/forklift/v1beta1/ref"
+	plancontext "github.com/kubev2v/forklift/pkg/controller/plan/context"
+	liberr "github.com/kubev2v/forklift/pkg/lib/error"
+)
+
+// Network types
+const (
+	Pod     = "pod"
+	Multus  = "multus"
+	Ignored = "ignored"
+)
+
+// NetworkMappingMatcher is a callback function that providers implement
+// to handle their specific network matching logic.
+// It should return true if the mapping applies to the VM's networks.
+// The vm parameter is the VM object retrieved by the provider.
+type NetworkMappingMatcher func(vm interface{}, mapping *api.NetworkPair) (bool, error)
+
+// VMRetriever is a callback function that providers implement
+// to retrieve VM data in their provider-specific way.
+type VMRetriever func(vmRef ref.Ref) (vm interface{}, err error)
+
+// ValidateNetworkMapping provides shared network mapping validation logic
+// that all providers can use. It validates that:
+// 1. Only one Pod network can be mapped per VM
+// 2. Multus networks must have unique destination names
+// 3. Other network types (like "ignored") have no restrictions
+//
+// The retriever function handles provider-specific VM retrieval.
+// The matcher function allows each provider to implement their specific
+// network matching logic while sharing the common validation rules.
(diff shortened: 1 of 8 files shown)


mansam:
@yaacov >.< I was in the middle of going through the OCP logic for this, it does not look correct to me. 

@Hazanel Why does the OCP validation check everything in the mapping? It's entirely possible for an OCP network mapping to have multiple source networks that target the same destination network, and it should only be a problem if a VM has two networks that target the same destination, just like every other provider.

yaacov:
> was in the middle of going through the OCP logic for this, it does not look correct to me.

@mansam , oow, sorry

@mansam , @Hazanel hi, do we need to revert ?

yaacov:
@Hazanel can you address @mansam concern in a followup ?

mnecas:
@Hazanel, I don't see any follow-up fix for the @mansam comments. Should I revert this PR?
The only edge case that I can think of is that you should not be able to migrate multiple networks to the Pod network.

pkg/controller/plan/adapter/vsphere/validator.go
```diff
@@ -59,38 +59,36 @@ func (r *Validator) NetworksMapped(vmRef ref.Ref) (ok bool, err error) {
...
-	mapping := r.Plan.Referenced.Map.Network.Spec.Map
-	podMapped := 0
-	for i := range mapping {
-		mapped := &mapping[i]
-		ref := mapped.Source
+// Validate that no more than one of a VM's networks is mapped to the pod network,
+// and that Multus networks have unique names.
+func (r *Validator) NetworkMapping(vmRef ref.Ref) (ok bool, err error) {
+	// Create provider-specific VM retrieval function
+	retriever := func(vmRef ref.Ref) (interface{}, error) {
```
mnecas, line 66: I'm not a big fan of the generic `interface{}`

pkg/controller/plan/adapter/ocp/validator.go
```diff
@@ -90,33 +91,32 @@ func (r *Validator) MaintenanceMode(vmRef ref.Ref) (bool, error) {
...
-	for i := range mapping {
-		mapped := &mapping[i]
-		if mapped.Destination.Type == Pod {
-			podMapped++
-		}
+	// Create provider-specific network matching function
+	// For OCP, we validate all mappings since we're migrating VMs within the cluster
+	matcher := func(vmInterface interface{}, mapping *api.NetworkPair) (bool, error) {
+		// For OCP provider, all mappings apply since we're validating intra-cluster migrations
+		return true, nil
```
mnecas, line 115: Not necessarily, what if you do multiple networks to pod

pkg/controller/plan/adapter/base/network_validation.go
```diff
@@ -0,0 +1,90 @@
...
+
+// NetworkMappingMatcher is a callback function that providers implement
+// to handle their specific network matching logic.
+// It should return true if the mapping applies to the VM's networks.
+// The vm parameter is the VM object retrieved by the provider.
+type NetworkMappingMatcher func(vm interface{}, mapping *api.NetworkPair) (bool, error)
+
+// VMRetriever is a callback function that providers implement
+// to retrieve VM data in their provider-specific way.
+type VMRetriever func(vmRef ref.Ref) (vm interface{}, err error)
```
mnecas, line 25: don't like fetching the VM in the validation and then piping it back to the
The validators are meant to do the provider-specific operations

PR #2397, opened 2025-08-04
MTV-1664 | Warm migration with pre-existing snapshot will not work.

Hazanel:
Issue:
if a VM has a pre-existing snapshot, then it's generally considered to be unsafe to migrate such a VM with warm migration.

Fix:
MTV should not allow to migrate such VMs - and raise a warning.

Ref: https://issues.redhat.com/browse/MTV-1664


Hazanel:
image

pkg/controller/plan/adapter/vsphere/validator.go
```diff
@@ -384,6 +384,25 @@ func (r *Validator) ChangeTrackingEnabled(vmRef ref.Ref) (bool, error) {
 	return vm.ChangeTrackingEnabled, nil
 }
 
+// Validate that VM has no pre-existing snapshots for warm migration
+func (r *Validator) HasSnapshot(vmRef ref.Ref) (ok bool, msg string, category string, err error) {
+	if !r.Plan.Spec.Warm {
```
yaacov: warm is deprecated in favor of migration type, can you also test for migration type ?

pkg/controller/plan/adapter/vsphere/validator.go
```diff
@@ -373,7 +373,9 @@ func (r *Validator) StaticIPs(vmRef ref.Ref) (ok bool, err error) {
 
 // Validate that the vm has the change tracking enabled
 func (r *Validator) ChangeTrackingEnabled(vmRef ref.Ref) (bool, error) {
-	if !r.Plan.Spec.Warm {
+	// Check if this is a warm migration using both old and new fields for backward compatibility
+	isWarmMigration := r.Plan.Spec.Warm || r.Plan.Spec.Type == api.MigrationWarm
```
yaacov, line 377: :+1: cc:// @mansam

PR #2421, opened 2025-08-06
MTV-2710 | OVA provider crash when adding more ova files to inventory.

Hazanel:
Issue:
1. When the OVA provider already exists and new OVA files are added to the NFS server, the provider becomes unresponsive, and the OVA pod reports the following error: "Error processing OVF from OVA: unexpected EOF."
2.The OVA VM list became corrupted after adding files to an existing provider, resulting in an empty list of VMs.

Fix:
1. Added isFileComplete check before reading OVA/OVF files to avoid processing files that are still being copied to NFS.
2. Prevents refresh failures when deleting non-existent inventory records.

Ref: https://issues.redhat.com/browse/MTV-2710

## Summary by CodeRabbit

* **New Features**
  * Added a check to ensure files are fully copied and stable before processing, preventing errors from incomplete files.

* **Bug Fixes**
  * Improved error handling during deletion operations to ignore "not found" errors, ensuring smoother resource cleanup.
  * Enhanced logging to better distinguish between files still being copied and other processing errors.

### DIFF
--- a/cmd/ova-provider-server/ova-provider-server.go
+++ b/cmd/ova-provider-server/ova-provider-server.go
@@ -15,6 +15,7 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"time"
 	"unicode"
 
 	"github.com/kubev2v/forklift/pkg/lib/gob"
@@ -312,9 +313,14 @@ func scanOVAsOnNFS() (envelopes []Envelope, ovaPaths []string) {
 	for _, ovaFile := range ovaFiles {
 		fmt.Println("Processing OVA file:", ovaFile)
 
+		if !isFileComplete(ovaFile) {
+			log.Printf("Skipping %s: file still being copied\n", ovaFile)
+			continue
+		}
+
 		xmlStruct, err := readOVFFromOVA(ovaFile)
 		if err != nil {
-			fmt.Println("Error processing OVF from OVA:", err)
+			log.Printf("Error processing OVF from OVA %s: %v\n", ovaFile, err)
 			continue
 		}
 		envelopes = append(envelopes, *xmlStruct)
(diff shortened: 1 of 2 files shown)


cmd/ova-provider-server/ova-provider-server.go
```diff
@@ -382,7 +390,23 @@ func hasSuffixIgnoreCase(fileName, suffix string) bool {
 	return strings.HasSuffix(strings.ToLower(fileName), strings.ToLower(suffix))
 }
 
+func isFileComplete(filePath string) bool {
```
yaacov, line 400:
```suggestion
// isFileComplete checks that the file was not modified in the last 30s
func isFileComplete(filePath string) bool {
```

just to be clear we don't actually know it's complete, we just have a very good gues

yaacov (approved):
comment nitpick

cmd/ova-provider-server/ova-provider-server.go
```diff
@@ -314,7 +315,11 @@ func scanOVAsOnNFS() (envelopes []Envelope, ovaPaths []string) {
 
 		xmlStruct, err := readOVFFromOVA(ovaFile)
 		if err != nil {
-			fmt.Println("Error processing OVF from OVA:", err)
+			if strings.Contains(err.Error(), "still being copied") {
```
yaacov: can we just check for `isFileComplete` ?

yaacov: checking error string is not nice ^^

cmd/ova-provider-server/ova-provider-server.go
Hazanel: @yaacov  - correct , among the other options this is the least time consuming along with the good guess...

PR #2431, opened 2025-08-07
MTV-2362 | Add cacert to virt-v2v

mnecas:
Issue:
The guest conversion fails when the provider uses self signed certificates.
```
virt-v2v: error: exception: libvirt: VIR_ERR_INTERNAL_ERROR: VIR_FROM_ESX: internal error: curl_easy_perform() returned an error: SSL peer certificate or SSH remote key was not OK (60) : SSL certificate problem:unable to get local issuer certificate
```

This fix requires fix in libvirt with the `cacert` parameter. This parameter allow us to specify cert for libvirt connection without requiring the root. As the /etc/pki/ca-trust/source/anchors, requires to runn update-ca-trust for which we need root but the conversion has QEMU (107) user.

Ref:
- https://issues.redhat.com/browse/MTV-2362
- https://issues.redhat.com/browse/RHEL-98292

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of SSL verification settings for vSphere connections, ensuring a default certificate authority bundle is used when TLS verification is enabled.

### DIFF
--- a/pkg/controller/plan/adapter/vsphere/builder.go
+++ b/pkg/controller/plan/adapter/vsphere/builder.go
@@ -399,9 +399,12 @@ func (r *Builder) getSourceDetails(vm *model.VM, sourceSecret *core.Secret) (lib
 		return
 	}
 
-	sslVerify := ""
+	var sslVerify string
 	if basecontroller.GetInsecureSkipVerifyFlag(sourceSecret) {
 		sslVerify = "no_verify=1"
+	} else {
+		// This path is created by linkCertificates in the v2v container containes either the provider cert or pod certs.
+		sslVerify = "cacert=/opt/ca-bundle.crt"
 	}
 
 	if hostDef, found := r.hosts[host.ID]; found {


pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -399,9 +399,11 @@ func (r *Builder) getSourceDetails(vm *model.VM, sourceSecret *core.Secret) (lib
 		return
 	}
 
-	sslVerify := ""
+	var sslVerify string
 	if basecontroller.GetInsecureSkipVerifyFlag(sourceSecret) {
 		sslVerify = "no_verify=1"
+	} else {
+		sslVerify = "cacert=/opt/ca-bundle.crt"
```
Hazanel, line 407: Is this path in the virt-v2v pod env?
Can you explain a bit where it resides , and I guess we should have an actual crt added, no?

mnecas: Good point, added comment where it's getting generated.
It's in the v2v pod `linkCertificates`.

Hazanel:
lgtm

PR #2448, opened 2025-08-11
Add only conversion mode

mnecas:
# Add only conversion mode
## Describtion
Right now, Forklift always manages the PVC creation, population and guest conversion.
If there is some external storage vendor who could provide the disk to the Forklift would need to perform just the guest conversion and VM creation. For this, I have reused the `type` on the spec plan, which was introduced as part of Live OCP migration https://github.com/kubev2v/forklift/pull/1579. The `type` can be  `cold`, `warm`, `live`, and with this feature, it can also contain `conversion`. I opted for a new `type` instead of adding new parameters as the migration would have a different itinerary.

## Requirements from vendor
**Labels:**
Forklift will look for the `vmID` and `vmUUID` in the PVC.

**Annotation:**
`forklift.konveyor.io/disk-source` => the vmdk disk name

## Note

- This feature is available only on the VMware provider, as the other providers do not perform the conversion.
- Recommendation: Keep the VM running right until the storage vendor migrates and have the provider accessible in the Forklift. Forklift will collect the IP information, allowing the persistent static IPs.

## Diagram
image

## Summary by CodeRabbit

* **New Features**
  * Added a new "conversion" migration type for Plans and Migrations.
  * New conversion-only workflow that performs guest conversion before VM creation.
  * Destination PVC matching in conversion mode now uses VM UUID to avoid accidental overwrites.

* **Bug Fixes / Validation**
  * Schema and validation updated to accept and enforce the new type; transfer logic will skip v2v when conversion-only is selected.

* **Tests**
  * Added test helpers to cover plan scenarios including the new type.

### DIFF
--- a/build/forklift-operator-bundle/Containerfile-downstream
+++ b/build/forklift-operator-bundle/Containerfile-downstream
@@ -13,7 +13,7 @@ ARG OCP_VERSIONS
 
 ARG API_IMAGE="registry.redhat.io/mtv-candidate/mtv-api-rhel9@sha256:63b10d7e986fe4052634f08fbc9227b7a217cd8998b76e4784f989e5c311fa66"
 
-ARG CONTROLLER_IMAGE="registry.redhat.io/mtv-candidate/mtv-controller-rhel9@sha256:77e4f020badd86b6a9595c6fb803215cbf247cb23ff1f2d4185f90c5485a915e"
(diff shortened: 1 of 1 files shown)


yaacov:
Q: how do you get the disk order ?

mnecas:
> Q: how do you get the disk order ?

we don't but afaik we don't need it for the guest conversion, but the provider could populate some annotation with disk name

PR #2453, opened 2025-08-11
MTV-3129 | Add ClusterRole for between cluster migrations

mnecas:
Issue:
Right now users have only documented the minimal permissions list. We should provide and automatically update the role for between cluster migrations.

Fix:
Create new roler which will be deployed with the forklift operator.

## Summary by CodeRabbit

- Chores
  - Added a new cluster-wide role granting access across core, storage, networking, and virtualization resources.
  - Consolidates permissions needed by migration/controller components to operate cluster-wide.
  - No changes to existing user-facing workflows; this enables broader, coordinated migration capabilities.


yaacov:
@mnecas hi, merging as is, if permissions need changing (see comments above, please fix in a followup)

PR #2467, opened 2025-08-13
Add support for vvol migration

borod108:


cmd/vsphere-xcopy-volume-populator/internal/pure/flashArray.go
```diff
@@ -118,12 +124,126 @@ func (f *FlashArrayClonner) CurrentMappedGroups(targetLUN populator.LUN, context
...
+		return populator.LUN{}, fmt.Errorf("failed to get target volume from Pure FlashArray: %w", err)
+	}
+
+	naa := FlashProviderID + strings.ToLower(v.Serial)
+
+	l := populator.LUN{
+		Name:         v.Name,
+		SerialNumber: v.Serial,
+		NAA:          naa,
 	}
```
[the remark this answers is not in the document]
xinzheyang: pick #2559 when rebasing

cmd/vsphere-xcopy-volume-populator/internal/populator/storage.go
```diff
@@ -1,7 +1,20 @@
...
+	"github.com/kubev2v/forklift/cmd/vsphere-xcopy-volume-populator/internal/vmware"
+)
+
+type StorageApi interface{}
+
+type VvolStorageApi interface {
+	// VvolCopy performs a direct copy operation using vSphere API to discover source volume
+	VvolCopy(vsphereClient vmware.Client, vmId string, sourceVMDKFile string, persistentVolume PersistentVolume, progress chan<- uint) error
+	// ResolvePVToLUN resolves PersistentVolume to LUN details
+	ResolvePVToLUN(persistentVolume PersistentVolume) (LUN, error)
```
xinzheyang, line 30: just re-use StorageResolver?

cmd/vsphere-xcopy-volume-populator/vsphere-xcopy-volume-populator.go
```diff
@@ -121,9 +121,20 @@ func main() {
...
+	case populator.VvolStorageApi:
+		p, err = populator.NewVvolPopulator(api, vsphereHostname, vsphereUsername, vspherePassword)
+		if err != nil {
+			klog.Fatalf("Failed to create a VVol populator: %s", err)
+		}
+	case populator.EsxiStorageApi:
+		p, err = populator.NewWithRemoteEsxcli(api, vsphereHostname, vsphereUsername, vspherePassword)
+		if err != nil {
+			klog.Fatalf("Failed to create a remote esxcli populator: %s", err)
+		}
```
xinzheyang: I'm new to golong so correct me if I'm wrong.
NewFlashArrayClonner implements all the required functions for both VvolStorageApi and EsxiStorageApi. so it satisfies both interfaces. then it will always match to the first case and create a VvolPopulator, and never use Esxi, but this will definitely fail if the the source VM is not vvol, right?

xinzheyang (changes requested):

cmd/vsphere-xcopy-volume-populator/internal/pure/flashArray.go
```diff
@@ -118,12 +124,126 @@
...
+				// Check if this is a VVol backing and matches our target VMDK
+				if backing.BackingObjectId != "" && f.matchesVMDKPath(backing.FileName, vmDisk) {
+					klog.Infof("Found VVol backing for VMDK %s with ID %s", vmDisk.VmdkFile, backing.BackingObjectId)
+
+					// Use REST client to find the volume by VVol ID
+					volumeName, err := f.restClient.FindVolumeByVVolID(backing.BackingObjectId)
+					if err != nil {
+						klog.Warningf("Failed to find volume by VVol ID %s: %w", backing.BackingObjectId, err)
+						continue
+					}
```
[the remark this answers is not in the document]
mkilavuz: This needs to be fixed.

cmd/vsphere-xcopy-volume-populator/vsphere-xcopy-volume-populator.go
mkilavuz: This is indeed what happens. The current implementation fails if the source VM is on VMFS instead of vvol. There should be an explicit decision to pick which StorageApi will be used.

Also, from the user experience side, we need to make a decision on how to pick which offloading method to use.
1) User decides whether to use xcopy offload or vvol copy by adding `offloadPlugin` field to the storage map. If `vsphereXcopyConfig` is added it will be xcopy, if `vVolCopyConfig` is added then it will be vvol copy. It is more explicit and gives flexibility to the user.
2) User decides to optimize migration by using an offload method. If it is VMFS, forklift chooses xcopy, if on vvols forklift chooses vvol copy. This will also depend on the vendor support.

Another thing is that everything related to vvol copy is put under files and resource fields with `xcopy` related names. We should either make separation in the code or change the names to be more generic.

mkilavuz (changes requested):
The vvol copy feature works. But this PR needs work with making it clear the correct method is picked.
1) How does the code picks the right optimization?
2) How does the user pick which optimization to use? Do we give them control over the choice or forklift picks the vvol copy over xcopy where applicable because it is always expected to be faster?

mkilavuz:
An update to my previous comments: We should always go with volume copy if it is a vvol datastore. If it is not then it will go with xcopy. The rest api should always be available for this operation. The only part missing here is to figure out the type of datastore. I hope that simplifies a few things in the workflow.

cmd/vsphere-xcopy-volume-populator/internal/pure/rest_client.go
```diff
@@ -0,0 +1,582 @@
...
+		Name string `json:"name"`
+	} `json:"source"`
+	Names string `json:"names"` // Changed from []string to string
+}
+
+// Host represents a Pure FlashArray host
+type Host struct {
+	Name string   `json:"name"`
+	Iqn  []string `json:"iqn"`
+	Wwn  []string `json:"wwn"`
```
snehal-paradkar: These should be "iqns" and "wwns". Is there a reason we don't collect "nqns"?

cmd/vsphere-xcopy-volume-populator/internal/pure/rest_client.go
```diff
@@ -0,0 +1,582 @@
...
+// VolumesResponse represents the response from volumes API
+type VolumesResponse struct {
+	Items []Volume `json:"items"`
+}
+
+// HostConnectionRequest represents a host connection request
+type HostConnectionRequest struct {
+	HostNames   []string `json:"host_names"`
+	VolumeNames []string `json:"volume_names"`
+}
```
snehal-paradkar, line 101: These should be of type String. In case of multiple values, it should be a comma separated list.

cmd/vsphere-xcopy-volume-populator/internal/pure/rest_client.go
```diff
@@ -0,0 +1,582 @@
...
+		return fmt.Errorf("copy volume request failed with status %d: %s", resp.StatusCode, string(body))
+	}
+
+	klog.Infof("Pure REST Client: Successfully copied volume from %s to %s", sourceVolumeName, targetVolumeName)
+	return nil
+}
+
+// ListHosts lists all hosts on the Pure FlashArray
+func (c *RestClient) ListHosts() ([]Host, error) {
+	url := fmt.Sprintf("https://%s/api/%s/hosts", c.hostname, c.apiV2)
```
snehal-paradkar: Pure FlashArray can have active cluster setup and in such cases the list hosts api will list remote hosts by default. If the same host is present on both FlashArrays we can select the remote host for connection for a local volume which is not supported. The query should use this filter to list only local hosts '?filter=is_local=true'

snehal-paradkar (changes requested):

cmd/vsphere-xcopy-volume-populator/internal/pure/rest_client.go
```diff
@@ -0,0 +1,582 @@
...
+}
+
+// ConnectHost connects a volume to a host
+func (c *RestClient) ConnectHost(hostName, volumeName string) error {
+	url := fmt.Sprintf("https://%s/api/%s/connections", c.hostname, c.apiV2)
+
+	requestBody := HostConnectionRequest{
+		HostNames:   []string{hostName},
+		VolumeNames: []string{volumeName},
+	}
```
snehal-paradkar, line 406: These should be String as mentioned in the previous comments

cmd/vsphere-xcopy-volume-populator/internal/pure/rest_client.go
```diff
@@ -0,0 +1,582 @@
...
+
+	jsonBody, err := json.Marshal(requestBody)
+	if err != nil {
+		return fmt.Errorf("failed to marshal disconnect host request: %w", err)
+	}
+
+	req, err := http.NewRequest("DELETE", url, bytes.NewBuffer(jsonBody))
```
snehal-paradkar: host_names and volume_names are query parameters. So the URL will be something like this - "url := fmt.Sprintf("https://%s/api/%s/connections?host_names=%s&volume_names=%s", c.hostname, c.apiV2, hostName, volumeName)"

cmd/vsphere-xcopy-volume-populator/internal/vmware/client.go
```diff
@@ -132,6 +144,125 @@ func (c *VSphereClient) GetDatastore(ctx context.Context, dc *object.Datacenter,
...
+				return nil, fmt.Errorf("error searching for VM in Datacenter '%s': %w", dc.Name(), err)
+			}
+		} else {
+			vm = result
+			break
+		}
+	}
+	if vm == nil {
+		moref := types.ManagedObjectReference{Type: "VirtualMachine", Value: vmId}
+		vm = object.NewVirtualMachine(c.Client.Client, moref)
```
mkilavuz, line 170: Isn't this enough? Why do we need to go through the above?

cmd/vsphere-xcopy-volume-populator/internal/vmware/client.go
```diff
@@ -132,6 +144,125 @@ func (c *VSphereClient) GetDatastore(ctx context.Context, dc *object.Datacenter,
...
+			klog.V(2).Infof("Disk %s is RDM-backed (DeviceName: %s)", vmdkPath, backing.DeviceName)
+			return &DiskBacking{
+				VVolId:     "",
+				IsRDM:      true,
+				DeviceName: backing.DeviceName,
+			}, nil
+		}
+	}
+
+	// If we couldn't find the disk, return default VMDK type
```
mkilavuz, line 241: This is probably not safe. It could be NFS or maybe something else as well. We should explicitly check for VMDK and fail if it is not.

cmd/vsphere-xcopy-volume-populator/internal/populator/disk_detector.go
```diff
@@ -0,0 +1,54 @@
...
+		// Virtual Volume (VVol) - has VVolId in backing
+		klog.Infof("Detected VVol disk (VVolId: %s)", backing.VVolId)
+		return DiskTypeVVol, nil
+
+	case backing.IsRDM:
+		// Raw Device Mapping
+		klog.Infof("Detected RDM disk (DeviceName: %s)", backing.DeviceName)
+		return DiskTypeRDM, nil
+
+	default:
```
mkilavuz, line 49: Defaults are not safe here unless we cover all possible cases. We should pick VMDK based on a signal. Default here should return a failure.

cmd/vsphere-xcopy-volume-populator/internal/vmware/client.go
rgolangh: we'll consider this in an upcoming change. This safety check is worth it.

PR #2473, opened 2025-08-13
MTV-3146 | Use Powerflex's systemId in the populator

rgolangh:
https://issues.redhat.com/browse/MTV-3146

## Summary by CodeRabbit

* **New Features**
  * Support for targeting a specific PowerFlex system by System ID, enabling operation in multi-system environments.
  * Added a required environment variable POWERFLEX_SYSTEM_ID to select the target system; clear error messaging guides configuration when it's missing.

### DIFF
--- a/cmd/vsphere-xcopy-volume-populator/internal/powerflex/powerflex.go
+++ b/cmd/vsphere-xcopy-volume-populator/internal/powerflex/powerflex.go
@@ -10,10 +10,14 @@ import (
 	"k8s.io/klog/v2"
 )
 
-const sdcIDContextKey string = "sdcId"
+const (
+	SYSTEM_ID_ENV_KEY        = "POWERFLEX_SYSTEM_ID"
+	sdcIDContextKey   string = "sdcId"
+)
 
 type PowerflexClonner struct {
-	Client *goscaleio.Client
+	Client   *goscaleio.Client
+	systemId string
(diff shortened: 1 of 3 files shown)


cmd/vsphere-xcopy-volume-populator/internal/powerflex/powerflex.go
```diff
@@ -103,7 +107,7 @@ func (p *PowerflexClonner) fetchSdcVolume(initatorGroup string, targetLUN popula
 
 	// TODO rgolan do we need an instanceID as part of the client?
 	// probably yes for multiple instances
-	system, err := p.Client.FindSystem("", "", "")
+	system, err := p.Client.FindSystem(p.systemId, "", "")
```
mnecas, line 110: While we are at it shouldn't we also make the other parameters configurable just in case?

rgolangh: the three arguments are $id, $name, $href. The only one we can extract from the CSI driver is the system ID. The name is not being used in the CSI driver config as far as I see, and the API expects the $id or $name. the $href is for multiple systems management and I am quite sure we don't need it. 
To summarize it looks like we always have the systemID from the the CSI driver and we only need it to trace the volumes the CSI created

PR #2523, opened 2025-08-17
Feauture/xcopy used indication

OrrBG:
## Summary by CodeRabbit

- New Features
  - Added clone progress bytes tracking, exposed via Prometheus metric (vsphere_xcopy_volume_populator_clone_progress_bytes) and in CR status (cloneProgressBytes).
  - Introduced a TLS-secured /metrics endpoint (self-signed) on port 8443 for safer metric scraping.
  - Improved progress reporting precision (64-bit) with concurrent percentage and byte-based updates.
  - Surfaced XCOPY usage indicators in task status to reflect hardware-accelerated clones when available.


cmd/vsphere-xcopy-volume-populator/internal/populator/remote_esxcli.go
```diff
@@ -26,11 +26,12 @@ type vmkfstoolsClone struct {
...
-	Pid      int    `json:"pid"`
-	ExitCode string `json:"exitCode"`
-	Stderr   string `json:"stdErr"`
-	LastLine string `json:"lastLine"`
-	TaskId   string `json:"taskId"`
+	Pid       int    `json:"pid"`
+	ExitCode  string `json:"exitCode"`
+	Stderr    string `json:"stdErr"`
+	LastLine  string `json:"lastLine"`
+	XcopyUsed string `json:"xcopyUsed"`
```
rgolangh: type bool?

cmd/vsphere-xcopy-volume-populator/vmkfstools-wrapper/vmkfstools_wrapper.py
```diff
@@ -51,6 +52,9 @@ def clone(args):
         with open(os.path.join(tmp_dir, "rdmfile"), "w") as rdmfile_file:
             rdmfile_file.write(f"{rdmfile}\n")
 
+        with open(os.path.join(tmp_dir, "targetLun"), "w") as target_lun_file:
+            target_lun_file.write(f"{os.path.basename(target)}\n")
```
rgolangh: perhaps this line feed is what caused the garbling of the response. I think we can saftly get rid of the \n in the end cause we have only a single entry.

cmd/vsphere-xcopy-volume-populator/vmkfstools-wrapper/vmkfstools_wrapper.py
```diff
@@ -84,10 +88,22 @@ def taskGet(args):
                   "exitCode": "1", "lastLine": line.rstrip(), "stdErr": e}
         print(XML.format("1", json.dumps(result)))
         return
+    try:
+        with open(os.path.join(tmp_dir, "targetLun"), "r") as target_lun_file:
+            target_lun = target_lun_file.read()
+        if was_xcopy_used(target_lun):
```
rgolangh: can be a boolean `xcopy_used = was_xcopy_used(target_lun)` no?

cmd/vsphere-xcopy-volume-populator/vmkfstools-wrapper/vmkfstools_wrapper.py
```diff
@@ -122,6 +138,32 @@ def extract_rdmdisk_file(rdm_file):
...
+
+    try:
+        target_lun_stats = subprocess.run(
+            ["vsish", "-e", "cat", stats_path],
+            capture_output=True,
+            text=True,
+            check=True
+        )
+    except subprocess.CalledProcessError:
+        print(f"Error: Unable to read stats for device {target_lun}")
```
rgolangh: I would return a None from that function in case that happens. I wouldn't want this to fail the taskGet

rgolangh: also, instead of print use `logging.info`

rgolangh: or logging.error

cmd/vsphere-xcopy-volume-populator/vmkfstools-wrapper/vmkfstools_wrapper.py
```diff
@@ -122,6 +138,32 @@ def extract_rdmdisk_file(rdm_file):
         logging.error(e)
     return ""
 
+def was_xcopy_used(target_lun):
+    stats_path = f"/storage/scsifw/devices/{target_lun.strip()}/stats"
+
+    try:
+        target_lun_stats = subprocess.run(
+            ["vsish", "-e", "cat", stats_path],
```
rgolangh: small improvement: add the `-r` flag so the name of the fields if machine friendly.

cmd/vsphere-xcopy-volume-populator/vmkfstools-wrapper/vmkfstools_wrapper.py
OrrBG: hahahaha true, didn't even notice

cmd/vsphere-xcopy-volume-populator/vmkfstools-wrapper/vmkfstools_wrapper.py
OrrBG: maybe I can raise the error since then the taskGet function will print the rest of the parameters without the xcopy used and clone writes paramters

rgolangh:
@OrrBG also follow the github check to see why they fail. The DCO one means you need to use `git commit -s` to add your signaure to the commit message

rgolangh:
Replaced by #3821

PR #2556, opened 2025-08-19
Backport 2.9.3 [2517]

yaacov:
Backport of:

  - #2517


solenoci:
We also need to generate operator/mtv-csv. If you don't know how, I can do it. We don't have anything to do that for you, so it will need to be done manually

PR #2564, opened 2025-08-19
Fix for Vsphere to work with powerflex

borod108:
1. For VsphereXcopyPopulator we should not restart the populator container when it fails, it makes no sense and hides root issues.
2. When a provider requires scinin adapter - we should fail when it is not found on the esxi
3. Fixed unmapping issues
4. Fixed mapping volume to multipel Sdcs (AllowMultiple)

Co-Authored with: Roy Golan <rgolan@redhat.com>

## Summary by CodeRabbit

* **New Features**
  * Scini-aware HBA discovery for ESXi, with Scini-driven mapping when required.
  * Two-phase cleanup: partial unmap by default and a full unmap-and-remap fallback.
  * Support for mapping to a specific SDC and optional “unmap all” behavior.
  * VSphere populator now emits a dedicated “VSphere xcopy populator failed (no retry)” event.

* **Bug Fixes**
  * Improved HBA UID detection and validation; explicit errors if none found.
  * Stricter handling when no ESXi SDC mappings exist.
  * Short delay added to stabilize CSI mapping timing and reduce transient failures.

### DIFF
--- a/cmd/vsphere-xcopy-volume-populator/internal/populator/remote_esxcli.go
+++ b/cmd/vsphere-xcopy-volume-populator/internal/populator/remote_esxcli.go
@@ -90,63 +90,79 @@ func (p *RemoteEsxcliPopulator) Populate(vmId string, sourceVMDKFile string, pv 
 	uniqueUIDs := make(map[string]bool)
 	hbaUIDs := []string{}
 
-	for i, a := range r {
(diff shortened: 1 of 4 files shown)


yaacov:
i am ok with the chagne to the lib ( e.g. fail fast on case of xcopy )

rgolangh:
LGTM. Need @yaacov to approve the volume controller part

PR #2583, opened 2025-08-20
MTV-3093 | Stop replacing first boot script

yaacov:
Ref:
https://issues.redhat.com/browse/MTV-3093

Issue:
As a fix for https://issues.redhat.com/browse/MTV-1299 we replaced the bootscript on windows machines, but as of RHEL 2.5.z and above this is not needed anymore.

Fix:
Stop replacing the bootscript on windows

## Summary by CodeRabbit

* **Chores**
  * Removed the Windows first-boot automation script from customization.
  * The Windows guest no longer uploads or executes firstboot.bat during customization; other Windows first-boot related uploads remain unchanged.
  * Users will no longer observe per-boot script execution or automatic reboot cycles previously driven by the removed script.


rwmjones:
I'm on [Chesterton's Fence](https://en.wikipedia.org/wiki/G._K._Chesterton#Chesterton's_fence) about why we ever uploaded `firstboot.bat` in the first place. As far as I can tell it was never necessary.

I compared `firstboot.bat` from the initial commit 1a4319e1c5e528601981591abe5fde51669dbadc with the one in virt-customize at the same time, and they seem completely identical.  Since virt-v2v always uploads a `firstboot.bat` to Windows guests, this appears like it was always unnecessary, but maybe @mnecas there is some reason I don't understand?

Anyway definitely overwriting an old copy of `firstboot.bat` over the current one that virt-v2v installs is very bad.  Therefore I think this commit is good, as far as I'm aware.

PR #2630, opened 2025-08-25
OCP Live: account for namespaced name in mapping

mansam:
Sources in OCP provider network mappings are named as `namespace/name` because Source mappings don't have a namespace field. This updates the OCP live migration builder to account for that when looking for matches in the network map.

## Summary by CodeRabbit

* **Bug Fixes**
  * Network mapping during live migration now respects the VM’s namespace for Multus networks, preventing incorrect remaps when identical network names exist across namespaces and reducing migration failures.

* **Improvements**
  * More reliable and predictable network resolution for migrated VMs in multi-namespace clusters, improving overall migration accuracy.

### DIFF
--- a/pkg/controller/provider/web/ocp/client.go
+++ b/pkg/controller/provider/web/ocp/client.go
@@ -101,10 +101,12 @@ func (r *Finder) ByRef(resource interface{}, ref base.Ref) (err error) {
 			err = r.Get(resource, id)
 			return
 		}
-		name := ref.Name
+		ns, name := r.resolve(ref)
 		if name != "" {
-			ns, name := path.Split(name)
-			ns = strings.TrimRight(ns, "/")
+			if ns == "" {
+				err = liberr.Wrap(RefNotUniqueError{Ref: ref})
+				break
+			}
 			list := []NetworkAttachmentDefinition{}
 			err = r.List(
 				&list,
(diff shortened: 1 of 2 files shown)


pkg/controller/plan/migrator/ocp/live.go
```diff
@@ -1198,11 +1198,11 @@ func (r *Builder) VirtualMachine(vm *planapi.VMStatus) (object *cnv.VirtualMachi
...
-	r.mapNetworks(object)
+	r.mapNetworks(vm, object)
 	return
 }
 
-func (r *Builder) mapNetworks(target *cnv.VirtualMachine) {
+func (r *Builder) mapNetworks(vm *planapi.VMStatus, target *cnv.VirtualMachine) {
 	networkMap := make(map[string]api.DestinationNetwork)
 	for _, network := range r.Map.Network.Spec.Map {
 		networkMap[network.Source.Name] = network.Destination
```
yaacov: Q: is it posible to have the namespace in the source ?

if it is then we also need to acount for source.namespace / source.name when creating the key value map
if it is not, maybe add a comment, `// source name always encode namespace in a namespace/name pattern`

yaacov: note that spource is `ref.ref` and it does have `namespace` attribute

pkg/controller/plan/migrator/ocp/live.go
```diff
@@ -1211,7 +1211,7 @@ func (r *Builder) mapNetworks(target *cnv.VirtualMachine) {
 		network := &target.Spec.Template.Spec.Networks[i]
 		switch {
 		case network.Multus != nil:
-			destination := networkMap[network.Multus.NetworkName]
+			destination := networkMap[path.Join(vm.Namespace, network.Multus.NetworkName)]
```
mnecas: This would apply to the NADs in the same namespace as the VM, but NADs can be in `default` namespace and are globally accessible. So this could override the `default/netowrk` NAD.

mansam: Ah, I had forgotten about that. Thanks, will fix.

pkg/controller/plan/migrator/ocp/live.go
mansam: If source is just a normal Ref, then we can resolve this by having the UI create the mappings in a way that uses the namespace field properly. I'll look at this again.

mansam:
@mnecas Can this be backported to 2.10? It's needed for https://issues.redhat.com/browse/MTV-3085.

PR #2635, opened 2025-08-26
Change legacy win iso path

solenoci:
Issue: There was a mismatch between filepath in code and in filepath in containerfile

Fix: Move the drivers to expected location


build/virt-v2v/Containerfile-downstream
```diff
@@ -73,7 +73,7 @@ COPY --from=builder /app/image-converter /usr/local/bin/image-converter
 
 COPY --from=builder /app/virt-v2v-wrapper /usr/bin/virt-v2v-wrapper
 
-COPY --from=win2008iso /usr/share/virtio-win/virtio-win.iso /usr/share/virtio-win/virtio-win-2008.iso
+COPY --from=winlegacyiso /usr/share/virtio-win/virtio-win.iso /usr/local/virtio-win-legacy.iso
```
Hazanel, line 76: Env variable id "/usr/local/virtio-win.iso" so we need to change also the code in https://github.com/kubev2v/forklift/blob/9965302073ecff71368e612a225175b241effb5f/pkg/controller/plan/adapter/vsphere/builder.go

Hazanel, line 76: Env variable is :  "/usr/local/virtio-win.iso", so we need to change also the code in https://github.com/kubev2v/forklift/blob/9965302073ecff71368e612a225175b241effb5f/pkg/controller/plan/adapter/vsphere/builder.go

Hazanel (changes requested):

PR #2637, opened 2025-08-26
2.9 | MTV-3227 | Change legacy win iso path

solenoci:
Issue: There was a mismatch between filepath in code and in filepath in containerfile

Fix: Move the drivers to expected location

Ref: https://issues.redhat.com/browse/MTV-3227


mnecas (approved):

solenoci:
Do not merge, we are missing mtv-csv.yaml which I will add once 2.8 is ready

PR #2691, opened 2025-09-01
Update .coderabbit.yaml

borod108:
Auto incremental reviews create too much noise, a review can be triggered by `@coderabbitai review`

### DIFF
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -47,5 +47,5 @@ reviews:
         - **coverage**: Make sure that the code has unit tests.
   auto_review:
     enabled: true
-    auto_incremental_review: true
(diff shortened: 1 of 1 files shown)


.coderabbit.yaml
```diff
@@ -47,5 +47,5 @@ reviews:
         - **coverage**: Make sure that the code has unit tests.
   auto_review:
     enabled: true
-    auto_incremental_review: true
-    drafts: false  # ignore draft PRs
\ No newline at end of file
+    auto_incremental_review: false
```
mnecas: Could we disable the auto_review?
it can be bit spammy and I would leave it up to devs to manage their own PRs

PR #2692, opened 2025-09-01
MTV-3264 | Allow dedicated ESXi host for storage offload

rgolangh:
https://issues.redhat.com/browse/MTV-3264

MOTIVATION:
In storage-offloaded migrations (XCOPY), the populator pod defaults to using the same ESXi host where the source VM is running to perform the disk cloning operation. This can lead to resource contention on the host, impacting the performance of the source VM and other workloads. To provide better control and load distribution, this change introduces the ability to specify a dedicated set of ESXi hosts for these migration tasks.

MODIFICATION:
- A new `dedicatedOffloadMigrationHosts` setting was added to the vSphere Provider spec. This setting accepts a comma-separated list of ESXi host IDs.
- The Plan builder was updated to read this list. If populated, it randomly selects a host and sets it in the `MigrationHost` field of the `VSphereXcopyVolumePopulator` CR.
- The `VSphereXcopyVolumePopulatorSpec` API was extended with the `MigrationHost` field.
- The volume populator binary now accepts a `--migration-host` flag.
- The core populator logic was modified:
  - If a `migrationHost` is provided, the populator finds the ESXi host by its ID and uses it to perform the XCOPY operation.
  - If no `migrationHost` is provided, it retains the existing behavior of finding and using the host where the source VM is registered.
- A new `GetEsxById` method was added to the vSphere client to facilitate finding hosts by their managed object ID.
- Tests and mocks were updated to cover the new functionality.
- The Makefile and build tooling were updated to use `go.uber.org/mock/mockgen`.

RESULT:
Users can now configure a dedicated pool of ESXi hosts for storage offload migrations, allowing for better resource management and load balancing. This prevents the migration process from impacting the source VM's host and provides greater flexibility in configuring the migration environment.

## Summary by CodeRabbit

* **New Features**
  * Add migrationHost option (CLI flag and CRD field) to target a specific ESX for storage-offload; controller can read provider settings to randomly assign a dedicated host per disk.

* **Documentation**
  * CRDs and manifests updated to expose migrationHost; installLegacyDrivers ISO path description updated.

* **Tests**
  * New unit tests covering migrationHost and CLI invocation paths.

* **Chores**
  * Add vendoring step and pin mockgen tooling; simplify Makefile and update test invocation to include migration-host.


cmd/populator-controller/populator-controller.go
```diff
@@ -149,6 +149,10 @@ func getVXPopulatorPodArgs(_ bool, u *unstructured.Unstructured, pvc corev1.Pers
 		"--secret-name=" + xcopy.Spec.SecretName,
 		"--storage-vendor-product=" + xcopy.Spec.StorageVendorProduct,
 	}
+	if xcopy.Spec.MigrationHost != "" {
+		args = append(args, "--migration-host="+xcopy.Spec.MigrationHost)
+
+	}
```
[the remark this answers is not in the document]
rgolangh: done

pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"math/rand"
```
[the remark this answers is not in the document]
rgolangh: debating on that. currently we'll start with rando selection

pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -1239,6 +1240,10 @@
 
 				r.Log.Info(fmt.Sprintf("getting storage mapping by storage class %q and datastore %v datastore name %s datastore", storageClass, disk.Datastore, disk.Datastore))
 				vsphereInstance := r.Context.Plan.Provider.Source.GetName()
+				migrationHosts := []string{}
+				if v, ok := r.Source.Provider.Spec.Settings[api.DedicatedOffloadMigrationHosts]; ok {
+					migrationHosts = strings.Split(v, ",")
+				}
 				storageVendorProduct := mapped.OffloadPlugin.VSphereXcopyPluginConfig.StorageVendorProduct
```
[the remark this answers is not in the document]
rgolangh: done

operator/config/crd/bases/forklift.konveyor.io_plans.yaml
```diff
@@ -1078,6 +1078,14 @@ spec:
...
                 type: object
                 x-kubernetes-map-type: atomic
+              dedicatedOffloadMigrationHosts:
+                description: |-
+                  DedicatedOffloadMigrationHosts is a list of ESXi host IDs dedicated
+                  to storage offload (XCOPY) migrations. Overrides the provider-level
+                  setting if set.
+                items:
+                  type: string
+                type: array
```
mnecas, line 1088: Think the `VSphereXcopyPluginConfig` inside StorageMapping  would be better fit for this.
Because that is tied to the storage mapping and we can say that in "this specific mapping these host can be used for xcopy". Also this new parameter would only affect the xcopy not other plans

pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -1484,6 +1489,16 @@ func (r *Builder) PopulatorVolumes(vmRef ref.Ref, annotations map[string]string,
...
 				pvc.Annotations[planbase.AnnSecret] = diskSecretName
 				pvcs = append(pvcs, &pvc)
+
+				migrationHost := ""
+				if len(migrationHosts) > 0 {
+					// randomly select one
+					// rgolan todo - preffer the current host of the VM if in the list 
+					i := rand.Intn(len(migrationHosts))
+					migrationHost = migrationHosts[i]
+					// a stable method could be: hash(vm.id) % len(migrationHosts)
```
mnecas, line 1499: Think we should also add host information to the `Status` of the `VSphereXcopyVolumePopulator`. With this information we can could get the list of active `VSphereXcopyVolumePopulator` and their hosts and we could build a load balancer so we could distribute the migrations across non used hosts.
randint would work well on large sample but we could end up with just few disks returning the same host multiple times.

mnecas (changes requested):

operator/config/crd/bases/forklift.konveyor.io_plans.yaml
rgolangh: this means you want it scoped only for offload, and not even on the provider level, right?  
I thought this is a feature that will be useful not only for offload. 
Guess that feature could evolve later to something more generic. just that that will be an api change from the mapping to the plan.

pkg/controller/plan/adapter/vsphere/builder.go
rgolangh: The ESX node name is added already in `spec.MigrationHost` - did you have in mind anything else?

operator/config/crd/bases/forklift.konveyor.io_plans.yaml
mnecas: As we already have dediced `Host` for normal migration where we specify which ESXi NIC should be used for the communication, but we want to say "these hosts should be used for xcopy as they are attached to the needed storage". So we would need to either add the storage information to the `Host` or we need to add the host list to the storage mapping. As this is XCOPY specific i would move it to the `VSphereXcopyPluginConfig`

pkg/controller/plan/adapter/vsphere/builder.go
mnecas: hmm yeah we could use the `spec.MigrationHost` and list "active" `VSphereXcopyVolumePopulator` to know how many of them are running on each node so we could do this balancing.

operator/config/crd/bases/forklift.konveyor.io_plans.yaml
rgolangh: Ack, moving into VsphereXcopyPluginConfig

pkg/controller/plan/adapter/vsphere/builder.go
rgolangh: host could down aftger we set them on the storage map. Could be that that storage map is global and is copied info each plan.   
So not sure we need to validate in the moment we add them.

rgolangh:
@amitosw15 I want to hand you over this PR. Going to give you permissions on it

pkg/controller/plan/adapter/vsphere/builder.go
amitosw15: Since MigrationHost is already on the populator CR's Spec, we can query active populators and count per-host without adding a Status field. I'll open a follow-up ticket for this, it's a separate concern from the host list configuration itself

pkg/controller/plan/adapter/vsphere/builder.go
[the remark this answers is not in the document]
amitosw15: Agreed that `rand.Intn` can cause flapping across reconciliations and uneven distribution with small host pools. We're planning to replace this with proper load balancing — listing active `VSphereXcopyVolumePopulator` CRs and picking the least-loaded host — in a follow-up. Stable hashing would be an improvement over random but still wouldn't account for actual load, so we'll skip straight to the load-aware approach.

amitosw15:
@mnecas 
any comments?

operator/config/crd/bases/forklift.konveyor.io_plans.yaml
amitosw15: done

mnecas (approved):
lgtm

PR #2705, opened 2025-09-03
mtv - 2930 | Make SupportsVolumePopulators per-VM instead of per-plan

amitosw15:
Fix mixed datastore configuration issue with per-VM SupportsVolumePopulators

  - Change SupportsVolumePopulators interface to accept vmRef parameter
  - Implement per-VM logic in vSphere builder to check only VM's datastores
  - Update all provider implementations to match new interface
  - Update call sites in migration.go to pass vm.Ref

  This prevents the mixed configuration bug (MTV-2930) where VMs with
  disks on mixed datastores (some with offload, some without) would
  cause endless populator pod creation and resource exhaustion.


amitosw15:
adding unit-testing, meanwhile pr is draft

amitosw15:
open new pr instead

PR #2712, opened 2025-09-05
storage-offload: Fix device discovery timeout

mnecas:
Issue:
When the device is mapped to the ESXi host, it can take some time to get discovered by the kernel. Right now, we wait only if the rescan fails, but in cases where the scan passes, we immediately check for the device.

Fix:
Always add timeout for the next device discovery try.


rgolangh:
I think we started seeing those more and more, that the scan passes and the clone fails, because the path wasn't ready to be used yet. 
+1

PR #2743, opened 2025-09-09
MTV-3283 | Show all NADs in the cluster

mnecas:
Issue:
We need to show targetNamespace specific NADs. As we don't have it accessible the plan information in the inventory we need to show all NADs across all namespaces.

Fix:
Show all NADs in the cluster.

## Summary by CodeRabbit

- Refactor
  - Streamlined how network attachments are listed by consolidating multiple queries into a single request.
  - Removed redundant parameters in internal calls for cleaner, more maintainable code.
  - Preserved existing filtering and error handling to ensure consistent behavior.
  - Users may see modest performance improvements and more reliable listing results, with no changes required to workflows.


mnecas:
Before: 
Screenshot From 2025-09-09 10-32-01

After:
Screenshot From 2025-09-09 10-31-20

yaacov:
@mnecas the design was that we only show the provider namespace and default, IMHO we should not change that because that will open the security issue again, MTV-3283 is already fixed in main and in 2.9.z

mnecas:
Closing as this was fixed properly by https://github.com/kubev2v/forklift/pull/2615

PR #2777, opened 2025-09-11
MTV-3394 | storage-offload: fix the rescan handling

rgolangh:
Fix the rescan logic so it fails in case the rescan succeeds, but the
device is still not detected.

Also, fix the device argument to be only the naa.XYZ (or any other world
wide name) of the device, and not the /path/to/naa.XYZ as the 'esxcli
storage core device list -d XXX' doesn't handle the path.

https://issues.redhat.com/browse/MTV-3394
Blocks: https://issues.redhat.com/browse/MTV-3368

## Summary by CodeRabbit

- New Features
  - Added support for the PowerStore storage vendor in StorageMap configuration.

- Improvements
  - Encapsulated and improved ESXi rescan flow after volume mapping, with retries, clearer logging, and final verification to ensure devices are visible before proceeding.

- Tests
  - Added comprehensive tests covering rescan scenarios (immediate success, retries, eventual success, and failures).
  - Adjusted existing tests to reflect refined device argument handling.

### DIFF
--- a/cmd/vsphere-xcopy-volume-populator/internal/populator/remote_esxcli.go
+++ b/cmd/vsphere-xcopy-volume-populator/internal/populator/remote_esxcli.go
@@ -8,12 +8,17 @@ import (
 	"time"
 
 	"github.com/kubev2v/forklift/cmd/vsphere-xcopy-volume-populator/internal/vmware"
+	"github.com/vmware/govmomi/object"
 	"k8s.io/klog/v2"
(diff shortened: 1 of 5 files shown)


rgolangh:
@amitosw15 Please take a look

rgolangh:
there should be another Pure-specific fix here. Perhaps I'll create another ticket for that or maybe add it on top

PR #2779, opened 2025-09-12
MTV-3400 | Add NBDE clevis unlock

mnecas:
Add a new parameter to the forklift plan to allow us to migrate VMs with [NBDE](https://access.redhat.com/articles/6987053) (Network Bound Disk Encryption).
For this, the virt-v2v uses the `--key all:clevis` that calls clevis inside the guestfs to unlock the [disks](https://github.com/libguestfs/libguestfs/blob/master/daemon/clevis-luks.c#L41). The main requirement is to have access from the guest conversion pod to the Tang server. 
This option is under `plan.spec.vm[].nbdeClevis` the option is boolean value and it takes priority over the LUKS, ut we will give warning about it.

Ref: https://issues.redhat.com/browse/MTV-3400

## Summary by CodeRabbit

* **New Features**
  * Added Clevis (NBDE) disk decryption support for VM migrations and hosts.
  * New nbdeClevis option available in VM/Plan/Migration/Host specs and configurable via env var / CLI flag.

* **Improvements**
  * Clevis now takes precedence when both LUKS and nbdeClevis are set.
  * Validation warns when both LUKS and Clevis are configured.

* **Documentation**
  * Updated CRD descriptions and precedence notes (including network name template clarifications).

### DIFF
--- a/pkg/apis/forklift/v1beta1/plan/vm.go
+++ b/pkg/apis/forklift/v1beta1/plan/vm.go
@@ -43,6 +43,13 @@ type VM struct {
 	// Disk decryption LUKS keys
 	// +optional
 	LUKS core.ObjectReference `json:"luks" ref:"Secret"`
+	// Attempt passphrase-less unlocking for all devices with Clevis, over the network.
+	// Conversion pod running on target cluster will attempt to connect to a TANG server, make sure TANG
+	// server is available on target network.
+	// https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/8/html/security_hardening/configuring-automated-unlocking-of-encrypted-volumes-using-policy-based-decryption_security-hardening
+	// If both nbdeClevis and LUKS are configured, nbdeClevis takes precedence.
+	// +optional
+	NbdeClevis bool `json:"nbdeClevis,omitempty"`
 	// Choose the primary disk the VM boots from
 	// +optional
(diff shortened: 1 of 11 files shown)


pkg/apis/forklift/v1beta1/plan/vm.go
```diff
@@ -43,6 +43,10 @@ type VM struct {
 	// Disk decryption LUKS keys
 	// +optional
 	LUKS core.ObjectReference `json:"luks" ref:"Secret"`
+	// When set, disk decryption uses the Clevis (NBDE) client.
+	// If both nbdeClevis and LUKS are configured, nbdeClevis takes precedence.
```
yaacov:
```suggestion
	// Attempt passphrase-less unlocking for all devices with Clevis, over the network.
	// Conversion pod running on target cluster will attempt to connect to a TANG server, make sure TANG
	// server is available on target netwrok.
	// https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/8/html/security_hardening/configuring-automated-unlocking-of-encrypted-volumes-using-policy-based-decryption_security-hardening
	// If both nbdeClevis and LUKS are configured, nbdeClevis takes precedence.
```

mnecas: sounds good thank you, updated!

PR #2780, opened 2025-09-12
MTV-2858 | Add UDN IP address preservation

mnecas:
The UDN in the OCP 4.20.0 will have a new feature for setting IP addresses on the VMs. This will allow the forklift to specify which IP address the VM should have after migration. This way, we will be able to preserve the IP from the original provider. This requires that the IP address is from the same subnet. 
Forklift will preserve this with the `preserveStaticIPs` parameter. So, from the User Interface perspective, no configuration will change, but the flag will not only preserve static IPs but also set IP addresses in the UDNs.

Notes:
- Right now, we can not specify the default gateway, DNS,or  secondary UDNs

Ref: https://issues.redhat.com/browse/MTV-2858

## Summary by CodeRabbit

- New Features
  - Validates static IPs against User-Defined Network (UDN) subnets during plan checks and surfaces a warning when a VM’s IP doesn't match the primary UDN subnet.
  - Preserves and annotates per-interface IPv4 addresses when mapping to UDN networks with static IPs.

- Enhancements
  - Enriched vSphere VM NIC and guest-network metadata for improved network mapping and validation.
  - Added no-op UDN static-IP validation stubs across other providers to align APIs without changing behavior.

### DIFF
--- a/pkg/controller/plan/adapter/vsphere/builder.go
+++ b/pkg/controller/plan/adapter/vsphere/builder.go
@@ -3,6 +3,7 @@ package vsphere
 import (
 	"context"
 	"encoding/base64"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"net"
@@ -652,7 +653,6 @@ func (r *Builder) VirtualMachine(vmRef ref.Ref, object *cnv.VirtualMachineSpec, 
 	if err != nil {
 		return
 	}
-
 	if object.Template == nil {
 		object.Template = &cnv.VirtualMachineInstanceTemplateSpec{}
 	}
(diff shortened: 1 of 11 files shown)


pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -676,9 +676,27 @@ func (r *Builder) VirtualMachine(vmRef ref.Ref, object *cnv.VirtualMachineSpec,
 	return
 }
 
+func IsIPv4(address string) bool {
```
yaacov: do we need to export this functionality, maybe `isIPv4` is enough  ?

note, in another place we do it inline ?
https://github.com/kubev2v/forklift/blob/01a277d0cd04ae211ba6187b23012f92ce37699d/pkg/controller/plan/adapter/vsphere/builder.go#L323

yaacov (approved):
lgtm

pkg/controller/plan/adapter/vsphere/builder.go
mnecas: not sure if the net.IP.To4 will work with `nil`

PR #2782, opened 2025-09-12
MTV-3368 | storage-offload: Pure flasharray - fix the FC host lookup and matchup

rgolangh:
Pure flasharray stores the wwn on the host in this form:
21:00:00:0E:1E:1E:83:D0
While the adapter on the ESX is reporting:
fc.2100000e1e1e83d0:2100000e1e1e83d2

This fix handles that difference when matching. If the matching fails
and we don't match any hosts then the Populate function returns with an
err.

https://issues.redhat.com/browse/MTV-3368

## Summary by CodeRabbit

- Bug Fixes
  - Improved Fibre Channel host matching using WWPN normalization, added per-host diagnostics, stronger validation when mapping hosts, and tolerant behavior for already-existing connections.

- Tests
  - Added table-driven tests for FC UID → WWN conversion covering valid and various invalid inputs.

- Chores
  - Updated test invocation to use an SSH-based ESXi clone method and adjusted test identifiers; added new entries to .gitignore.

### DIFF
--- a/cmd/vsphere-xcopy-volume-populator/internal/pure/flashArray.go
+++ b/cmd/vsphere-xcopy-volume-populator/internal/pure/flashArray.go
@@ -41,31 +41,47 @@ func NewFlashArrayClonner(hostname, username, password string, skipSSLVerificati
 	return FlashArrayClonner{client: client, clusterPrefix: clusterPrefix}, nil
 }
 
-// EnsureClonnerIgroup creates or updates an initiator group with the clonnerIqn
+// EnsureClonnerIgroup creates or updates an initiator group with the ESX adapters
 // Named hgroup in flash terminology
-func (f *FlashArrayClonner) EnsureClonnerIgroup(initiatorGroup string, clonnerIqn []string) (populator.MappingContext, error) {
+func (f *FlashArrayClonner) EnsureClonnerIgroup(initiatorGroup string, esxAdapters []string) (populator.MappingContext, error) {
 	// pure does not allow a single host to connect to 2 separae groups. Hence
 	// we must connect map the volume to the host, and not to the group
-	hostNames := []string{}
 	hosts, err := f.client.Hosts.ListHosts(nil)
 	if err != nil {
 		return nil, err
 	}
 	for _, h := range hosts {
+		klog.Infof("checking host %s, iqns: %v, wwns: %v", h.Name, h.Iqn, h.Wwn)
 		for _, iqn := range h.Iqn {
-			if slices.Contains(clonnerIqn, iqn) {
+			if slices.Contains(esxAdapters, iqn) {
 				klog.Infof("adding host to group %v", h.Name)
-				hostNames = append(hostNames, h.Name)
+				return populator.MappingContext{"hosts": []string{h.Name}}, nil
 			}
 		}
 		for _, wwn := range h.Wwn {
-			if slices.Contains(clonnerIqn, wwn) {
-				klog.Infof("adding host to group %v", h.Name)
-				hostNames = append(hostNames, h.Name)
+			for _, hostAdapter := range esxAdapters {
+				if !strings.HasPrefix(hostAdapter, "fc.") {
+					continue
+				}
+				adapterWWPN, err := fcUIDToWWPN(hostAdapter)
+				if err != nil {
+					klog.Warningf("failed to extract WWPN from adapter %s: %s", hostAdapter, err)
+					continue
+				}
+
+				// The WWN from the Pure API needs to be formatted consistently for comparison
+				formattedHostWwn := strings.ReplaceAll(strings.ToUpper(wwn), ":", "")
+				formattedAdapterWwpn := strings.ReplaceAll(adapterWWPN, ":", "")
+
+				klog.Infof("comparing ESX adapter WWPN %s with Pure host WWN %s", formattedAdapterWwpn, formattedHostWwn)
+				if formattedAdapterWwpn == formattedHostWwn {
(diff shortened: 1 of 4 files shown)


cmd/vsphere-xcopy-volume-populator/internal/pure/flashArray.go
```diff
@@ -129,3 +143,31 @@ func (f *FlashArrayClonner) ResolvePVToLUN(pv populator.PersistentVolume) (popul
...
+// return the WWNN
+func fcUidToWWN(fcUid string) (string, error) {
+	formattedWwn, ok := strings.CutPrefix(fcUid, "fc.")
+	if !ok {
+		return "", fmt.Errorf("fcUid %q doesn't start with 'fc.'", fcUid)
+	}
+	if len(formattedWwn) == 0 {
+		return "", fmt.Errorf("fcUid %q should be in the form of 'fc.2020202020202020:2020202020202020' or 'fc.2020202020202020'", fcUid)
+	}
+	formattedWwn = strings.Split(formattedWwn, ":")[0]
```
mkilavuz: I checked a setup with FC. FlashArray's `purehost list` will list WWPNs only which is the second part of the FC UID. So it should be `strings.Split(formattedWwn, ":")[1]` instead.

The UIDs on esxi host was in `fc.wwnn.wwpn` format. To be honest I'm not sure if it could be fc.wwpn` format or not. In that case, it might be safe to accept that format and extract it.

cmd/vsphere-xcopy-volume-populator/internal/pure/flashArray.go
```diff
@@ -41,53 +41,67 @@ func NewFlashArrayClonner(hostname, username, password string, skipSSLVerificati
...
 		return nil, err
 	}
 	for _, h := range hosts {
+		klog.Infof("checking host %s, iqns: %v, wwns: %v", h.Name, h.Iqn, h.Wwn)
 		for _, iqn := range h.Iqn {
-			if slices.Contains(clonnerIqn, iqn) {
+			if slices.Contains(esxAdapters, iqn) {
 				klog.Infof("adding host to group %v", h.Name)
-				hostNames = append(hostNames, h.Name)
+				return populator.MappingContext{"hosts": []string{h.Name}}, nil
```
mkilavuz, line 58: Why are we returning here instead of collecting the list of hosts with matching adapters?
Same below for wwn.

cmd/vsphere-xcopy-volume-populator/internal/pure/flashArray.go
rgolangh: is a port name unique to a node name?

mkilavuz: [Wikipedia](https://en.wikipedia.org/wiki/World_Wide_Port_Name) says WWPNs are unique in the network, and they are like MAC addresses. I've also seen in other sources that these are universally unique. I assume `World Wide` also emphasizes that. Without the WWNN, WWPN by itself should be enough to refer to the FC port.

MikeAnders08:
What's the status on this PR?

cmd/vsphere-xcopy-volume-populator/internal/pure/flashArray.go
```diff
@@ -41,53 +41,67 @@ func NewFlashArrayClonner(hostname, username, password string, skipSSLVerificati
...
 			}
 		}
 		for _, wwn := range h.Wwn {
-			if slices.Contains(clonnerIqn, wwn) {
-				klog.Infof("adding host to group %v", h.Name)
-				hostNames = append(hostNames, h.Name)
+			for _, hostAdapter := range esxAdapters {
+				if !strings.HasPrefix("fc.", hostAdapter) {
```
MikeAnders168: Should be:

`if !strings.HasPrefix(hostAdapter, "fc.") {`

Otherwise it will would always return false?

cmd/vsphere-xcopy-volume-populator/internal/pure/flashArray_test.go
```diff
@@ -0,0 +1,81 @@
...
+			fcUid:       "fc.2020202020202020:2020202020202020",
+			expectedWwn: "20:20:20:20:20:20:20:20",
+			expectError: false,
+		},
+		{
+			name:        "valid fcUid with no WWNP",
+			fcUid:       "fc.2020202020202020",
+			expectedWwn: "20:20:20:20:20:20:20:20",
+			expectError: false,
+		},
```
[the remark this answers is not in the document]
rgolangh: done

cmd/vsphere-xcopy-volume-populator/internal/pure/flashArray.go
rgolangh: In the end we have a single host, a single ESX, that needs to be matched and not multiple.

cmd/vsphere-xcopy-volume-populator/internal/pure/flashArray.go
rgolangh: done

cmd/vsphere-xcopy-volume-populator/internal/pure/flashArray.go
rgolangh: In that case, where Pure stores the PoreName, then I make sure to match it with PortName I parse from the HBA, which is like you mentioned, the last part of the split.

mkilavuz: Yes. I will try to get a setup to test this end-to-end soon. Unfortunately, I don't have anything readily available.

PR #2806, opened 2025-09-15
Fix: powerstore map and unmap functions

amitosw15:
Fix PowerStore mapping to use real host names instead of logical names
    
    The Map and UnMap functions were ignoring the initiatorGroup parameter
    and always operating on the ESXi host. This caused the xcopy process
    to leave volumes mapped to the ESXi host instead of the intended target,
    which is unwanted behavior.
    
    The root cause was that PowerStore initiators can only be mapped to one
    host, but the logical host name used in remote_excli (the initiator
    group name) may differ from the actual PowerStore host name used in
    production environments.
    
    Solution: we store the logical name and the real name as follows:
    - esxLogicalHostName: The initiator group name in esxi
    - esxRealHostName: The actual PowerStore host name for ESXi
    
    Now Map and UnMap resolve hosts by their real PowerStore host names
    rather than using logical names. The CurrentMappedGroups function also
    returns early if a volume is already mapped to avoid unnecessary work.
    
    Resolves: MTV-3501

### DIFF
--- a/cmd/vsphere-xcopy-volume-populator/internal/powerstore/powerstore.go
+++ b/cmd/vsphere-xcopy-volume-populator/internal/powerstore/powerstore.go
@@ -3,7 +3,6 @@ package powerstore
 import (
 	"context"
 	"fmt"
-	"slices"
 	"strings"
 
 	"github.com/dell/gopowerstore"
@@ -12,9 +11,9 @@ import (
 )
 
 const (
-	initiatorGroupContextKey string = "initiatorGroup"
-	hostIDContextKey         string = "hostID"
-	hostNameContextKey       string = "hostName"
+	hostIDContextKey      string = "hostID"
+	esxLogicalHostNameKey string = "esxLogicalHostName"
+	esxRealHostNameKey    string = "esxRealHostName"
 )
 
 type PowerstoreClonner struct {
@@ -34,6 +33,7 @@ func (p *PowerstoreClonner) CurrentMappedGroups(targetLUN populator.LUN, mapping
 	}
 
 	mappedHosts := make([]string, 0, len(mappings))
+
 	for _, mapping := range mappings {
 		host, err := p.Client.GetHost(ctx, mapping.HostID)
 		if err != nil {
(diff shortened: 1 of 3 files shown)


cmd/vsphere-xcopy-volume-populator/internal/powerstore/powerstore.go
```diff
@@ -34,27 +34,21 @@ func (p *PowerstoreClonner) CurrentMappedGroups(targetLUN populator.LUN, mapping
...
+
 	for _, mapping := range mappings {
 		host, err := p.Client.GetHost(ctx, mapping.HostID)
 		if err != nil {
 			klog.Warningf("Failed to get host info for host ID %s: %s", mapping.HostID, err)
 			continue
 		}
 		mappedHosts = append(mappedHosts, host.Name)
+		if _, ok := mappingContext[ocpRealHostNameKey]; !ok {
+			mappingContext[ocpRealHostNameKey] = host.Name
```
rgolangh: this looks like it is going to be overwritten over an over again. why not just return right after that line?

rgolangh: I mention this because I see mappingContext[ocpRealHNostNameKey] will eventually be only 1 entry.

cmd/vsphere-xcopy-volume-populator/internal/powerstore/powerstore.go
```diff
@@ -218,19 +217,30 @@ func (p *PowerstoreClonner) Map(initiatorGroup string, targetLUN populator.LUN,
...
 
-	hostID, ok := mappingContext[hostIDContextKey].(string)
-	if !ok || hostID == "" {
-		return targetLUN, fmt.Errorf("host ID not found in mapping context")
-	}
 	klog.Infof("mapping volume %s to initiator-group %s", targetLUN.Name, initiatorGroup)
 
 	ctx := context.Background()
+	hostName := ""
+	if initiatorGroup != mappingContext[esxLogicalHostNameKey] {
```
sid-dell: Hi team, I noticed the use of esxLogicalHostNameKey, esxRealHostNameKey, and ocpRealHostNameKey in the powerstore.go logic. These concepts don’t exist in Dell PowerStore’s native API. Could you clarify the reasoning behind introducing them? Are they intended purely for internal abstraction within Forklift to handle multi-platforms (like ESXi and OCP) mapping logic, or is there a specific integration scenario that necessitated this separation?
Thanks!

amitosw15: Hey @sid-dell 
I used those concepts because of the limitation of the array , where an initiator can be mapped to one host. 
In the file [remote_esxcli.go](https://github.com/kubev2v/forklift/blob/796e7ddcd97f2f73816a2a855c920e00476d57ab/cmd/vsphere-xcopy-volume-populator/internal/populator/remote_esxcli.go#L15) we have a fixed variable (`xcopy-esxs`) for the required host name in powerstore whereas the real name used can be anything that the administrator has defined.
The usage of this abstraction allow us to use the existing name without creating new hosts in powerstore or changing mapping.

cmd/vsphere-xcopy-volume-populator/internal/powerstore/powerstore.go
```diff
@@ -12,9 +11,10 @@ import (
 )
 
 const (
-	initiatorGroupContextKey string = "initiatorGroup"
-	hostIDContextKey         string = "hostID"
-	hostNameContextKey       string = "hostName"
+	hostIDContextKey      string = "hostID"
+	esxLogicalHostNameKey string = "esxLogicalHostName"
+	esxRealHostNameKey    string = "esxRealHostName"
+	ocpRealHostNameKey    string = "ocpRealHostName"
```
rgolangh: like you mentioned, not really needed

sid-dell: are you guys planning to change this PR?

amitosw15:
@sid-dell @rgolangh 
PR ready, fixed with new comments

cmd/vsphere-xcopy-volume-populator/internal/powerstore/powerstore.go
amitosw15: yep, please check the changes

PR #2873, opened 2025-09-17
MTV-2427 | Add preflight inspection step

solenoci:
Adds a step to the warm migration pipeline so it fails early instead of only after the disk transfer.

Resolves: MTV-2427

Ref: https://issues.redhat.com/browse/MTV-2427

## Summary by CodeRabbit

* **New Features**
  * Optional preflight inspection before the first disk transfer for warm migrations (Plan toggle, default: true).
  * Remote disk inspection mode via CLI flag / environment variables to inspect remote VM disks before migration.
  * New migration phase and pod workflow to run and report preflight inspections prior to conversion/customization.

* **Improvements**
  * vSphere disk metadata now includes parent-file information for clearer disk lineage.
  * VM customization now runs after inspection data is obtained.

### DIFF
--- a/pkg/controller/plan/kubevirt.go
+++ b/pkg/controller/plan/kubevirt.go
@@ -2135,6 +2135,259 @@ func (r *KubeVirt) guestConversionPod(vm *plan.VMStatus, vmVolumes []cnv.Volume,
 	return
 }
 
+func (r *KubeVirt) EnsureGuestInspectionPod(vm *plan.VMStatus) (pod *core.Pod, err error) {
+	var volumes []core.Volume
+	var mounts []core.VolumeMount
+
+	labels := r.vmLabels(vm.Ref)
+
+	// Create secret for virt-v2v inspection pod
+	v2vSecret, err := r.ensureSecret(vm.Ref, r.secretDataSetterForCDI(vm.Ref), labels)
+	if err != nil {
+		return
+	}
+
+	// Create vddk aio configmap if desired
+	var vddkConfigMap *core.ConfigMap
+	if r.Source.Provider.UseVddkAioOptimization() {
+		vddkConfigMap, err = r.ensureVddkConfigMap()
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	if vddkConfigMap != nil {
+		mounts = append(mounts,
+			core.VolumeMount{
+				Name:      VddkConf,
+				MountPath: fmt.Sprintf("/mnt/%s", VddkConf),
+			},
+		)
+		volumes = append(volumes, core.Volume{
+			Name: VddkConf,
+			VolumeSource: core.VolumeSource{
+				ConfigMap: &core.ConfigMapVolumeSource{
+					LocalObjectReference: core.LocalObjectReference{
+						Name: vddkConfigMap.Name,
+					},
+				},
+			},
+		})
+	}
+
+	// mount the secret for the password and CA certificate
+	volumes = append(volumes, core.Volume{
+		Name: "secret-volume",
+		VolumeSource: core.VolumeSource{
+			Secret: &core.SecretVolumeSource{
+				SecretName: v2vSecret.Name,
+			},
+		},
+	})
+
+	// add volume and mount for the libvirt domain xml config map.
+	// the virt-v2v pod expects to see the libvirt xml at /mnt/v2v/input.xml
+	// libvirtConfigMap, err := r.ensureLibvirtConfigMap(vm.Ref, vmCr, []*core.PersistentVolumeClaim{})
+	// if err != nil {
+	// 	return
+	// }
+
+	// volumes = append(volumes, core.Volume{
+	// 	Name: "libvirt-domain-xml",
+	// 	VolumeSource: core.VolumeSource{
+	// 		ConfigMap: &core.ConfigMapVolumeSource{
+	// 			LocalObjectReference: core.LocalObjectReference{
+	// 				Name: libvirtConfigMap.Name,
+	// 			},
+	// 		},
+	// 	},
+	// })
+
+	// Temporary space for VDDK library
+	volumes = append(volumes, core.Volume{
+		Name: VddkVolumeName,
+		VolumeSource: core.VolumeSource{
+			EmptyDir: &core.EmptyDirVolumeSource{},
+		},
+	})
+
+	mounts = append(mounts,
+		core.VolumeMount{
+			Name:      "secret-volume",
+			ReadOnly:  true,
+			MountPath: "/etc/secret",
+		},
+		// core.VolumeMount{
+		// 	Name:      "libvirt-domain-xml",
+		// 	MountPath: "/mnt/v2v",
+		// },
+		core.VolumeMount{
+			Name:      "vddk-vol-mount",
+			MountPath: "/opt",
+		},
+	)
+
+	extraConfigMapExists := len(Settings.Migration.VirtV2vExtraConfConfigMap) > 0
+	if extraConfigMapExists {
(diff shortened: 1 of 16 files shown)


pkg/controller/plan/migrator/base/migrator.go
```diff
@@ -264,6 +275,13 @@ func (r *BaseMigrator) Step(status *plan.VMStatus) (step string) {
 		} else {
 			step = Initialize
 		}
+	case api.PhasePreflightInspection:
+		if !r.Context.Plan.Spec.SkipGuestConversion &&
+			r.Context.Plan.Spec.RunPreflightInspection {
+			step = PreflightInspection
+		} else {
+			step = Initialize
+		}
```
mnecas: Think we can have  `step = PreflightInspection` and hide this step in the flags. Please take a look at 
https://github.com/kubev2v/forklift/blob/e2eeccf04e30297b09bf15eecb2a4cc2d5517afd/pkg/controller/plan/migrator/base/migrator.go#L365

pkg/controller/plan/migrator/base/migrator.go
```diff
@@ -194,6 +194,17 @@ func (r *BaseMigrator) Pipeline(vm plan.VM) (pipeline []*plan.Step, err error) {
 						Progress:    libitr.Progress{Total: 1},
 					},
 				})
+		case api.PhasePreflightInspection:
+			pipeline = append(
+				pipeline,
+				&plan.Step{
+					Task: plan.Task{
+						Name:        PreflightInspection,
+						Description: "Inspect disks before migration.",
```
mnecas: Inspect VM before migration

pkg/virt-v2v/config/variables.go
```diff
@@ -28,6 +29,8 @@ const (
 	EnvHostName                   = "V2V_HOSTNAME"
 	EnvNbdeClevis                 = "V2V_NBDE_CLEVIS"
 	EnvMultipleIpsPerNicName      = "V2V_multipleIPsPerNic"
+	EnvRemoteInspection           = "V2V_remoteInspection"
+	EnvRemoteInspectionDisk       = "V2V_remoteInspectDisk_"
```
mnecas, line 33: I don't see where this is specified in the controller

pkg/controller/plan/migration.go
```diff
@@ -1189,6 +1189,57 @@ func (r *Migration) execute(vm *plan.VMStatus) (err error) {
 			if step.MarkedCompleted() && !step.HasError() {
 				r.NextPhase(vm)
 			}
+		case api.PhasePreflightInspection:
+			// Skip this phase if conversion is also skipped or preflight check is disabled
+			if r.Plan.Spec.SkipGuestConversion || !r.Context.Plan.Spec.RunPreflightInspection {
```
mnecas: This should be done in the step flags

pkg/virt-v2v/conversion/conversion.go
```diff
@@ -224,3 +224,27 @@ func (c *Conversion) RunCustomize(osinfo utils.InspectionOS) error {
 	custom := customize.NewCustomize(c.AppConfig, disks, osinfo)
 	return custom.Run()
 }
+
+func (c *Conversion) RunRemoteV2vInspection() (err error) {
+	v2vCmdBuilder := c.CommandBuilder.New("virt-v2v-inspector").
```
mnecas, line 229: try to add sleep before this to wait for vmware might help with the nbdkit issues.

pkg/controller/plan/kubevirt.go
```diff
@@ -2125,6 +2125,259 @@ func (r *KubeVirt) guestConversionPod(vm *plan.VMStatus, vmVolumes []cnv.Volume,
...
+
+	// Add disks to be inspected
+	for i, disk := range virtualMachine.Disks {
+		environment = append(environment, core.EnvVar{
+			Name:  fmt.Sprintf("V2V_remoteInspectDisk_%d", i),
+			Value: fmt.Sprintf("\"%s\"", disk.ParentFile),
+		})
+	}
+
+	pod = &core.Pod{
```
mnecas: could we reuse the existing parts? looking at this now i'm worried that we will miss one of these configs
I know we will need to refactor it in future. but worried to miss something now or in future

pkg/controller/plan/migration.go
solenoci: Modified

pkg/controller/plan/migrator/base/migrator.go
solenoci: Modified

pkg/controller/plan/migrator/base/migrator.go
solenoci: Modified

pkg/virt-v2v/config/variables.go
solenoci: Modified

pkg/virt-v2v/conversion/conversion.go
solenoci: NO-OP

pkg/controller/plan/kubevirt.go
solenoci: Modified

solenoci:
I had to force push over the changes for some reason, but at least I separated the commit that addresses the review comments.

pkg/controller/plan/migration.go
```diff
@@ -1237,6 +1237,51 @@ func (r *Migration) execute(vm *plan.VMStatus) (err error) {
...
+			pod, err = r.getInspectionPod(vm)
+
+			if err != nil {
+				step.AddError(err.Error())
+				err = nil
+				break
+			}
+
+			if pod == nil {
+				return
```
mnecas, line 1273: please add log

mnecas (approved):

pkg/controller/plan/kubevirt.go
```diff
@@ -974,17 +986,20 @@ func (r *KubeVirt) getListOptionsNamespaced() (listOptions *client.ListOptions)
 	}
 }
 
-// Ensure the guest conversion (virt-v2v) pod exists on the destination.
-func (r *KubeVirt) EnsureGuestConversionPod(vm *plan.VMStatus, vmCr *VirtualMachine, pvcs []*core.PersistentVolumeClaim) (err error) {
```
mansam, line 978: I would recommend that rather than adding branches to this method (and the methods that it calls), to create new methods explicitly for inspection. The branching makes it harder to follow the logic and it will be harder to eventually refactor into the migrator pattern. A small amount of code duplication is worth it to avoid extra branching.

pkg/controller/plan/migration.go
```diff
@@ -1237,6 +1237,52 @@ func (r *Migration) execute(vm *plan.VMStatus) (err error) {
 			if step.MarkedCompleted() && !step.HasError() {
 				r.NextPhase(vm)
 			}
+		case api.PhasePreflightInspection:
+			step, found := vm.FindStep(r.migrator.Step(vm))
+			if !found {
+				vm.AddError(fmt.Sprintf("Step '%s' not found", r.migrator.Step(vm)))
+				break
+			}
+			step.MarkStarted()
```
mansam, line 1246: Manually marking the step started and setting the phase should be unnecessary. `NextPhase()` should take care of marking steps started and completed.

pkg/controller/plan/kubevirt.go
solenoci: We agreed with @mnecas that it would be better to do it this way for now and refactor little bit later so we don't have a huge PR that adds a new feature and also refactors the old one.

PR #2879, opened 2025-09-18
MTV-3473 | Remove the fetch before plan update

mnecas:
Issue:
When the plan starts to reconcile it adds condition `The plan is EXECUTING`. But when we try to update the new resrouce the old condition the statuis gets removed.
So in the next reconcile the condintion is missing as we don't add it to the condition.
```
if snapshot.HasAnyCondition(api.ConditionExecuting, api.ConditionSucceeded, api.ConditionFailed, api.ConditionCanceled) {
	return
}
r.Plan.Status.Migration.MarkReset()
r.Plan.Status.Migration.MarkStarted()
snapshot.SetCondition(
	libcnd.Condition{
		Type:     api.ConditionExecuting,
		Status:   True,
		Category: api.CategoryAdvisory,
		Message:  "The plan is EXECUTING.",
		Durable:  true,
	})
```
This causes the update to remove this condition and restart the whole plan.

Fix:
Remove the fetch before plan update.

## Summary by CodeRabbit

- Bug Fixes
  - Reduced intermittent failures when updating plan status, improving reliability and consistency of status reporting.
  - More accurate status generation tracking to prevent stale or conflicting state.

- Refactor
  - Simplified status update flow to apply changes directly from local state, decreasing update latency and complexity.

- Chores
  - Improved error logging to include plan context for easier troubleshooting.

### DIFF
--- a/pkg/controller/plan/controller.go
+++ b/pkg/controller/plan/controller.go
@@ -276,32 +276,15 @@ func (r *Reconciler) updatePlanStatus(plan *api.Plan) error {
 	// Record events.
 	r.Record(plan, plan.Status.Conditions)
 
-	// Try to fetch latest version first to get current ResourceVersion
-	latest := &api.Plan{}
-	key := client.ObjectKeyFromObject(plan)
-	if err := r.Get(context.TODO(), key, latest); err != nil {
-		r.Log.V(1).Info("Failed to fetch latest plan version, trying update with current version",
-			"plan", plan.Name, "error", err.Error())
-
-		// Fallback: try update with current plan
-		plan.Status.ObservedGeneration = plan.Generation
-		if updateErr := r.Status().Update(context.TODO(), plan.DeepCopy()); updateErr != nil {
-			r.Log.Error(updateErr, "Failed to update plan status with current version")
-			return updateErr
-		}
-		return nil
-	}
-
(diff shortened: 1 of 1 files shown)


mnecas:
Pertial revert of https://github.com/kubev2v/forklift/pull/2348

mnecas:
This PR switches back to the reconciliation mechanism if there are problems with the update

pkg/controller/plan/controller.go
```diff
@@ -276,32 +276,14 @@ func (r *Reconciler) updatePlanStatus(plan *api.Plan) error {
...
-	latest.Status.ObservedGeneration = latest.Generation
+	// Apply changes.
+	plan.Status.ObservedGeneration = plan.Generation
 
 	// At this point, the plan contains data that is not persisted by design, like the Referenced data
 	// and the staged flags in the status, and more data that has been loaded in the validate function,
 	// like the name of the VMs in the spec section, therefore we don't want the plan to be overridden
 	// by data from the server (even the spec section is overridden) and so we pass a copy of the plan
-	if err := r.Status().Update(context.TODO(), latest.DeepCopy()); err != nil {
-		r.Log.Error(err, "Failed to update plan status with latest ResourceVersion")
```
yaacov, line 304: not sure we need to remove the log, maybe just rephrase it ?

PR #2940, opened 2025-09-25
Only attempt to update Plan status once per reconcile

mansam:
The Plan status was getting updated before the call to `execute()`, and then once during or after `execute()` as well. This frequently results in conflicts during the second attempt to update the status, and updating the status after `execute()` returns with error can result in the Plan entering an inconsistent state. (for instance, if the migration runner returned with an error after the `Executing` condition had been set, but before `begin()` had completed successfully.)

The check for VMs that have been deleted from the inventory while the plan is being processed should occur outside of the scheduler so that the scheduler implementations don't need to be responsible for setting the `Canceled` condition on the VM.

This also fixes a bug where the `Canceled` condition was being set on the missing VM's status, but the `NotFound` error from the inventory was still being passed to the caller, causing the reconcile to fail. The `NotFound` error should be set to `nil` after the VM's status is set so that plan execution can continue. 

## Summary by CodeRabbit

* New Features
  * None

* Bug Fixes
  * Migrations no longer abort when a VM is missing; the VM is marked as canceled with an advisory message (“VM was not found in inventory.”), allowing the run to continue.

* Refactor
  * Streamlined plan status updates to occur once upon successful completion, reducing redundant updates and improving reliability.

### DIFF
--- a/pkg/controller/plan/controller.go
+++ b/pkg/controller/plan/controller.go
@@ -248,19 +248,11 @@ func (r Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (r
 	// End staging conditions.
 	plan.Status.EndStagingConditions()
 
-	if err = r.updatePlanStatus(plan); err != nil {
-		r.Log.Error(err, "failed to update plan status")
-		return
-	}
-
 	//
 	// Execute.
 	// The plan is updated as needed to reflect status.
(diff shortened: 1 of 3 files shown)


mnecas:
related to https://github.com/kubev2v/forklift/pull/2096

PR #2962, opened 2025-09-28
NO-JIRA | Fix downstream dependency versions in go.mod

AvielSegev:
[MTV-3499](https://issues.redhat.com/browse/MTV-3499)


.konflux/validation/go.mod
```diff
@@ -1,98 +1,99 @@
 module github.com/kubev2v/forklift
 
-go 1.24.4
-
-require github.com/open-policy-agent/opa v0.70.0
```
yaacov, line 5: i am missing the new require ?

AvielSegev: I just combined it with the larger `require` block.
Is there any reason to keep it separate?

yaacov: yes, the larget block is for indirect require, e.g. packages we need because they are used by the backages we need
the smaller group, usually before is for the packages we explicitly want, for example opa

yaacov: https://www.pixelstech.net/article/1702120803-in-depth-exploration-of-direct-and-indirect-dependency-management-in-golang

solenoci:
@AvielSegev Hi, please also update `FROM go-toolset` in validation Containerfile-downstream to `1.24.6-1758501173`

.konflux/validation/go.mod
AvielSegev: oops

AvielSegev:
> @AvielSegev Hi, please also update FROM go-toolset in validation Containerfile-downstream to 1.24.6-1758501173

@solenoci 
Done

PR #2963, opened 2025-09-29
MTV-2666 | VM migration with single interface from OSP fails with "more than one interface is connected to a pod network in spec.template.spec.interfaces"

Hazanel:
Issue:
When migrating an OpenStack VM that has a single interface but multiple NICs attached to it, the migration fails with an error saying there are multiple interfaces connected to the same pod.

Fix:
During destination VM creation in the mapping network phase, use only the first NIC per interface to avoid the duplication.

Ref: https://issues.redhat.com/browse/MTV-2666

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevents duplicate network interface mappings during OpenStack VM planning by using only the primary NIC per interface.
  * Ensures safer handling when interfaces have no NICs, avoiding errors during plan generation.
  * Improves consistency of network assignments, resulting in more predictable provisioning outcomes and cleaner plans.


yaacov (approved):

mnecas:
This is in the release-2.10, but the issue is not targeted to 2.10.
We did not ask QE to verify it, nor have the capacity.

PR #2972, opened 2025-09-29
offload: implement proper cleanup for tmp resources

amitosw15:
### feat(vsphere): implement comprehensive cleanup for VSphere xcopy volume populator resources

Implement proper resource lifecycle management for VSphereXcopyVolumePopulator CRs
to prevent resource leakage during migration cleanup operations.

Key changes:
- Add owner references from VSphereXcopyVolumePopulator CRs to PVCs for automatic
  garbage collection when PVCs are deleted
- Implement DeletePopulatorDataSource() to properly clean up populator CRs during
  migration cancel/archive operations
- Add comprehensive logging and error handling for cleanup operations
- Fetch PVC UID after creation to ensure accurate owner reference setup

This addresses temporary resource leakage issues where populator pods, PVCs, and
related resources were left orphaned after migration completion, particularly
during migration cancellation or archival.

Related to: https://github.com/kubev2v/forklift/pull/223

### DIFF
--- a/pkg/controller/plan/adapter/vsphere/destinationclient.go
+++ b/pkg/controller/plan/adapter/vsphere/destinationclient.go
@@ -1,8 +1,18 @@
 package vsphere
 
 import (
+	"context"
+	"path"
+
+	"github.com/kubev2v/forklift/pkg/apis/forklift/v1beta1"
 	"github.com/kubev2v/forklift/pkg/apis/forklift/v1beta1/plan"
 	plancontext "github.com/kubev2v/forklift/pkg/controller/plan/context"
+	liberr "github.com/kubev2v/forklift/pkg/lib/error"
+	core "k8s.io/api/core/v1"
+	k8serr "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/labels"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 type DestinationClient struct {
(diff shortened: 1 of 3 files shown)


pkg/controller/plan/adapter/vsphere/destinationclient.go
```diff
@@ -1,20 +1,213 @@
...
+			continue
+		}
+
+		r.Log.Info("Found matching PVC",
+			"populatorName", populatorCr.Name,
+			"pvcName", pvc.Name,
+			"pvcNamespace", pvc.Namespace)
+
+		populatorCrCopy := populatorCr.DeepCopy()
+		err = k8sutil.SetOwnerReference(pvc, &populatorCr, r.Scheme())
```
rgolangh: consider maybe directly setting this when we create the pvc and CR on builder.go

amitosw15: done

amitosw15:
@rgolangh  can you please review?

pkg/controller/plan/adapter/vsphere/destinationclient.go
```diff
@@ -1,20 +1,163 @@
...
+
+	pvcList := core.PersistentVolumeClaimList{}
+	err = r.Destination.Client.List(
+		context.TODO(),
+		&pvcList,
+		&client.ListOptions{
+			Namespace: r.Plan.Spec.TargetNamespace,
+			LabelSelector: labels.SelectorFromSet(map[string]string{
+				"migration": string(r.Plan.Status.Migration.ActiveSnapshot().Migration.UID),
+				"vmdkKey":   cr.Labels["vmdkKey"],
```
rgolangh: just worth changing the log at line 117 because the search is by vmdkKey and not by vmdkPath
another point- worth checking why we don't set the label of vmdk path on the PVC and then later use it here

pkg/controller/plan/adapter/vsphere/destinationclient.go
```diff
@@ -1,20 +1,163 @@
...
+			}),
+		})
+
+	if err != nil {
+		r.Log.Error(err, "Failed to list PVCs for populator CR", "populatorName", cr.Name)
+		err = liberr.Wrap(err)
+		return
```
rgolangh: please no naked returns

pkg/controller/plan/adapter/vsphere/destinationclient.go
```diff
@@ -1,20 +1,163 @@
...
+
+	if len(pvcList.Items) == 0 {
+		err = liberr.New("PVC not found", "vmdkPath", cr.Spec.VmdkPath)
+		r.Log.Error(err, "No PVC found for populator CR", "populatorName", cr.Name)
+		return
+	}
+
+	if len(pvcList.Items) > 1 {
+		err = liberr.New("Multiple PVCs found", "vmdkPath", cr.Spec.VmdkPath)
+		r.Log.Error(err, "Multiple PVCs found for populator CR",
```
rgolangh, line 168: if that is the case then perhaps the usage of get with label selector (if exists) is simpler. please check

pkg/controller/plan/adapter/vsphere/destinationclient.go
```diff
@@ -1,20 +1,213 @@
...
+
+	populatorCrList = v1beta1.VSphereXcopyVolumePopulatorList{}
+	err = r.Destination.Client.List(
+		context.TODO(),
+		&populatorCrList,
+		&client.ListOptions{
+			Namespace:     r.Plan.Spec.TargetNamespace,
+			LabelSelector: labels.SelectorFromSet(map[string]string{"migration": string(r.Plan.Status.Migration.ActiveSnapshot().Migration.UID)}),
+		})
+
```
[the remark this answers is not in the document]
rgolangh: try to find out if this could happen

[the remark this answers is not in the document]
rgolangh: @amitosw15 another use case - say a plan never started, or started then failed early before there is history. then we archive and delete it. Could that code path still be reached, and just blow on that nil? if yes then I guess a protection is worth it

pkg/controller/plan/adapter/vsphere/destinationclient.go
```diff
@@ -1,20 +1,213 @@
...
+			r.Log.Error(err, "Failed to delete object",
+				"type", objType,
+				"name", object.GetName())
+			return liberr.Wrap(err)
+		}
+	} else {
+		r.Log.Info(message,
+			objType,
+			path.Join(object.GetNamespace(), object.GetName()),
+			"vm", vm.String())
```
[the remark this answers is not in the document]
rgolangh: @amitosw15  The Log.Info API is `Info(msg string, kv ...interface{})` so that comment makes sense. 
`Log("message...", "key1", var1, "key2", var2) `

pkg/controller/plan/adapter/vsphere/destinationclient.go
rgolangh: that was a stupid comment, as labels by nature designed to return multiple objects. ignore

pkg/controller/plan/adapter/vsphere/destinationclient.go
amitosw15: we cannot set a label to exact vmdk path because it contains illegal chars (e.g. '[' ) but we put it as an annotation under the key 'copy-offload', i added this filter for more accurate match.

rgolangh: ahh right...

pkg/controller/plan/adapter/vsphere/destinationclient.go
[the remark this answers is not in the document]
amitosw15: okay, added protection

rgolangh (commented):
@amitosw15 when this is verified I'll leave a +1 and let @mnecas review

rgolangh:
LGTM
@mnecas please review

pkg/controller/plan/adapter/vsphere/destinationclient.go
```diff
@@ -1,20 +1,180 @@
...
+	err = r.Destination.Client.List(
+		context.TODO(),
+		&pvcList,
+		&client.ListOptions{
+			Namespace: r.Plan.Spec.TargetNamespace,
+			LabelSelector: labels.SelectorFromSet(map[string]string{
+				"migration": migUID,
+			}),
+			FieldSelector: fields.SelectorFromSet(map[string]string{
+				"metadata.annotations.copy-offload": cr.Spec.VmdkPath,
```
mnecas, line 146: Why do we have this annotation we already have `forklift.konveyor.io/disk-source` @rgolangh ?

rgolangh: could be historic reasons, but we need to double check if we still use it. 
I wonder if we use that to distinguish between disk copies.

amitosw15: @rgolangh 
Right now we are using this annotation to determined `isCopyOffload`

rgolangh: ack. we can't get rid of that function - it set the usage of virtv2v-inplace which is mandatory for offloaded disks that already exists on the conversion pod:

```
	if !useV2vForTransfer || r.IsCopyOffload(pvcs) {
		environment = append(environment,
			core.EnvVar{
				Name:  "V2V_inPlace",
				Value: "1",
			})
	}

```

rgolangh: unless we refactor it to something like - check if copy offload feature flag is on, check the storage map and the VM disks and see they have offload matching. which is something we want to do for excluding the vddk validator in a plan if it is offload copy. 
For now let's keep it and a small refactor could help remove it.

mnecas (approved):

PR #3164, opened 2025-10-07
ECOPROJECT-3359 | fix: BTRFS filesystem concern ID shouldn't include btrfs_disks list index

nirarg:
The concern ID should be the same for all the concerns of this type
Align to the way its done in other concerns


nirarg:
> Please, leave out all of the `.tekton/` changes. It's okay if the validate-tekton check will fail at this time.

Thanks @solenoci , done

PR #3172, opened 2025-10-10
MTV-3597 | Feature/IBM flashsystem xcopy

borod108:
The main difference for this provider is that we do not use a host
cluster - instead we map directly to a host, this required a bit of
special handeling in the mapping and cleaning after, where the right
host to map to is given in the context, rather than being the
xcopyInitiator group.

Also note that duplication protection is more strict in this provider,
so if we have even one HBA WWPN in a host we will use that host and ignore
other WWPNs. But if there are no hosts with any of the WWPNs we will
create a host with all of them and use that.

Assisted-by: Cursor
Signed-off-by: borod108 <boris.od@gmail.com>

## Summary by CodeRabbit

* **New Features**
  - FlashSystem (and Infinibox) added as supported storage vendor options; FlashSystem end-to-end support and a new test target for FlashSystem scenarios.

* **Bug Fixes**
  - Improved cleanup: skips unmap when LUN resolution fails to avoid unnecessary operations.

* **Documentation**
  - Example test plan updated to show FlashSystem configuration and sample host settings.


cmd/vsphere-xcopy-volume-populator/vsphere-xcopy-volume-populator.go
```diff
@@ -87,6 +88,12 @@ func main() {
 			klog.Fatalf("failed to initialize Ontap storage mapper with %s", err)
 		}
 		storageApi = &sm
+	case "flashsystem":
```
mnecas: Please make it a const

cmd/vsphere-xcopy-volume-populator/internal/flashsystem/flashsystem.go
```diff
@@ -0,0 +1,738 @@
...
+
+// FlashSystemProviderIDPrefix is the standard NAA prefix for IBM LUNs.
+const FlashSystemProviderIDPrefix = "naa.6005076"
+
+// isHexWWPN checks if a string looks like a hexadecimal WWPN
+func isHexWWPN(s string) bool {
+	// Remove common separators and prefixes
+	cleaned := strings.ReplaceAll(strings.ReplaceAll(s, ":", ""), "-", "")
+	cleaned = strings.TrimPrefix(cleaned, "fc.")
+
```
mnecas: Please share the example of the string in comment

cmd/vsphere-xcopy-volume-populator/internal/flashsystem/flashsystem.go
```diff
@@ -0,0 +1,738 @@
...
+	// Check if it's 16 hex characters (typical WWPN length)
+	hexPattern := regexp.MustCompile(`^[0-9a-fA-F]{16}$`)
```
mnecas: Think we could have a regex to match all of it without the string manipulations. We could also reuse the regex in the `extractWWPNsFromFCFormat` to get necessary group from the string

cmd/vsphere-xcopy-volume-populator/internal/flashsystem/flashsystem.go
```diff
@@ -0,0 +1,738 @@
...
+	req, err := http.NewRequest(method, fullURL, reqBody)
+	if err != nil {
+		return nil, 0, fmt.Errorf("failed to create request for %s: %w", fullURL, err)
+	}
+
+	req.Header.Set("Content-Type", "application/json")
+	req.Header.Set("Accept", "application/json")
+
+	// Thread-safe token access
+	c.authMutex.RLock()
```
mnecas: Why RLock here? This is running in separate go routines?

cmd/vsphere-xcopy-volume-populator/internal/flashsystem/flashsystem.go
```diff
@@ -0,0 +1,738 @@
...
+		// Retry the request with new token
+		return c.doRequest(method, path, payload)
+	}
+
+	return respBodyBytes, statusCode, err
+}
+
+// doRequest performs the actual HTTP request
+func (c *FlashSystemAPIClient) doRequest(method, path string, payload interface{}) ([]byte, int, error) {
+	fullURL := fmt.Sprintf("https://%s:7443/rest/v1%s", c.ManagementIP, path)
```
mnecas, line 178: Can the API port be changed in the IBM flashsystem?

borod108: it is constant and not modifiable.

PR #3192, opened 2025-10-15
MTV-3259 | | OCP > OCP migrate VM failed on failed to verify certificate: x509 with prehook.

Hazanel:
Issue:    
Pre-hooks create ConfigMaps with hook data (plan.yml, playbook.yml, workload.yml).
Later when ensureConfigMap() is called to prepare for DataVolume creation,
it finds this existing hook ConfigMap and reuses it as the certConfigMap for
CDI importer (avoiding duplicate ConfigMaps for the same VM),
causing TLS verification failures: "x509: certificate signed by unknown authority"

Fix:
Adding a new "resource" label to the configMaps to distinguish between the hook and the vm configMaps.

Ref: https://issues.redhat.com/browse/MTV-3259

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resource filtering to only include VM-related items that are not already marked with a step, preventing duplicate or unintended processing.
  * Enhances reliability of plan execution and reduces noise in logs and UI listings by excluding already-handled resources.

### DIFF
--- a/pkg/controller/plan/kubevirt.go
+++ b/pkg/controller/plan/kubevirt.go
@@ -103,6 +103,14 @@ const (
 	kDV = "isDV"
 	// Populator secret
 	kPopulator = "isPopulator"
+	// Resource label
+	kResource = "resource"
+)
+
+// Resource labels
+const (
+	ResourceVMConfig   = "vm-config"
+	ResourceVDDKConfig = "vddk-config"
 )
 
 // User
@@ -2425,6 +2433,7 @@ func (r *KubeVirt) ensureConfigMap(vmRef ref.Ref) (configMap *core.ConfigMap, er
 	if err != nil {
 		return
 	}
+
 	list := &core.ConfigMapList{}
 	err = r.Destination.Client.List(
 		context.TODO(),
@@ -2624,6 +2633,7 @@ func (r *KubeVirt) inspectionLabels(vmRef ref.Ref) (labels map[string]string) {
 func (r *KubeVirt) vmLabels(vmRef ref.Ref) (labels map[string]string) {
 	labels = r.planLabels()
 	labels[kVM] = vmRef.ID
+	labels[kResource] = ResourceVMConfig
 	return
 }
 
@@ -2632,6 +2642,7 @@ func (r *KubeVirt) vmLabels(vmRef ref.Ref) (labels map[string]string) {
 func (r *KubeVirt) vddkLabels() (labels map[string]string) {
 	labels = r.planLabels()
 	labels[kUse] = VddkConf
+	labels[kResource] = ResourceVDDKConfig
 	return
 }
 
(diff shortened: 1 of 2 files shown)


pkg/controller/plan/kubevirt.go
```diff
@@ -2519,12 +2520,26 @@ func (r *KubeVirt) ensureConfigMap(vmRef ref.Ref) (configMap *core.ConfigMap, er
...
+
+	// Create "step doesn't exist" requirement
+	requiredLabels, err := k8slabels.NewRequirement(kStep, selection.DoesNotExist, nil)
+	if err != nil {
+		err = liberr.Wrap(err)
+		return
+	}
+
+	// Combine base selector with "step doesn't exist" requirement {migration, plan, vmID, !step}
+	labelsSelector := k8slabels.SelectorFromSet(selector).Add(*requiredLabels)
```
mnecas: So what will happen if the configmap with labels exist? You are just excluding the kStep but what if there are two with the same labels and no kstep?

mnecas:
I dont see how this could be solving the issue

pkg/controller/plan/kubevirt.go
```diff
@@ -2519,12 +2520,26 @@ func (r *KubeVirt) ensureConfigMap(vmRef ref.Ref) (configMap *core.ConfigMap, er
...
+	//Get base selector for the VM
+	selector := r.vmLabels(vmRef)
+
+	// Create "step doesn't exist" requirement
+	requiredLabels, err := k8slabels.NewRequirement(kStep, selection.DoesNotExist, nil)
+	if err != nil {
+		err = liberr.Wrap(err)
+		return
+	}
+
```
mnecas: Why excluding the kStep and not just adding new label for configmap or hooks or whatever we need? This goes againts the patterns that are inside the forklift

Hazanel: Kstep appears  only in the hook Config, so we can leverage that instead of adding a new label.

pkg/controller/plan/kubevirt.go
Hazanel: Exactly what happens before the change... it takes the first one, that works for every scenario without preehook

pkg/controller/plan/kubevirt.go
mnecas: What if there will be another config with kstep? It's easier to distinguesh them by adding new label

Hazanel: OK,
I thought about adding a kHookId label using : r.hook.UID
and then we can filter it knowing that other config map don't have it. 

What do you think?

mnecas: UID sounds good aswell also thinking to add there what is it used for so we could see what the configmap is used for

Hazanel: Adding what it used for for every config map or just for the hook?
UID is a quick change. also adding a "resource" label just for Hook, the 2nd option will take longer

mnecas: for every configmap, so we can directly the needed configmap and distinguish them

Hazanel:
image

image

PR #3261, opened 2025-10-21
MTV-3362 | Migration plan stuck in 'Not Ready' due to NAD conflict between default namespace and target namespace.

Hazanel:
Issue:
Migration plans  fail to reach the Ready state when NADs with identical names exist across different namespaces (restricted providers).

Fix:
Added a query for the requested namespace that fixes the issue where GET commands with a specific namespace were ignored .

Ref: https://issues.redhat.com/browse/MTV-3362

## Summary by CodeRabbit

* **New Features**
  * Requests can include a query parameter to target a specific namespace for NetworkAttachmentDefinitions. Specified namespaces are validated against provider restrictions. If omitted, restricted providers search the provider + default namespaces; unrestricted providers search across all namespaces. Per-namespace listing behavior and existing filtering/error handling remain unchanged.

### DIFF
--- a/pkg/controller/provider/web/ocp/base.go
+++ b/pkg/controller/provider/web/ocp/base.go
@@ -273,9 +273,20 @@ func (h Handler) NetworkAttachmentDefinitions(ctx *gin.Context, provider *api.Pr
 		return
 	}
 
+	requestedNS := ctx.Request.URL.Query().Get(NsParam)
 	// Determine which namespaces to query
 	var namespacesToQuery []string
-	if provider != nil && provider.IsRestrictedHost() {
+
+	if requestedNS != "" {
+		if provider != nil && provider.IsRestrictedHost() {
+			// Validate the requested namespace is in allowed list
+			if requestedNS != provider.GetNamespace() && requestedNS != core.NamespaceDefault {
+				return nil, liberr.New("Namespace '%s' not allowed for restricted provider. Allowed namespaces: ['%s', 'default']",
+					requestedNS, provider.GetNamespace())
+			}
+		}
+		namespacesToQuery = []string{requestedNS}
+	} else if provider != nil && provider.IsRestrictedHost() {
 		// For restricted host providers, query specific namespaces
 		namespacesToQuery = []string{provider.GetNamespace(), core.NamespaceDefault}
 	} else {


pkg/controller/provider/web/ocp/base.go
```diff
@@ -273,9 +273,20 @@ func (h Handler) NetworkAttachmentDefinitions(ctx *gin.Context, provider *api.Pr
...
+		if provider != nil && provider.IsRestrictedHost() {
+			// Validate the requested namespace is in allowed list
+			if requestedNS != provider.GetNamespace() && requestedNS != core.NamespaceDefault {
+				return nil, liberr.New("Namespace '%s' not allowed for restricted provider. Allowed namespaces: ['%s', 'default']",
+					requestedNS, provider.GetNamespace())
+			}
+		}
+		namespacesToQuery = []string{requestedNS}
+	} else if provider != nil && provider.IsRestrictedHost() {
 		// For restricted host providers, query specific namespaces
```
yaacov, line 290:
```suggestion
        providerNS = provider.GetNamespace()
        defaultNS = core.NamespaceDefault
        queryNS = ctx.Request.URL.Query().Get(NsParam)
        
        // check if query ns is allowed
        if (queryNS != "" && queryNS != providerNS && queryNS != defaultNS) { // error ... }
        
		// For restricted host providers, query allowed namespaces
		namespacesToQuery = []string{providerNS, defaultNS}
```

this logic is identical to the logic you added from lines 279 to 289, it makes the issue clear:
when using restricted provider we can only get items from the provider ns and default ns

so the logic say:
```
if restricted provider:
     if qury ns is defined and not allowed:
          error
     fetch from allowed ns
```

yaacov: @Hazanel hi, after our of line discussion I think I understand the issue better,
I think the issue is not in the list NADs method here, where we should get all available NADs from all available namespaces, IMHO the issue is in methods that use the list of all available NADs and search for specific NAD using it's name only, while they should look for NAD using it's name and namespace.

@coderabbitai, hi, can you verify this method is used in places that expect all the NADs from all available namespaces ?
can you look for places we search the list of all the NADs, but filter using only name, without taking into account the namespace, fetching NADs that may have the same name but wrong namespaces ?

yaacov: @coderabbitai thank you for your analisys, I am missing something,
the method NetworkAttachmentDefinitions is only used where users except namespaced search, are you sure, it is not used in places user expect to get all namespace the provider have access to ?

did you look at the code and did not find places we filter NADs only by name, without checking the namespace ?

yaacov: @coderabbitai hi, I am missing something, in pkg/controller/provider/web/ocp/client.go in the method Network, do we filter correctly ?

pkg/controller/provider/web/ocp/base.go
```diff
@@ -273,9 +273,20 @@ func (h Handler) NetworkAttachmentDefinitions(ctx *gin.Context, provider *api.Pr
 		return
 	}
 
+	requestedNS := ctx.Request.URL.Query().Get(NsParam)
 	// Determine which namespaces to query
 	var namespacesToQuery []string
-	if provider != nil && provider.IsRestrictedHost() {
+
+	if requestedNS != "" {
```
yaacov, line 280: OK, I agree, in the case a user specify a specific namespace in the HTTP query, we should only return the NADs from this namespace.

so the logic should be:
a. did the user specify a namespace, if they did, check if it's valid, then search only this namespace
b. if no namespace was provided and we use restricted provider, search provider namespace + default
c. if no namespace was provided and we use root provider, search all namespace ( e.g. namespace = "" )

we need to add a comment for each case, so it will be clear that we support query namespaced lists.

pkg/controller/provider/web/ocp/base.go
yaacov: Thank you for the explanation, I agree, the list NADs method is the place to restrict the NADs by namespace provided the query include a namespace, I do think we need to be explicit about how this work and add clear comments because the namespace is not one of the method explict parameters, and it's not clear from the method name that we limit the search for NADs by namespace in cases it is provided

PR #3317, opened 2025-10-23
MTV-3569 | Add error message when missing parent backing

solenoci:
Issue: Sometimes a situation can happen where parent file/disk/backing of a created snapshot is empty. This results in failed inspection step with a missleading error message in virt-v2v logs.

Fix: Add validation of parent file/disk/backing before creating inspection pod.

Resolves: MTV-3569

Before:
Screenshot From 2025-10-23 17-22-31

After (works with multiple disks):
Screenshot From 2025-10-29 12-12-06

## Summary by CodeRabbit

* **New Features**
  * Configurable retry limit (default 10) for handling missing parent disk/backing; exposed via controller configuration and CRD.
  * Pod provisioning now receives step context so migration steps can track and control retries per-step.

* **Bug Fixes**
  * Inspection halts when parent-disk info is missing, returns clear errors, and marks affected steps as not ready rather than proceeding blindly.
  * Per-step retry counters prevent premature failures and enforce configured retry limits.


solenoci:
The error message can be modified if desired.

mrnold:
Is the lack of a parent file a temporary condition, like is it possible to re-check after a few seconds?

I definitely prefer seeing this error message over the VixDiskLib error, but when I try to read the second sentence from a user point of view I don't really understand what action I can take to be helpful. If there's something to be retried, it seems like it's forklift that should be doing that retry, unless I am supposed to just retry the whole migration.

solenoci:
We did not have the ability to check in the provided environment but we saw that those were missing sporadically.

solenoci:
> Is the lack of a parent file a temporary condition, like is it possible to re-check after a few seconds?
> 
> I definitely prefer seeing this error message over the VixDiskLib error, but when I try to read the second sentence from a user point of view I don't really understand what action I can take to be helpful. If there's something to be retried, it seems like it's forklift that should be doing that retry, unless I am supposed to just retry the whole migration.

Recheck is a good idea. However it seems the issue happens at a random. So I will remove the "retry" part of the error message and a retry mechanism for the inspection pod.

pkg/controller/plan/kubevirt.go
```diff
@@ -2158,12 +2164,48 @@ func (r *KubeVirt) getVirtV2vPod(vm *plan.VMStatus, vmVolumes []cnv.Volume, vddk
...
+				retries += 1
+				step.Annotations[ParentBackingRetriesAnnotation] = strconv.Itoa(retries)
+				errMsg := fmt.Sprintf("Parent disk of %s was not found, will retry on next attempt", disk.File)
+				r.Log.Info(errMsg,
+					"vm", vm.Ref.String())
+				return
+			}
+		}
+		if limitExceeded {
+			return nil, err
```
mnecas: NP: Could we move this to separate function? Seeing that the whole function and the case VirtV2vInspectionPod is getting larger and larger

solenoci: Good point, I will extract some things out, thanks

solenoci:
@coderabbit review

PR #3357, opened 2025-10-28
chore(fix typo): Makefile task build image does not use the platform var

yaacov:
Issue:
When building the operator index image we don't use the platform variable, resulting in an image built for current machine architecture.

Fix:
Fix the type to use the PLATFORM_FLAG var, resulting in using the platform flag for the build

## Summary by CodeRabbit

* **Chores**
  * Standardized platform handling across container image builds to ensure consistent multi-architecture tagging.
  * Included bundle image reference (with platform-specific suffix) as an explicit build argument so bundle and index images align by platform.
  * Build configuration updates only; no runtime behavior changes.


solenoci (approved):
lgtm

PR #3358, opened 2025-10-28
MTV-3745 | Hit unexpected MacConflicts error.

Hazanel:
Issue:
Completed migration plans continue to be re-validated (before the refactoring the
 mac address conflicts were handled in the provider builder).
If the destination VM is deleted and the same source VM is migrated again, the old completed plan may incorrectly flag MAC address conflicts with the new destination VM, leading to false validation errors.

Fix:
Skip all validation upon succeeded plans.
This avoids unnecessary reconciliation work and prevents false positives, such as MAC address conflicts from completed plans interfering with new migrations.

Ref: https://issues.redhat.com/browse/MTV-3475

### DIFF
--- a/pkg/controller/plan/controller.go
+++ b/pkg/controller/plan/controller.go
@@ -198,6 +198,14 @@ func (r Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (r
 		return
 	}
 
+	// Don't reconcile succeeded plans unless they need archiving.
+	if plan.Status.HasCondition(Succeeded) && !plan.Spec.Archived {
+		r.Log.V(1).Info("Skipping reconcile of succeeded plan.", "plan", plan.Name)
+		result.RequeueAfter = 0
+		err = nil
+		return
+	}
+
 	// Postpone as needed.
 	postpone, err := r.postpone()
 	if err != nil {


pkg/controller/plan/controller.go
```diff
@@ -192,6 +192,12 @@ func (r Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (r
 		r.Log.V(2).Info("Conditions.", "all", plan.Status.Conditions)
 	}()
 
+	// Don't reconcile succeeded plans.
+	if plan.Status.HasCondition(Succeeded) {
```
mansam: I don't think you can do this so early in the reconcile loop. By doing it so early it becomes impossible to archive a successful plan, and I don't know if there are other side effects.

Hazanel: It has to be before the validation , so I can move that right after the archive

mansam: I see you moved it after the check to see if the plan has the Archived condition, but it's still before the method call that archives the plan which happens on lines 233/234.

Hazanel:
@mansam @mnecas 
I came up with a slightly different suggestion , describing the logic flow bellow:

Logic Flow:
**Scenario 1: Normal succeeded plan**
    Line 202: Succeeded=true && !Archived=true → Skip 
    No periodic reconciliation, only watches
**Scenario 2: User archives succeeded plan**
    First reconciliation (after user sets spec.archived=true):
    Line 196: Archived=true && HasCondition(Archived)=false → Continue
    Line 202: Succeeded=true && !Archived=false → Continue
    Line 221: Validation runs → populates plan.Referenced
    Line 241: r.archive(plan) → Sets Archived condition
    Next reconciliation (from watch):
    Line 196: Archived=true && HasCondition(Archived)=true → Abort 
**Scenario 3: Already archived plan**
    Line 196: Both true → Abort immediately

pkg/controller/plan/controller.go
```diff
@@ -198,6 +198,12 @@ func (r Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (r
 		return
 	}
 
+	// Don't reconcile succeeded plans unless they need archiving.
+	if plan.Status.HasCondition(Succeeded) && !plan.Spec.Archived {
+		r.Log.V(1).Info("Skipping reconcile of succeeded plan.", "plan", plan.Name)
+		return reconcile.Result{RequeueAfter: 0}, nil
```
mnecas: NP: please use the `result.RequeueAfter = 0 ` and `return`
Just keeping it consistent across the code

mnecas (commented):
I was originally thinking to move it right before the `result.RequeueAfter, err = r.execute(plan)`. But think this coudl work aswell

PR #3359, opened 2025-10-28
chore: Set memory limits for prefetch-input task

Kyl-Bempah:
Added compute resource requests and limits for prefetch-input task.


solenoci:
Missing config for push pipeline

PR #3369, opened 2025-10-28
MTV-3537 | CBT not detected when snapshot exist of VM in VMware

gwencasey96:
@https://issues.redhat.com/browse/MTV-3537

Resolves: MTV-3550

Fixed CBT detection for VMs with snapshots by preserving per-disk CBT status when the disk array is rebuilt during fDevices property updates.

## Problem

When snapshots are created/deleted on a VM, VMware's PropertyCollector sends `fSnapshot` and `fDevices` updates, but **NOT** `fExtraConfig` (because ExtraConfig values didn't actually change). When `updateDisks()` rebuilds the disk array from `fDevices`, it creates fresh disk objects with `ChangeTrackingEnabled=false` (default value), losing the previously-set CBT status from ExtraConfig.This caused MTV to show "Disk - scsi0:0 does not have CBT enabled" warnings even when CBT was actually enabled in VMware. Power cycling the VM forced a full property refresh which fixed the issue temporarily.

## Solution

Before rebuilding the disk array in `updateDisks()`, we now save the CBT status from existing disks into a map keyed by disk file path. After creating each new disk object, we restore its CBT status from the map if available. This preserves the CBT status across `fDevices` updates when `fExtraConfig` is not sent.

## Changes

Added logic to preserve `ChangeTrackingEnabled` status across disk array rebuilds in `model.go`- Added test coverage for CBT detection with snapshots in `changed_block_tracking_test.rego`## TestingValidated with customer scenario: VM with CBT enabled, create snapshot → CBT warning

### DIFF
--- a/pkg/controller/plan/adapter/vsphere/builder.go
+++ b/pkg/controller/plan/adapter/vsphere/builder.go
@@ -667,12 +667,19 @@ func (r *Builder) VirtualMachine(vmRef ref.Ref, object *cnv.VirtualMachineSpec, 
 				vmRef.String()))
 		return
 	}
-	if r.Plan.IsWarm() && !vm.ChangeTrackingEnabled {
-		err = liberr.New(
-			fmt.Sprintf(
-				"Changed Block Tracking (CBT) is disabled for VM %s",
-				vmRef.String()))
-		return
+	// MTV-3537: Check per-disk CBT status for warm migrations.
+	// Don't use vm.ChangeTrackingEnabled as it can be stale when snapshots exist.
+	if r.Plan.IsWarm() {
+		for _, disk := range vm.Disks {
+			if !disk.ChangeTrackingEnabled {
+				err = liberr.New(
+					fmt.Sprintf(
+						"Changed Block Tracking (CBT) is disabled for disk %s on VM %s",
+						disk.File,
+						vmRef.String()))
+				return
+			}
+		}
 	}
 	if !r.Context.Plan.Spec.MigrateSharedDisks {
 		sharedPVCs, missingDiskPVCs, err := findSharedPVCs(r.Destination.Client, vm, r.Plan.Spec.TargetNamespace)
--- a/pkg/controller/plan/adapter/vsphere/validator.go
+++ b/pkg/controller/plan/adapter/vsphere/validator.go
@@ -619,7 +619,25 @@ func (r *Validator) ChangeTrackingEnabled(vmRef ref.Ref) (bool, error) {
 	if err != nil {
 		return false, liberr.Wrap(err, "vm", vmRef)
 	}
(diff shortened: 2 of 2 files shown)


pkg/controller/provider/container/vsphere/model.go
```diff
@@ -1003,6 +1005,36 @@ func isCBTEnabledForDisks(ctkPerDisk map[string]bool, disks []model.Disk) {
 	}
 }
 
+// isCBTEnabledForDisksWithSnapshotCheck performs snapshot-aware CBT detection.
+// This function uses ChangeId from disk backing information as the primary method
+// for CBT detection, which works reliably even when snapshots exist.
+// ExtraConfig-based detection is used as a fallback when ChangeId is not available.
+func isCBTEnabledForDisksWithSnapshotCheck(ctkPerDisk map[string]bool, disks []model.Disk, hasSnapshots bool) {
```
mrnold: Why does this need to be a whole new function? The only new code could just as easily go in the existing isCBTEnabledForDisks. And hasSnapshots doesn't seem to do anything.

pkg/controller/provider/container/vsphere/model.go
```diff
@@ -1003,6 +1005,36 @@ func isCBTEnabledForDisks(ctkPerDisk map[string]bool, disks []model.Disk) {
...
+	for i := range disks {
+		disk := &disks[i]
+
+		// Primary method: Use ChangeId from disk backing information
+		// This is the most reliable method and works with snapshots
+		if disk.ChangeTrackingEnabled {
```
mrnold: Maybe I am not understanding exactly which disk object is getting updated, but it looks like this will make it so disk.ChangeTrackingEnabled can never get un-set? Since this is the function that is responsible for setting it in the first place? It needs to accurately reflect whatever the VMware API says.

pkg/controller/provider/container/vsphere/model.go
```diff
@@ -1152,6 +1184,7 @@ func (v *VmAdapter) updateDisks(devArray *types.ArrayOfVirtualDevice) {
 					Bus:            bus,
 					Serial:         backing.Uuid,
 					WinDriveLetter: winDriveLetter,
+					ChangeTrackingEnabled: backing.ChangeId != "",
```
mrnold: I don't like indirectly assuming the state of ctkEnabled from the existence of a change ID.

mrnold (changes requested):
Does this really work? I would really rather fix the inventory to reflect whatever the VMware API says instead of assuming a change ID implies ctkEnabled. It could be okay, maybe, but I think it will need way more proof than the comments in these changes.

mnecas:
closing as stale

PR #3370, opened 2025-10-28
MTV-1904 | Warm migration failed when mtv is installed on normal namespace

gwencasey96:
Resolves: MTV-1904

@https://issues.redhat.com/browse/MTV-1904

Summary
When MTV is installed in a namespace with restrictive Pod Security Standards (e.g. pod-security.kubernetes.io/enforce=restricted) or OpenShift restricted-v2 SCC, warm migration can fail because conversion pods need security context settings that violate those policies. This change does not modify conversion pod security context; it adds detection and a non-blocking warning so users are informed before running a migration.
Changes

Detection: During plan validation, check the controller namespace (from POD_NAMESPACE) for the pod-security.kubernetes.io/enforce=restricted label.

Warning: If restricted policy is detected, add a non-blocking warning condition to the plan status. The warning does not block migration but tells users they may need to disable SCC label synchronization and apply the privileged label.

Scope: Only pkg/controller/plan/validation.go is modified. No changes to conversion pod creation or security context.

User impact: Users see a warning on the migration plan when the MTV controller runs in a restricted namespace.

Warning message: "Namespace 'X' where MTV is installed may be restricted, causing migration to fail. Disable SCC label synchronization and apply the privileged label."
Migrations are not blocked; the warning is advisory so users can adjust the namespace policy before running a migration if needed.

### DIFF
--- a/pkg/controller/plan/validation.go
+++ b/pkg/controller/plan/validation.go
@@ -2003,27 +2003,6 @@ func (r *Reconciler) validatePodSecurity(plan *api.Plan) error {
 			"plan", plan.Name)
 	}
 
-	// For OpenShift, also check if restricted-v2 SCC is indicated
-	// The restricted-v2 SCC typically sets specific annotations on the namespace
-	if settings.Settings.OpenShift && !isRestricted {
-		// Check for restricted-v2 SCC indicator
-		// When restricted-v2 SCC is used, the namespace may have specific annotations
-		// or the enforce label may be set to restricted
-		if ns.Annotations["openshift.io/sa.scc.uid-range"] != "" {
-			// Check if the UID range indicates restricted (restricted-v2 uses specific ranges)
-			// restricted-v2 typically uses UID ranges like 1000790000-1000799999
-			uidRange := ns.Annotations["openshift.io/sa.scc.uid-range"]
-			if strings.Contains(uidRange, "100079") {
-				// This looks like restricted-v2 SCC range
-				isRestricted = true
-				r.Log.Info("Detected restricted-v2 SCC from namespace annotation",
-					"namespace", controllerNamespace,
-					"uidRange", uidRange,
-					"plan", plan.Name)
-			}
-		}
-	}
-
 	// If restricted policies detected, add a warning condition (non-blocking)
 	if isRestricted {
 		restrictedPodSecurity := libcnd.Condition{


pkg/controller/plan/adapter/converter.go
```diff
@@ -146,14 +147,20 @@ func createConvertJob(pvc *v1.PersistentVolumeClaim, dv *cdi.DataVolume, srcForm
...
+						psc := &v1.PodSecurityContext{
+							RunAsNonRoot: ptr.To(true),
+						}
+						// Only set seccomp profile on OpenShift
+						// On Kubernetes, seccomp may be forbidden by restricted Pod Security Standards
+						if settings.Settings.OpenShift {
+							psc.SeccompProfile = &v1.SeccompProfile{
+								Type: v1.SeccompProfileTypeRuntimeDefault,
+							}
+						}
```
mnecas: RuntimeDefault applies also to kubernetes, if we would have specific seccomp profile like unshare for virt-v2v that would be different story.
+ don't see how this fixes the problem, for our tests it still would be the same

pkg/controller/plan/kubevirt.go
```diff
@@ -2338,16 +2343,11 @@ func (r *KubeVirt) getVirtV2vPod(vm *plan.VMStatus, vmVolumes []cnv.Volume, vddk
 			GenerateName: podName,
 		},
 		Spec: core.PodSpec{
-			SecurityContext: &core.PodSecurityContext{
-				FSGroup:        &fsGroup,
-				RunAsUser:      &user,
-				RunAsNonRoot:   &nonRoot,
```
mnecas, line 2344: We do need these users othervise the v2v will fail.

pkg/controller/plan/kubevirt.go
```diff
@@ -2259,24 +2258,30 @@ func (r *KubeVirt) getVirtV2vPod(vm *plan.VMStatus, vmVolumes []cnv.Volume, vddk
...
 		unshare := "profiles/unshare.json"
-		seccompProfile = core.SeccompProfile{
+		psc.SeccompProfile = &core.SeccompProfile{
 			Type:             core.SeccompProfileTypeLocalhost,
 			LocalhostProfile: &unshare,
 		}
-	} else {
-		seccompProfile = core.SeccompProfile{
-			Type: core.SeccompProfileTypeRuntimeDefault,
-		}
```
mnecas, line 2272: Why removing the default runtime?

gwencasey96:
restarted this-the logs showed that the migration failure happened on a restricted namespace. In openshift-mtv > details, there are namespace labels and one of the labels will show if the namespace is "baseline", "privileged" or "restricted". restricted needs to run as root but we need to runAsUser as qemuUser 107 for virtv2v. We can only support migrations with "baseline" or "privileged". There are other factors that can make the cluster behave as restricted so we cannot detect perfectly all the time-but now we can check UID ranges, the label, the pod-security.kubernetes.io/enforce and create a non-blocking warning to redeploy the cluster with more permissive settings.

pkg/controller/plan/validation.go
```diff
@@ -1909,6 +1917,106 @@ func (r *Reconciler) validateConversionTempStorage(plan *api.Plan) error {
...
+	// The restricted-v2 SCC typically sets specific annotations on the namespace
+	if settings.Settings.OpenShift && !isRestricted {
+		// Check for restricted-v2 SCC indicator
+		// When restricted-v2 SCC is used, the namespace may have specific annotations
+		// or the enforce label may be set to restricted
+		if ns.Annotations["openshift.io/sa.scc.uid-range"] != "" {
+			// Check if the UID range indicates restricted (restricted-v2 uses specific ranges)
+			// restricted-v2 typically uses UID ranges like 1000790000-1000799999
+			uidRange := ns.Annotations["openshift.io/sa.scc.uid-range"]
+			if strings.Contains(uidRange, "100079") {
```
mrnold: I think the format is more important than the specific value, so I am not sure checking for "100079" will work all the time. My cluster shows: `openshift.io/sa.scc.uid-range: 1000740000/10000`

mrnold:
I see a release-2.11 branch, sounds like this should be safe to merge after a rebase?

PR #3394, opened 2025-10-30
Fix coderabbit config

borod108:
Fix indentation for the auto-reviews part


borod108:
@mnecas hi, I used the wrong indentation so it did not work, this time validated with this fun tool https://docs.coderabbit.ai/configuration/yaml-validator

PR #3405, opened 2025-10-30
MTV-3606 | Avoid duplicate checkpoints on DataVolumes.

mrnold:
Issue: in a warm migration test, a duplicate checkpoint was added to a DataVolume's list of checkpoints, causing the migration to stall. There should never be duplicates in this list.

Fix: I have not found the root cause for how this happened, but for an immediate speculative fix it is harmless enough to avoid adding duplicates.

Resolves: MTV-3606 | https://issues.redhat.com/browse/MTV-3606


yaacov (approved):

mnecas (approved):

mnecas (commented):
think this will fix the issue, however I don't think this is the root cause, something smels fishi around this one

PR #3457, opened 2025-11-05
MTV-3700 | Look for default reoute in the NAD config

yaacov:
Ref: 
https://issues.redhat.com/browse/MTV-3700

Issue:
Today MTV get the transfer network default route only using the NADs `forklift.konveyor.io/route` annotation, if users forget to set the default route explicitly, migration may fail silently.

Fix:
In case the explicit default route is not set via the annotation, MTV will try to fetch the default gateway via the NADs configuration.

Note:
A warning the route is not set explicitly will still be available to users even if we can get the route implicitly - https://github.com/kubev2v/forklift/pull/3453

### DIFF
--- a/pkg/controller/plan/kubevirt.go
+++ b/pkg/controller/plan/kubevirt.go
@@ -154,6 +154,23 @@ type KubeVirt struct {
 	Ensurer adapter.Ensurer
 }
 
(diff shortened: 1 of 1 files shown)


yaacov:
@mansam @mnecas please review

PR #2418, opened 2025-08-06
WIP | ECOPROJECT-3106| Add Infinibox storage support for vsphere-xcopy-volume-populator

amitosw15:

PR #3460, opened 2025-11-05
[release-2.10] MTV-3113 | Add Infinibox storage support for vsphere-xcopy-volume-populator 

amitosw15:
Backport: https://github.com/kubev2v/forklift/pull/2418

Add InfiniBox provider

The mapping behavior of this provider is the following: InfiniBox uses a direct host-to-volume mapping approach where each host can be mapped to multiple volumes through the storage array's native API. It also limits (not as hard limit but as best practices) one host to initiators, we handle this by implementing a three-tier host name abstraction system.

The provider will ensure that:

A logical ESXi host name ('xcopy-esxs') is maintained consistently across the migration process
Real ESXi host names are used when mapping from ESXi context (esxRealHostNameKey)
Real OpenShift host names are used when mapping from OpenShift context (ocpRealHostNameKey)
Host name resolution adapts automatically based on the initiator group context
To map a volume: call MapVolumeToHost with the appropriate host name based on platform context.
To unmap a volume: call UnMapVolumeFromHost with the corresponding host name.

New Features

Added support for InfiniBox storage systems in the vSphere Xcopy Volume Populator, enabling integration and migration workflows with InfiniBox as a storage backend.

Extended configuration and CRD schemas to recognize "infinibox" as a valid storage vendor option.


rgolangh:
@mnecas @yaacov PTAL

solenoci:
@amitosw15 Please remove changes to `.tekton` and then ignore the failing  validate-tekton gh action.

PR #3489, opened 2025-11-06
Resolves: MTV-2811 | Warm migration with RHV provider fails if Skipping certificate validation.

Hazanel:
Issue:
Warm migrations fail when "Skip certificate validation" is enabled. CDI's ImageIO importer doesn't support insecureSkipVerify and always validates TLS, causing "x509: certificate signed by unknown authority" errors when no CA cert is provided.

Fix:
Added insecureSkipVerify field to the DataVolume ImageIO source. When set to true, this flag is propagated via the
cdi.kubevirt.io/storage.import.insecureSkipVerify annotation to the importer pod, allowing the CDI importer to skip TLS certificate verification.

Ref: https://issues.redhat.com/browse/MTV-2811


Hazanel:
@coderabbitai review

pkg/controller/plan/adapter/ovirt/builder.go
```diff
@@ -162,14 +163,25 @@ func (r *Builder) DataVolumes(vmRef ref.Ref, secret *core.Secret, configMap *cor
...
+
+				imageioSource := &cdi.DataVolumeSourceImageIO{
+					URL:       url,
+					DiskID:    da.Disk.ID,
+					SecretRef: secret.Name,
+				}
+
+				if insecure {
+					// Use CDI's insecureSkipVerify flag
+					imageioSource.InsecureSkipVerify = &insecure
```
mnecas, line 227: How will this act on the cluster that does not have this paramter?
Bit worried that we could fail to create the DataVolume

Hazanel: Thanks @mnecas for the catch.
I added the version support and the fetch certificate (The complementary solution I had before ) as a fallback so we could still perform warm migration with skip tls

pkg/controller/plan/adapter/ovirt/builder.go
```diff
@@ -120,10 +123,58 @@ type Builder struct {
...
+		} else if len(cacert) > 0 {
+			object.BinaryData["ca.pem"] = cacert
+		}
+	}
 	return
 }
 
+// Fetches the CA certificate from the oVirt engine URL.
+// This is needed for CNV < 4.21 in insecure mode as a fallback when InsecureSkipVerify field is not supported.
+func (r *Builder) fetchOVirtCACert() (cert []byte, err error) {
```
mnecas, line 149: we shouldn't fetch it for them, the users should provide valid certificate for us

pkg/controller/plan/adapter/ovirt/builder.go
mnecas: what were the results of the testing? @Hazanel

pkg/controller/plan/kubevirt.go
```diff
@@ -1453,17 +1453,25 @@ func (r *KubeVirt) dataVolumes(vm *plan.VMStatus, secret *core.Secret, configMap
...
 
+	if r.Plan.IsWarm() {
+		if r.Builder.SupportsVolumePopulators() {
+			// For storage offload, tie DataVolume to pre-imported PVC
+			annotations[planbase.AnnAllowClaimAdoption] = "true"
+			annotations[planbase.AnnPrePopulated] = "true"
+		} else {
+			// For warm migrations that use traditional  (ImageIO, VDDK) import sources (not populators),
+			// explicitly disable CDI's populator auto-detection to avoid webhook validation errors
+			annotations[planbase.AnnUsePopulator] = "false"
```
mnecas, line 1464: Why do we need to set these? not sure if i follow why it's related to the bug

pkg/controller/plan/adapter/ovirt/builder.go
Hazanel: It hits the same  error message from webhook :
"admission webhook "datavolume-validate.cdi.kubevirt.io" denied the request: spec.source.Imageio source Imageio is not valid"

pkg/controller/plan/adapter/ovirt/builder.go
Hazanel: Even on skipTls = true?
I'm probably missing something.
That's a temporary fallback for skipTls=true  , so user thinks he doesn't need to supply a valid certificate

pkg/controller/plan/kubevirt.go
Hazanel: This issue is indeed independent of the insecureSkipVerify but I encountered it also as part of the webhook validation error.
Do you want that on a different PR?

yaacov:
@mnecas @Hazanel 

In this PR we patched the vendor dir directly without changing the go.mod !!!!

every call to `go mod vendor` will break the code, what do you suggest to do ?

yaacov:
https://github.com/kubev2v/forklift/pull/3678 - updating go.mod

PR #3537, opened 2025-11-09
MTV-3703 | fix: populator rescan fails in final rescan  

amitosw15:
rescan adapters command to all devices caused an error. After the unmap a specific
storage device enters apd mode, which cause the adapter to be busy, as
it tries to reconnect to the storage device.
We'll target in the deletion the specific adapter(s) which we used, and
combine it with a retry mechanism.


cmd/vsphere-xcopy-volume-populator/vsphere-xcopy-volume-populator_test.go
rgolangh: here

amitosw15: removed

PR #3539, opened 2025-11-10
MTV-2973 | Fix MAX_VM_IN_FLIGHT with storage offload

gwencasey96:
This commit fixes two critical issues that caused migration failures when using storage offload with MAX_VM_IN_FLIGHT to limit concurrent migrations:

1. Made createPodToBindPVCs() idempotent to prevent duplicate consumer pod creation during reconciliation loops. Previously, every reconciliation would create a new consumer pod, causing populator pods to restart mid-transfer and resulting in vmkfstools conflicts and device locks.

2. Fixed scheduler cost calculation for cold migrations using storage offload. VMs in PhaseCreateDataVolumes now count as 1 VM in flight (not per-disk cost), preventing the scheduler from incorrectly preventing new VMs from starting.

These changes ensure that MAX_VM_IN_FLIGHT correctly throttles concurrent storage offload migrations without causing populator pod restarts or vmkfstools device lock conflicts.

Jira: https://issues.redhat.com/browse/MTV-2973
Resolves: MTV-2973

### DIFF
--- a/pkg/controller/plan/scheduler/vsphere/scheduler.go
+++ b/pkg/controller/plan/scheduler/vsphere/scheduler.go
@@ -223,18 +223,13 @@ func (r *Scheduler) cost(vm *model.VM, vmStatus *plan.VMStatus) int {
 			// By setting the cost to 0 other VMs can start migrating
 			return 0
 		default:
-			// For cold migrations using storage offload (volume populators),
-			// PhaseCreateDataVolumes is where the actual disk copy happens via
-			// populator pods. These are NOT tracked as individual disk tasks,
-			// so we should not count them in the cost calculation to avoid
-			// preventing other VMs from starting.
-			// Once the VM moves to PhaseCopyDisks, the populator work is complete.
-			if vmStatus.Phase == api.PhaseCreateDataVolumes && !r.Plan.IsWarm() {
-				// Cold migration with storage offload - populator pods do the work
-				// Cost should be based on whether populators are actively running
-				return 1 // Count as 1 VM in flight, not per-disk
-			}
-			// CDI transfers the disks in parallel by different pods
+			// CDI transfers the disks in parallel by different pods.
+			// This calculation represents the number of disks being transferred,
+			// which is used for any disk parallel migration: warm, storage offload,
+			// raw copy, anything with CDI -> anything except v2v.
+			// For storage offload, PhaseCreateDataVolumes is where populator pods
+			// start copying disks, and the cost should reflect the number of disks
+			// being transferred, not just count as 1 VM.
 			return len(vm.Disks) - r.finishedDisks(vmStatus)
 		}
 	}


pkg/controller/plan/scheduler/vsphere/scheduler.go
```diff
@@ -223,6 +223,17 @@ func (r *Scheduler) cost(vm *model.VM, vmStatus *plan.VMStatus) int {
...
+			// For cold migrations using storage offload (volume populators),
+			// PhaseCreateDataVolumes is where the actual disk copy happens via
+			// populator pods. These are NOT tracked as individual disk tasks,
+			// so we should not count them in the cost calculation to avoid
+			// preventing other VMs from starting.
+			// Once the VM moves to PhaseCopyDisks, the populator work is complete.
+			if vmStatus.Phase == api.PhaseCreateDataVolumes && !r.Plan.IsWarm() {
+				// Cold migration with storage offload - populator pods do the work
+				// Cost should be based on whether populators are actively running
+				return 1 // Count as 1 VM in flight, not per-disk
```
rgolangh: That is problematic for offload.
PhaseCopyDisk is while the disk copy in progress and there we increment the progress counter till the PVC is bouns to the conversion pod. it still needs to be counted as 1 per disk

gwencasey96: As far as I understand, this should not affect PhaseCopyDisk. Behavior for Phase CopyDisk should stay the same.

pkg/controller/plan/scheduler/vsphere/scheduler.go
```diff
@@ -223,6 +223,17 @@ func (r *Scheduler) cost(vm *model.VM, vmStatus *plan.VMStatus) int {
 			// By setting the cost to 0 other VMs can start migrating
 			return 0
 		default:
+			// For cold migrations using storage offload (volume populators),
+			// PhaseCreateDataVolumes is where the actual disk copy happens via
+			// populator pods. These are NOT tracked as individual disk tasks,
+			// so we should not count them in the cost calculation to avoid
```
mnecas: This function is for cost valuation, so if we will calculate the transfer as 1 that means that other VMs could get started at that point. We should be still doing the `len(vm.Disks) - r.finishedDisks(vmStatus)` calculation as that represents the nubmer of disks. This is used for any disk parallel migration which is warm, storage offload, raw copy, anything with CDI -> anything except v2v.

gwencasey96: acked, now using (len(vm.Disks) - r.finishedDisks(vmStatus)) instead of returning 1 for PhaseCreateDataVolumes

pkg/controller/plan/kubevirt.go
```diff
@@ -914,6 +914,33 @@ func (r *KubeVirt) createPodToBindPVCs(vm *plan.VMStatus, pvcNames []string) (er
...
+	// If a consumer pod already exists and is not failed/succeeded, don't create another
+	for _, pod := range podList.Items {
+		if pod.Status.Phase != core.PodFailed && pod.Status.Phase != core.PodSucceeded {
+			r.Log.V(1).Info(
+				"Consumer pod already exists, skipping creation.",
+				"pod", pod.Name,
+				"phase", pod.Status.Phase)
+			return nil
+		}
+	}
```
mnecas, line 942: I don't see how this fixes the problem of scheduler, could you please expalin

gwencasey96: This PR fixes idempotent consumer pod creation as well as the cost calculation for the scheduler

PR #3541, opened 2025-11-10
MTV-3673 | fix: VDDK upload works only in openshift-mtv namespace

AvielSegev:
Signed-off-by: Aviel Segev <asegev@redhat.com>

### DIFF
--- a/operator/config/rbac/forklift-controller_role.yaml
+++ b/operator/config/rbac/forklift-controller_role.yaml
@@ -211,9 +211,11 @@ rules:
     - update
 - apiGroups:
     - image.openshift.io
+    - ""
   resources:
     - imagestreamtags
     - imagestreams
+    - imagestreams/layers
   verbs:
     - get
     - list
--- a/pkg/controller/plan/validation.go
+++ b/pkg/controller/plan/validation.go
@@ -11,6 +11,9 @@ import (
 	"strconv"
 	"strings"
 
+	rbacv1 "k8s.io/api/rbac/v1"
+	"k8s.io/apimachinery/pkg/types"
+
 	k8snet "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
 	api "github.com/kubev2v/forklift/pkg/apis/forklift/v1beta1"
 	refapi "github.com/kubev2v/forklift/pkg/apis/forklift/v1beta1/ref"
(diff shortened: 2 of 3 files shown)


solenoci:
@AvielSegev please, also run `make generate` and `make generate-manifests`

yaacov:
@AvielSegev @solenoci 

I have concerns about openning a cluster role without a good reason [1], can you consult security before applying this security relaxation ?

[1] you can fix the bug on the UI side without needing to relax the cluster security

yaacov (changes requested):
see comment above

solenoci:
Agreed @yaacov, I am also not sure about the relaxation. Just wanted to highlight that we are missing manifests.

mnecas:
@yaacov @solenoci could you please re-review?

operator/config/rbac/allow-all-image-pullers.yaml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  name: allow-all-image-pullers
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: system:image-puller
+subjects:
+  - kind: Group
+    apiGroup: rbac.authorization.k8s.io
+    name: system:authenticated
```
yaacov: I have a concern about letting everyone ( e.g. system:authenticated ) in all namespace ( e.g. ClusterRole ) to pull any image.
Another concern is that we create this cluster wide open for all ingress surface without notifying users or allowing them to opt out of it.

solenoci:
@AvielSegev Could you please take a look if it would be possible to limit the role only to the used namespace? For example: we could try to create the role programmatically (although, that may create problems on it's own because of RBAC validation...)

AvielSegev:
> @AvielSegev Could you please take a look if it would be possible to limit the role only to the used namespace? For example: we could try to create the role programmatically (although, that may create problems on it's own because of RBAC validation...)

IIRC, at least for the VDDK init pod that uses the VDDK image, we are using the default role from the namespace. This role has minimal permissions by default and cannot pull the image we build in openshift-mtv (or whichever namespace Forklift is installed in). 

We can consider two solutions here:

1. I can spend some time on Sunday to check how we can create a specific role that has only the permissions required to pull the image. As far as I remember, it’s not possible to extend the default role.

2. We can limit this feature to the Forklift installation namespace only.

AvielSegev:
I've changed the PR. 
PTAL @yaacov @solenoci

pkg/controller/plan/validation.go
```diff
@@ -1590,6 +1593,74 @@ func (r *Reconciler) cancelOtherActiveVddkCheckJobs(plan *api.Plan) (err error)
...
+	if k8serr.IsNotFound(err) {
+		// RoleBinding doesn't exist yet, create it
+		roleBinding := &rbacv1.RoleBinding{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      roleBindingName,
+				Namespace: sourceImageNamespace,
+			},
+			Subjects: []rbacv1.Subject{newSubject},
+			RoleRef: rbacv1.RoleRef{
+				Kind:     "ClusterRole",
```
yaacov, line 1631: we are oppeining `ClusterRole` for all` system:image-puller` this is better done using the operator yaml

My concern is that we open a cluster role for all the images without alerting the users.

AvielSegev: What’s different here is the subjects. We are not giving these permissions to `system:authenticated` users, only to the default users in the target namespace of a plan.

In addition, while reading the OpenShift docs right now, I see that this is actually the common procedure:
[5.4.1. Allowing pods to reference images across projects](https://docs.redhat.com/en/documentation/openshift_container_platform/4.10/html/images/managing-images#images-allow-pods-to-reference-images-across-projects_using-image-pull-secrets)

> When using the OpenShift image registry, to allow pods in project-a to reference images in project-b, a service account in project-a must be bound to the system:image-puller role in project-b.

yaacov (approved):
@AvielSegev thanks for the explaination

pkg/controller/plan/validation.go
yaacov: thank you for the clarification

operator/config/rbac/forklift-controller_role.yaml
```diff
@@ -211,9 +211,11 @@ rules:
     - update
 - apiGroups:
     - image.openshift.io
+    - ""
```
mnecas, line 214: @coderabitai why do we need "" api group?

mnecas:
closing as stale

PR #3551, opened 2025-11-11
MTV-3491 | Add IPV6 single/dual stack support

Hazanel:
Issue:
vSphere environments using IPv6-only or dual-stack configurations cannot migrate VMs because host address selection only supports IPv4. ESXi hosts with IPv6 addresses cause migration failures when constructing URLs for direct host connections during disk transfer.

Fix:
- Added IPv6 address field to VNIC model to capture IPv6 addresses from vSphere API
- Enhanced getHostAddress() to select IPv4 or IPv6 based on provider preference
- Added formatHostAddress() to wrap IPv6 addresses in brackets for URL compatibility
- Updated all host URL construction points to properly format IPv6 addresses
- Added comprehensive tests for IPv6 address selection and formatting

The implementation matches provider IP family preference when available, with  fallback to the other family if needed. Supports IPv4-only, IPv6-only, and dual-stack environments.

Ref: https://issues.redhat.com/browse/MTV-3491


Hazanel:
@coderabbitai review

yaacov:
@Hazanel hi, 

can you rename the PR from:
`Resolves: MTV-3491 | Add IPV6 single/dual stack support`
to 
`MTV-3491 | Add IPV6 single/dual stack support`
?

pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -344,17 +344,90 @@ func isWindows(vm *model.VM) bool {
 	return strings.Contains(vm.GuestID, WindowsPrefix) || strings.Contains(vm.GuestName, WindowsPrefix)
 }
 
-// Retrieve the IP address of an ESXI host from its Management Network VNIC or fall back to the hostname.
-func getHostAddress(host *model.Host) string {
+// Extract provider IP family detection
+func detectProviderIPFamily(provider *api.Provider) bool {
```
mnecas: NP: Please rename it to `providerUsesIPv6` or `providerHostnameIsIPv6` or `isProviderHostnameIPv6`.
The "detect" does not explicitly say anything a about IPv6 or IPv4 and which of them is true/false

pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -344,17 +344,90 @@ func isWindows(vm *model.VM) bool {
...
+			for _, ipv6 := range vnic.IpV6Address {
+				if ipv6 == "" {
+					continue
+				}
+				if ip := net.ParseIP(ipv6); ip != nil && !ip.IsLinkLocalUnicast() {
+					ipv6Address = ipv6
+					break
+				}
+			}
+		}
```
mnecas: This gets bit complicated as we are relaying on multiple iterations to set both variables, please make it into two functions.

mnecas (commented):
Don't we need to get the `HostIpConfigIpV6AddressConfiguration` from the mgmt nic?
The users can also specify the host network lot of it is implemented in the console plugin. How are we looking there with the integration?

pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -396,7 +473,7 @@ func (r *Builder) getSourceDetails(vm *model.VM, sourceSecret *core.Secret) (lib
 	} else if r.Source.Provider.Spec.Settings[api.SDK] == api.ESXI {
 		libvirtURL = liburl.URL{
 			Scheme:   "esx",
-			Host:     getHostAddress(host),
+			Host:     getHostAddress(host, r.Source.Provider),
```
mnecas: What if we change to `r.Source.Provider.Spec.url`?

mnecas (commented):
Recap from our internal conversation:
Issue: we override the provider spec url to managment IP
Fix: we should remove the override

pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -347,17 +347,15 @@ func isWindows(vm *model.VM) bool {
...
-func getHostAddress(host *model.Host) string {
-	for _, vnic := range host.Network.VNICs {
-		if vnic.PortGroup == ManagementNetwork {
-			if vnic.IpAddress != "" && net.ParseIP(vnic.IpAddress) != nil {
-				return vnic.IpAddress // Return the IP address if found
-			}
-		}
+// formatHostAddress wraps IPv6 addresses in brackets for URL compatibility.
+func formatHostAddress(address string) string {
+	ip := net.ParseIP(address)
```
mnecas, line 352: Just double checking what would happen if we provide `[IPv6]`
Just so we don't add the double brackes by accident

mnecas: To answer my question: https://pkg.go.dev/net/url#URL.Hostname
`If the result is enclosed in square brackets, as literal IPv6 addresses are, the square brackets are removed from the result.`

mnecas (approved):
lgtm

PR #3554, opened 2025-11-11
Don't compile regex for every dynamic customization script

jonner:
Currently we're looking up customization scripts and checking if they
match against a regex and returning the list of matching script names.
After we have the list of scripts, we then (for linux guests) recompile
the regex for each script and get the subgroup matches in order to
extract the action.

This patch provides a minor optimization by compiling the regex only
once and returning the match groups during the initial regex match.
We can then use these match groups later instead of recompiling and
matching in the script iteration loop.


jonner:
closing because I'm submitting it with a larger set of related patches

PR #3570, opened 2025-11-11
Fix issues related to dynamic customization scripts

jonner:
- **Don't compile regex for every dynamic customization script**
- **Fix virt-customize invocation for dynamic scripts**
- **Make virt-customize execution errors fatal**

### DIFF
--- a/pkg/virt-v2v/customize/customize.go
+++ b/pkg/virt-v2v/customize/customize.go
@@ -22,6 +22,8 @@ const (
 	LinuxDynamicRegex       = `^([0-9]+_linux_(run|firstboot)(([\w\-]*).sh))$`
 	ShellSuffix             = ".sh"
 	UploadCmd               = "--upload"
+	RunCmd                  = "--run"
+	FirstbootCmd            = "--firstboot"
 )
 
 //go:embed scripts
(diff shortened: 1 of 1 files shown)


cmd/virt-v2v/entrypoint.go
```diff
@@ -82,6 +82,7 @@ func main() {
 		err = convert.RunCustomize(inspection.OS)
 		if err != nil {
 			fmt.Println("Failed to customize the VM", err)
+			os.Exit(1)
```
mnecas: This was on purpose not failing with 1, the customize is running after the disk transfers so on large disks this could cause not user firendly errors. Could we propagate the errors as warnings?

mnecas:
rest of the PR lgmt

cmd/virt-v2v/entrypoint.go
jonner: I was wondering about that.  But I wonder how many customization failures are going undetected because of this? 

When you suggest to propagate the errors as warnings, what did you have in mind? I thought about adding a new `/warnings` endpoint to the http server within the conversion pod which can be queried by the controller and used to set warning conditions on the migration object. But the http server is only started and accessible for local migrations. So with that approach the warnings would still be silently ignored for remote migrations. Do you have other ideas about how to do this?

jonner:
I pushed the branch without the third patch. I'll look into propagating the customization warnings in a different PR.

cmd/virt-v2v/entrypoint.go
jonner: @mnecas  here is what I described above: https://github.com/kubev2v/forklift/pull/3595

mnecas (approved):

PR #3595, opened 2025-11-13
Propagate customization errors up to controller as warnings

jonner:
The virt-v2v container exposes a new '/warnings' endpoint and when
requested, it returns a json-encoded list of warnings that have occured:
`[ { reason: "MyWarningReason", message: "detailed description of the warning"}]`

After conversion and before issuing the `/shutdown` request, the
controller queries `/warnings` and uses the response to set conditions
on the Migration vm statuses.

### DIFF
--- a/cmd/virt-v2v/entrypoint.go
+++ b/cmd/virt-v2v/entrypoint.go
@@ -81,7 +81,12 @@ func main() {
 		// virt-customize
 		err = convert.RunCustomize(inspection.OS)
 		if err != nil {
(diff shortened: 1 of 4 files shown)


jonner:
I should note that although this propagates warnings up from the virt-v2v container to the controller for local migrations, it doesn't work for remote migrations. This is because the web server in the virt-v2v container is not started at all for remote migrations since the controller will not be able to connect to the remote conversion pod.

mnecas (approved):

PR #3606, opened 2025-11-16
Add support for "none" keyword in transfer network route annotation

grandeit:
Allows to explicitly specify no gateway for transfer networks by setting forklift.konveyor.io/route to "none". This enables use of OVN-Kubernetes localnet topologies where the transfer network has no gateway but direct layer 2 connectivity to ESXi hosts is available and vCenter is not reachable from within the transfer network.

Resolves: None

### DIFF
--- a/pkg/controller/plan/kubevirt.go
+++ b/pkg/controller/plan/kubevirt.go
@@ -64,6 +64,9 @@ const (
 	// Annotation to specify the default route for the transfer network.
 	// To be set on the transfer network NAD by the end user.
 	AnnForkliftNetworkRoute = "forklift.konveyor.io/route"
+	// Special value for AnnForkliftNetworkRoute to explicitly request no gateway.
+	// Use this to enable modern k8s.v1.cni.cncf.io/networks annotation without default-route.
+	AnnForkliftRouteValueNone = "none"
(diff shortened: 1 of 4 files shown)


yaacov:
@coderabbitai please review

yaacov:
@grandeit hi, can we make fallback to network annotation without gw, the default fallback ? then we can remove the fallback to the legacy annotation ? it will simplify the code, and make the new annotatio allowing network without gw unnecessary as it will be always true ?

grandeit:
> @grandeit hi, can we make fallback to network annotation without gw, the default fallback ? then we can remove the fallback to the legacy annotation ? it will simplify the code, and make the new annotatio allowing network without gw unnecessary as it will be always true ?

hey @yaacov
i decided to create the "none" sentinel value because of two reasons:
1. i'm not sure what implications it has if one would remove the code path for the legacy `v1.multus-cni.io/default-network` fallback.
2. there is code that creates a warning when the `forklift.konveyor.io/route` annotation is missing from the NAD that is used for the transfer network.

especially because of the second point i decided to give the user the option to explicitly opt-out from the default route, without impacting the current warning logic. afaik there is also a lot of talk in the docs saying you absolutely have to set the `forklift.konveyor.io/route` annotation in the NAD.

PR #3607, opened 2025-11-16
Resolves: MTV-3611 | ESXi provider connection fails when certificate verification is enabled.

Hazanel:
Issue:
When creating an ESXi provider with certificate verification enabled using a hostname URL (e.g., https://esxi-host.example.com/sdk), TLS validation fails because ESXi certificates only contain IP addresses in their Subject Alternative Name (SAN), not DNS hostnames. Go's TLS library validates the certificate hostname against the connection hostname, causing validation to fail even though the certificate is valid and matches the server's IP address.

Fix:
Enhanced VerifyTLSConnection() to handle hostname/IP mismatches: when the certificate doesn't match the URL hostname, resolve the hostname to an IP address and check if it matches the certificate's IP addresses. If a match is found, use the IP address for the TLS connection instead of the hostname.

Ref:https://issues.redhat.com/browse/MTV-3611

### DIFF
--- a/pkg/controller/base/controller.go
+++ b/pkg/controller/base/controller.go
@@ -133,7 +133,7 @@ func VerifyTLSConnection(rawURL string, secret *core.Secret) (*x509.Certificate,
 		return nil, fmt.Errorf("invalid URL: %w", err)
 	}
 
-	// Attempt to get certificate
+	// Get the TLS certificate from the server
 	cert, err := util.GetTlsCertificate(parsedURL, secret)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


Hazanel:
@coderabbitai review

Hazanel:
It's still arguable whether it's a BUG or a customer responsibility to supply a valid certificate (We tent to customer because the certificate basically lacks host information).
Moving to Draft

mnecas:
Yeah customer should be resposible for the certificates.
Closing as we don't need this change

PR #3609, opened 2025-11-16
MTV-3837 | add resources requests and limit to populator containers 

amitosw15:
Currently, the populator container created during the DiskAllocation phase is spun up without any Kubernetes resource requests or limits.

This poses two main problems:

Cluster Instability (Noisy Neighbor): An unbounded populator container can consume an excessive amount of CPU or memory, potentially starving other critical workloads on the same node and leading to instability.

No Quality of Service (QoS): Without requests, the populator pod gets a BestEffort QoS class, making it the first to be evicted under resource pressure. This can interrupt and delay disk operations (like VM migrations) unexpectedly.

Proposed Solution
Modify the populator container specification to include configurable resources definitions. This involves adding:

resources.requests (e.g., CPU, memory): To guarantee a minimum amount of resources for the populator to function reliably. This will likely move its QoS class to Burstable.

resources.limits (e.g., CPU, memory): To cap the maximum resources the populator can consume, preventing it from impacting other services.

These values should be made configurable to allow cluster administrators to tune them based on their specific node sizes and workloads.

Resolve: MTV-3837

### DIFF
--- a/cmd/populator-controller/populator-controller.go
+++ b/cmd/populator-controller/populator-controller.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/kubev2v/forklift/pkg/apis/forklift/v1beta1"
 	populator_machinery "github.com/kubev2v/forklift/pkg/lib-volume-populator/populator-machinery"
+	"github.com/kubev2v/forklift/pkg/settings"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -67,9 +68,13 @@ func main() {
 	flag.StringVar(&masterURL, "master", "", "The address of the Kubernetes API server. Overrides any value in kubeconfig. Only required if out-of-cluster.")
 	// Metrics args
 	flag.StringVar(&metricsPath, "metrics-path", "/metrics", "The HTTP path where prometheus metrics will be exposed. Default is `/metrics`.")
-
 	klog.InitFlags(nil)
 	flag.Parse()
+
+	resources, err := getResources()
+	if err != nil {
+		klog.Fatalf("Failed to parse resources: %v", err)
+	}
 
 	sigs := make(chan os.Signal, 1)
 	signal.Notify(sigs, syscall.SIGINT, syscall.SIGTERM)
@@ -91,7 +96,7 @@ func main() {
 		metricsEndpoint := populator.metricsEndpoint
 		go func() {
 			populator_machinery.RunController(masterURL, kubeconfig, imageName, metricsEndpoint, metricsPath,
-				prefix, gk, gvr, mountPath, devicePath, controllerFunc)
+				prefix, gk, gvr, mountPath, devicePath, controllerFunc, resources)
 			<-stop
 		}()
 	}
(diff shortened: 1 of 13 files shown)


scripts/validate-commits.sh
```diff
@@ -1,298 +0,0 @@
-#!/bin/bash
```
mnecas, line 1: Please don't remove this file

pkg/lib-volume-populator/populator-machinery/controller.go
```diff
@@ -143,11 +143,13 @@ type controller struct {
 	metrics           *metricsManager
 	recorder          record.EventRecorder
 	httpClient        *http.Client
+	resources         *corev1.ResourceRequirements
 }
 
 func RunController(masterURL, kubeconfig, imageName, httpEndpoint, metricsPath, prefix string,
 	gk schema.GroupKind, gvr schema.GroupVersionResource, mountPath, devicePath string,
 	populatorArgs func(bool, *unstructured.Unstructured, corev1.PersistentVolumeClaim) ([]string, error),
+	resources *corev1.ResourceRequirements,
```
mnecas, line 152: Don't see update of the RunController where you pass the updated resoruces

amitosw15: added https://github.com/kubev2v/forklift/blob/29e1cbbfaea13943c5452e2de9b099055fd01109/cmd/populator-controller/populator-controller.go#L109

solenoci:
You can ignore the konflux failure. We are aware of this and working on resolution.

rgolangh (commented):
- I saw the usage of command line flags. I wonder if they are really needed because
  you added ENV variables that will be added to the volume-popoulator-controller deployment - see /operator/roles/forkliftcontroller/templates/populator/deployment-populator-controller.yml.j2
- The requests you used are 512Mi mem and 0.1 core. What numbers did you see on those pods today?

amitosw15:
> * I saw the usage of command line flags. I wonder if they are really needed because
>   you added ENV variables that will be added to the volume-popoulator-controller deployment - see /operator/roles/forkliftcontroller/templates/populator/deployment-populator-controller.yml.j2
> * The requests you used are 512Mi mem and 0.1 core. What numbers did you see on those pods today?

Thanks, i changed it to use the values from the env var, and deleted the usage of flags.
On PowerStore with disk of 50Gi it uses  (at most) CPU(cores): 46m   MEMORY(bytes):  21Mi

cmd/populator-controller/populator-controller.go
```diff
@@ -159,3 +165,40 @@ func getVolumePath(rawBlock bool) string {
 		return mountPath + "disk.img"
 	}
 }
+
+func getResources() (*corev1.ResourceRequirements, error) {
+	cpuLimit := settings.Lookup(settings.PopulatorContainerLimitsCpu, settings.DefaultPopulatorContainerLimitsCpu)
+	memoryLimit := settings.Lookup(settings.PopulatorContainerLimitsMemory, settings.DefaultPopulatorContainerLimitsMemory)
```
mnecas: This default handeling is done inside the setitngs so we dont' need it here

amitosw15: fixed

pkg/settings/migration.go
```diff
@@ -262,6 +278,26 @@ func (r *Migration) Load() (err error) {
 	} else {
 		r.OvaContainerRequestsMemory = "512Mi"
 	}
+	if val, found := os.LookupEnv(PopulatorContainerLimitsCpu); found {
```
mnecas, line 281: Please take a look at example setting https://github.com/kubev2v/forklift/blob/9e98d71a145fd183865f01a7330b614b0eb085ec/pkg/settings/providers.go#L42

cmd/populator-controller/populator-controller.go
mnecas: you still do lookup, the settings are used to do all the configurations, if there is some problem we fail during the setting incialization which is right at the start of the controller.

mnecas (changes requested):

cmd/populator-controller/populator-controller.go
```diff
@@ -159,3 +165,24 @@ func getVolumePath(rawBlock bool) string {
 		return mountPath + "disk.img"
 	}
 }
+
+func getResources() (*corev1.ResourceRequirements, error) {
+	cpuLimit := settings.Settings.Migration.PopulatorContainerLimitsCpu
+	memoryLimit := settings.Settings.Migration.PopulatorContainerLimitsMemory
+	cpuRequest := settings.Settings.Migration.PopulatorContainerRequestsCpu
+	memoryRequest := settings.Settings.Migration.PopulatorContainerRequestsMemory
+	parsedCpuLimit, _ := resource.ParseQuantity(cpuLimit)
```
mnecas: You still don't need to do this `resource.ParseQuantity` as it's done inside the settigns

amitosw15: So where to parse the string into a quantity? In settings? because at some point i need to convert the type

amitosw15: okay i think i get you, hope it's okay now :)

PR #3612, opened 2025-11-17
MTV-3591 | SSH Conditions

borod108:
Add SSHReadiness condition on the provider and plan level.
In the condition add instructions on how to properly enable ssh on the ESXi.
A change to the restricted command code has been made to accommodate the fact we do not have a datastore when we do the validation of the esxi, and so can't run the python script (since it is uploaded to a specific datastore).

Resolves: https://issues.redhat.com/browse/MTV-3590
Resolves: https://issues.redhat.com/browse/MTV-3591


pkg/controller/plan/ssh.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+
+	// Check source provider for SSH readiness issues
+	sourceProvider := plan.Referenced.Provider.Source
+	if sourceProvider == nil {
+		return nil // This would be caught by other validation
+	}
+
+	// Only check vSphere providers with SSH method enabled
+	if sourceProvider.Type() != api.VSphere {
```
rgolangh, line 26: worth making this the first check

pkg/controller/provider/validation.go
```diff
@@ -398,3 +420,516 @@ func isValidNFSPath(nfsPath string) bool {
...
+	}
+
+	// Log detailed host information for debugging
+	r.Log.V(3).Info("SSH validation: analyzing host inventory", "provider", provider.Name)
+	var hostsToTest []hostInfo
+	hostsWithIP := 0
+	hostsWithoutIP := 0
+	for i := range hosts {
+		hostIP := getHostIP(&hosts[i])
+		r.Log.Info("SSH validation: host details",
```
rgolangh: other log level so the log doesn't get cluttered

pkg/controller/provider/validation.go
```diff
@@ -398,3 +420,516 @@ func isValidNFSPath(nfsPath string) bool {
...
+			"managementIPs", hosts[i].ManagementIPs,
+			"resolvedIP", hostIP,
+			"productVersion", hosts[i].ProductVersion,
+			"status", hosts[i].Status)
+		if hostIP != "" {
+			hostsWithIP++
+			hostsToTest = append(hostsToTest, hostInfo{name: hosts[i].Name, ip: hostIP})
+		} else {
+			hostsWithoutIP++
+			hostsToTest = append(hostsToTest, hostInfo{name: hosts[i].Name, ip: ""})
```
rgolangh: revise

pkg/controller/provider/validation.go
```diff
@@ -398,3 +420,516 @@ func isValidNFSPath(nfsPath string) bool {
...
+		return nil
+	}
+
+	r.Log.Info("SSH validation: starting validation for vSphere provider with SSH method enabled",
+		"provider", provider.Name,
+		"namespace", provider.Namespace,
+		"providerType", provider.Type())
+
+	// Check if SSH keys exist (they should be created by ensureSSHKeys)
+	privateSecretName := fmt.Sprintf("offload-ssh-keys-%s-private", provider.Name)
```
rgolangh: can you check if you can share this with the code that reconciles the secret

pkg/controller/provider/validation.go
```diff
@@ -398,3 +420,516 @@ func isValidNFSPath(nfsPath string) bool {
...
+	// Check for expected responses that indicate SSH is working
+	outputStr := strings.TrimSpace(string(output))
+
+	// Check for simple connectivity test response
+	if outputStr == "SSH_OK" {
+		r.Log.V(2).Info("Received connectivity test response - SSH key configured correctly", "hostIP", hostIP)
+		return true
+	}
+
+	// Check for XML response from Python script (in case someone tests with a datastore)
```
rgolangh: needed?

pkg/controller/provider/validation.go
```diff
@@ -398,3 +420,516 @@ func isValidNFSPath(nfsPath string) bool {
...
+	if strings.Contains(outputStr, "<?xml version=") {
+		r.Log.V(2).Info("Received XML response from script - SSH working correctly", "hostIP", hostIP)
+		return true
+	}
+
+	if strings.Contains(outputStr, "No such file or directory") && strings.Contains(outputStr, ".py") {
```
rgolangh: needed?:

rgolangh:
- LGTM only very small comments
- reminder to add the vendor
- reminder to rebase
@mnecas please review the controller parts.

rgolangh:
Code Duplication: RestrictedSSHCommandTemplate

I noticed that `RestrictedSSHCommandTemplate` is defined identically in two locations:

1. `pkg/controller/provider/validation.go:71`
2. `cmd/vsphere-xcopy-volume-populator/internal/vmware/ssh_client.go:32`

Both definitions are:
```go
const RestrictedSSHCommandTemplate = `sh -c 'DS=$(echo "$SSH_ORIGINAL_COMMAND" | sed -n "s/^DS=\\([^;]*\\);.*/\\1/p"); CMD=$(echo "$SSH_ORIGINAL_COMMAND" | sed -n "s/^DS=[^;]*;CMD=\\(.*\\)/\\1/p"); if [ -z "$DS" ]; then echo "SSH_OK"; else SSH_ORIGINAL_COMMAND="$CMD" exec python /vmfs/volumes/$DS/secure-vmkfstools-wrapper.py; fi'`
```

pkg/controller/plan/ssh.go
mnecas, line 1: This should go inside the validation.go

pkg/controller/plan/ssh.go
```diff
@@ -0,0 +1,79 @@
...
+	// Only check when ESXiCloneMethod is set to "ssh"
+	esxiCloneMethod, methodSet := sourceProvider.Spec.Settings[api.ESXiCloneMethod]
+	if !methodSet || esxiCloneMethod != "ssh" {
+		// Remove any existing SSH readiness conditions since SSH method is not enabled
+		plan.Status.DeleteCondition(SSHReadiness)
+		return nil
+	}
+
+	// Check if the source provider has SSH readiness conditions
+	sshReadinessCondition := sourceProvider.Status.FindCondition("SSHReadiness")
```
mnecas: NP: make it SSHReadiness const

pkg/controller/plan/ssh.go
```diff
@@ -0,0 +1,79 @@
...
+	// Only validate SSH readiness for plans using VSphere xcopy volume populators
+	if !r.planUsesVSphereXcopyPopulator(plan) {
+		// Remove any existing SSH readiness conditions since xcopy is not used
+		plan.Status.DeleteCondition(SSHReadiness)
+		return nil
+	}
+
+	// Only check when ESXiCloneMethod is set to "ssh"
+	esxiCloneMethod, methodSet := sourceProvider.Spec.Settings[api.ESXiCloneMethod]
+	if !methodSet || esxiCloneMethod != "ssh" {
```
mnecas: Do we have some costs type for the methods?

pkg/controller/provider/validation.go
```diff
@@ -400,3 +412,429 @@ func isValidNFSPath(nfsPath string) bool {
...
+			Status:   False,
+			Reason:   "SSHKeysNotFound",
+			Category: Warn,
+			Message: fmt.Sprintf(
+				"SSH method is enabled but SSH keys are being generated. "+
+					"After keys are created, you must manually install the public key on each ESXi host. "+
+					"Expected secrets: %s, %s",
+				privateSecretName, publicSecretName),
+		})
+		return nil
```
mnecas, line 697: please add 
```
			//nolint:nilerr
```

pkg/controller/provider/validation.go
```diff
@@ -400,3 +412,429 @@ func isValidNFSPath(nfsPath string) bool {
...
+		}
+		message.WriteString("\n")
+	}
+
+	message.WriteString("SETUP INSTRUCTIONS:\n\n")
+	message.WriteString("1. Enable SSH on each ESXi host:\n")
+	message.WriteString("   vim-cmd hostsvc/enable_ssh\n")
+	message.WriteString("   vim-cmd hostsvc/start_ssh\n\n")
+	message.WriteString("2. Add the following line to /etc/ssh/keys-root/authorized_keys on each ESXi host:\n\n")
+	message.WriteString(restrictedKey + "\n\n")
```
mnecas: This should be in the MTV docs not in status

pkg/controller/provider/validation.go
```diff
@@ -400,3 +412,429 @@ func isValidNFSPath(nfsPath string) bool {
...
+	var message strings.Builder
+	if len(successHosts) == 0 {
+		message.WriteString(fmt.Sprintf("SSH connectivity failed for all %d ESXi host(s). Manual SSH key installation required.\n\n", len(hostsToTest)))
+	} else {
+		message.WriteString(fmt.Sprintf("SSH connectivity: %d succeeded, %d failed (of %d total host(s)).\n\n",
+			len(successHosts), len(failedHosts), len(hostsToTest)))
+	}
+
+	if len(successHosts) > 0 {
+		message.WriteString(fmt.Sprintf("✓ SSH WORKING (%d hosts):\n", len(successHosts)))
```
mnecas: Please don't include any emojis just ASCII characters

pkg/controller/provider/validation.go
```diff
@@ -400,3 +412,429 @@ func isValidNFSPath(nfsPath string) bool {
...
+		message.WriteString("\n")
+	}
+
+	if len(failedHosts) > 0 {
+		message.WriteString(fmt.Sprintf("✗ SSH SETUP NEEDED (%d hosts):\n", len(failedHosts)))
+		for _, host := range failedHosts {
+			message.WriteString(fmt.Sprintf("  • %s\n", host))
+		}
+		message.WriteString("\n")
+	}
```
mnecas: You specify the hosts in the condition items do we need to have it in the message aswell? This applies to the passed hosts, do we need to mention the working hosts? It goes against how we build the condition for other validations.

pkg/controller/provider/validation.go
```diff
@@ -400,3 +412,429 @@ func isValidNFSPath(nfsPath string) bool {
...
+
+	// Build detailed message with both success and failure lists
+	var message strings.Builder
+	if len(successHosts) == 0 {
+		message.WriteString(fmt.Sprintf("SSH connectivity failed for all %d ESXi host(s). Manual SSH key installation required.\n\n", len(hostsToTest)))
+	} else {
+		message.WriteString(fmt.Sprintf("SSH connectivity: %d succeeded, %d failed (of %d total host(s)).\n\n",
+			len(successHosts), len(failedHosts), len(hostsToTest)))
+	}
```
mnecas: I'm okay with logging it but don't think condition is right place for this

pkg/controller/provider/validation.go
```diff
@@ -400,3 +412,429 @@ func isValidNFSPath(nfsPath string) bool {
...
+		r.Log.Error(nil, "SSH validation: collector does not have parity yet - SSH readiness check failed", "provider", provider.Name)
+		provider.Status.SetCondition(libcnd.Condition{
+			Type:     SSHReadiness,
+			Status:   False,
+			Reason:   "InventoryNotReady",
+			Category: Warn,
+			Message:  "Cannot validate SSH readiness: inventory collection has not completed (parity not reached). Wait for inventory to finish loading.",
+		})
+		return nil
+	}
```
mnecas: This is checked in previous validations, we should not start validating this if the DB is not ready

pkg/controller/provider/validation.go
```diff
@@ -400,3 +412,429 @@ func isValidNFSPath(nfsPath string) bool {
...
+
+	// Helper function to get host IP from ManagementIPs
+	getHostIP := func(host *vsphere.Host) string {
+		if len(host.ManagementIPs) > 0 {
+			r.Log.V(3).Info("SSH validation: using ManagementIP for host",
+				"provider", provider.Name,
+				"hostName", host.Name,
+				"managementIPs", host.ManagementIPs,
+				"selectedIP", host.ManagementIPs[0])
+			return host.ManagementIPs[0]
```
mnecas: **Important**: In MTV customers can also specify host migration network where we get the host IP address.
This is mainly if they create isolated network for connection with OCP. The maangment IP in those cases would not be accessible.
https://github.com/kubev2v/forklift/blob/29b60e21c388420e6e0e79a55802c2c0a201ab45/pkg/apis/forklift/v1beta1/host.go#L32

borod108: So just to be clear: the IpAddress is always populated and is the field I need? Or if it that if it is specified - use that one?

pkg/controller/provider/validation.go
borod108: I have hit this several times, weird.

pkg/controller/provider/validation.go
borod108: when all hosts pass the condition disappears, I found it very useful, because we may not want to have all hosts with the ssh, we may move the vm to one of the hosts that passed

pkg/controller/provider/validation.go
borod108: ok

pkg/controller/provider/validation.go
borod108: why? this is so so nice to use, and part of the fields are dynamic

pkg/controller/provider/validation.go
borod108: ok

pkg/controller/provider/validation.go
rgolangh: @mnecas against in what way?

pkg/controller/provider/validation.go
mnecas: you can use that ip, it should be populated by the UI

pkg/controller/provider/validation.go
mnecas: In that case we should add check if the provider does not have other Critical condition before we start doing these checks

pkg/controller/provider/validation.go
mnecas: the condition is made of 3 main things
1. type -> if it's error 
2. message -> what's the problems, should be short describtion.
3. items -> List of items to which this applies to, mainly used for VMs but can be anything. UI team would need to finish the visualization of that for any other item list.

So if we will want to fit everything inside the message that will go agaist the design.

PR #3622, opened 2025-11-18
MTV-3630 | storage-offload: use leases, serialize rescans

rgolangh:
Motivation
Storage offload and in particular adding copying on the ESXi in the
vsphere requires storage rescan, per a disk copy process. Also, VMWare
states there is a maximum cost of 2 clone operations on a single ESXi
they allow before the system will start to experience in stability.

Currently there is not mechanism to serialize or regulate the number of
concurrent requests, even with MAX_VM_INFLIGHT, because those operations
originates from the populator pod which is created by reacting to a
relevant PVC created per disk of a VM in the same stage.

Also, vib related operations like get and list, on the ESX are very
expensive and will fail quickly if called from different processes.

Modification
To orchestrate those operations and regulate them per-host we introduce
the usage of kubernetes leases form coordination.k8s.io package.
Per each host there are 2 leases named after the host and a slot. For
example for a host with managed object host-100 in VSphere, the
populator there should be up to 2 leases in the name:
esxi-lock-hostsystem-host-100-slot-0
esxi-lock-hostsystem-host-100-slot-1

Leases are exclusively under openshift-mtv namespace, so all plans use
the same leases.

When a populator code section takes the lease it updates it every 15
seconds and the lock is acquired for 15. When the critical section is
done other contenders are allowed to take it after the 15 second period
expires. In other words the lease expires automatically. There is no
need to delete it, but other way round-the holder needs to keep updating
it.

To sync a critical section in the populator execution the remote esxcli
can now use a HostLocker inteface:

	err := hostLocker.WithLock(context.Background(), hostID,
            func() error {
                // ... critical/heavy section, e.g. rescan
            }
       )

Result
Multiple disks VMs can be served in a plan, currently tested 10 without
a problem. Looks like the heavy part of all was rescans.


cmd/vsphere-xcopy-volume-populator/internal/populator/host_lease.go
```diff
@@ -0,0 +1,282 @@
...
+
+	if leaseNs := os.Getenv("HOST_LEASE_NAMESPACE"); leaseNs != "" {
+		h.namespace = leaseNs
+	}
+
+	if durationStr := os.Getenv("HOST_LEASE_DURATION_SECONDS"); durationStr != "" {
+		if duration, err := time.ParseDuration(durationStr + "s"); err == nil {
+			h.leaseDuration = duration
+		}
+	}
```
mnecas, line 59: Please make any env variable configurable from operator and ForkliftController CR. 
So it will be correctyl configure on the u/s with konveyor-forklift

cmd/vsphere-xcopy-volume-populator/internal/populator/remote_esxcli.go
```diff
@@ -274,7 +274,12 @@ func (p *RemoteEsxcliPopulator) Populate(vmId string, sourceVMDKFile string, pv
 	targetLUN := fmt.Sprintf("/vmfs/devices/disks/%s", lun.NAA)
 	klog.Infof("resolved lun with IQN %s to lun %s", lun.IQN, targetLUN)
 
-	err = rescan(p.VSphereClient, host, lun.NAA)
+	err = hostLocker.WithLock(context.Background(), strings.ReplaceAll(strings.ToLower(host.String()), ":", "-"),
```
mnecas: Please add a validation if the hostname follows the RFC 1123

mnecas (commented):
please take a look at the CI

cmd/vsphere-xcopy-volume-populator/internal/populator/host_lease.go
rgolangh: done and pushed - added controller_host_lease_namespace and controller_host_lease_duration_seconds

rgolangh:
@mnecas this is green now.

mnecas (approved):

PR #3650, opened 2025-11-19
MTV-3412 | label pvc with ds volume details 

amitosw15:
This change adds a storage affinity hint to PVCs created during vSphere migrations using the xcopy populator. The hint provides the NAA identifier of the source disk to help CSI drivers make optimal placement decisions.

## What this does

When creating PVCs for xcopy migrations, we now:
1. Query the vSphere API to retrieve the NAA ID from the VMFS datastore
2. Add a  label to the PVC with the NAA value
3. Pass this hint to the storage provider via PVC labels

## Why this matters

The NAA identifier is a vendor-neutral way to reference physical storage devices. By providing this information, CSI drivers can:
- Create volumes on the same physical storage as the source disk
- Optimize data locality and reduce migration overhead
- Improve performance by minimizing data movement across storage arrays

This is purely advisory - CSI drivers may choose to use or ignore this hint based on their capabilities and configuration. It's the storage provider's responsibility to implement support for this label if beneficial.

## Implementation details

- Added  method in vsphere/client.go to query VMFS extent info
- Created helper method in builder.go to establish vSphere connection and retrieve NAA
- Added  constant with standardized label key
- NAA lookup integrated into  function during PVC creation
- Errors in NAA retrieval are logged but don't block PVC creation 
Resolve: MTV-3412

### DIFF
--- a/pkg/controller/plan/adapter/vsphere/builder.go
+++ b/pkg/controller/plan/adapter/vsphere/builder.go
@@ -81,6 +81,7 @@ const (
 	TemplateOSLabel       = "os.template.kubevirt.io/%s"
 	TemplateWorkloadLabel = "workload.template.kubevirt.io/server"
 	TemplateFlavorLabel   = "flavor.template.kubevirt.io/medium"
+	TemplateNAALabel      = "volume.csi.k8s.io/affinity-source-naa"
 )
 
 // Operating Systems
@@ -1335,11 +1336,12 @@ func (r *Builder) PopulatorVolumes(vmRef ref.Ref, annotations map[string]string,
 	sortedDisks := r.sortedDisksAsVmware(vm.Disks)
 
 	dsMapIn := r.Context.Map.Storage.Spec.Map
+	dsNaaMap := make(map[string]string)
 	for i := range dsMapIn {
 		mapped := &dsMapIn[i]
-		ref := mapped.Source
+		sourceRef := mapped.Source
 		ds := &model.Datastore{}
-		fErr := r.Source.Inventory.Find(ds, ref)
+		fErr := r.Source.Inventory.Find(ds, sourceRef)
 		if fErr != nil {
 			err = fErr
 			return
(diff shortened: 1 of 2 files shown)


pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -1349,7 +1350,10 @@ func (r *Builder) PopulatorVolumes(vmRef ref.Ref, annotations map[string]string,
 		for diskIndex, disk := range sortedDisks {
 			if disk.Datastore.ID == ds.ID {
 				storageClass := mapped.Destination.StorageClass
-
+				naa, err := r.getNAAFromDatastore(context.TODO(), ref.Ref{ID: ds.ID, Name: ds.Name})
+				if err != nil {
+					r.Log.Error(err, "failed to get NAA from datastore", "datastore", ds.Name)
```
rgolangh: I would add to the message that we ignore this failure and continue.

pkg/controller/plan/adapter/vsphere/client.go
```diff
@@ -577,3 +577,58 @@ func (r *Client) DetachDisks(vmRef ref.Ref) (err error) {
 	// no-op
 	return
 }
+
+func (r *Client) GetNAAFromDatastore(ctx context.Context, datastoreRef ref.Ref) (string, error) {
```
rgolangh: I don't see the need for an extra function, and an exported one. This could go into the getNAAFromDatastore as is.

rgolangh: now just make sure it is not public (change capital G to g)

pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -1349,7 +1350,16 @@ func (r *Builder) PopulatorVolumes(vmRef ref.Ref, annotations map[string]string,
 		for diskIndex, disk := range sortedDisks {
 			if disk.Datastore.ID == ds.ID {
 				storageClass := mapped.Destination.StorageClass
-
+				vsphereClient := &Client{Context: r.Context}
+				err = vsphereClient.connect()
+				if err != nil {
+					r.Log.Error(err, "failed to connect to vSphere client, continue without storage affinity label")
+				}
+				naa, err := vsphereClient.GetNAAFromDatastore(context.TODO(), ref.Ref{ID: ds.ID, Name: ds.Name})
```
rgolangh: btw I would like to see memoization used here, using a map. A VM with multiple disks on the same datastore is worth it. 
But that could go in a subsequent PR and is not necessary.

pkg/lib/condition/zz_generated.deepcopy.go
```diff
@@ -21,8 +21,6 @@ limitations under the License.
 
 package condition
 
-import ()
```
rgolangh, line 24: no clue why this popped up. Please revert it, and make sure your controller-gen library isn't different than wha't set in the Make file's `controller-gen` target  (around line 567)

rgolangh:
@mnecas that work is ready (expect small changes) is an important one for 2.10.1 (and relatively small and safe) - please review

pkg/lib/condition/zz_generated.deepcopy.go
amitosw15: fixed

rgolangh:
@mnecas ping

PR #3657, opened 2025-11-21
MTV-3709 | Provide an option to customize the ephemeral storage used …

gwencasey96:
https://issues.redhat.com/browse/MTV-3709

This commit adds support for configuring custom ephemeral storage for virt-v2v conversion pods, preventing pod eviction due to storage pressure on nodes with limited ephemeral storage.

Changes:
- Added ConvertorTempStorageClass field to PlanSpec for specifying the storage class to use for temporary conversion storage
- Added ConvertorTempStorageSize field to PlanSpec for specifying the size of the temporary storage PVC (e.g., "30Gi", "1Ti")
- Implemented Generic Ephemeral Volume support in virt-v2v pod creation that automatically provisions and cleans up temporary PVCs
- Added volume mount at /var/tmp/virt-v2v for the temporary storage
- Set TMPDIR environment variable to direct virt-v2v to use the custom scratch directory
- Added validation to ensure both ConvertorTempStorageClass and ConvertorTempStorageSize are specified together
- Added validation to ensure ConvertorTempStorageSize is a valid Kubernetes resource quantity
- Added proper error messages via Plan conditions when validation fails

Use Cases:
- Large VM migrations (10+ TB disks) where fstrim operations create large overlays
- OVA imports that require full uncompressed disk copies in temporary storage
- Nodes with limited ephemeral storage that may cause pod eviction

Implementation Details:
- Uses Kubernetes Generic Ephemeral Volumes for automatic PVC lifecycle management
- Temporary PVC is created when the conversion pod starts and deleted when it completes
- Backward compatible: existing plans without these fields continue to use node ephemeral storage
- Follows existing convertor configuration patterns (ConvertorLabels, ConvertorNodeSelector, ConvertorAffinity)

Resolves: MTV-3709

### DIFF
--- a/operator/config/crd/bases/forklift.konveyor.io_plans.yaml
+++ b/operator/config/crd/bases/forklift.konveyor.io_plans.yaml
@@ -57,6 +57,28 @@ spec:
               archived:
                 description: Whether this plan should be archived.
                 type: boolean
+              conversionTempStorageClass:
+                description: |-
+                  ConversionTempStorageClass specifies the storage class to use for temporary conversion storage.
+                  When specified, virt-v2v conversion pods will use a temporary PVC from this storage class
+                  instead of using the node's ephemeral storage for the conversion scratch space.
+                  This is useful for:
+                    - Large VM migrations (10+ TB disks) where fstrim operations create large overlays
+                    - OVA imports that require full uncompressed disk copies in temporary storage
+                    - Nodes with limited ephemeral storage that may cause pod eviction due to storage pressure
+                  The temporary PVC is automatically created and deleted with the conversion pod.
+                type: string
+              conversionTempStorageSize:
+                description: |-
+                  ConversionTempStorageSize specifies the size of the temporary conversion storage PVC.
+                  Only used when ConversionTempStorageClass is specified.
+                  User specification allows for buffer space beyond the largest disk size to accommodate:
+                    - Temporary files during conversion
+                    - Multiple concurrent conversions
+                    - OVA imports requiring full uncompressed copies
+                  Recommended minimum: size of the largest VM disk being migrated.
+                  Format: standard Kubernetes resource quantity (e.g., "30Gi", "1Ti")
+                type: string
               convertorAffinity:
                 description: |-
                   ConvertorAffinity allows specifying hard- and soft-affinity for virt-v2v convertor pods.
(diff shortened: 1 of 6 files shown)


mnecas (approved):
lgtm please rebase the PR

operator/config/crd/bases/forklift.konveyor.io_plans.yaml
```diff
@@ -1005,6 +1005,25 @@ spec:
                   See Pod NodeSelector documentation for more details,
                   https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#nodeselector
                 type: object
+              convertorTempStorageClass:
```
mrnold: Kind of nitpicky but I think "converter" is more common. Can you change the spelling, or maybe just make it "conversionTempStorageClass" instead?

operator/config/crd/bases/forklift.konveyor.io_plans.yaml
```diff
@@ -1005,6 +1005,25 @@ spec:
...
+                  ConvertorTempStorageClass specifies the storage class to use for temporary conversion storage.
+                  When specified, virt-v2v convertor pods will use a temporary PVC from this storage class
+                  instead of using the node's ephemeral storage for the conversion scratch space.
+                  This is useful for:
+                    - Large VM migrations (10+ TB disks) where fstrim operations create large overlays
+                    - OVA imports that require full uncompressed disk copies in temporary storage
+                    - Nodes with limited ephemeral storage that may cause pod eviction due to storage pressure
+                  The temporary PVC is automatically created and deleted with the convertor pod.
+                type: string
+              convertorTempStorageSize:
```
mrnold: I am curious if it is better to make users specify this size or to have forklift calculate it from the VM automatically? If all we really want to do is take the size of the largest disk from the VM, it would be simpler to have forklift figure it out.

Am I missing some other reason that someone would want to specify this size?

pkg/controller/plan/kubevirt.go
```diff
@@ -2062,6 +2062,45 @@ func (r *KubeVirt) getVirtV2vPod(vm *plan.VMStatus, vmVolumes []cnv.Volume, vddk
 		MountPath: "/etc/secret",
 	})
 
+	// Add temporary conversion storage if configured
+	if r.Plan.Spec.ConvertorTempStorageClass != "" && r.Plan.Spec.ConvertorTempStorageSize != "" {
+		// Use Generic Ephemeral Volume for temporary conversion storage
+		// This creates a temporary PVC that is automatically deleted with the pod
+		storageClass := r.Plan.Spec.ConvertorTempStorageClass
+		volumeMode := core.PersistentVolumeFilesystem
```
mrnold, line 2108: Is it necessary to specify the volumeMode? The documentation for ephemeral volumes doesn't show it, so I am wondering if including it will prevent using block volumes if the storage class supports them.

mrnold (changes requested):

operator/config/crd/bases/forklift.konveyor.io_plans.yaml
gwencasey96: acked

operator/config/crd/bases/forklift.konveyor.io_plans.yaml
gwencasey96: added comment, allowing for buffer space beyond largest disk size

pkg/controller/plan/kubevirt.go
gwencasey96: added comment, avoiding default block mode

PR #3664, opened 2025-11-23
MTV-3834 | ARM64: Support multi arch OPA downloads, upstream only

yaacov:
Ref:
https://issues.redhat.com/browse/MTV-3834

Issue:
Upstream build of validation container image download pre build binaries, hardcoded to amd64

Fix:
Use env var to control the OPA download image, default to amd64 in case no var is used


yaacov:
@coderabbitai please review, check that the default build will be amd64, check that in case of other arch, when we use the make build system we pass the correct env var to the build

PR #3672, opened 2025-11-24
MTV-3892 | Add Req for provider validation

mnecas:
Resolves: https://issues.redhat.com/browse/MTV-3892

Issue: When the validation fails we don't req. Additionally I noticed that we have update of the CR at two places so I moved it to one defer same as we have in plan controller.

### DIFF
--- a/pkg/controller/provider/controller.go
+++ b/pkg/controller/provider/controller.go
@@ -180,7 +180,6 @@ func (r Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (r
 		return
 	} else {
 		r.catalog.add(request, provider)
-
 	}
 
 	defer func() {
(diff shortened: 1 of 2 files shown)


pkg/controller/provider/controller.go
```diff
@@ -208,6 +207,14 @@ func (r Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (r
 	provider.Status.Phase = Staging
 	provider.Status.BeginStagingConditions()
 
+	defer func() {
+		if err == nil {
```
Hazanel, line 211: It shouldn't be  if err != nil ?

mnecas: nope, when we want to update the provider status we return err nil and set the status.
other errors should not update the resoruce expamle https://github.com/kubev2v/forklift/blob/a6a70979535115658000acc70923f564dafb8088/pkg/controller/provider/validation.go#L151-L160

pkg/controller/provider/validation.go
```diff
@@ -267,6 +267,7 @@ func (r *Reconciler) validateSecret(provider *api.Provider) (secret *core.Secret
 				Category: Critical,
 				Message:  err.Error(),
 			})
+			err = nil
```
mnecas, line 270: @mansam the req is in the end of the controller but it did not got to that phase as the validation was returning error and critical condition. https://github.com/kubev2v/forklift/blob/b170050734902a37b5e3daa56d433d7d23012c5f/pkg/controller/provider/controller.go#L297-L304

PR #3681, opened 2025-11-24
MTV-2970 | Fix secret override

amitosw15:
When a VM has multiple disks backed by different storage systems, the migration fails because a single shared secret is overwritten with conflicting credentials.

Root cause: One secret was created per plan. When iterating over the origin VM's disks to create matching PVCs, the code updated the same secret for each disk. If two disks used different storage backends, each iteration overwrote the secret with different credentials, causing the wrong details to be used.

Solution

Approach: Create one secret per disk during the volume populator stage, rather than trying to create them earlier in the plan stage.

 Why this approach:
  - Creating secrets per-disk in the plan stage would require access to VM disk details, adding complexity
  - The volume populator stage has convenient access to individual disk information
  - More practical and maintainable solution

 Implementation:
  1. Create additional secrets (one per disk) in the volume populator stage
  2. Each disk-specific secret is owned by its corresponding PVC (via ownerReference)
  3. The original plan-level secret remains owned by the first PVC created (using the Plan or Migration object as owner would be more semantically correct, but requires cross-namespace ownership which is more complex)

Changes:
  - Added logic to create disk-specific secrets during PVC creation
  - Set PVC as owner reference for disk secrets (ensures cleanup when PVC is deleted)

### DIFF
--- a/pkg/controller/plan/adapter/vsphere/builder.go
+++ b/pkg/controller/plan/adapter/vsphere/builder.go
@@ -1448,7 +1448,7 @@ func (r *Builder) PopulatorVolumes(vmRef ref.Ref, annotations map[string]string,
 
 				// Update DataSourceRef to point to the volume populator
 				pvc.Spec.DataSourceRef.Name = populatorName
-
+				diskSecretName := fmt.Sprintf("%s-%d", secretName, diskIndex)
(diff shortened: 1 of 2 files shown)


amitosw15:
@solenoci can i still ignore this failure?

PR #3683, opened 2025-11-25
MTV-3834 | ARM64: fix makefile target opa-bin for none x86 machines

yaacov:
Issue:
Running `make opa-bin` on none x86 machine will install the wrong binaries.

Fix:
  - Use go lang automatic platform detection to install the correct binaries on any platform.
  - Update OPA version from v0.65.0 to v1.8.0 to match build version


yaacov:
cc:// @gwencasey96

PR #3691, opened 2025-11-25
MTV-3645 | RFE: Allow setting transfer network to the forklift controller pod

Hazanel:
Issue:
Currently, MTV allows setting transfer network in the provider and plan, which adds secondary network to importer and populator pods using the k8s.v1.cni.cncf.io/networks annotation. However, when the OCP default node network lacks connectivity to VMware/RHV providers, customers cannot add providers directly since the forklift controller cannot contact them. The current workaround requires cluster-wide changes (IP on second interface + routingViaHost), which is not ideal.

Fix:
Add controller_transfer_network field to ForkliftController CRD to allow customers to assign a secondary network to the forklift controller pod, similar to how it's assigned to importer/populator pods. This enables connectivity to providers via transfer network without cluster-wide changes.

Ref: https://issues.redhat.com/browse/MTV-3645

### DIFF
--- a/hack/validate_forklift_controller_crd.py
+++ b/hack/validate_forklift_controller_crd.py
@@ -195,6 +195,7 @@ def validate_forklift_controller_crd(crd_file, tasks_file):
     # Properties that are allowed to be in CRD even if not directly used in tasks
     # These are valid CRD fields that may be used indirectly or are configuration values
     allowed_crd_only_properties = {
+        'controller_transfer_network',
         'inventory_route_timeout',
         'metric_interval',
         'ova_proxy_route_timeout',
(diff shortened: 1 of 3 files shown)


Hazanel:
@coderabbitai review

operator/roles/forkliftcontroller/tasks/main.yml
```diff
@@ -79,6 +79,23 @@
...
+          namespace: "{{ app_namespace }}"
+        spec:
+          template:
+            metadata:
+              annotations:
+                k8s.v1.cni.cncf.io/networks: null
+      state: present
+      merge_type: "strategic-merge"
+    when: controller_transfer_network | default('') == ''
+
```
mnecas: Please checkout other jinja templates how they deal with the empty variables

mnecas: https://github.com/kubev2v/forklift/blob/9e98d71a145fd183865f01a7330b614b0eb085ec/operator/roles/forkliftcontroller/templates/controller/deployment-controller.yml.j2#L50-L53

PR #3721, opened 2025-11-27
chore(docs): create vib installer helper

rgolangh:
- create a vib installer tool which is shipped with the container
  instead of the ansible playbook
- update the README with the install method
- restructure the README


rgolangh:
@solenoci I had to create a small change in the copy-offload container file and changed also the downstream one

PR #3733, opened 2025-11-30
MTV-3703 | extend storageApi interface with get adapter id

amitosw15:
Adding to the offload storage provider interface an option to get the adapter id.
A specific clonnerIgroup ( abstraction of a security group of hosts from the storage perspective)  is chosen during the `EnsureClonnerIgroup`, but outside of this function the populator pod has no idea what adapter it uses. 
This causes problems when attempting to remove the dead devices of the adapter used.
The solution includes adding two functions to the interface: `getAdapterId` and `setAdapterId`, and return slice of strings, as for some providers use more than one adapter.


amitosw15:
https://github.com/kubev2v/forklift/pull/4221

PR #3734, opened 2025-11-30
MTV-2930 | Prevent mixed copy methods in migration plans

amitosw15:
## Description
Add validation before plan execution to ensure all disks use a consistent copy method (either VDDK or XCOPY/Offload).

## Logic
- **Multi-VM Plans:** If one VM cannot use offload while others can, the plan must be split (all VMs in a plan must share the copy method).
- **Multi-Disk VMs:** If a VM has multiple disks and one does not support offload, the system requires VDDK for all disks attached to that VM.

Resolves: MTV-2930

### DIFF
--- a/pkg/controller/plan/validation.go
+++ b/pkg/controller/plan/validation.go
@@ -92,6 +92,7 @@ const (
 	VMPowerStateUnsupported         = "VMPowerStateUnsupported"
 	VMMigrationTypeUnsupported      = "VMMigrationTypeUnsupported"
 	GuestToolsIssue                 = "GuestToolsIssue"
+	VDDKAndOffloadMixedUsage        = "VDDKAndOffloadMixedUsage"
 )
 
 // Categories
@@ -762,10 +763,24 @@ func (r *Reconciler) validateVM(plan *api.Plan) error {
 		Message:  "LUKS keys and Clevis cannot be configured together; Clevis will be used.",
 		Items:    []string{},
 	}
+	vddkAndOffloadMixedUsage := libcnd.Condition{
+		Type:     VDDKAndOffloadMixedUsage,
+		Status:   True,
+		Reason:   NotSupported,
+		Category: api.CategoryCritical,
+		Message:  "Copy offload is enabled. MTV does not support mixed copy methods. Each migration plan can use one migration strategy, either VDDK or copy offload. Check your storage map and VMs to ensure they are using the same migration strategy.",
+		Items:    []string{},
+	}
 
 	var sharedDisksConditions []libcnd.Condition
 	setOf := map[string]bool{}
 	setOfTargetName := map[string]bool{}
+
+	// Check if plan uses storage offload (vSphere only)
+	source := plan.Referenced.Provider.Source
+	checkMixedUsage := source != nil && source.Type() == api.VSphere && settings.Settings.Features.CopyOffload
+	planUsesOffload := checkMixedUsage && plan.IsUsingOffloadPlugin()
+
 	//
 	// Referenced VMs.
 	for i := range plan.Spec.VMs {
@@ -852,6 +867,23 @@ func (r *Reconciler) validateVM(plan *api.Plan) error {
 				}
 				if len(pvcs) != len(vm.Disks) {
 					missingPvcForOnlyConversion.Items = append(missingPvcForOnlyConversion.Items, ref.String())
+				}
+			}
+		}
+
+		// Check for mixed VDDK/Offload usage (vSphere only)
+		// If plan uses offload, add VMs with VDDK disks to the condition
+		if planUsesOffload {
+			if vsphereVM, ok := v.(*vsphere.VM); ok {
+				storageMap := plan.Referenced.Map.Storage
+				if storageMap != nil {
+					curVMHasVddk, err := r.vmUsesVddk(storageMap, vsphereVM, vm.Name)
+					if err != nil {
+						return err
+					}
+					if curVMHasVddk {
+						vddkAndOffloadMixedUsage.Items = append(vddkAndOffloadMixedUsage.Items, ref.String())
+					}
 				}
 			}
 		}
@@ -1169,6 +1201,12 @@ func (r *Reconciler) validateVM(plan *api.Plan) error {
 	if len(luksAndClevisIncompatibility.Items) > 0 {
 		plan.Status.SetCondition(luksAndClevisIncompatibility)
 	}
+
+	// Set the condition if any VMs with VDDK disks were found when plan uses offload
+	if len(vddkAndOffloadMixedUsage.Items) > 0 {
+		plan.Status.SetCondition(vddkAndOffloadMixedUsage)
+	}
+
 	return nil
 }
 
@@ -1869,4 +1907,19 @@ func (r *Reconciler) validateConversionTempStorage(plan *api.Plan) error {
 	}
 
 	return nil
+}
+
+// vmUsesVddk checks if the VM requires VDDK for migration (i.e., if any disk doesn't use storage offload)
+func (r *Reconciler) vmUsesVddk(storageMap *api.StorageMap, vsphereVM *vsphere.VM, vmName string) (bool, error) {
+	for _, disk := range vsphereVM.Disks {
+		mapping, found := storageMap.FindStorage(disk.Datastore.ID)
+		if !found {
+			continue // Another validation will handle this
+		}
+		if mapping.OffloadPlugin == nil || mapping.OffloadPlugin.VSphereXcopyPluginConfig == nil {
+			return true, nil
+		}
+	}
+
+	return false, nil
 }
--- a/pkg/controller/plan/validation_test.go
+++ b/pkg/controller/plan/validation_test.go
@@ -8,6 +8,8 @@ import (
 	"github.com/kubev2v/forklift/pkg/apis/forklift/v1beta1/provider"
 	"github.com/kubev2v/forklift/pkg/apis/forklift/v1beta1/ref"
 	"github.com/kubev2v/forklift/pkg/controller/base"
+	vspheremodel "github.com/kubev2v/forklift/pkg/controller/provider/model/vsphere"
+	"github.com/kubev2v/forklift/pkg/controller/provider/web/vsphere"
 	libcnd "github.com/kubev2v/forklift/pkg/lib/condition"
 	"github.com/kubev2v/forklift/pkg/lib/logging"
 	ginkgo "github.com/onsi/ginkgo/v2"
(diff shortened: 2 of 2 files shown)


pkg/controller/plan/validation.go
```diff
@@ -1837,3 +1842,83 @@ func (r *Reconciler) IsValidTargetName(targetName string) error {
...
+
+		for _, disk := range vsphereVM.Disks {
+			var mapping *api.StoragePair
+			for i := range storageMap.Spec.Map {
+				m := &storageMap.Spec.Map[i]
+				if m.Source.ID == disk.Datastore.ID {
+					mapping = m
+					break
+				}
+			}
```
mnecas: NP: please move this into separate function to get the mapping

pkg/controller/plan/validation.go
```diff
@@ -1837,3 +1842,83 @@ func (r *Reconciler) IsValidTargetName(targetName string) error {
...
+	if !settings.Settings.Features.CopyOffload {
+		return nil
+	}
+
+	mixedUsage := libcnd.Condition{
+		Type:     VDDKAndOffloadMixedUsage,
+		Status:   True,
+		Reason:   NotSupported,
+		Category: api.CategoryCritical,
+		Message:  "VM has mixed disk migration methods (VDDK and copy offload). All disks in a VM must use the same migration method.",
```
rgolangh: Users likely not understand immediately that the method is driven by the storage map. Lets help them by referring them to that storage map

rgolangh: Also the VM name(s) would help

pkg/controller/plan/validation.go
```diff
@@ -1837,3 +1842,83 @@ func (r *Reconciler) IsValidTargetName(targetName string) error {
...
+
+			// Check if this disk uses copy offload (matches SupportsVolumePopulators logic)
+			if mapping != nil && mapping.OffloadPlugin != nil && mapping.OffloadPlugin.VSphereXcopyPluginConfig != nil {
+				hasOffloadDisks = true
+			} else {
+				hasVddkDisks = true
+			}
+		}
+
+		if hasVddkDisks && hasOffloadDisks {
```
rgolangh: This early return is a missed opportunity to return an aggregated message with list of affected VMs. Let the users know what they need to address

pkg/controller/plan/validation.go
```diff
@@ -1837,3 +1842,83 @@ func (r *Reconciler) IsValidTargetName(targetName string) error {
...
+		return nil
+	}
+
+	mixedUsage := libcnd.Condition{
+		Type:     VDDKAndOffloadMixedUsage,
+		Status:   True,
+		Reason:   NotSupported,
+		Category: api.CategoryCritical,
+		Message:  "VM has mixed disk migration methods (VDDK and copy offload). All disks in a VM must use the same migration method.",
+		Items:    []string{},
```
rgolangh: Exactly this ^^ use it for the affected vm names affected

rgolangh:
Any update on this?

amitosw15:
> Any update on this?

Yep, sorry, I dealt with other issues first. I pushed a new version addressing your comments, which was tested with the following cases:


**should pass:**
1.  one pure vddk
2.  many pure vddk
3.  one pure offload
4.  many pure offload

**should fail:**
1.  one mixed vm
2.  mix of pure vddk and pure offload
3.  mix of mixed and pures

mixed == VM with both offload and VDDK disks

amitosw15:
image

Single mixed VM

image

Mix of pure VDDK and pure offload

image

Mix of mixed VM and pure offload

amitosw15:
@rgolangh @mnecas 
Can you please review?

pkg/controller/plan/validation.go
```diff
@@ -137,6 +138,22 @@ const (
 	Shareable = "shareable"
 )
 
+// +k8s:enum
```
mnecas: Do we need this comment?

pkg/controller/plan/validation.go
```diff
@@ -1843,3 +1864,205 @@ func (r *Reconciler) IsValidTargetName(targetName string) error {
...
+			"diskDetails", diskDetails,
+		)
+
+		mixedUsage := libcnd.Condition{
+			Type:     VDDKAndOffloadMixedUsage,
+			Status:   True,
+			Reason:   NotSupported,
+			Category: api.CategoryCritical,
+			Message:  message,
+			Items:    items,
```
mnecas: Items represents the list of VM references, UI takes this list and shows to which of the VMs the condition applies to. Pleasetake a look at other validations.

pkg/controller/plan/validation.go
```diff
@@ -1843,3 +1864,205 @@ func (r *Reconciler) IsValidTargetName(targetName string) error {
...
+			var pureVMs []string
+			if len(pureVddkVms) > 0 {
+				pureVMs = pureVddkVms
+				message = basicMessage + " The following VMs are planned to use VDDK: " + strings.Join(pureVddkVms, ", ") +
+					". The following VMs are planned to use mixed usage of VDDK and copy offload: " + strings.Join(mixedVms, ", ") + "."
+			} else if len(pureOffloadVms) > 0 {
+				pureVMs = pureOffloadVms
+				message = basicMessage + " The following VMs are planned to use copy offload: " + strings.Join(pureOffloadVms, ", ") +
+					". The following VMs are planned to use mixed usage of VDDK and copy offload: " + strings.Join(mixedVms, ", ") + "."
+			}
```
mnecas: Lets just keep the message constant, something that MTV does not supporte mixed copy methods. WE don't need to show the whole list in the message. Imagine that it will scale to 1000 VMs.

pkg/controller/plan/validation.go
```diff
@@ -1843,3 +1864,205 @@ func (r *Reconciler) IsValidTargetName(targetName string) error {
...
+
+	// validation of mixed usage of VDDK and copy offload on the plan level
+	var (
+		pureVddkVms    []string
+		pureOffloadVms []string
+		mixedVms       []string
+		diskDetails    []diskMigrationInfo // Track individual disk info
+	)
+
+	for _, vm := range plan.Spec.VMs {
```
mnecas: If we are going throught the VMs, this should be part of `validateVM`

pkg/controller/plan/validation.go
amitosw15: deleted

pkg/controller/plan/validation.go
amitosw15: I changed to static message

amitosw15:
@mnecas ping

rgolangh:
@mnecas Please re-review the messages. looks like the things you wanted to fix are handled.

pkg/controller/plan/validation.go
```diff
@@ -1169,6 +1220,21 @@ func (r *Reconciler) validateVM(plan *api.Plan) error {
...
 	}
+
+	// Check for mixed VDDK/Offload usage (vSphere only)
+	// Problem exists if: mixed VMs OR both pure VDDK and pure offload VMs in same plan
+	if checkMixedUsage {
+		mixedUsage := len(mixedVms) > 0 || (len(pureVddkVms) > 0 && len(pureOffloadVms) > 0)
+		if mixedUsage {
+			vddkAndOffloadMixedUsage.Items = append(vddkAndOffloadMixedUsage.Items, mixedVms...)
+			vddkAndOffloadMixedUsage.Items = append(vddkAndOffloadMixedUsage.Items, pureVddkVms...)
+			vddkAndOffloadMixedUsage.Items = append(vddkAndOffloadMixedUsage.Items, pureOffloadVms...)
```
mnecas: It's condition showing the mixed VMs we should not show the pure vddk or pure offload

pkg/controller/plan/validation.go
```diff
@@ -855,6 +889,23 @@ func (r *Reconciler) validateVM(plan *api.Plan) error {
...
+			if vsphereVM, ok := v.(*vsphere.VM); ok {
+				storageMap := plan.Referenced.Map.Storage
+				if storageMap != nil {
+					_, curVMHasOffload, curVMHasVddk, _ := r.checkDiskMigrationTypesDetailed(storageMap, vsphereVM, vm.Name)
+					if curVMHasOffload && curVMHasVddk {
+						mixedVms = append(mixedVms, ref.String())
+					} else if curVMHasOffload {
+						pureOffloadVms = append(pureOffloadVms, ref.String())
+					} else if curVMHasVddk {
+						pureVddkVms = append(pureVddkVms, ref.String())
```
mnecas: Please add the logic of adding the `vddkAndOffloadMixedUsage.Items` here, just trying to keep it as consistent as possible

pkg/controller/plan/validation.go
```diff
@@ -1870,3 +1936,39 @@ func (r *Reconciler) validateConversionTempStorage(plan *api.Plan) error {
...
+		diskDetails = append(diskDetails, diskMigrationInfo{
+			VMName:      vmName,
+			DiskKey:     disk.Key,
+			DatastoreID: disk.Datastore.ID,
+			Type:        diskType,
+		})
+	}
+
+	return diskDetails, hasOffloadDisks, hasVddkDisks, nil
+}
```
mnecas: We need to check if thje VM has mixed disks. Think this should return just on bool.

pkg/controller/plan/validation.go
amitosw15: But if we have a plan with one vm with offload and one vm with vddk we'' have the same error
Any combination of vddk & offload causes this error

pkg/controller/plan/validation.go
amitosw15: Consider the following:
VM A: two disks both are vddk
VM B: two disks both are offload

If we check on one vm:
VM A: valid
VM B: valid
Result: plan is valid
But the plan should be blocked

pkg/controller/plan/validation.go
```diff
@@ -1870,3 +1920,57 @@ func (r *Reconciler) validateConversionTempStorage(plan *api.Plan) error {
...
+		return false
+	}
+	for i := range storageMap.Spec.Map {
+		m := &storageMap.Spec.Map[i]
+		if m.OffloadPlugin != nil && m.OffloadPlugin.VSphereXcopyPluginConfig != nil {
+			return true
+		}
+	}
+	return false
+}
```
mnecas: No need for this, we already have
```
		plan.IsUsingOffloadPlugin()
```

pkg/controller/plan/validation.go
```diff
@@ -1870,3 +1920,57 @@ func (r *Reconciler) validateConversionTempStorage(plan *api.Plan) error {
...
+// getStorageMappingForDatastore finds the storage mapping for a given datastore ID
+func (r *Reconciler) getStorageMappingForDatastore(storageMap *api.StorageMap, datastoreID string) *api.StoragePair {
+	for i := range storageMap.Spec.Map {
+		m := &storageMap.Spec.Map[i]
+		if m.Source.ID == datastoreID {
+			return m
+		}
+	}
+	return nil
+}
```
mnecas: Think this can be replaced by 
```
storageMap.FindStorage(datastoreID)
```

pkg/controller/plan/validation.go
```diff
@@ -136,6 +137,21 @@ const (
...
+)
+
+// DiskMigrationInfo tracks migration type for individual disks
+type diskMigrationInfo struct {
+	VMName      string
+	DiskKey     int32
+	DatastoreID string
+	Type        diskType `json:"Type"`
+}
+
```
mnecas: We don't use the result so we don't need new struct with this informaiton nor new diskType.

pkg/controller/plan/validation.go
```diff
@@ -1870,3 +1920,57 @@ func (r *Reconciler) validateConversionTempStorage(plan *api.Plan) error {
...
+		diskDetails = append(diskDetails, diskMigrationInfo{
+			VMName:      vmName,
+			DiskKey:     disk.Key,
+			DatastoreID: disk.Datastore.ID,
+			Type:        diskType,
+		})
+	}
+
+	return diskDetails, hasOffloadDisks, hasVddkDisks, nil
+}
```
mnecas: We use just `curVMHasVddk` so the function shoudl be just that, we don't need any other things.

rgolangh:
any news?

PR #3773, opened 2025-12-03
MTV-3632 | Expand vsphere osmap

mnecas:
Issue: When migrating from VMware lot of the OS are not mapped to the preferences.

Fix: This commit expands the osmap with all available preferences in latest kubevirt. Additionally it adds a removal of the minor version inside the osmap, alternative to this we would need to upadte the osmap every time when there would be new minor release.

### DIFF
--- a/operator/roles/forkliftcontroller/templates/controller/configmap-osmap-vsphere.yml.j2
+++ b/operator/roles/forkliftcontroller/templates/controller/configmap-osmap-vsphere.yml.j2
@@ -8,22 +8,54 @@ metadata:
     app: {{ app_name }}
     service: {{ controller_service_name }}
 data:
-  centos7_64Guest: "centos.7"
-  centos7Guest: "centos.7"
-  centos8_64Guest: "centos.stream8"
-  centos9_64Guest: "centos.stream9"
+  genericLinuxGuest: "linux"
+
+  ubuntu64Guest: "ubuntu"
+  ubuntuGuest: "ubuntu"
+
   fedora64Guest: "fedora"
   fedoraGuest: "fedora"
+
+  centos7_64Guest: "centos.7"
+  centos8_64Guest: "centos.stream8"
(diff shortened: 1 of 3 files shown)


operator/roles/forkliftcontroller/templates/controller/configmap-osmap-vsphere.yml.j2
```diff
@@ -8,22 +8,52 @@ metadata:
...
+  windows12_64Guest: "windows.12"
+
+  windows7Server64Guest: "windows.2k8"
+  windows8Server64Guest: "windows.2k12"
+  windows9Server64Guest: "windows.2k16"
   windows2019srv_64Guest: "windows.2k19"
-  windows2019srvNext_64Guest: "windows.2k19"
-  windows2022srvNext_64Guest: "windows.2k22"
\ No newline at end of file
+  windows2019srvNext_64Guest: "windows.2k22"
```
solenoci, line 60: Let's put a comment here, that this is correct and according to the vmware naming for our future sanity.

pkg/controller/plan/adapter/vsphere/inspectionparser.go
```diff
@@ -40,15 +48,21 @@ var osV2VMap = map[string]string{
...
+	"win11": "windows11_64Guest",
+	"win12": "windows12_64Guest",
+
+	"win2k8r2":  "windows7Server64Guest",
+	"win2k12":   "windows8Server64Guest",
+	"win2k12r2": "windows8Server64Guest",
+	"win2k16":   "windows9Server64Guest",
+	"win2k19":   "windows2019srv_64Guest",
+	"win2k22":   "windows2019srvNext_64Guest",
+	"win2k25":   "windows2022srvNext_64Guest",
```
solenoci: Same as the other comment if this is what it really should be.

solenoci:
you can safely ignore the konflux failure

PR #3781, opened 2025-12-04
MTV-3049 | RFE: Interfaces' mac addresses are changed after migrating with preserve static ip - btrfs build.

Hazanel:
Issue:
SUSE VMs migrated from vSphere had swapped network interface MAC addresses because udev rules were not created. The script only checked RHEL's /etc/sysconfig/network-scripts/, missing SUSE's /etc/sysconfig/network/.

Fix:
Add SUSE network config path support to udev_from_ifcfg()
 Extract device name from filename for SUSE (ifcfg-eth0 -> eth0)
Only skip rule creation if existing udev file is non-empty.

Ref: https://issues.redhat.com/browse/MTV-3049


Hazanel:
False positive for Sonar comments: POSIX sh compatibility required; [ ] is correct here.

Hazanel:
Added **ifcfg-suse-test.d** to test SUSE's network configuration path (/etc/sysconfig/network/) which differs from RHEL/CentOS (/etc/sysconfig/network-scripts/).
This is SUSE's with Wicked default  path  so distros without Wicked are already covered in the NetworkManager-test

PR #3797, opened 2025-12-08
MTV-4061 | Allow storage offlaod with skip guest conversion

mnecas:
Resolves: MTV-4061

Issue: Right now we have RCM validation that blocks users from running the migraiton without VDDK. This is true for CDI migraiton which requires VDDK, however storage offload does not use the VDDK.


rgolangh:
lgtm

PR #3876, opened 2025-12-15
MTV-3635 | HyperV provider

Hazanel:
- Add HyperVProviderServer CR and controller for managing provider server pods
- Integrate SMB CSI driver for mounting Windows SMB shares (supports UNC paths)
- Refactor OVA/Hyper-V to use shared ovfbase adapter for OVF-based migrations
- Add Hyper-V inventory collection via OVF file scanning on SMB shares
- Add cold migration support for Hyper-V VMs (VHDX to KubeVirt conversion)
- Use existing virt-v2v conversion pipeline (no changes to conversion logic)

Ref:https://issues.redhat.com/browse/MTV-3635


Hazanel:
https://github.com/user-attachments/assets/ccb3606f-4ac0-4d5c-8b91-c123879a65a2

image

operator/roles/forkliftcontroller/defaults/main.yml
```diff
@@ -10,7 +10,7 @@ feature_copy_offload: false
 feature_ocp_live_migration: false
 feature_vmware_system_serial_number: true
 feature_cli_download: true
-feature_ova_appliance_management: false
```
mnecas, line 13: The feature gate should not change this could break backwards compatibilty, in cases when customers already enabled it in the spec it would stop working as they need to use new variable.

operator/config/rbac/forklift-controller_role.yaml
```diff
@@ -82,6 +82,10 @@ rules:
   - get
   - list
   - watch
+  - create
+  - update
+  - patch
+  - delete
```
mnecas: Why do we need to manipulat storage classes?

operator/roles/forkliftcontroller/tasks/main.yml
```diff
@@ -113,6 +113,77 @@
...
+    debug:
+      msg: >
+        {% if smb_operator_catalog.resources | length == 0 %}
+        SMB CSI Driver Operator not available in cluster catalog (OpenShift < 4.20).
+        HyperV provider will not be available.
+        {% elif smb_operator_installed.resources | default([]) | length > 0 %}
+        SMB CSI Driver Operator already installed.
+        {% else %}
+        SMB CSI Driver Operator installed successfully from cluster catalog.
+        {% endif %}
```
mnecas: No installation from our side, they should configure it.

mnecas: We can have some example for testing but we should not install other components

pkg/controller/plan/adapter/ovfbase/firmware_test.go
```diff
@@ -37,7 +37,7 @@ func TestMapFirmware(t *testing.T) {
 		},
 		{
 			name:       "UEFI with SecureBoot disabled",
-			firmware:   "efi",
+			firmware:   "uefi",
```
mnecas, line 40: did we change the OVF structure to have uefi instead of efi?

pkg/controller/plan/kubevirt.go
```diff
@@ -3214,6 +3301,129 @@ func getEntityPrefixName(resourceType, providerName, planName string) string {
...
+					},
+					NodeStageSecretRef: &core.SecretReference{
+						Name:      secretName,
+						Namespace: secretNamespace,
+					},
+				},
+			},
+			MountOptions: []string{
+				"dir_mode=0555",
+				"file_mode=0444",
```
mnecas: These look strange, is it some recommended default from SMB CSI?

operator/roles/forkliftcontroller/defaults/main.yml
Hazanel: I wonder if in the deployment we can leave
`{% if feature_ova_appliance_management|bool %}
        - name: FEATURE_OVF_APPLIANCE_MANAGEMENT
          value: "true"
{% endif %}`

So backward compatible for the user but in the code itself we align to the new ovf based

operator/roles/forkliftcontroller/tasks/main.yml
Hazanel: I'll omit the entire section

pkg/controller/plan/adapter/ovfbase/firmware_test.go
Hazanel: https://github.com/kubev2v/forklift/blob/44b2ef7770b9ce9ddeb93972a34d850ee36d2e57/pkg/controller/plan/adapter/ova/builder.go#L31
That's the original ova structure
 And vsphere API returns : https://github.com/kubev2v/forklift/blob/44b2ef7770b9ce9ddeb93972a34d850ee36d2e57/pkg/controller/plan/adapter/vsphere/builder.go#L57
 
 Both values map to the same  EFI configuration in the end - they're just different source formats using different terminology.

pkg/controller/plan/kubevirt.go
Hazanel: No, I wanted to prevent modification permissions, it can work regardless.
Removed - in case you want it I'll return it back

operator/config/rbac/forklift-controller_role.yaml
Hazanel: I need to double check again - probably was added temporarily while working with CRC

mnecas (approved):

Hazanel:
Error from kustomized-manifests isn't related to HyperV changes

PR #3892, opened 2025-12-16
MTV-3931 |  Move VIB validation from populator to provider

amitosw15:
Change VIB validation from per-populator check to provider-level validation
with a minimum 15-minute interval between checks to improve scalability.

The VIB check is an expensive operation and ESXi has limited capacity for concurrent VIB operations. Previously, each populator pod ran the VIB check independently, often executing many checks in parallel. This caused pod failures and prevented scaling beyond 5 concurrent disk migrations.
Move VIB validation to provider reconciliation with the following behavior:
  1. Check all ESXi hosts during provider reconcile (maximum once per 15 min)
  2. Query VIB version on each host when cache expires
  3. Attempt VIB installation if version does not match desired version
  4. Set warning condition on installation failure (non-blocking)
This eliminates parallel VIB check contention and enables migration of more than 5 disks concurrently.
Resolves: MTV-3931


amitosw15:
@rgolangh Can you please review?

cmd/vsphere-xcopy-volume-populator/cmd/vib-installer.go
```diff
@@ -206,7 +206,7 @@ func discoverESXiHostsFromVCenter(ctx context.Context, config *Config) ([]string
 
 	hostIPs := make([]string, 0, len(hosts))
 	for _, host := range hosts {
-		ip, err := vmware.GetHostIPAddress(ctx, host)
+		ip, err := vsphere_offload.GetHostIPAddress(ctx, host)
```
rgolangh: host is not relevant to vib queries whikch are all done through govmomi API. The ip address of a host is needed only for SSH kind of interactions.

build/forklift-controller/Containerfile
```diff
@@ -5,14 +5,16 @@ COPY --chown=1001:0 ./ ./
 ENV GOFLAGS="-mod=vendor -tags=strictfipsruntime"
 ENV GOEXPERIMENT=strictfipsruntime
 ENV GOCACHE=/go-build/cache
-RUN --mount=type=cache,target=${GOCACHE},uid=1001 go build -buildvcs=false -ldflags="-w -s" -o manager github.com/kubev2v/forklift/cmd/forklift-controller
+ARG VIB_VERSION
+RUN --mount=type=cache,target=${GOCACHE},uid=1001 go build -buildvcs=false -ldflags="-w -s -X github.com/kubev2v/forklift/pkg/lib/vsphere_offload.VibVersion=${VIB_VERSION}" -o manager github.com/kubev2v/forklift/cmd/forklift-controller
```
rgolangh: I want the VIB VERSION to be read from `vmkfstools-wrapper/version.mk` under this REPO. I don't want that to be passed by the CI tools. We set the version undre the version control. 

What I would do is follow the same pattern the `cmd/vsphere-xcopy-volume-populator/Makefile` does. Which is to include the version.mk and then use that for the build and in the Containerfile invoke the `make build`

pkg/lib/vsphere_offload/vib.go
```diff
@@ -0,0 +1,225 @@
+package vsphere_offload
```
rgolangh, line 1: I think offload in the name is unecessary and is a bit of a mix. It is true that this is used in the context of offload, however it can be in used in other places in forklift. 
In fact I think this kind of module does not belong under /pkg , which is something you want the module to expose to users of kubev2v/forklift. 
This needs to go under /internal/vmware , which doesn't exist at the moment, but should be created.
@mnecas @yaacov please comment here.

rgolangh: @mnecas @yaacov  for more context - we want to make the validation early, instead of in the population process, so we push our validation logic into the plan/provider label. Hence we need some library code to exist in an /internal module.
I want to believe that there should other vmware specific library code that can find its way there.

cmd/vsphere-xcopy-volume-populator/vmkfstools-wrapper/version.mk
```diff
@@ -0,0 +1 @@
+VIB_VERSION := 0.1.2
```
rgolangh: This is a mix of conerns. This doesn't belong here.

rgolangh: I would locate that under the Makefile, or leave it under cmd/vsphere-xcopy-volume-popoulator/vmkfstools-wrapper and just include thta from that directory.

pkg/controller/plan/vib.go
```diff
@@ -0,0 +1,94 @@
...
+	if !usesXcopy {
+		// Remove any existing VIB readiness conditions since xcopy is not used
+		plan.Status.DeleteCondition(VIBReady)
+		plan.Status.DeleteCondition(VIBNotReady)
+		return nil
+	}
+
+	// Check ESXiCloneMethod setting - VIB is the default when not set or explicitly set to "vib"
+	esxiCloneMethod, methodSet := sourceProvider.Spec.Settings[api.ESXiCloneMethod]
+	useVIBMethod := !methodSet || esxiCloneMethod != api.ESXiCloneMethodSSH
```
rgolangh: Since Settings is of type map[string]string, trying to get a key which doesn't exist would return "" as zero value. so you can narrow all of this to something like
      cloneMethod, _ := sourceProvider.Spec.Settings[api.ESXiCloneMethod]

next you just compare it to the VIB method, and not the SSH method const value
and remove the intermediate variable:

`if cloneMethod != api.EsxiCloneMethodVIB`

Since SSH is not mentioned here there's less confusing, no intermediate variable, no methodSet check

rgolangh: Now I realize that the clone method could be empty since there is no default value for VIB? 

If that is the case then I would set a helper function for that.

cmd/vsphere-xcopy-volume-populator/cmd/vib-installer.go
amitosw15: i missed this file, refactoring it....

amitosw15: change it to use `EnsureVIB` logic

cmd/vsphere-xcopy-volume-populator/vmkfstools-wrapper/version.mk
amitosw15: moved

pkg/controller/plan/vib.go
amitosw15: I dont sure i understand the concern here, assuming Ill change line 39 to:
`useVIBMethod := !methodSet || esxiCloneMethod == api.ESXiCloneMethodVIB`

So either methodSet is empty string, or it equals to `VIB`. 

Do we really need a helper function for that?

pkg/controller/provider/controller.go
```diff
@@ -215,6 +215,24 @@ func (r Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (r
 		}
 	}()
 
+	// Update provider metadata (annotations) before status
+	// Defers execute in LIFO order, so this runs before the status update
+	defer func() {
```
rgolangh: What specifically you wanted to update here? this looks like general purpose save to fix some issue?

rgolangh: also there is already an r.updateProviderStatus in line 212. 
Please explain all this. It looks like a work in progress of something

pkg/controller/plan/vib.go
```diff
@@ -0,0 +1,94 @@
...
+)
+
+// Types
+const (
+	VIBReady    = "VIBReady"
+	VIBNotReady = "VIBNotReady"
+)
+
+// validateVIBReadiness validates VIB readiness for migration plans using xcopy volume populators
+func (r *Reconciler) validateVIBReadiness(plan *api.Plan) error {
```
rgolangh, line 17: I don't see any `err` returned here, only nil.
 If there is no way this would fail then I would remove the error type from the return and even change the method name to something that would express "we are checking/setting/updating the vib readiness" or somethig similar

pkg/controller/plan/validation.go
```diff
@@ -224,6 +224,11 @@ func (r *Reconciler) validate(plan *api.Plan) error {
 		return err
 	}
 
+	// Validate VIB readiness for plans using xcopy with VIB-enabled providers
+	if err = r.validateVIBReadiness(plan); err != nil {
```
rgolangh, line 233: according to the implementation this can not fail. Maybe your intention was to fail it at some situation?

pkg/controller/provider/validation.go
```diff
@@ -444,7 +455,7 @@ func (r *Reconciler) getHostsForSSHValidation(provider *api.Provider) []hostInfo
 				Status:   True,
 				Reason:   "InvalidProviderURL",
 				Category: Warn,
-				Message:  fmt.Sprintf("Cannot validate SSH readiness (checked because 'esxiCloneMethod' setting is set to 'ssh'): failed to parse provider URL: %v", err),
```
rgolangh, line 492: please don't change this

pkg/controller/provider/validation.go
```diff
@@ -457,7 +468,7 @@ func (r *Reconciler) getHostsForSSHValidation(provider *api.Provider) []hostInfo
 				Status:   True,
 				Reason:   "NoHostIPInURL",
 				Category: Warn,
-				Message:  fmt.Sprintf("Cannot validate SSH readiness (checked because 'esxiCloneMethod' setting is set to 'ssh'): no host/IP found in provider URL: %s", provider.Spec.URL),
```
rgolangh, line 505: same here

pkg/controller/provider/validation.go
```diff
@@ -491,7 +502,7 @@ func (r *Reconciler) getHostsForSSHValidation(provider *api.Provider) []hostInfo
 			Status:   True,
 			Reason:   "HostListError",
 			Category: Warn,
-			Message:  fmt.Sprintf("Cannot validate SSH readiness (checked because 'esxiCloneMethod' setting is set to 'ssh'): failed to list hosts from inventory: %v", err),
```
rgolangh, line 539: same here

pkg/controller/provider/validation.go
```diff
@@ -507,7 +518,7 @@ func (r *Reconciler) getHostsForSSHValidation(provider *api.Provider) []hostInfo
 			Status:   True,
 			Reason:   "NoHostsFound",
 			Category: Warn,
-			Message:  "Cannot validate SSH readiness (checked because 'esxiCloneMethod' setting is set to 'ssh'): no ESXi hosts found in inventory.",
```
rgolangh, line 555: same here

pkg/controller/provider/validation.go
```diff
@@ -567,14 +578,27 @@ func (r *Reconciler) getHostsForSSHValidation(provider *api.Provider) []hostInfo
 			Status:   True,
 			Reason:   "NoHostIP",
 			Category: Warn,
-			Message:  fmt.Sprintf("Cannot validate SSH readiness (checked because 'esxiCloneMethod' setting is set to 'ssh'): no ESXi hosts with IP address found (found %d hosts total).", len(inventoryHosts)),
```
rgolangh, line 615: same here

pkg/controller/provider/validation.go
```diff
@@ -567,14 +578,27 @@ func (r *Reconciler) getHostsForSSHValidation(provider *api.Provider) []hostInfo
...
+			Message:  fmt.Sprintf("Cannot validate SSH readiness: no ESXi hosts with IP address found (found %d hosts total).", len(inventoryHosts)),
 		})
 		return nil
 	}
 
 	return hostsToTest
 }
 
+func (r *Reconciler) getHostsForVIBValidation(provider *api.Provider, client vmware.Client) ([]*object.HostSystem, error) {
```
rgolangh: doesn't look like you need a provider here.  not sure why you need this method at all if you can get the hsots directly from vmware.Client?

pkg/controller/provider/validation.go
```diff
@@ -833,7 +857,7 @@ func (r *Reconciler) validateSSHReadiness(provider *api.Provider, secret *core.S
 			if len(parts) == 3 {
 				successSuggestion.WriteString(fmt.Sprintf("  - %s (%s)\n", parts[1], parts[2]))
 			} else {
-				successSuggestion.WriteString(fmt.Sprintf("  - %s\n", item))
```
rgolangh, line 881: perhaps leave those changes to a different refactoring, if needed. same as in line 863

pkg/lib/vsphere_offload/vib.go
amitosw15: @mmartinv @yaacov 
ping

pkg/controller/plan/validation.go
amitosw15: I dont think i want to fail at any situation, but i can see many validate functions with this pattern. For example same behavior for `validateSSHReadiness` or `validateNetworkNameTemplate`

pkg/controller/provider/validation.go
amitosw15: you are right, deleted

pkg/controller/plan/vib.go
amitosw15: It seems like the common practice and name pattern in validate function, so i copied this behavior.  I agree with you that I can express the meaning better. What you prefer?

build/forklift-controller/Containerfile
amitosw15: done

amitosw15: When I changed the Containerfile to use RUN make forklift-controller (to follow the same pattern as vsphere-xcopy-volume-populator and read VIB_VERSION from version.mk), it introduced a dependency chain issue. The forklift-controller target depends on generate, which requires installing controller-gen via go install. However, this installation fails because the Containerfile sets GOFLAGS="-mod=vendor" earlier, preventing external package downloads.

  Solution options:
  1. Install tools before vendor mode (current approach): Install controller-gen without vendor restrictions, then set vendor mode for the actual build
  2. Remove generate dependency: Change the forklift-controller target to only depend on fmt and vet (like vsphere-xcopy-volume-populator does), since generated code is already committed to the repository
  3. Pre-compile outside container: Build the binary locally and copy it into the container (defeats the purpose of using make in the Containerfile)

pkg/controller/provider/controller.go
amitosw15: I initially misunderstood how Kubernetes optimistic concurrency control works with the reconcile update pattern.

**The problem**: VIB validation was updating annotations directly to Kubernetes, changing the `resourceVersion`. Then the defer tried to update status with the stale provider object → conflict error.

**The fix**: VIB validation now only sets in-memory state (like `SSH` validation does). The `defer` at line 212 handles all Kubernetes updates (annotations + status) with retry logic. I refactored `updateProvider` to handle both instead of just the in-memory collector.

amitosw15:
@mnecas @yaacov @mrnold  ping

build/forklift-controller/Containerfile
```diff
@@ -5,14 +5,15 @@ COPY --chown=1001:0 ./ ./
...
-RUN --mount=type=cache,target=${GOCACHE},uid=1001 go build -buildvcs=false -ldflags="-w -s" -o manager github.com/kubev2v/forklift/cmd/forklift-controller
+RUN --mount=type=cache,target=${GOCACHE},uid=1001 make forklift-controller
 
 FROM registry.access.redhat.com/ubi9-minimal:9.6-1752587672
 # RUN microdnf -y update && microdnf -y clean all
 # Required to be able to get files from within the pod
 RUN microdnf -y install tar && microdnf clean all
 
-COPY --from=builder /app/manager /usr/local/bin/forklift-controller
+COPY --from=builder /app/bin/forklift-controller /usr/local/bin/forklift-controller
```
rgolangh, line 15: this looks different than the source, please verify this path change is needed

cmd/vsphere-xcopy-volume-populator/vendor/github.com/kubev2v/forklift/pkg/lib/vsphere_offload/ssh_client.go
```diff
@@ -134,15 +135,13 @@ func (c *ESXiSSHClient) ExecuteCommand(datastore, sshCommand string, args ...str
 		return "", fmt.Errorf("SSH command timed out after 60 seconds: %s", fullCommand)
 	}
 
-	outputStr := string(output)
```
rgolangh, line 137: changes here seem unrelated.

cmd/vsphere-xcopy-volume-populator/vendor/github.com/vmware/govmomi/fault/fault.go
```diff
@@ -1,5 +1,5 @@
 // © Broadcom. All Rights Reserved.
-// The term “Broadcom” refers to Broadcom Inc. and/or its subsidiaries.
+// The term "Broadcom" refers to Broadcom Inc. and/or its subsidiaries.
```
rgolangh, line 2: WAT?

pkg/controller/plan/vib.go
```diff
@@ -0,0 +1,97 @@
...
+	if !useVIBMethod(sourceProvider.Spec.Settings[api.ESXiCloneMethod]) {
+		// Remove any existing VIB readiness conditions since VIB method is not enabled
+		plan.Status.DeleteCondition(VIBReady)
+		plan.Status.DeleteCondition(VIBNotReady)
+		return nil
+	}
+
+	// Check for provider VIB ready condition (advisory - hosts that passed validation)
+	vibReadyCondition := sourceProvider.Status.FindCondition(VIBReady)
+	if vibReadyCondition != nil && vibReadyCondition.Status == libcnd.True && len(vibReadyCondition.Items) > 0 {
```
amitosw15, line 40: use function, look at mixed validation

pkg/lib/vsphere_offload/vib.go
yaacov: my concern is that this PR moves offload stuff into MTV, AFAIU MTV does is agonistic to offload, we allow it as an external plugin but we should not add offload code inside MTV that is why it's called "offload" e.g. MTV lets other application do the storage movement. any reason for moving offload code inside MTV ?

pkg/lib/vsphere_offload/vib.go
```diff
@@ -0,0 +1,225 @@
...
+		if vibsSkipped, ok := r[0]["VIBsSkipped"]; ok && len(vibsSkipped) > 0 {
+			message := "unknown reason"
+			if msg, ok := r[0]["Message"]; ok && len(msg) > 0 {
+				message = msg[0]
+			}
+			skippedVib := ""
+			if len(vibsSkipped) > 0 {
+				skippedVib = vibsSkipped[0]
+			}
+			return fmt.Errorf("VIB installation was skipped by ESXi (host already has '%s' installed, desired version is '%s'). ESXi message: %s", skippedVib, VibVersion, message)
```
amitosw15, line 179: how to make sure this .vib version and the desired version are aligned?

build/forklift-controller/Containerfile
amitosw15: In makefile we have -o /bin/forklift-controlelr

Makefile
```diff
@@ -182,7 +184,12 @@ integration-test: generate fmt vet manifests envtest
 
 # Build forklift-controller binary
 forklift-controller: generate fmt vet
-	go build -o bin/forklift-controller github.com/kubev2v/forklift/cmd/forklift-controller
+	go build -ldflags="-w -s -X github.com/kubev2v/forklift/pkg/lib/vsphere_offload.VibVersion=$(VIB_VERSION)" -o bin/forklift-controller github.com/kubev2v/forklift/cmd/forklift-controller
+
+# Build forklift-controller for container builds (skips vet which requires libvirt-devel for virt-v2v package)
+# CI runs full vet separately with proper dependencies installed
+forklift-controller-build: generate fmt
```
amitosw15, line 191: I couldnt make this work with vet, because vsphere-xcopy-volume-populator has its own go.mod (separate   
  module), so when it runs `make build`, vet only checks that module.                                      
                                                                                                           
  But forklift-controller is part of the root module, so `go vet ./pkg/... ./cmd/...` checks ALL packages  
  including virt-v2v, which requires libvirt-devel (not available in UBI9).

pkg/lib/vsphere_offload/vib.go
mnecas: there is lot of offload logic inside the MTV, and if we will start working on the HyperV offload than there will be even more, I think having validations that are provider for offload makes sense

amitosw15:
new PRs instead: #6134  #6136

PR #3908, opened 2025-12-17
MTV-4135 | RFE:  Failed migration when no CA attached or selected 'Skip certificate validation'

Hazanel:
Issue:
vSphere migrations failed with SSL cert error when cacert was missing but insecureSkipVerify=false. 
Provider validation passed incorrectly.

Fix:
Modified validateConnectionStatus() to detect a missing cacert, resulting in a clear keyList validation error with Items: ["cacert"].

Ref: https://issues.redhat.com/browse/MTV-4135

### DIFF
--- a/pkg/controller/provider/validation.go
+++ b/pkg/controller/provider/validation.go
@@ -187,8 +187,8 @@ func (r *Reconciler) validateURL(provider *api.Provider) error {
 	return nil
 }
 
-func (r *Reconciler) validateConnectionStatus(provider *api.Provider, secret *core.Secret) bool {
-	if base.GetInsecureSkipVerifyFlag(secret) {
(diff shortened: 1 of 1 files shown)


pkg/controller/provider/validation.go
```diff
@@ -196,18 +196,28 @@ func (r *Reconciler) validateConnectionStatus(provider *api.Provider, secret *co
...
-			})
-		}
+		return true
 	}
+
+	// Check if cacert is provided when TLS verification is enabled
+	if len(secret.Data["cacert"]) == 0 {
+		// Return false to signal caller to skip further validation (GetTlsCertificate)
+		// The keyList validation will handle the error
+		return false
```
mnecas: This should be separate function it's not realted to the validateConnectionStatus.

PR #3929, opened 2025-12-21
MTV-3931 | Convert vmkfstool-wrapper script from python to shell

amitosw15:
Replace Python vmkfstools-wrapper with shell to reduce memory usage
Migrate vmkfstools-wrapper from Python to shell, reducing memory consumption from 14 MB to 4 MB per execution (71% reduction).
  - Preserve all existing functionality and logic
  - Maintain identical XML output format
  - BusyBox compatible for ESXi environment
  - Verified with comprehensive test suite

### DIFF
--- a/cmd/vsphere-xcopy-volume-populator/README.md
+++ b/cmd/vsphere-xcopy-volume-populator/README.md
@@ -81,7 +81,7 @@ spec:
 
 ## vmkfstools-wrapper
 An ESXi CLI extension that exposes the vmkfstools clone operation to API interaction.
-The folder vmkfstools-wrapper has a script to create a VIB to wrap the vmkfstools_wrapper.py
+The folder vmkfstools-wrapper has a script to create a VIB to wrap the vmkfstools_wrapper.sh
 to be a proxy to perform vmkfstools commands and more.
 The VIB should be installed on every ESXi that is connected to the datastores which
 are holds migratable VMs.
@@ -535,7 +535,7 @@ The system requires command restrictions for security. Create the restricted key
 ```bash
 # The public key needs to be prefixed with command restrictions
 # The system now uses dynamic datastore routing - a single key works for all datastores
-echo 'command="sh -c '\''DS=$(echo \"$SSH_ORIGINAL_COMMAND\" | sed -n \"s|.*/vmfs/volumes/\\([^/]*\\)/.*|\\1|p\"); exec python /vmfs/volumes/$DS/secure-vmkfstools-wrapper.py'\''",no-port-forwarding,no-agent-forwarding,no-X11-forwarding '$(cat esxi_public_key.pub) > restricted_key.pub
+echo 'command="sh -c '\''DS=$(echo \"$SSH_ORIGINAL_COMMAND\" | sed -n \"s|.*/vmfs/volumes/\\([^/]*\\)/.*|\\1|p\"); exec sh /vmfs/volumes/$DS/secure-vmkfstools-wrapper.sh'\''",no-port-forwarding,no-agent-forwarding,no-X11-forwarding '$(cat esxi_public_key.pub) > restricted_key.pub
 
 # View the final restricted key
 cat restricted_key.pub
--- a/cmd/vsphere-xcopy-volume-populator/internal/populator/secure_script.go
+++ b/cmd/vsphere-xcopy-volume-populator/internal/populator/secure_script.go
@@ -7,18 +7,18 @@ import (
 	"time"
 
 	"github.com/kubev2v/forklift/cmd/vsphere-xcopy-volume-populator/internal/vmware"
-	"github.com/kubev2v/forklift/cmd/vsphere-xcopy-volume-populator/vmkfstools-wrapper"
+	vmkfstoolswrapper "github.com/kubev2v/forklift/cmd/vsphere-xcopy-volume-populator/vmkfstools-wrapper"
 	"github.com/vmware/govmomi/object"
 	"k8s.io/klog/v2"
 )
 
 const (
-	secureScriptName = "secure-vmkfstools-wrapper"
+	scriptName = "secure-vmkfstools-wrapper"
 )
 
 // writeSecureScriptToTemp writes the embedded script to a temporary file
 func writeSecureScriptToTemp() (string, error) {
-	tempFile, err := os.CreateTemp("", "secure-vmkfstools-wrapper-*.py")
+	tempFile, err := os.CreateTemp("", "secure-vmkfstools-wrapper-*")
 	if err != nil {
 		return "", fmt.Errorf("failed to create temp file: %w", err)
 	}
@@ -34,6 +34,7 @@ func writeSecureScriptToTemp() (string, error) {
 }
 
 // ensureSecureScript ensures the secure script is uploaded and available on the target ESX
+// Returns the script path
 func ensureSecureScript(ctx context.Context, client vmware.Client, esx *object.HostSystem, datastore string) (string, error) {
 	klog.Infof("ensuring secure script on ESXi %s", esx.Name())
 
@@ -49,7 +50,6 @@ func ensureSecureScript(ctx context.Context, client vmware.Client, esx *object.H
 	if err != nil {
 		return "", fmt.Errorf("failed to upload the secure script to ESXi %s: %w", esx.Name(), err)
 	}
-	// Script will execute directly from datastore - no need for shell commands
 	klog.Infof("uploaded secure script to ESXi %s at %s - ready for execution", esx.Name(), scriptPath)
 
 	return scriptPath, nil
@@ -71,7 +71,6 @@ func uploadScript(ctx context.Context, client vmware.Client, dc *object.Datacent
 	}
 	defer os.Remove(tempScriptPath) // Clean up temp file
 
-	scriptName := fmt.Sprintf("%s.py", secureScriptName)
 	klog.Infof("Uploading embedded script to datastore as %s", scriptName)
 
 	// Upload the file with timeout
--- a/cmd/vsphere-xcopy-volume-populator/internal/powerstore/powerstore.go
+++ b/cmd/vsphere-xcopy-volume-populator/internal/powerstore/powerstore.go
@@ -3,9 +3,9 @@ package powerstore
 import (
 	"context"
 	"fmt"
-	"strings"
 	"net/http"
-	
+	"strings"
+
 	"github.com/dell/gopowerstore"
 	"github.com/kubev2v/forklift/cmd/vsphere-xcopy-volume-populator/internal/fcutil"
 	"github.com/kubev2v/forklift/cmd/vsphere-xcopy-volume-populator/internal/populator"
@@ -317,9 +317,9 @@ func NewPowerstoreClonner(hostname, username, password string, sslSkipVerify boo
 	}
 
 	client.SetCustomHTTPHeaders(http.Header{
-		"Application-Type": { "MTV" },
+		"Application-Type": {"MTV"},
 	})
-	
+
 	ctx := context.Background()
 	_, err = client.GetCluster(ctx)
 	if err != nil {
--- a/cmd/vsphere-xcopy-volume-populator/internal/vmware/ssh_client.go
+++ b/cmd/vsphere-xcopy-volume-populator/internal/vmware/ssh_client.go
@@ -134,13 +134,15 @@ func (c *ESXiSSHClient) ExecuteCommand(datastore, sshCommand string, args ...str
 		return "", fmt.Errorf("SSH command timed out after 60 seconds: %s", fullCommand)
 	}
 
+	outputStr := string(output)
+
 	if cmdErr != nil {
-		klog.Warningf("SSH command failed: %s, output: %s, error: %v", fullCommand, string(output), cmdErr)
-		return string(output), cmdErr
+		klog.Warningf("SSH command failed: %s, output: %s, error: %v", fullCommand, outputStr, cmdErr)
+		return outputStr, cmdErr
 	}
 
-	klog.V(2).Infof("SSH command succeeded: %s, output: %s", fullCommand, string(output))
-	return string(output), nil
+	klog.V(2).Infof("SSH command succeeded: %s, output: %s", fullCommand, outputStr)
+	return outputStr, nil
 }
 
 func (c *ESXiSSHClient) Close() error {
@@ -189,6 +191,9 @@ func EnableSSHAccess(ctx context.Context, vmwareClient Client, host *object.Host
 	klog.Errorf("Manual SSH key installation required. Please add the following line to /etc/ssh/keys-root/authorized_keys on the ESXi host:")
 	klog.Errorf("")
 	klog.Errorf("  %s", restrictedPublicKey)
+	klog.Errorf("")
+	klog.Errorf("The template extracts datastore from commands (DS=<datastore>;CMD=<command>)")
+	klog.Errorf("and executes: /vmfs/volumes/$DS/secure-vmkfstools-wrapper")
 	klog.Errorf("")
 	klog.Errorf("Steps to manually configure SSH key:")
 	klog.Errorf("1. SSH to the ESXi host: ssh root@%s", hostIP)
--- a/cmd/vsphere-xcopy-volume-populator/vmkfstools-wrapper/script.go
+++ b/cmd/vsphere-xcopy-volume-populator/vmkfstools-wrapper/script.go
@@ -2,5 +2,10 @@ package vmkfstoolswrapper
 
 import _ "embed"
 
-//go:embed vmkfstools_wrapper.py
+//go:embed vmkfstools_wrapper.sh
 var Script []byte
+
+// Version is the expected version of the vmkfstools wrapper script
+// This is set at build time via ldflags from version.mk
+// If not set via ldflags, defaults to "dev"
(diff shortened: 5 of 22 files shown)


amitosw15:
@rgolangh ready for review

cmd/vsphere-xcopy-volume-populator/internal/populator/secure_script.go
```diff
@@ -18,7 +18,7 @@ const (
 
 // writeSecureScriptToTemp writes the embedded script to a temporary file
 func writeSecureScriptToTemp() (string, error) {
-	tempFile, err := os.CreateTemp("", "secure-vmkfstools-wrapper-*.py")
+	tempFile, err := os.CreateTemp("", "secure-vmkfstools-wrapper-*.sh")
```
rgolangh: I think we can remove this .sh file extension. things can get simpler without it. Also it will be identical to how we install the script using vib - where we also install it without the file extension, just vmkfstools-wrapper

rgolangh: it is okay however to leave it as `.sh` on our source code.

cmd/vsphere-xcopy-volume-populator/README.md
```diff
@@ -465,7 +465,7 @@ The system requires command restrictions for security. Create the restricted key
 ```bash
 # The public key needs to be prefixed with command restrictions
 # The system now uses dynamic datastore routing - a single key works for all datastores
-echo 'command="sh -c '\''DS=$(echo \"$SSH_ORIGINAL_COMMAND\" | sed -n \"s|.*/vmfs/volumes/\\([^/]*\\)/.*|\\1|p\"); exec python /vmfs/volumes/$DS/secure-vmkfstools-wrapper.py'\''",no-port-forwarding,no-agent-forwarding,no-X11-forwarding '$(cat esxi_public_key.pub) > restricted_key.pub
+echo 'command="sh -c '\''DS=$(echo \"$SSH_ORIGINAL_COMMAND\" | sed -n \"s|.*/vmfs/volumes/\\([^/]*\\)/.*|\\1|p\"); exec sh /vmfs/volumes/$DS/secure-vmkfstools-wrapper.sh'\''",no-port-forwarding,no-agent-forwarding,no-X11-forwarding '$(cat esxi_public_key.pub) > restricted_key.pub
```
rgolangh, line 538: what happens on installation that has the old 'command' form with the python keyword? can we catch that and warn users?

cmd/vsphere-xcopy-volume-populator/vmkfstools-wrapper/vmkfstools_wrapper.sh
```diff
@@ -0,0 +1,571 @@
...
+    printf '%s' "$str"
+}
+
+validate_path() {
+    local path="$1"
+
+    log_info "vmkfstools-wrapper version ${SCRIPT_VERSION} validating path: ${path}"
+
+    # CRITICAL: Normalize FIRST before security checks
+    path=$(echo "${path}" | sed 's|//*|/|g')
```
rgolangh, line 75: TIL that sed can use '|' in the expression instead of '/' - nice!

cmd/vsphere-xcopy-volume-populator/vmkfstools-wrapper/vmkfstools_wrapper.sh
```diff
@@ -0,0 +1,571 @@
...
+        # Fallback to /dev/random
+        local hex
+        hex=$(dd if=/dev/random bs=16 count=1 2>/dev/null | od -A n -t x1 | sed 's/ //g' | sed 's/\n//g')
+
+        local p1=$(echo "${hex}" | cut -c1-8)
+        local p2=$(echo "${hex}" | cut -c9-12)
+        local p3=$(echo "${hex}" | cut -c13-16)
+        local p4=$(echo "${hex}" | cut -c17-20)
+        local p5=$(echo "${hex}" | cut -c21-32)
+
```
rgolangh: more efficient and short in one go, no intermediate vars:

```
echo ${hex:0:8}-${hex:8:4}-${hex:12:4}-${hex:16:4}-${hex:20:12}
```

cmd/vsphere-xcopy-volume-populator/vmkfstools-wrapper/vmkfstools_wrapper.sh
```diff
@@ -0,0 +1,571 @@
...
+        echo "${p1}-${p2}-${p3}-${p4}-${p5}"
+    else
+        # Last resort fallback: timestamp-pid-hash
+        # Use md5sum if available to create more randomness
+        local seed="$(date +%s)-$$-$(date +%N 2>/dev/null || echo $$)"
+        if command -v md5sum >/dev/null 2>&1; then
```
rgolangh: not sure if we need all those fallbacks. I'm fine with the first two methods.

cmd/vsphere-xcopy-volume-populator/vmkfstools-wrapper/vmkfstools_wrapper.sh
```diff
@@ -0,0 +1,571 @@
...
+    fi
+
+    echo ""
+}
+
+# Generate UUID-compatible ID for ESXi
+# Format: xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
+generate_task_id() {
+    # Use /dev/urandom (available on ESXi)
+    if [ -r /dev/urandom ]; then
```
rgolangh, line 146: the code for /dev/urandom and /dev/random is identical. I would use a variable for the device to use and use that for `dd` command , this should shrink that implementation further down

cmd/vsphere-xcopy-volume-populator/vmkfstools-wrapper/vmkfstools_wrapper.sh
```diff
@@ -0,0 +1,571 @@
...
+
+    local xcopy_used="null"
+    local target_lun
+    if [ -f "${task_dir}/targetLun" ]; then
+        target_lun=$(cat "${task_dir}/targetLun" 2>/dev/null)
+        if [ -n "${target_lun}" ]; then
+            # Try to determine xcopy usage, but keep as null if vsish fails
+            local xcopy_result
+            # Don't redirect stderr - was_xcopy_used logs to stderr, we only want stdout
+            if xcopy_result=$(was_xcopy_used "${target_lun}"); then
```
rgolangh: I suspect this always be evaluated as true no?

amitosw15: yep you are right this is ambiguous, fixing it

cmd/vsphere-xcopy-volume-populator/README.md
amitosw15: It will use the version to detect that.
After the upload succeeded, if the runtime version of the script is lower than expected, it means we run different script than we thought we would.

amitosw15:
@rgolangh Ready for review

cmd/vsphere-xcopy-volume-populator/vmkfstools-wrapper/vmkfstools_wrapper.sh
```diff
@@ -0,0 +1,550 @@
+#!/bin/sh
+
+TMP_PREFIX="/tmp/vmkfstools-wrapper-"
+SCRIPT_VERSION="0.3.0"
```
rgolangh, line 6: good, thanks

cmd/vsphere-xcopy-volume-populator/internal/vmware/ssh_client.go
```diff
@@ -236,3 +244,90 @@ func GetHostIPAddress(ctx context.Context, host *object.HostSystem) (string, err
 
 	return ips[0].String(), nil
 }
+
+func CheckScriptVersion(sshClient SSHClient, datastore, embeddedVersion string, publicKey []byte) error {
+	output, err := sshClient.ExecuteCommand(datastore, "--version")
+	if err != nil {
+		return fmt.Errorf("old script format detected (likely Python-based). Update script on datastore %s to version %s or newer: %w", datastore, embeddedVersion, err)
```
rgolangh: if this is python answering here then it could be a public key 'command' directive error as well.

But perhaps a nitpic here

cmd/vsphere-xcopy-volume-populator/internal/vmware/ssh_client.go
```diff
@@ -236,3 +244,90 @@ func GetHostIPAddress(ctx context.Context, host *object.HostSystem) (string, err
...
+	}
+
+	var resp struct {
+		XMLName   xml.Name `xml:"output"`
```
rgolangh: you have those structures in populator package. see if you can reuse them. I would hate to see that code duplicated.

rgolangh: I guess you can add the version as a method there. You'll get a lot of the handling already there

cmd/vsphere-xcopy-volume-populator/vmkfstools-wrapper/script.go
```diff
@@ -2,5 +2,10 @@ package vmkfstoolswrapper
...
 import _ "embed"
 
-//go:embed vmkfstools_wrapper.py
+//go:embed vmkfstools_wrapper.sh
 var Script []byte
+
+// Version is the expected version of the vmkfstools wrapper script
+// This is set at build time via ldflags from version.mk
+// If not set via ldflags, defaults to "dev"
+var Version = "dev"
```
rgolangh, line 11: on one hand reusing VIB_VERSION makes sesnse, but it looks odd. Because VIB_VERSION is used for both the 'script version' AND the VIB VERSION.

I'm okay with this after all

cmd/vsphere-xcopy-volume-populator/internal/powerstore/powerstore.go
```diff
@@ -317,9 +317,9 @@ func NewPowerstoreClonner(hostname, username, password string, sslSkipVerify boo
 	}
 
 	client.SetCustomHTTPHeaders(http.Header{
-		"Application-Type": { "MTV" },
+		"Application-Type": {"MTV"},
```
rgolangh, line 320: go vet fix that I probably missed from a review

cmd/vsphere-xcopy-volume-populator/vmkfstools-wrapper/vmkfstools_wrapper.sh
```diff
@@ -0,0 +1,550 @@
...
+}
+
+log_warning() {
+    local message="$1"
+    echo "$(date '+%Y-%m-%d %H:%M:%S') - WARNING - ${message}" >> "${LOG_FILE}"
+    echo "$(date '+%Y-%m-%d %H:%M:%S') - WARNING - ${message}" >&2
+}
+
+# Escape special characters for JSON strings
+escape_json() {
```
rgolangh, line 39: The only values in the json we construct which are not of type integer are 'lastLine' and 'err' . Is this handling is specifically for those?

cmd/vsphere-xcopy-volume-populator/vmkfstools-wrapper/vmkfstools_wrapper.sh
```diff
@@ -0,0 +1,550 @@
...
+    local str="$1"
+
+    # IMPORTANT: Escape backslashes FIRST, before other escapes
+    # Otherwise we'll double-escape the escape sequences!
+
+    # 1. Escape backslashes first ( \ -> \\ )
+    str="${str//\\/\\\\}"
```
rgolangh, line 46: again not sure why is this needed? in which case?

cmd/vsphere-xcopy-volume-populator/vmkfstools-wrapper/vmkfstools_wrapper.sh
```diff
@@ -0,0 +1,550 @@
...
+
+    # 2. Escape double quotes ( " -> \" )
+    str="${str//\"/\\\"}"
```
rgolangh, line 49: ok I guess all of this was handled for us by python json.dumps, which is probably one of the reason I picked python in the begging.

amitosw15: Yep its all attempts to implement shell version for `json dump`

cmd/vsphere-xcopy-volume-populator/vmkfstools-wrapper/vmkfstools_wrapper.sh
```diff
@@ -0,0 +1,550 @@
...
+    else
+        echo "false ${write_ops}"
+    fi
+}
+
+task_get() {
+    local task_id="$1"
+
+    if [ -z "${task_id}" ]; then
+        local json_result="{\"error\": \"-i (task-id) is required for task-get\"}"
```
rgolangh: we don't have 'error' field in the backend vmkfstoolsTask structure. We do have StdError and ExitCode

cmd/vsphere-xcopy-volume-populator/vmkfstools-wrapper/vmkfstools_wrapper.sh
```diff
@@ -0,0 +1,550 @@
...
+        xml_output "1" "${json_result}"
+        return 1
+    fi
+
+    local task_dir="${TMP_PREFIX}${task_id}"
+
+    if [ ! -d "${task_dir}" ]; then
+        local json_result="{\"taskId\": \"${task_id}\", \"error\": \"Task directory ${task_dir} not found\"}"
```
rgolangh: same comment as line 316

cmd/vsphere-xcopy-volume-populator/vmkfstools-wrapper/vmkfstools_wrapper.sh
```diff
@@ -0,0 +1,550 @@
...
+        xml_output "1" "${json_result}"
+        return 1
+    fi
+
+    local pid
+    pid=$(cat "${task_dir}/pid" 2>/dev/null) || {
+        xml_output "1" "{\"taskId\": \"${task_id}\", \"error\": \"Failed to read PID\"}"
```
rgolangh: same comment as line 316

rgolangh: while at itI think its worth having a function just like xml_output for json_output?

cmd/vsphere-xcopy-volume-populator/vmkfstools-wrapper/vmkfstools_wrapper.sh
```diff
@@ -0,0 +1,550 @@
...
+    fi
+
+    xml_output "0" "${json_result}"
+}
+
+task_clean() {
+    local task_id="$1"
+
+    if [ -z "${task_id}" ]; then
+        local json_result="{\"error\": \"-i (task-id) is required for task-clean\"}"
```
rgolangh: same as line 316

cmd/vsphere-xcopy-volume-populator/vmkfstools-wrapper/vmkfstools_wrapper.sh
```diff
@@ -0,0 +1,550 @@
...
+        xml_output "1" "${json_result}"
+        return 1
+    fi
+
+    local task_dir="${TMP_PREFIX}${task_id}"
+
+    if [ ! -d "${task_dir}" ]; then
+        local json_result="{\"taskId\": \"${task_id}\", \"error\": \"Task directory ${task_dir} not found\"}"
```
rgolangh: same as line 316

cmd/vsphere-xcopy-volume-populator/vmkfstools-wrapper/vmkfstools_wrapper.sh
```diff
@@ -0,0 +1,553 @@
+#!/bin/sh
+
+set -e
```
rgolangh, line 3: yesss, good

cmd/vsphere-xcopy-volume-populator/vmkfstools-wrapper/vmkfstools_wrapper.sh
```diff
@@ -0,0 +1,553 @@
...
+        <field name="status"><string>${status}</string></field>
+        <field name="message"><string>${message}</string></field>
+    </structure>
+</output>
+EOF
+}
+
+log_info() {
+    local message="$1"
+    logger -t "${LOGIN_TAG}" -p INFO "${message}"
```
rgolangh, line 25: We would have to create a release note for this. and perhaps even update the docs.

amitosw15: Add to this pr? how do i do that?

cmd/vsphere-xcopy-volume-populator/vmkfstools-wrapper/vmkfstools_wrapper.sh
amitosw15: Took this internally, will not implement json func

PR #3950, opened 2025-12-23
chore(security): apply security updates for MTV images

Hazanel:
Uses `dnf update -y --security` when `dnf` is available (full UBI images) Falls back to `microdnf update -y` when only `microdnf` is available (minimal UBI images)

### DIFF
--- a/build/forklift-api/Containerfile
+++ b/build/forklift-api/Containerfile
@@ -9,7 +9,10 @@ RUN --mount=type=cache,target=${GOCACHE},uid=1001 go build -buildvcs=false -ldfl
 
 FROM registry.access.redhat.com/ubi9-minimal:9.6-1752587672
 # Required to be able to get files from within the pod
-RUN microdnf -y install tar && microdnf clean all
+RUN microdnf -y install tar && \
+    # Update packages to address security vulnerabilities (CVE fixes)
+    microdnf update -y && \
+    microdnf clean all
 COPY --from=builder /app/forklift-api /usr/local/bin/forklift-api
 
 ENTRYPOINT ["/usr/local/bin/forklift-api"]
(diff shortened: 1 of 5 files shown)


build/forklift-api/Containerfile
```diff
@@ -9,7 +9,17 @@ RUN --mount=type=cache,target=${GOCACHE},uid=1001 go build -buildvcs=false -ldfl
 
 FROM registry.access.redhat.com/ubi9-minimal:9.6-1752587672
 # Required to be able to get files from within the pod
-RUN microdnf -y install tar && microdnf clean all
+# Update packages to address security vulnerabilities (CVE fixes)
+# Use dnf --security if available (full UBI), otherwise microdnf update (minimal UBI)
+RUN if command -v dnf &> /dev/null; then \
```
yaacov: in line 10 we define the image as ubi9-minimal, UBI minimal images, use the microdnf command [1].

[1] https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/9/html/building_running_and_managing_containers/assembly_adding-software-to-a-ubi-container_building-running-and-managing-containers

Hazanel: right so the workaround here is either use dnf and apply only security updates or use minimal microdnf and apply all the updates.
That's something also @mnecas looked on - whatever we'll decide, most importantly, that way  we'll have security issues up to date

Hazanel:
@mnecas do we still need this or I can close?

solenoci:
For upstream containerfiles this seems fine, downstream will not work as during the build there is no internet access (builds are hermetic). Please remove modifications for d/s containerfiles.

Hazanel:
@solenoci  - Done

solenoci:
@Hazanel Hi, let's remove the switch between dnf and microdnf from the containerfiles so it's easier to read. We try to get so d/s builds as close as possible and d/s can't usse anything other than minimal. Thanks

PR #3952, opened 2025-12-23
MTV-4161 | Validate NAA label values before adding to PVC labels

borod108:
Use k8svalidation.IsValidLabelValue() to check if the NAA value from vSphere datastores is valid for Kubernetes labels. If invalid, omit the label rather than failing resource creation.

Resolves: MTV-4161

### DIFF
--- a/pkg/controller/plan/adapter/vsphere/builder.go
+++ b/pkg/controller/plan/adapter/vsphere/builder.go
@@ -1382,9 +1382,12 @@ func (r *Builder) PopulatorVolumes(vmRef ref.Ref, annotations map[string]string,
 				labels := map[string]string{
 					"migration": string(r.Migration.UID),
 					// we need uniqness and a value which is less than 64 chars, hence using vmRef.id + disk.key
-					"vmdkKey":        fmt.Sprint(disk.Key),
-					"vmID":           vmRef.ID,
-					TemplateNAALabel: naa,
+					"vmdkKey": fmt.Sprint(disk.Key),
+					"vmID":    vmRef.ID,
+				}
+				// Only add the NAA label if it's a valid Kubernetes label value
+				if errs := k8svalidation.IsValidLabelValue(naa); len(errs) == 0 {
+					labels[TemplateNAALabel] = naa
 				}
 
 				r.Log.Info("target namespace for migration", "namespace", namespace)


pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -84,6 +84,24 @@ const (
 	TemplateNAALabel      = "volume.csi.k8s.io/affinity-source-naa"
 )
 
+var labelValueSanitizer = regexp.MustCompile(`[^A-Za-z0-9\-_.]`)
```
rgolangh: didn't we have colons `:`  in that naa as well?
I don't understand how all those chars are part of naa to begin with. 
Couuld it be that we have a bug in the query?

rgolangh: According to the bug report the bad naa was something like `mpx.vmhba0:00:00:00`  (I replaced the real chars with zeros)

pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -84,6 +84,24 @@ const (
...
+
+func sanitizeLabelValue(value string) string {
```
rgolangh: I think there must be a kubernetes function for that,

pkg/controller/plan/adapter/vsphere/builder.go
rgolangh: see k8s.io/apimachinery/pkg/util/validation https://pkg.go.dev/k8s.io/apimachinery/pkg/util/validation#IsDNS1035Label

pkg/controller/plan/adapter/vsphere/builder.go
borod108: They have validation functions in the lib you sent, I did not see sanitization ones.

borod108:
I think moving into an annotation instead is the best solution - if you agree I will do that.
MPX is not useful right now, but maybe in the future, so I think preserving data is best.
The solution in this PR was quick just to unblock testing, and since MPX format is not useful to us at the moment, I thought doing this for now will give us enough to debug this in the future, and not hurt any current flows we are testing.
Anyway - your idea of moving to annotations is just better. Let me know.

borod108:
Update (as we discussed): MPX is useful because when we try to vmkfs clone from mpx to array at least it runs the copy on the host (fallback from xcopy) - waiting for reply from provider csi team regarding moving to annotations, but in any case - I think this should be merged and improved on when we get a reply.

PR #3980, opened 2025-12-24
MTV-3789 | Use uncached Get in Plan reconciliation.

mrnold:
Issue: sometimes two Plan reconciliations get queued to run one after the other with no requeue delay, and an Update to the Plan from the first reconciliation does not get propagated to the Get in the second reconciliation. That causes the second reconciliation to re-run the same phase, which causes problems if that phase needs to work with external resources like VMware snapshots. So far I have seen this cause duplicate snapshot creations, duplicate snapshot removals, and duplicate VM poweroffs.

Fix: for a short-term fix, change the Plan reconciliation to use an uncached Get. On my test cluster, I was able to reproduce the bug consistently within about three hours, and with the fix I successfully ran a warm migration for 14 hours with a snapshot interval of 1. I believe it is still possible to hit this, but this fix makes it much rarer. I am not certain that it will fix all the related warm migration bugs listed in MTV-3789, but it at least makes it possible to work on them without running into this every time.

Resolves: [MTV-3789](https://issues.redhat.com/browse/MTV-3789), partially. Longer term work involves carefully making sure the reconciles are idempotent, finding out exactly what else is triggering extra Plan reconciliations, and figuring out why so many migrations started hitting this bug all of a sudden.

### DIFF
--- a/pkg/controller/plan/controller.go
+++ b/pkg/controller/plan/controller.go
@@ -64,6 +64,7 @@ func Add(mgr manager.Manager) error {
 			Client:        mgr.GetClient(),
 			Log:           log,
 		},
+		APIReader: mgr.GetAPIReader(),
 	}
 	cnt, err := controller.New(
 		Name,
@@ -160,6 +161,7 @@ var _ reconcile.Reconciler = &Reconciler{}
 // Reconciles a Plan object.
 type Reconciler struct {
 	base.Reconciler
+	APIReader client.Reader
 }
 
 // Reconcile a Plan CR.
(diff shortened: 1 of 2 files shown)


rgolangh:
@mrnold could it be that those two reconciliations are actually running concurrently and not one after another?

mrnold:
> @mrnold could it be that those two reconciliations are actually running concurrently and not one after another?

@rgolangh I didn't think it was possible to get concurrent reconciliations on the same resource, but I don't know for sure. I can try some kind of mutex and see if it makes a difference.

mnecas (approved):
We have just one pod and one controller process running at the time.
We do have `MaxConcurrentReconciles` however that runs only on one resrouce at the time.
From the logs I could see that this issue happens when the reconcile is started due to updates to watched resource ie NetworkMap StorageMap Migraiton etc. This causes the plan reconcile to trigger right after the previous reconile without any wait in between.
When Matthew was doing the testing we no longer could see the issue
LGTM

PR #3994, opened 2025-12-30
MTV-3745 | RFE: Mac conflict post successful migration

Hazanel:
This is a complementary PR for https://github.com/kubev2v/forklift/pull/3358

So by skipping  validating VMs that the Plan considers to have been successfully migrated, we will also prevent this false mac conflict error.

Ref: https://issues.redhat.com/browse/MTV-4151
https://issues.redhat.com/browse/MTV-3475

### DIFF
--- a/pkg/controller/plan/validation.go
+++ b/pkg/controller/plan/validation.go
@@ -775,6 +775,14 @@ func (r *Reconciler) validateVM(plan *api.Plan) error {
 	for i := range plan.Spec.VMs {
 		vm := &plan.Spec.VMs[i]
 		ref := &vm.Ref
+
+		// Skip VMs that have already succeeded - no validation needed
+		if status, found := plan.Status.Migration.FindVM(*ref); found {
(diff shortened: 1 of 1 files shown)


pkg/controller/plan/validation.go
```diff
@@ -831,6 +831,15 @@ func (r *Reconciler) validateVM(plan *api.Plan) error {
...
 		}
+
+		// We don't need to revalidate a VM that was already successfully
+		// migrated by this plan, but we still want to consider it for
+		// name and ID uniqueness checks.
+		if status, found := plan.Status.Migration.FindVM(*ref); found {
+			if status.HasCondition(api.ConditionSucceeded) {
+				continue
+			}
+		}
```
mnecas: This is to skip the validation for succedded, could we ommit the VM before the validation step?

Hazanel: I pushed it up , so now we don't have redundant validation steps for succeeded VM's

PR #4001, opened 2025-12-31
MTV-3932 | Fix race condition in parallel script uploads using UUID-based filenames

amitosw15:
Change SSH script upload from using a constant filename to UUID-based unique filenames to prevent race conditions during parallel populator execution.

Problem:
- Multiple pods upload to the same filename: secure-vmkfstools-wrapper.py
- ds.UploadFile() is not atomic: it truncates the file to 0 bytes first, then streams content in chunks
- Due to the short length of the secure script, parallel uploads can cause:
  * Pod P2 executing script while Pod P1 is uploading → empty file execution
  * Partial/corrupt script execution leading to failures
- Evidence: 8 instances of 0-byte reads, 118 instances of partial/corrupt reads in stress tests

Solution:
- Each process uploads script to unique UUID-based filename: secure-vmkfstools-wrapper-{UUID}.py
- Script is executed directly from the UUID-named file via SSH
- Updated SSH command template to extract UUID from command format: DS=<datastore>;UUID=<uuid>;CMD=<command>
- SSH authorized_keys template parses UUID and executes: /vmfs/volumes/$DS/secure-vmkfstools-wrapper-$UUID.py

Changes:
- secure_script.go: Generate UUID for each upload, return path and UUID
- ssh_client.go: Add SetScriptUUID() method, include UUID in SSH commands
- sshutils.go: Update RestrictedSSHCommandTemplate to parse and use UUID
- remote_esxcli.go: Extract UUID from ensureSecureScript return value
- secure_script_test.go: Add tests for UUID generation and format validation
- vsphere-xcopy-volume-populator.go: Add DISABLE_METRICS_SERVER flag

This ensures each parallel execution operates on its own unique script instance, eliminating race conditions without requiring atomic move operations.

Resolves: MTV-3932

### DIFF
--- a/cmd/vsphere-xcopy-volume-populator/internal/populator/remote_esxcli.go
+++ b/cmd/vsphere-xcopy-volume-populator/internal/populator/remote_esxcli.go
@@ -146,7 +146,8 @@ func (p *RemoteEsxcliPopulator) Populate(vmId string, sourceVMDKFile string, pv 
 	// as the possible clonner identifier
 	if isSciniRequired {
 		klog.Infof("scini is required for the storage api")
-		sciModule, err := p.VSphereClient.RunEsxCommand(context.Background(), host, []string{"system", "module", "parameters", "list", "-m", "scini"})
+		sciModule, errScini := p.VSphereClient.RunEsxCommand(context.Background(), host, []string{"system", "module", "parameters", "list", "-m", "scini"})
+		err = errScini
 		if err != nil {
 			klog.Infof("failed to fetch the scini module parameters %s: ", err)
 			return err
@@ -329,14 +330,26 @@ func (p *RemoteEsxcliPopulator) Populate(vmId string, sourceVMDKFile string, pv 
 		defer sshCancel()
 
 		// Setup secure script
-		finalScriptPath, err := ensureSecureScript(sshSetupCtx, p.VSphereClient, host, vmDisk.Datastore)
+		scriptPath, scriptUUID, err := ensureSecureScript(sshSetupCtx, p.VSphereClient, host, vmDisk.Datastore)
 		if err != nil {
 			return fmt.Errorf("failed to ensure secure script: %w", err)
 		}
-		klog.V(2).Infof("Secure script ready at path: %s", finalScriptPath)
+		klog.V(2).Infof("Secure script ready: %s (UUID: %s)", scriptPath, scriptUUID.String())
 
-		// Enable SSH access
-		err = vmware.EnableSSHAccess(sshSetupCtx, p.VSphereClient, host, p.SSHPrivateKey, p.SSHPublicKey, finalScriptPath)
+		// Extract script filename for cleanup
+		scriptName := fmt.Sprintf("%s-%s.py", secureScriptName, scriptUUID.String())
+		dc, errDC := getHostDC(host)
+		if errDC == nil {
+			// Cleanup script file when done (best-effort, non-blocking)
+			defer func() {
+				cleanupCtx, cleanupCancel := context.WithTimeout(context.Background(), contextTimeout)
+				defer cleanupCancel()
+				cleanupSecureScript(cleanupCtx, p.VSphereClient, dc, vmDisk.Datastore, scriptName)
+			}()
+		}
+
+		// Enable SSH access (pass full path for logging, but UUID is what matters)
+		err = vmware.EnableSSHAccess(sshSetupCtx, p.VSphereClient, host, p.SSHPrivateKey, p.SSHPublicKey, scriptPath)
 		if err != nil {
 			return fmt.Errorf("failed to enable SSH access: %w", err)
 		}
@@ -347,15 +360,14 @@ func (p *RemoteEsxcliPopulator) Populate(vmId string, sourceVMDKFile string, pv 
 			return fmt.Errorf("failed to get host IP address: %w", err)
 		}
 
-		// Create SSH client with background context (no timeout for long-running operations)
-		sshClient := vmware.NewSSHClient()
+		sshClient := vmware.NewSSHClient(scriptUUID.String())
 		err = sshClient.Connect(context.Background(), hostIP, "root", p.SSHPrivateKey)
 		if err != nil {
 			return fmt.Errorf("failed to connect via SSH: %w", err)
 		}
 		defer sshClient.Close()
 
-		klog.V(2).Infof("SSH connection established with restricted commands")
+		klog.V(2).Infof("SSH connection established with restricted commands (script UUID: %s)", scriptUUID.String())
 		executor = NewSSHTaskExecutor(sshClient)
 	} else {
 		executor = NewVIBTaskExecutor(p.VSphereClient)
--- a/cmd/vsphere-xcopy-volume-populator/internal/populator/secure_script.go
+++ b/cmd/vsphere-xcopy-volume-populator/internal/populator/secure_script.go
@@ -4,16 +4,19 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"strings"
 	"time"
 
+	"github.com/google/uuid"
 	"github.com/kubev2v/forklift/cmd/vsphere-xcopy-volume-populator/internal/vmware"
-	"github.com/kubev2v/forklift/cmd/vsphere-xcopy-volume-populator/vmkfstools-wrapper"
+	vmkfstoolswrapper "github.com/kubev2v/forklift/cmd/vsphere-xcopy-volume-populator/vmkfstools-wrapper"
 	"github.com/vmware/govmomi/object"
 	"k8s.io/klog/v2"
 )
 
 const (
 	secureScriptName = "secure-vmkfstools-wrapper"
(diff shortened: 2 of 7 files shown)


cmd/vsphere-xcopy-volume-populator/internal/populator/remote_esxcli.go
```diff
@@ -329,14 +329,21 @@ func (p *RemoteEsxcliPopulator) Populate(vmId string, sourceVMDKFile string, pv
 		defer sshCancel()
 
 		// Setup secure script
-		finalScriptPath, err := ensureSecureScript(sshSetupCtx, p.VSphereClient, host, vmDisk.Datastore)
+		scriptPathWithUUID, err := ensureSecureScript(sshSetupCtx, p.VSphereClient, host, vmDisk.Datastore)
```
borod108: I think it may make more sense to change the output of this function to return 2 strings, then we will not need post processing, what do you think?

cmd/vsphere-xcopy-volume-populator/internal/populator/remote_esxcli.go
```diff
@@ -355,7 +362,12 @@ func (p *RemoteEsxcliPopulator) Populate(vmId string, sourceVMDKFile string, pv
 		}
 		defer sshClient.Close()
 
-		klog.V(2).Infof("SSH connection established with restricted commands")
+		// Set the script UUID in the SSH client so it can be used in commands
+		if sshClientImpl, ok := sshClient.(*vmware.ESXiSSHClient); ok {
+			sshClientImpl.SetScriptUUID(scriptUUID)
```
borod108: can we just add it to when we create the client instead of doing this here separately?

cmd/vsphere-xcopy-volume-populator/internal/populator/secure_script.go
```diff
@@ -45,43 +46,47 @@ func ensureSecureScript(ctx context.Context, client vmware.Client, esx *object.H
...
 		return "", fmt.Errorf("failed to upload the secure script to ESXi %s: %w", esx.Name(), err)
 	}
-	// Script will execute directly from datastore - no need for shell commands
-	klog.Infof("uploaded secure script to ESXi %s at %s - ready for execution", esx.Name(), scriptPath)
+	// Script will execute directly from datastore using UUID filename
+	klog.Infof("uploaded secure script to ESXi %s at %s (UUID: %s) - ready for execution", esx.Name(), scriptPath, guid)
 
-	return scriptPath, nil
+	// Return both path and UUID - UUID will be used in SSH commands
+	return scriptPath + ":" + guid, nil
```
borod108: mentioned in another comment - let's just return two strings?

cmd/vsphere-xcopy-volume-populator/internal/vmware/ssh_client.go
```diff
@@ -153,6 +156,11 @@ func (c *ESXiSSHClient) Close() error {
 	return nil
 }
 
+// This UUID is used in SSH commands to identify which script file to execute
+func (c *ESXiSSHClient) SetScriptUUID(uuid string) {
```
borod108: I think we may want to drop that and just set in constructor.

amitosw15:
> Great PR, small comments. You added tests - awesome. Also - do we cleanup the script file?

good catch, i added a cleanup mechanism.

cmd/vsphere-xcopy-volume-populator/internal/populator/remote_esxcli.go
amitosw15: sounds good

cmd/vsphere-xcopy-volume-populator/internal/populator/remote_esxcli.go
```diff
@@ -146,7 +146,8 @@ func (p *RemoteEsxcliPopulator) Populate(vmId string, sourceVMDKFile string, pv
 	// as the possible clonner identifier
 	if isSciniRequired {
 		klog.Infof("scini is required for the storage api")
-		sciModule, err := p.VSphereClient.RunEsxCommand(context.Background(), host, []string{"system", "module", "parameters", "list", "-m", "scini"})
+		sciModule, errScini := p.VSphereClient.RunEsxCommand(context.Background(), host, []string{"system", "module", "parameters", "list", "-m", "scini"})
+		err = errScini
```
borod108, line 150: this seems over complicated somehow

borod108:
awesome!
More small comments:
"%s-%s.py" repeats 6 times - makes me feel like we can have a small refactor - maybe put it in utils? (I noticed it now that it is used in cleanup - I fear that we may change the format in one place and then not in another one day)
also - I would say lets change all "guid" to "uuid" unless you have a strong reason to call it guid - this is not critical at all - but if you read who made "guid"...

rgolangh:
Now that the key needs UUID in the ssh command, how do we expire those old authorized key that are no longer valid? 
That is hard to manage... That is not a new problem. We always had that and we need to minimize it.

I suggest an alternative that keeps the old name and does not add UUID:
- upload the script with the $TMP_NAME, here the uuid is a good use.
- move $TMP_NAME to the final name. This acts similar to the linux 'mv' command and is atomic. See `FileManager.MoveDataStoreFile`
- if the file exists then success or no other error then success. Remove $TMP_NAME file
- What happens if we want to force upload? like for new version? need to think about that

amitosw15:
We can change the `ds.UploadFile` to be conditional, if the file doesnt exist upload it. It will create a technical debt for the provider to verify that the file on the host is there and updated. I did some tests on this, and when the script doesnt exist we'll still have a race, but for the normal case it will solve the problem

@borod108 @rgolangh

cmd/vsphere-xcopy-volume-populator/internal/populator/secure_script.go
```diff
@@ -34,54 +36,89 @@ func writeSecureScriptToTemp() (string, error) {
...
+	defer dsCancel()
+	ds, err := client.GetDatastore(dsCtx, dc, datastore)
+	if err != nil {
+		klog.Warningf("Failed to get datastore for cleanup: %v", err)
+		return
+	}
+
+	fileManager := ds.NewFileManager(dc, false)
+
+	delCtx, delCancel := context.WithTimeout(ctx, 10*time.Second)
```
rgolangh: give it more time to delete, like 5 minutes

rgolangh: constant it away and look at the parent context

cmd/vsphere-xcopy-volume-populator/internal/populator/secure_script.go
```diff
@@ -34,54 +36,89 @@ func writeSecureScriptToTemp() (string, error) {
...
 	if err != nil {
-		return "", fmt.Errorf("failed to write embedded script to temp file: %w", err)
+		return "", "", fmt.Errorf("failed to write embedded script to temp file: %w", err)
 	}
 	defer os.Remove(tempScriptPath) // Clean up temp file
 
-	scriptName := fmt.Sprintf("%s.py", secureScriptName)
-	klog.Infof("Uploading embedded script to datastore as %s", scriptName)
+	guid := uuid.New().String()
+	scriptName := fmt.Sprintf("%s-%s.py", secureScriptName, guid)
```
rgolangh: Boris said to call it uuid

rgolangh:
@borod108 thanks for the review
@amitosw15 nice code coverage
Waiting for the rest of the comments fixes - after those we are ready to merge

cmd/vsphere-xcopy-volume-populator/internal/populator/remote_esxcli.go
```diff
@@ -329,14 +330,26 @@ func (p *RemoteEsxcliPopulator) Populate(vmId string, sourceVMDKFile string, pv
...
-		klog.V(2).Infof("Secure script ready at path: %s", finalScriptPath)
+		klog.V(2).Infof("Secure script ready: %s (UUID: %s)", scriptPath, scriptUUID.String())
+
+		// Extract script filename for cleanup
+		scriptName := fmt.Sprintf("%s-%s.py", secureScriptName, scriptUUID.String())
+		dc, errDC := getHostDC(host)
+		if errDC == nil {
+			// Cleanup script file when done (best-effort, non-blocking)
+			defer func() {
+				cleanupCtx, cleanupCancel := context.WithTimeout(context.Background(), 10*time.Second)
```
rgolangh: still too short of a parent context

amitosw15: sorry missed that, fixed

yaacov:
@amitosw15 @rgolangh @mnecas  @solenoci  hi, this PR breaks the CI for me :-(

does the tests run for you after this patch ?

amitosw15:
> @amitosw15 @rgolangh @mnecas @solenoci hi, this PR breaks the CI for me :-(
> 
> does the tests run for you after this patch ?

https://github.com/kubev2v/forklift/pull/4128

PR #4002, opened 2026-01-01
MTV-4194 | Fix prime PVCs stuck in Terminating state after migration

amitosw15:
Prime PVCs get stuck in Terminating state and are never deleted after migration completes. This issue affects all migrations using volume populators (xcopy, storage offload) since October 2023.

Root Cause:
Prime PVCs have the kubernetes.io/pvc-protection finalizer which prevents deletion. Currently, this finalizer is never removed because:
1. deleteCorrespondingPrimePVC() calls DeleteObject() but doesn't remove finalizers (unlike deletePopulatedPVC() which does)
2. Prime PVC cleanup only runs during plan archiving - it never runs after migration completes

This doesn't align with the intended role of prime PVCs as temporary resources. They should be cleaned up immediately after migration finishes (success or failure), not wait for plan archiving.

Evidence:
- Found 50+ prime PVCs stuck in Terminating state on test cluster
- Manual finalizer removal causes immediate deletion, confirming the fix
- Bug introduced in commit 18bc35cd5 when DeletePVCs was refactored

The Fix:
1. Add finalizer removal to deleteCorrespondingPrimePVC() to match deletePopulatedPVC() behavior
2. Add new DeletePrimePVCs() function to delete only prime PVCs while preserving target PVCs
3. Trigger deletePrimePVCs() immediately after migration completes (PhaseCompleted) for both successful and failed migrations
4. Add safety net in cleanup() for plan archiving in case cleanup failed

Testing:
- Added 4 comprehensive unit tests (24 total tests, all passing)
- Manually verified fix on production cluster
- No breaking changes, backwards compatible

Resolves: MTV-3039

### DIFF
--- a/pkg/controller/plan/kubevirt.go
+++ b/pkg/controller/plan/kubevirt.go
@@ -1426,6 +1426,20 @@ func (r *KubeVirt) DeletePopulatedPVCs(vm *plan.VMStatus) error {
 	return nil
 }
 
+// Deletes only the prime PVCs, leaving the target PVCs intact
+func (r *KubeVirt) DeletePrimePVCs(vm *plan.VMStatus) error {
+	pvcs, err := r.getPVCs(vm.Ref)
+	if err != nil {
+		return err
+	}
+	for _, pvc := range pvcs {
+		if err = r.deleteCorrespondingPrimePVC(pvc, vm); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
 func (r *KubeVirt) deleteCorrespondingPrimePVC(pvc *core.PersistentVolumeClaim, vm *plan.VMStatus) error {
 	primePVC := core.PersistentVolumeClaim{}
 	err := r.Destination.Client.Get(context.TODO(), client.ObjectKey{Namespace: r.Plan.Spec.TargetNamespace, Name: fmt.Sprintf("prime-%s", string(pvc.UID))}, &primePVC)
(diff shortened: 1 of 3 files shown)


pkg/controller/plan/kubevirt.go
```diff
@@ -1437,6 +1451,12 @@ func (r *KubeVirt) deleteCorrespondingPrimePVC(pvc *core.PersistentVolumeClaim,
 		if err != nil && !k8serr.IsNotFound(err) {
 			return err
 		}
+		primePVCCopy := primePVC.DeepCopy()
+		primePVC.Finalizers = nil
+		patch := client.MergeFrom(primePVCCopy)
+		if err = r.Destination.Client.Patch(context.TODO(), &primePVC, patch); err != nil {
+			return err
+		}
```
mnecas, line 1459: oh no, we can't do this, dangerous to leave leftover volume on the storage array

pkg/controller/plan/migration.go
```diff
@@ -1380,6 +1390,13 @@ func (r *Migration) execute(vm *plan.VMStatus) (err error) {
 			})
 	}
 
+	// Clean up prime PVCs after migration completes (success or failure)
+	if vm.Phase == api.PhaseCompleted {
+		if err := r.deletePrimePVCs(vm); err != nil {
```
mnecas, line 1395: thought that CDI automatically removes the prime pvcs

PR #4042, opened 2026-01-06
MTV-4209 | Add token-based authentication for Pure FlashArray

borod108:
Added STORAGE_TOKEN environment variable to allow token based authentication for Pure. Token takes precedence over username/password. Includes validation logic, tests, and documentation.

Authentication behavior:
- If STORAGE_TOKEN is provided: uses token-based auth
- If STORAGE_TOKEN is empty: requires both STORAGE_USERNAME and STORAGE_PASSWORD
- Token takes precedence when both token and credentials are provided

Resolves: MTV-4209


rgolangh:
@borod108 LGTM - I want to merge it but CI fails. Can you check?

PR #4043, opened 2026-01-06
MTV-3645 | operator: Configure transfer network with default route for controller/API pods

Hazanel:
Issue:
Controller/API pods had simple string annotation without gateway configuration.

Fix:
Controller/API pods have proper annotation with default-route. The fix ensures controller and API pods can properly configure their transfer network interface with the correct gateway.

Ref: https://issues.redhat.com/browse/MTV-3645

### DIFF
--- a/operator/roles/forkliftcontroller/tasks/main.yml
+++ b/operator/roles/forkliftcontroller/tasks/main.yml
@@ -73,53 +73,99 @@
       state: present
       definition: "{{ lookup('template', 'api/service-services.yml.j2') }}"
 
+  # controller_transfer_network can be provided either as:
+  # 1) 'namespace/name'
+  # 2) full JSON definition (used as-is)
   - name: "Build transfer network annotation with default route for controller and API"
     when: controller_transfer_network is defined
     block:
-      - name: "Validate transfer network format"
+      - name: "Validate that controller_transfer_network is a string"
         assert:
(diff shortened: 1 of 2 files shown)


operator/roles/forkliftcontroller/tasks/main.yml
```diff
@@ -73,6 +73,99 @@
...
+      - name: "Validate structured controller_transfer_network format"
+        when: controller_transfer_network_is_json_string
+        assert:
+          that:
+            - controller_transfer_network is iterable
+            - controller_transfer_network is not string
+            - controller_transfer_network | length > 0
+            - controller_transfer_network[0].name is defined
+            - controller_transfer_network[0].namespace is defined
+          fail_msg: "JSON controller_transfer_network must be a non-empty list with items containing 'name' and 'namespace' fields"
```
mnecas, line 115: i'm not sure if I follow `controller_transfer_network[0].name`  but then we do `transfer_network_name: "{{ controller_transfer_network.split('/')[1] }}"` those are different operations on the struct

Hazanel: The .split('/') only runs when it's a string, and [0].name only runs after JSON parsing converts it to a list

mnecas (approved):

PR #4094, opened 2026-01-07
fix(manifests): Re-generate manifests to include Condition.suggestion

jonner:
The 'suggestion' field was added in 12e24f09a but the manifests were not re-generated. Re-generate them so they are up to date.


yaacov (approved):
Thanks

PR #4128, opened 2026-01-08
MTV-3932 | fix duplicate RunSpecs in tests

amitosw15:
Fix Ginkgo duplicate RunSpecs error in populator tests Ginkgo doesn't support multiple RunSpecs calls in the same package. Both secure_script_test.go and remote_esxcli_test.go had their own test suite entry points, causing CI to fail. Removed the duplicate TestSecureScript function and unused testing import. Both test files now share the existing TestRemoteEsxcli entry point, allowing all 22 specs to run successfully.
https://github.com/kubev2v/forklift/pull/4001
Resolves: MTV-3932


borod108:
LGTM

PR #4144, opened 2026-01-08
fix(build): cgo enable for cli image

solenoci:


yaacov (approved):

solenoci:
We need it to be FIPS compliant. However, builds fail with CGO enabled, so we will request an exception for FIPS. This PR is non functional and I will close it.

PR #4183, opened 2026-01-09
MTV-4321 | Prefer to query volume by id in ResolvePvToLun()

snehal-pure:
Resolves: MTV-4321

Change the logic to first query by volume id and
fallback to query by volume name in case of
failure due to invalid volume id.

### DIFF
--- a/cmd/vsphere-xcopy-volume-populator/internal/pure/flashArray.go
+++ b/cmd/vsphere-xcopy-volume-populator/internal/pure/flashArray.go
@@ -129,6 +129,7 @@ func (f *FlashArrayClonner) Map(
 // UnMap is responsible to unmapping an initiator group from a populator.LUN
 func (f *FlashArrayClonner) UnMap(initatorGroup string, targetLUN populator.LUN, context populator.MappingContext) error {
 	hosts, ok := context["hosts"]
+
 	if ok {
 		hs, ok := hosts.([]string)
 		if ok && len(hs) > 0 {
@@ -154,17 +155,22 @@ func (f *FlashArrayClonner) CurrentMappedGroups(targetLUN populator.LUN, context
 
 // ResolvePVToLUN resolves a PersistentVolume to Pure FlashArray LUN details
 func (f *FlashArrayClonner) ResolvePVToLUN(pv populator.PersistentVolume) (populator.LUN, error) {
-	klog.Infof("Resolving target volume for PV %s", pv.Name)
-
-	volumeName := fmt.Sprintf("%s-%s", f.clusterPrefix, pv.Name)
-	klog.Infof("Target volume name: %s", volumeName)
-
-	v, err := f.restClient.GetVolume(volumeName)
+	pvVolumeHandle := pv.VolumeHandle
+	v, err := f.restClient.GetVolumeById(pvVolumeHandle)
 	if err != nil {
-		return populator.LUN{}, fmt.Errorf("failed to get target volume from Pure FlashArray: %w", err)
+		if strings.Contains(err.Error(), "Volume does not exist.") {
+			klog.Errorf("Volume with handle %s does not exist: %v. Trying with Volume Name", pvVolumeHandle, err)
+			volumeName := fmt.Sprintf("%s-%s", f.clusterPrefix, pv.Name)
+			v, err = f.restClient.GetVolume(volumeName)
+			if err != nil {
+				return populator.LUN{}, fmt.Errorf("failed to get volume by name %s: %w", volumeName, err)
+			}
+		}
 	}
+
 	klog.Infof("volume %+v\n", v)
 	l := populator.LUN{Name: v.Name, SerialNumber: v.Serial, NAA: fmt.Sprintf("naa.%s%s", FlashProviderID, strings.ToLower(v.Serial))}
+
 	return l, nil
 }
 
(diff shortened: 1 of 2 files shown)


rgolangh:
This will not work on our lab, with CSI driver version 25.5.0. 
I see that the `VolumeHandle` is not a uuid, and is useless in the list and filter volumes query. 
Maybe in later versions it will work.

For now, I have a workaround that doesn't require any code changes:

since we are using CLUSTER_PREFIX to build the volume name, if I understand correctly the flashArray POD comes before the it.
So given we use a specific storage class with a POD set in it,  say with value "USER-POD" and the cluster id is "px_xyz" , we can set the cluster prefix to include the pod name.
For example:  "USER-POD::px_xyz" will be the cluster prefix and with that we can locate the volume by name.

rgolangh:
And to the fallback behaviour - this won't work in case a POD is being used, because the k8s volume doesn't have any information about the pod itself.

cmd/vsphere-xcopy-volume-populator/internal/pure/flashArray.go
```diff
@@ -154,17 +155,23 @@ func (f *FlashArrayClonner) CurrentMappedGroups(targetLUN populator.LUN, context
...
-
-	v, err := f.restClient.GetVolume(volumeName)
+	pvVolumeHandle := pv.VolumeHandle
+	v, err := f.restClient.GetVolumeById(pvVolumeHandle)
 	if err != nil {
-		return populator.LUN{}, fmt.Errorf("failed to get target volume from Pure FlashArray: %w", err)
+		if strings.Contains(err.Error(), "Volume does not exist.") {
+			klog.Errorf("Volume with handle %s does not exist: %v. Trying with Volume Name", pvVolumeHandle, err)
+			volumeName := fmt.Sprintf("%s-%s", f.clusterPrefix, pv.Name)
+			v, err = f.restClient.GetVolume(volumeName)
```
mkilavuz, line 164: If we can get the volume here, we want to return at line 173, right? But we return empty at line 169 instead.

snehal-pure:
> This will not work on our lab, with CSI driver version 25.5.0. I see that the `VolumeHandle` is not a uuid, and is useless in the list and filter volumes query. Maybe in later versions it will work.
> 
> For now, I have a workaround that doesn't require any code changes:
> 
> since we are using CLUSTER_PREFIX to build the volume name, if I understand correctly the flashArray POD comes before the it. So given we use a specific storage class with a POD set in it, say with value "USER-POD" and the cluster id is "px_xyz" , we can set the cluster prefix to include the pod name. For example: "USER-POD::px_xyz" will be the cluster prefix and with that we can locate the volume by name.


@rgolangh the intent behind the change was to support migration into User_pod from CSI driver version 25.8.0 while not changing the current behavior with CSI driver < 25.8.

We can change the behavior for CSI driver < 25.8, once we find a way to get the User_pod name. 

We can document this behavior and add your suggestion as a workaround for CSI driver < 25.8.

Also the workaround works only if all the vms are migrated into the same pod i.e a storage map and migration can have only one destination storage class.

cmd/vsphere-xcopy-volume-populator/internal/pure/flashArray.go
snehal-pure: Fixed

cmd/vsphere-xcopy-volume-populator/internal/pure/flashArray.go
```diff
@@ -154,17 +155,22 @@ func (f *FlashArrayClonner) CurrentMappedGroups(targetLUN populator.LUN, context
...
 // ResolvePVToLUN resolves a PersistentVolume to Pure FlashArray LUN details
 func (f *FlashArrayClonner) ResolvePVToLUN(pv populator.PersistentVolume) (populator.LUN, error) {
-	klog.Infof("Resolving target volume for PV %s", pv.Name)
-
-	volumeName := fmt.Sprintf("%s-%s", f.clusterPrefix, pv.Name)
-	klog.Infof("Target volume name: %s", volumeName)
-
-	v, err := f.restClient.GetVolume(volumeName)
+	pvVolumeHandle := pv.VolumeHandle
+	v, err := f.restClient.GetVolumeById(pvVolumeHandle)
```
rgolangh, line 159: This also means that CLUSTER_PREFIX is irrelevant for CSI >= 25.8

rgolangh:
@snehal-pure  thanks, last small but mandatory steps:
 1. please fetch and rebase
 2.  please add 'Resolves: MTV-4321' in the first line of the commit body by doing this 
    `git commit -s --amend -m "Resolves: MTV-4321" -m "$(git log -1 --format=%B)" `

cmd/vsphere-xcopy-volume-populator/internal/pure/flashArray.go
snehal-pure: yes

snehal-pure:
> @snehal-pure thanks, last small but mandatory steps:
> 
> 1. please fetch and rebase
> 2. please add 'Resolves: MTV-4321' in the first line of the commit body by doing this
>    `git commit -s --amend -m "Resolves: MTV-4321" -m "$(git log -1 --format=%B)" `

Thanks @rgolangh. Done.

PR #4286, opened 2026-01-12
MTV-4150 | Fix disk transfer completed when importer pod is pending

gwencasey96:
## Summary

Fixes MTV-4150: Disk Transfer was marked **Completed** while import was not actually done.

The original QE report showed a **blank** DataVolume phase (not `Succeeded`), pending PVCs, and a pending importer pod while nfs.csi provisioning was stuck. That can happen when CDI/storage never advances status cleanly under CSI failure — MTV should not treat disk transfer as done in those cases.

This PR adds defensive checks in Forklift (in addition to whatever CDI reports):

1. **`cdi.Succeeded`** — verify before completing:
   - Main PVC is `Bound`
   - Prime PVC is `Bound` (if present)
   - Importer pod is not `Pending` (if present)
2. **Blank / unknown DV phases** — explicit `default` case keeps the task **Pending** (addresses Matthew's blank-phase observation)
3. **`PendingPopulation` on warm migrations** — no longer auto-completed (warm still needs the importer)
4. **`PendingPopulation` on cold conversion** — still completes disk transfer (conversion pod is first consumer / WFFC)

## Why both MTV and CDI?

CDI/env issues (e.g. broken nfs.csi) can leave DV phase blank or misleading. MTV on `main` still trusts `Succeeded` unconditionally. This change makes Disk Transfer reflect **actual** PVC/importer readiness.

### Unit tests
- [x] `go test ./pkg/controller/plan/ -run 'TestVerifyDataVolumeCompletion|TestPendingPopulation'`

### Manual / lab (recommended before merge)

**Regression — happy path (warm vSphere → OCP)**
1. Run a normal warm migration with working storage.
2. Confirm Disk Transfer eventually shows **Completed** only after DVs reach `Succeeded` and PVCs are bound.
3. Confirm migration completes successfully.

**Bug reproduction — stuck CSI (MTV-4150 scenario)**
1. Use a warm migration plan similar to the Jira repro (vSphere → OCP, nfs-csi or storage that can be made to fail provisioning).
2. During disk transfer, confirm PVCs stay `Pending` and/or importer pod stays `Pending`.
3. Confirm Disk Transfer stays **Pending/Running**, **not Completed**.
4. Check `oc get dv` — phase may be blank, `Pending`, or `Succeeded`; UI should not show completed regardless.

**Blank / unknown DV phase**
1. If possible, capture forklift-controller logs during disk transfer showing DV phase when the step updates.
2. Confirm blank/unknown phases keep the disk task pending with a clear reason.

**Cold conversion (sanity)**
1. Run a cold migration that uses conversion (e.g. vSphere Windows).
2. Confirm disk transfer can still progress when DV is `PendingPopulation` (conversion pod is first consumer).

Resolves: MTV-4150

### DIFF
--- a/pkg/controller/plan/migration.go
+++ b/pkg/controller/plan/migration.go
@@ -1786,6 +1786,134 @@ func (r *Migration) setTaskCompleted(task *plan.Task) {
 	task.MarkCompleted()
 }
 
+// verifyDataVolumeCompletion checks if a DataVolume that reports Succeeded phase
+// is actually ready for completion by verifying:
+// - The main PVC is bound
+// - The prime PVC (if it exists) is bound
+// - The importer pod (if it exists) is not pending
+//
+// Forklift may treat PendingPopulation as Succeeded for cold conversion migrations;
+// this verifies PVC and importer state before marking disk transfer complete.
+//
+// Returns true if the DataVolume can be marked as completed, false otherwise.
+// Also returns a reason string explaining why completion was blocked (if applicable).
+func (r *Migration) verifyDataVolumeCompletion(dv *cdi.DataVolume, vm *plan.VMStatus) (canComplete bool, reason string) {
+	// If no claim name, we can't verify - allow completion
+	if dv.Status.ClaimName == "" {
+		r.Log.V(4).Info("DataVolume has no ClaimName, allowing completion",
+			"vm", vm.String(),
+			"dv", path.Join(dv.Namespace, dv.Name),
+			"dvPhase", dv.Status.Phase)
+		return true, ""
+	}
+
+	// Get the main PVC
+	pvc := &core.PersistentVolumeClaim{}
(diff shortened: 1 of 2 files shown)


mnecas (commented):
This is hard to read, please change the logic so there are not that many nested conditions. [ref](https://youtu.be/CFRhGnuXG-4?si=NiXRM-ouBh_72iKD)

Please take a look at CDI, maybe it does not update the DataVolume status correctly.

mrnold:
Where does the DataVolume get set to Succeeded? From the bug, it looks like the phase is just blank:

```
# oc get dv
NAME                              PHASE   PROGRESS   RESTARTS   AGE
warm-win2019-test-vm-1056-x2zzw           N/A                   23m
```

The patch is kind of hard to read, but I am more wondering if it fixes the wrong symptom. If you have a controller log or something that shows the DataVolume itself is Succeeded can you attach it to the Jira ticket?

PR #4287, opened 2026-01-12
MTV-4178 | Fail to create VMware provider if VDDK is skipped

gwencasey96:
When creating a VMware provider without VDDK, users were seeing an error about 'UseVddkAioOptimization' being non-configurable and can't be deleted. This prevented them from completing provider creation even though VDDK is optional for cold migrations.

This fix adds automatic cleanup of VDDK-specific settings in the provider mutating webhook when VDDK is not set:
- Removes UseVddkAioOptimization setting
- Removes VddkConfig setting

This allows users to create VMware providers without VDDK, with only a warning that VDDK is recommended (which is expected behavior).

### DIFF
--- a/pkg/forklift-api/webhooks/mutating-webhook/mutators/provider-mutator.go
+++ b/pkg/forklift-api/webhooks/mutating-webhook/mutators/provider-mutator.go
@@ -26,6 +26,7 @@ func (mutator *ProviderMutator) Mutate(ar *admissionv1.AdmissionReview) *admissi
 	}
 
 	specChanged := mutator.setSdkEndpointIfNeeded()
+	specChanged = mutator.cleanupVddkSettingsIfNeeded() || specChanged
 	metadataChanged := mutator.ar.Request.Operation == admissionv1.Create && mutator.setFinalizers()
 
 	var patches []util.PatchOperation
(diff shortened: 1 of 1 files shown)


mnecas (commented):
My big question about this whole issue: where is the UseVddkAioOptimization coming from?

mrnold:
I tried this out, and I think it might actually be a UI bug. If you create a Provider with the CLI it doesn't seem to care if you have the VDDK optimization set with no VDDK image. Does this actually fix the bug when you run it?

mnecas:
I belive that this would fix the bug, however I agree that this looks more like UI issue

pkg/forklift-api/webhooks/mutating-webhook/mutators/provider-mutator.go
```diff
@@ -80,6 +81,43 @@ func (mutator *ProviderMutator) setSdkEndpointIfNeeded() bool {
...
+func (mutator *ProviderMutator) cleanupVddkSettingsIfNeeded() bool {
+	var providerChanged bool
+
+	// Only apply to vSphere providers
+	if mutator.provider.Type() != api.VSphere {
+		return false
+	}
+
+	// Check if VDDK is set
+	vddkImage, hasVddk := mutator.provider.Spec.Settings[api.VDDK]
```
yaacov, line 93: note that the vddk image can be set from the operator too:
https://github.com/kubev2v/forklift/blob/a6a70979535115658000acc70923f564dafb8088/pkg/settings/settings.go#L134

so if not finding the vddk image in the provider we may find it in the settings

-------

Notes:
  - I agree we should use mutators sparingly , and not do changes if we don't really need to
  - did the user set the flag ?
  - can a user remove the flag ?
  - looks like a UI bug, e.g. they should not set the flag / remove it

mnecas:
fixed in UI, closing

PR #4289, opened 2026-01-12
MTV-2208 | Change disk sharing status on vCenter, migration plan don'…

gwencasey96:
…t detect the change

Resolves: MTV-2208

When a disk sharing status is changed on vCenter from Multi-Write to nosharing, refreshing the migration plan did not detect the change and clear the warning condition. The plan would continue to show a warning about missing shared disks PVC even though the disk was no longer shared.

This fix ensures that when SharedDisks validation passes (ok=true), any existing shared disk warning conditions for that VM are explicitly cleared. This allows the plan status to update correctly when disk sharing status changes on vCenter.

The fix tracks VMs that pass shared disk validation and explicitly deletes both SharedWarnDisks and SharedDisks conditions for those VMs after validation.

### DIFF
--- a/pkg/controller/plan/adapter/vsphere/validator.go
+++ b/pkg/controller/plan/adapter/vsphere/validator.go
@@ -351,12 +351,12 @@ func (r *Validator) SharedDisks(vmRef ref.Ref, client client.Client) (ok bool, m
 	if err != nil {
 		return false, msg, "", liberr.Wrap(err, "vm", vmRef)
 	}
-	
+
 	// Log disk sharing status from inventory for debugging
 	sharedDisks := []string{}
 	for _, disk := range vm.Disks {
 		if disk.Shared {
-			sharedDisks = append(sharedDisks, fmt.Sprintf("%s (file: %s)", disk.Key, disk.File))
+			sharedDisks = append(sharedDisks, fmt.Sprintf("%d (file: %s)", disk.Key, disk.File))
 		}
 	}
 	if len(sharedDisks) > 0 {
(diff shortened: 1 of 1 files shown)


pkg/controller/plan/validation.go
```diff
@@ -992,6 +994,9 @@ func (r *Reconciler) validateVM(plan *api.Plan) error {
 			}
 			sharedDisks.Type = fmt.Sprintf("%s-%s", sharedDisks.Type, ref.ID)
 			sharedDisksConditions = append(sharedDisksConditions, sharedDisks)
+		} else {
+			// Validation passed - mark this VM as validated so we can clear any old conditions
+			sharedDisksValidatedVMs[ref.ID] = true
```
mnecas: Each reconcile starts with clean list of validations and we add to that list. We don't need t odo any cleanup.

mrnold:
This kind of reminds me of #3369. Do inventory updates include disk sharing status? And if I understand Martin correctly, all these validations are always calculated from the inventory state, so the code to remove conditions won't really have any effect.

pkg/controller/plan/adapter/vsphere/validator.go
```diff
@@ -351,6 +351,25 @@ func (r *Validator) SharedDisks(vmRef ref.Ref, client client.Client) (ok bool, m
...
+	if len(sharedDisks) > 0 {
+		r.Log.V(3).Info("VM has shared disks according to inventory",
+			"vm", vmRef.String(),
+			"sharedDisks", sharedDisks,
+			"totalDisks", len(vm.Disks))
+	} else {
+		r.Log.V(4).Info("VM has no shared disks according to inventory",
+			"vm", vmRef.String(),
+			"totalDisks", len(vm.Disks))
+	}
```
mnecas, line 371: All this does is extra logs

pkg/controller/plan/validation.go
```diff
@@ -974,6 +974,24 @@ func (r *Reconciler) validateVM(plan *api.Plan) error {
...
+					"vm", ref.String(),
+					"vmID", ref.ID)
+			}
+		} else {
+			r.Log.V(3).Info("Shared disk validation failed",
+				"vm", ref.String(),
+				"vmID", ref.ID,
+				"category", category,
+				"message", msg)
+		}
```
mnecas, line 994: All this does is extra logs

PR #4290, opened 2026-01-12
MTV-2419 | VM spec should be preserved if possible

gwencasey96:
MTV-2419 | VM spec should be preserved if possible

When migrating a VM from OCP to OCP, if the source VM had a dataVolumeTemplate, the target VM will now preserve it. Previously, the target VM only referenced the target PVC, but some user workflows may expect the VM's DataVolume to be present.

This fix:
1. Updates the OCP builder to copy DataVolumeTemplates from the source VM spec
2. Updates vmTemplate to preserve DataVolumeTemplates when source is OCP

Resolves: MTV-2419

### DIFF
--- a/pkg/controller/plan/adapter/ocp/builder.go
+++ b/pkg/controller/plan/adapter/ocp/builder.go
@@ -282,9 +282,136 @@ func (r *Builder) VirtualMachine(vmRef ref.Ref, object *cnv.VirtualMachineSpec, 
 	// Preserve DataVolumeTemplates from source VM to maintain user workflows
 	// that may expect the VM's DataVolume to be present
 	object.DataVolumeTemplates = targetVmSpec.DataVolumeTemplates
+
+	// Sanitize DataVolumeTemplates to prevent conflicts with Forklift-created DataVolumes:
+	// 1. Match template names to the PVC names (Forklift creates DataVolumes with Name = pvc.Name)
+	// 2. Clear spec.source to prevent KubeVirt from trying to create new DataVolumes with invalid sources
+	// 3. Ensure namespace is set correctly (will be set by KubeVirt to match VM namespace)
+	// 4. Update volume references to match renamed DataVolumeTemplates
+	r.sanitizeDataVolumeTemplates(vmRef, object.DataVolumeTemplates, object)
+
 	r.mapNetworks(sourceVm, targetVmSpec)
 
 	return nil
+}
+
+// sanitizeDataVolumeTemplates ensures DataVolumeTemplates are compatible with Forklift-created DataVolumes.
+// Forklift creates DataVolumes with Name = pvc.Name, so templates must match these names.
+// We also clear spec.source to prevent KubeVirt from trying to create new DataVolumes with invalid source URLs.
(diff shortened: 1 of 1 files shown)


mrnold:
I am trying to understand what this does. It looks like if you have a VM with a DataVolumeTemplate on one OpenShift cluster, and then migrate it to another cluster, the resulting VM spec will keep a copy of that DataVolumeTemplate. And this DataVolumeTemplate will create a DataVolume that adopts the existing destination PVC that forklift created during the migration?

Do I have it right, or is there more to it than this? It seems simple enough but I am not clear on whether or not there are other side effects. Would the PVC namespace in the DataVolumeTemplate need to be adjusted for the destination migration namespace?

pkg/controller/plan/adapter/ocp/builder.go
```diff
@@ -279,6 +279,9 @@ func (r *Builder) VirtualMachine(vmRef ref.Ref, object *cnv.VirtualMachineSpec,
 
 	targetVmSpec := sourceVm.Spec.DeepCopy()
 	object.Template = targetVmSpec.Template
+	// Preserve DataVolumeTemplates from source VM to maintain user workflows
+	// that may expect the VM's DataVolume to be present
+	object.DataVolumeTemplates = targetVmSpec.DataVolumeTemplates
```
mnecas, line 284: This preserves the spec, but what about the disks themselves referenced in the spec?

gwencasey96: thank you-there were a few things in my initial PR I needed to think through more. Now the DataVolumeTemplate names are mapped to the PVC names that Forklift uses when creating DataVolumes and sanitized. The volume references will also be updated when template names change

mnecas: The disk references are preserved in line above so we don't need anything else
```
	object.Template = targetVmSpec.Template
```
but still question is: Are we migrating the datavolume referenced in the VM? 
If you look at the [export api](https://kubevirt.io/user-guide/storage/export_api/) which is used for the cold migraiton it support datavolumes. Are we using it already in a way that the datavolumes data are copied?

pkg/controller/plan/adapter/ocp/builder.go
```diff
@@ -279,11 +279,141 @@ func (r *Builder) VirtualMachine(vmRef ref.Ref, object *cnv.VirtualMachineSpec,
...
+
+	// Sanitize DataVolumeTemplates to prevent conflicts with Forklift-created DataVolumes:
+	// 1. Match template names to the PVC names (Forklift creates DataVolumes with Name = pvc.Name)
+	// 2. Clear spec.source to prevent KubeVirt from trying to create new DataVolumes with invalid sources
+	// 3. Ensure namespace is set correctly (will be set by KubeVirt to match VM namespace)
+	// 4. Update volume references to match renamed DataVolumeTemplates
+	r.sanitizeDataVolumeTemplates(vmRef, object.DataVolumeTemplates, object)
```
mnecas, line 291: don't think we need this we are trying to persist the VM spec as much as possible to the source, my main question was about the disks themselves and the data on them

mnecas:
closing as stale

PR #4292, opened 2026-01-13
MTV-3635 | Add local HyperV OVF generator tool for  VM metadata extraction.

Hazanel:
- Lightweight Windows tool that runs locally on HyperV host
- Queries VMs via PowerShell and generates OVF metadata files
- Uses KVP Exchange for guest OS detection (no credentials needed)
- Supports multi-disk VMs with VHDX virtual size extraction
- Included in hyperv-provider-server image at /usr/share/forklift/tools/
- Customers extract and run locally to generate OVFs for SMB-shared VHDXs

Resolves: None

Ref: https://issues.redhat.com/browse/MTV-3635

### DIFF
--- a/cmd/ovf-generator/main.go
+++ b/cmd/ovf-generator/main.go
@@ -1,63 +1,208 @@
-// Standalone OVF Generator for HyperV
 package main
 
 import (
+	"context"
 	"encoding/json"
 	"flag"
 	"fmt"
 	"os"
-	"os/exec"
+	"os/signal"
 	"path/filepath"
 	"strings"
+	"syscall"
 
 	hypervovf "github.com/kubev2v/forklift/pkg/hyperv-ovf"
 )
 
+const (
+	toolName    = "ovf-generator"
+	toolVersion = "1.0.0"
+)
+
+type Generator struct {
+	executor  PSExecutor
+	validator *Validator
+	rootPath  string
+}
+
+func NewGenerator(executor PSExecutor, rootPath string) *Generator {
+	return &Generator{
+		executor:  executor,
+		validator: NewValidator(executor, rootPath),
+		rootPath:  rootPath,
+	}
+}
+
 func main() {
-	// Parse flags
-	rootPath := flag.String("path", "", "Optional: only process VMs with disks under this path")
+	rootPath := flag.String("path", "", "Only process VMs with disks under this path")
+	showHelp := flag.Bool("help", false, "Show help information")
+	skipValidation := flag.Bool("skip-validation", false, "Skip pre-flight validation checks (not recommended)")
+
+	// Custom usage function
+	flag.Usage = func() {
+		printHelp()
+	}
+
 	flag.Parse()
 
+	if *showHelp {
+		printHelp()
+		os.Exit(0)
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	sigChan := make(chan os.Signal, 1)
+	signal.Notify(sigChan, os.Interrupt, syscall.SIGTERM)
+
+	go func() {
+		sig := <-sigChan
+		fmt.Printf("\nReceived signal %v, shutting down gracefully...\n", sig)
+		cancel()
+	}()
+
+	executor := &RealPSExecutor{}
+	generator := NewGenerator(executor, *rootPath)
+
+	exitCode := 0
+	if err := generator.RunWithValidation(ctx, !*skipValidation); err != nil {
+		if ctx.Err() != nil {
+			fmt.Fprintln(os.Stderr, "Operation cancelled by user")
+			exitCode = 130
+		} else {
+			fmt.Fprintf(os.Stderr, "Error: %v\n", err)
+			exitCode = 1
+		}
+	}
+
+	os.Exit(exitCode)
+}
+
+func printHelp() {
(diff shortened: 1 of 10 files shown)


pkg/hyperv-ovf/envelope.go
```diff
@@ -0,0 +1,110 @@
+package hypervovf
+
+import "encoding/xml"
+
+const xmlHeader = "<?xml version='1.0' encoding='UTF-8'?>\n"
+
+// Envelope is the root element of an OVF file
+type Envelope struct {
```
mnecas, line 7: Do you think it would be possible to combine it with the existing OVF Envelope so we don't have duplicite code?

pkg/hyperv-ovf/formatter.go
```diff
@@ -0,0 +1,253 @@
...
+	)
+
+	// --- CPU ---
+	cpuCount := int64(1)
+	if val, ok := vmMap["ProcessorCount"].(float64); ok {
+		cpuCount = int64(val)
+	}
+	hardwareItems = append(hardwareItems, Item{
+		InstanceID:      strconv.Itoa(itemInstanceID),
+		ResourceType:    3,
```
mnecas: Please create enum and new type for ResourceType so we have list of used.

pkg/hyperv-ovf/vhdx.go
```diff
@@ -0,0 +1,172 @@
...
+
+var metadataRegionGUID = []byte{
+	0x06, 0xA2, 0x7C, 0x8B, 0x90, 0x47, 0x9A, 0x4B,
+	0xB8, 0xFE, 0x57, 0x5F, 0x05, 0x0F, 0x88, 0x6E,
+}
+
+var virtualDiskSizeGUID = []byte{
+	0x24, 0x42, 0xA5, 0x2F, 0x1B, 0xCD, 0x76, 0x48,
+	0xB2, 0x11, 0x5D, 0xBE, 0xD8, 0x3B, 0xF4, 0xB8,
+}
```
mnecas, line 32: What are these numbers?

pkg/hyperv-ovf/vhdx.go
```diff
@@ -0,0 +1,172 @@
...
+type metadataTableEntry struct {
+	ItemID   [16]byte
+	Offset   uint32
+	Length   uint32
+	Flags    uint32
+	Reserved uint32
+}
+
+// GetVHDXVirtualSize reads the virtual disk size from a VHDX file
+func GetVHDXVirtualSize(path string) (uint64, error) {
```
mnecas, line 63: why do we need to readh vhdx to get the size?
there is not command to get the disk size?

pkg/hyperv-ovf/formatter.go
```diff
@@ -0,0 +1,253 @@
...
+
+	// --- IDE Controller ---
+	ideControllerID := strconv.Itoa(itemInstanceID)
+	hardwareItems = append(hardwareItems, Item{
+		InstanceID:   ideControllerID,
+		ResourceType: 5,
+		Address:      "0",
+		Description:  "IDE Controller",
+		ElementName:  "VirtualIDEController 0",
+	})
```
mnecas: Is it possible to have other controllers like scsi or multiple of them?

cmd/ovf-generator/main.go
```diff
@@ -0,0 +1,229 @@
...
+	}
+	return result, nil
+}
+
+// getGuestOSInfo tries to get OS info from running VM via KVP exchange
+// Returns defaults if VM is off or integration services unavailable
+func getGuestOSInfo(vmName string) map[string]interface{} {
+	cmd := fmt.Sprintf(`
+		$ErrorActionPreference = 'SilentlyContinue'
+		$vm = Get-WmiObject -Namespace root\virtualization\v2 -Class Msvm_ComputerSystem -Filter "ElementName='%s'"
```
mnecas, line 311: just idea: Maybe we could use some package to make it more golang friendly like https://github.com/yusufpapurcu/wmi.

pkg/hyperv-ovf/formatter.go
```diff
@@ -0,0 +1,253 @@
...
+			})
+			itemInstanceID++
+		}
+	}
+
+	// --- Operating System ---
+	vmName := "VM"
+	if n, ok := vmMap["Name"].(string); ok {
+		vmName = n
+	}
```
mnecas: The name sounds like it should be requirement. There should be validations for this.

mnecas (changes requested):
Please add unit test with example input and ouputs of each commands.
Recommend creating interface on top of the command execution something like `CommandBuilder` what we have for virt-v2v, that will allow you to easily create mock inputs and outputs of your expected commands to easily unit test it. Think also how we could test this inside the CI.

Also add "help" which will describe information about the tool. I see that we have one flag, but think also general descibtion would be nice.

Add more validations.
All CLIs should work in a flow that once we start them:
1. We need to watchout as this is interacting directly with users, we should have `context.WithCancel` so all commands can be stopped and we don't end up with hanning process.
2. Perform validations on the input before doing anything
3. Perform validations on outside resources, HyperV permissions, VM requirements, OVF output file. The goal is always to fail early.
4. Do what is needed.

Hazanel:
We moved to different approach, hence ovf generator isn't required anymore - closing.

PR #4299, opened 2026-01-14
MTV-4180 | Fix ovirt disk size validation

mnecas:
Fixes: MTV-4180

Issue: We are blocking the migration if the VM has LUN attached.

Fix: Check if the diks is LUN and validate the the LUN size instead of the disk size.

### DIFF
--- a/pkg/controller/provider/web/ovirt/disk.go
+++ b/pkg/controller/provider/web/ovirt/disk.go
@@ -19,6 +19,13 @@ const (
 	DiskRoot       = DisksRoot + "/:" + DiskParam
 )
 
+type StorageType string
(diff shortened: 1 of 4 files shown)


pkg/controller/provider/web/ovirt/disk.go
```diff
@@ -174,3 +174,8 @@ func (r *Disk) Content(detail int) interface{} {
 
 	return r
 }
+
+// is the disk LUN
+func (r *Disk) IsLun() bool {
+	return r.StorageType == "lun"
```
solenoci: NP: maybe extract to a const?

PR #4304, opened 2026-01-14
MTV-3704 | Update migrator cluster role

mnecas:
Resolves: MTV-3704

Issue: Right now we are not providing list of permissions whcih Forklfit requires to migrate to remote clusters.

Fix: Add the role to operator RBAC so it will get deployed automatically after installing 

How to create new SA and token for testing
```
1. oc create serviceaccount migrator-service-account -n default
2. oc create clusterrolebinding migrator-sa-binding \
  --clusterrole=forklift-migrator-role \
  --serviceaccount=default:migrator-service-account
3. oc apply -f - <<EOF
apiVersion: v1
kind: Secret
metadata:
  name: migrator-sa-token-secret
  namespace: default
  annotations:
    kubernetes.io/service-account.name: "migrator-service-account"
type: kubernetes.io/service-account-token
EOF
4. oc get secret migrator-sa-token-secret -n default -o jsonpath='{.data.token}' | base64 --decode
```

### DIFF
--- a/operator/config/rbac/kustomization.yaml
+++ b/operator/config/rbac/kustomization.yaml
@@ -14,6 +14,7 @@ resources:
 - forklift-controller_service_account.yaml
 - forklift-controller_role.yaml
 - forklift-controller_role_binding.yaml
+- forklift-controller_migrator_role.yml
 
 # forklift-api service account
 - api/service_account.yaml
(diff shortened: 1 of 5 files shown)


operator/config/rbac/kustomization.yaml
```diff
@@ -14,6 +14,7 @@ resources:
 - forklift-controller_service_account.yaml
 - forklift-controller_role.yaml
 - forklift-controller_role_binding.yaml
+- forkliftcontroller_migrator_role.yml
```
solenoci: NP: can we rename the file to `forklift-controller_migrator_role.yml` for consistency?

PR #4308, opened 2026-01-14
MTV-4338 | fix fc support for Infinibox

amitosw15:
### TLDR
The vsphere-xcopy-volume-populator failed to match ESXi Fibre Channel (FC) adapter IDs with Infinibox hosts when using FC storage protocol. 

### Root Cause
The EnsureClonnerIgroup function in the Infinibox implementation was performing a direct string comparison between:
ESXi FC adapter IDs in the format fc.WWNN:WWPN (e.g., fc.20000024ff33016c:21000024ff33016c)
Infinibox port addresses, which store only the WWPN (possibly with different formatting)
This direct comparison always failed because the formats don't match.

### Solution
Updated the comparison logic to:
Detect FC adapters by checking for the fc. prefix
Extract the WWPN from the FC adapter ID using fcutil.ExtractWWPN
Compare the extracted WWPN with the Infinibox port address using fcutil.CompareWWNs, which normalizes both values to handle formatting differences (colons, case, etc.)
For non-FC adapters (iSCSI, NVMe-oF), continue using direct string comparison
This approach aligns with other storage implementations (Powerstore, Pure FlashArray) and correctly matches ESXi FC adapter IDs with Infinibox host port addresses.

Resolves: MTV-4338

### DIFF
--- a/cmd/vsphere-xcopy-volume-populator/internal/infinibox/infinibox.go
+++ b/cmd/vsphere-xcopy-volume-populator/internal/infinibox/infinibox.go
@@ -3,9 +3,11 @@ package infinibox
 import (
 	"fmt"
 	"strconv"
+	"strings"
 
 	"github.com/go-logr/logr"
 	"github.com/infinidat/infinibox-csi-driver/iboxapi"
+	"github.com/kubev2v/forklift/cmd/vsphere-xcopy-volume-populator/internal/fcutil"
 	"github.com/kubev2v/forklift/cmd/vsphere-xcopy-volume-populator/internal/populator"
 	"k8s.io/klog/v2"
 )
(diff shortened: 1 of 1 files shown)


cmd/vsphere-xcopy-volume-populator/internal/infinibox/infinibox.go
```diff
@@ -101,8 +103,27 @@ func (c *InfiniboxClonner) EnsureClonnerIgroup(initiatorGroup string, adapterIds
...
+						matched = true
+					}
+				} else {
+					// For iSCSI or other protocols, do direct comparison
+					if port.Address == adapterId {
+						matched = true
+					}
+				}
+
+				if matched {
```
rgolangh: I think we can just return early with no temp variable. If we got here it should mean we didn't find a host with that adapter id

rgolangh (approved):

PR #4316, opened 2026-01-14
revert ssh uuid in script name

amitosw15:
Reasons for the revert:

we plan another change to the exact area
that will break the SSH key 'command=' section
the change we plan will render this change unnecessar


amitosw15:
duplicated #4293

PR #4320, opened 2026-01-14
MTV-4340 | Add HPE Primera/3PAR secret configuration example

tshefi:
Add documentation for configuring HPE Primera/3PAR storage credentials including the required STORAGE_HOSTNAME format with full URL and WSAPI port. Also reorganized Dell vendor sections so as to fit the order in which they showup on the vendor table.

Resolves: MTV-4340

### DIFF
--- a/cmd/vsphere-xcopy-volume-populator/README.md
+++ b/cmd/vsphere-xcopy-volume-populator/README.md
@@ -165,6 +165,24 @@ See [README](internal/vantara/README.md)
 | --- | --- | --- |
 | ONTAP_SVM | string | the SVM to use in all the client interactions. Can be taken from trident.netapp.io/v1/TridentBackend.config.ontap_config.svm resource field. |
 
+### HPE Primera/3PAR
+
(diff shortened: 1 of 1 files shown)


tshefi:
@rgolangh It's about time I fixed something, p(lease)review away.

cmd/vsphere-xcopy-volume-populator/README.md
```diff
@@ -165,6 +165,24 @@ See [README](internal/vantara/README.md)
...
+```yaml
+apiVersion: v1
+kind: Secret
+metadata:
+  name: hpe-3par-secret
+  namespace: openshift-mtv
+type: Opaque
+stringData:
+  STORAGE_HOSTNAME: "https://192.168.1.1:8080"
+  STORAGE_USERNAME: "admin"
```
rgolangh, line 183: weird combination, https and 8080, but I see this is what works.

rgolangh:
The only change you need is to the commit message
- Run `git commit --amend -s`
-  Change the opening 'docs:' to 'chore(docs):' - I'll do it for the PR title
- the `-s` in the git command will add your git signoff (name and email) to the git commit. This is a must. 

I suggest setting that globally so you wont' see this again
```
git config set --global format.signOff true
```

cmd/vsphere-xcopy-volume-populator/README.md
tshefi: Yeah I was confused by this too, but this works. 

The default port for the HPE 3PAR Web Services API (WSAPI) is 8080 for HTTP **and also used for HTTPS** connections on 3PAR systems, although newer Primera and Alletra 9000/MP models often use 443 (SSL/HTTPS) by default for better security, which can be verified with the showwsapi command in the CLI.

OK I guess I should mention:
Use command cli% showwsapi on the 3PAR  to determine configured HTTP / HTTPS ports.

PR #4346, opened 2026-01-18
MTV-4349 | Add interface for EC2 client calls

yaacov:
Issue:
Using EC2 client without an interface makes the code less testable

Fix:
Add interface between us and the EC2 client lib

Ref: https://issues.redhat.com/browse/MTV-4349


yaacov:
@Hazanel @solenoci @mnecas please review

PR #4347, opened 2026-01-18
MTV-4349 | Add interface for AWS STS client calls

yaacov:
Issue:
Using STS client without an interface makes the code less testable

Fix:
Add interface between us and the STS client lib

Ref: https://issues.redhat.com/browse/MTV-4349

Resolves: MTV-4349


yaacov:
@Hazanel @solenoci @mnecas please review

PR #4349, opened 2026-01-18
Revert "MTV-3932 | Fix race condition in parallel script uploads using UUID-based filenames (#4001)"

rgolangh:
This reverts commit baae02560741017360e0f07da84908df9d6ea3eb.

Reason for the revert is that there's another work which changes the key command and we want to avoid changing it twice for no good reason. Downstream release-2.10 was reverted already and that will help to avoid backport conflicts

Related-to: MTV-3931


rgolangh:
This commit enables #3929

PR #4370, opened 2026-01-20
MTV-4237 | Use per-host igroups to prevent LUN ID reuse conflicts

borod108:
When multiple ESXi hosts share one igroup, ONTAP reuses LUN IDs too quickly, causing "UID of device has changed" errors and failed migrations. Fix: Generate unique igroup per ESXi host (xcopy-hostsystem-host-12345) instead of shared xcopy-esxs.

resolves: https://issues.redhat.com/browse/MTV-4237


cmd/vsphere-xcopy-volume-populator/vsphere-xcopy-volume-populator_test.go
```diff
@@ -149,7 +149,7 @@ var _ = Describe("Populator", func() {
 				storageClient.EXPECT().EnsureClonnerIgroup(gomock.Any(), gomock.Any()).Return(nil, nil)
 				storageClient.EXPECT().ResolvePVToLUN(gomock.Any()).Return(populator.LUN{NAA: "naa.616263"}, nil)
 				storageClient.EXPECT().CurrentMappedGroups(gomock.Any(), gomock.Any()).Return([]string{}, nil)
-				storageClient.EXPECT().Map("xcopy-esxs", gomock.Any(), nil).Return(populator.LUN{NAA: "naa.616263"}, nil)
+				storageClient.EXPECT().Map(gomock.Any(), gomock.Any(), nil).Return(populator.LUN{NAA: "naa.616263"}, nil)
```
rgolangh, line 152: it would have been good to actually test that based on the host id. 
can do in later PR

PR #4407, opened 2026-01-21
MTV-3934 | Handle changes to Provider secrets

jonner:
Previously, when a Provider secret was rotated, the reconciliation request didn't pick up the changed token if the provider was already reconciled. This is because when we get a reconcile request, we skip updating the container if HasReconciled() returns true, but this only checks if the Provider generation is different than the observedGeneration (i.e. whether the provider object itself has changed). If the provider secret has changed, but the provider spec has not, the container will therefore not be updated. This resulted in the Collector's kubernetes client continuing to use the old authentication token that is cached in memory in its rest.Config.

To solve this, we keep track of the observed resource version of the secret and store it in the provider's status when the container is updated. If the resource version in the provider's status doesn't match the secret's current resource version, we update the container even if the provider hasn't changed since the previous reconciliation.

### DIFF
--- a/operator/.downstream_manifests
+++ b/operator/.downstream_manifests
@@ -6003,6 +6003,11 @@ spec:
               phase:
                 description: Current life cycle phase of the provider.
                 type: string
+              secretResourceVersion:
+                description: |-
+                  The ResourceVersion of the secret referenced by this provider.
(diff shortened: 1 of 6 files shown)


mnecas (approved):
lgtm

PR #4439, opened 2026-01-22
MTV-4315 | Fix PVC identifier resolution and improve populator naming

amitosw15:
- Fix ResolvePersistentVolumeClaimIdentifier to handle empty AnnImportBackingFile annotation
- Replace UUID-based populator naming with deterministic hash-based suffix
- Move createdPVC declaration outside if block to fix scope issues
- Add logging for populator creation and PVC existence checks
- Ensure populator creation happens for both new and existing PVCs

Resolves: MTV-4315

### DIFF
--- a/pkg/controller/plan/adapter/vsphere/builder.go
+++ b/pkg/controller/plan/adapter/vsphere/builder.go
@@ -2,7 +2,9 @@ package vsphere
 
 import (
 	"context"
+	"crypto/sha256"
 	"encoding/base64"
+	"encoding/hex"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -17,7 +19,6 @@ import (
 
 	"k8s.io/klog/v2"
 
-	"github.com/google/uuid"
 	api "github.com/kubev2v/forklift/pkg/apis/forklift/v1beta1"
 	"github.com/kubev2v/forklift/pkg/apis/forklift/v1beta1/plan"
 	"github.com/kubev2v/forklift/pkg/apis/forklift/v1beta1/ref"
@@ -1168,6 +1169,9 @@ func (r *Builder) ResolveDataVolumeIdentifier(dv *cdi.DataVolume) string {
 
 // Return a stable identifier for a PersistentDataVolume.
 func (r *Builder) ResolvePersistentVolumeClaimIdentifier(pvc *core.PersistentVolumeClaim) string {
+	if source, ok := pvc.Annotations[planbase.AnnDiskSource]; ok {
+		return baseVolume(source, r.Plan.IsWarm())
+	}
 	return baseVolume(pvc.Annotations[planbase.AnnImportBackingFile], r.Plan.IsWarm())
 }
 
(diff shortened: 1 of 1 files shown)


rgolangh:
@mnecas @yaacov The jira ticket and what @amitosw15  is trying to fix here is due to a probably race cause by 2 reconcilers for the same plan running together. Any chance you started seeing 2 reconcilers per plan? the adapters are not idempotent all the way, and this is a step in trying to make  them so. Is there a reason the design of the adatper to always create new PVC by new names for each reconcile? I assume a restart of a plan would be affected by that, could be for the better.

pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -1429,17 +1433,15 @@ func (r *Builder) PopulatorVolumes(vmRef ref.Ref, annotations map[string]string,
 				if err := r.setColdMigrationDefaultPVCName(&pvc.ObjectMeta, vm, diskIndex, disk); err != nil {
 					r.Log.Info("Failed to set PVC name from template for populator volume, using default name", "error", err)
 				}
-
-				// PVC name is the name of the populator, and we can't use generateName for the populator
-				// if generateName is used, we need to remove it and set a deterministic name
 				if pvc.ObjectMeta.GenerateName != "" {
-					// For GenerateName, use the generateName prefix as the populator name
-					pvc.ObjectMeta.Name = strings.TrimSuffix(pvc.ObjectMeta.GenerateName, "-") + "-" + uuid.New().String()[:8]
+					suffix := r.generatePopulatorSuffix(string(r.Migration.UID), vmRef.ID, disk.Key, disk.File, diskIndex)
```
yaacov, line 1437: @coderabbitai hi, the new suffix can be longer then 8 chars, 
a. what are the risks of hash collision with uuid ?
b. what are the risks of long suffix ?

pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -1168,7 +1169,10 @@ func (r *Builder) ResolveDataVolumeIdentifier(dv *cdi.DataVolume) string {
 
 // Return a stable identifier for a PersistentDataVolume.
 func (r *Builder) ResolvePersistentVolumeClaimIdentifier(pvc *core.PersistentVolumeClaim) string {
-	return baseVolume(pvc.Annotations[planbase.AnnImportBackingFile], r.Plan.IsWarm())
+	if source, ok := pvc.Annotations[planbase.AnnImportBackingFile]; ok {
```
yaacov: @coderabbitai, hi, AnnImportBackingFile is only used for warm migration, AnnDiskSource should be set always ?
why do we use AnnImportBackingFile here ?

yaacov: please retry

pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -2246,12 +2255,49 @@ func (r *Builder) addSSHKeysToSecret(secret *core.Secret) error {
...
+	return nil
+}
+
+func (r *Builder) generatePopulatorSuffix(migrationUID, vmID string, diskKey int32, diskFile string, diskIndex int) string {
+	input := fmt.Sprintf("%s-%s-%d-%s-%d", migrationUID, vmID, diskKey, diskFile, diskIndex)
+	hash := sha256.Sum256([]byte(input))
+	return hex.EncodeToString(hash[:])[:8]
+}
+
+func (r *Builder) ensureVolumePopulator(vp *api.VSphereXcopyVolumePopulator) error {
```
yaacov: nitpick: we use other volume populator in addition to xCopy, can you rename the method to be clear it's only for xCopy populator ?

pkg/controller/plan/adapter/vsphere/builder.go
yaacov: @amitosw15 hi, WDYT ^^ ?

yaacov:
lgtm, 2 nitpicks

pkg/controller/plan/adapter/vsphere/builder.go
amitosw15: sounds good, changing it

amitosw15:
> lgtm, 2 nitpicks

fixed

yaacov:
need rebase

PR #4470, opened 2026-01-25
MTV-3646 | RFE: Allow setting transfer network to the forklift controller pod.

Hazanel:
Issue:
controller_transfer_network: namespace/name fails because "from_json filter" throws an exception on non-JSON input, and "failed_when: false" can't catch Jinja2 filter errors.

Fix:
Check if the string looks like JSON before attempting to parse

Ref: https://issues.redhat.com/browse/MTV-3646

### DIFF
--- a/operator/roles/forkliftcontroller/tasks/main.yml
+++ b/operator/roles/forkliftcontroller/tasks/main.yml
@@ -88,15 +88,21 @@
             - 'namespace/name' format (e.g., 'openshift-mtv/br-eno8403')
             - a JSON string representing a list of transfer networks
 
+      - name: "Detect if controller_transfer_network looks like JSON"
+        set_fact:
+          controller_transfer_network_is_json_string: "{{ controller_transfer_network | trim | regex_search('^[\\[\\{]') is not none }}"
+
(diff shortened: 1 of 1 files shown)


Hazanel:
```
1. oc get forkliftcontroller forklift-controller -n openshift-mtv -o yaml | grep -A20 "status:"

      message: |
            the field 'args' has an invalid value ({'controller_transfer_network_parsed': '{{ controller_transfer_network | from_json }}'}), and could not be converted to dict.. Expecting value: line 1 column 1 (char 0)
      
            The error appears to be in '/app/roles/forkliftcontroller/tasks/main.yml': line 91, column 9
```
2. Manual patching showed pods DO get recreated with transfer network

yaacov:
@coderabbitai hi, this PR adds a check for json format before parsing, and remove the fail and register lines from the parser,
what do you think, what is the impact of removing this lines, should we leave them ? can we make the json check more robust ?

yaacov:
@coderabbitai , thank you, now we have controller_transfer_network_is_json_string , but we don't use it, should we use it to update controller_transfer_network_parsed ? should we use it to raise a warning or error msg, what do you think ?

PR #4476, opened 2026-01-25
[release-2.10] MTV-2930 | Prevent mixed copy methods in migration plans

amitosw15:
**Backport:** https://github.com/kubev2v/forklift/pull/4475

## Description
Add validation before plan execution to ensure all disks use a consistent copy method (either VDDK or XCOPY/Offload).

## Logic
- **Multi-VM Plans:** If one VM cannot use offload while others can, the plan must be split (all VMs in a plan must share the copy method).
- **Multi-Disk VMs:** If a VM has multiple disks and one does not support offload, the system requires VDDK for all disks attached to that VM.

Resolves: MTV-2930


mnecas:
I don't want to add new validations to 2.10

PR #4482, opened 2026-01-27
Add volume tagging support for Pure Storage FlashArray migrations

waliu-ps:
Added Pure Storage FlashArray volume tagging support to the vsphere-xcopy-volume-populator. 
It will tag migrated volumes with metadata to help track and manage volumes in Pure Storage environments. Migrated volumes now get tagged with metadata like:

```
pureuser@mm630c-4> purevol list --tags --namespace=openshift-forklift
Name  Key             Value                                       Namespace   Copyable
xxxx  copyMethod      xcopy                                 openshift-forklift  False
xxxx  migrationType   cold                                  openshift-forklift  False
xxxx  migrationUID    0b3f525a-656b-4c6c-8590-89ad19e2feb8  openshift-forklift  False
xxxx  sourceProvider  vsphere                               openshift-forklift  False
xxxx  sourceVmId      vm-3324                               openshift-forklift  False
xxxx  targetProvider  kubevirt                              openshift-forklift  False
```

source/target providers: Track where VMs are migrating from/to
migrationType: Distinguish between warm and cold migrations
copyMethod: Track whether xcopy or volume copy was used
migrationUID + sourceVmId: Together, these help group all volumes from the same VM

### DIFF
--- a/cmd/vsphere-xcopy-volume-populator/internal/populator/storage.go
+++ b/cmd/vsphere-xcopy-volume-populator/internal/populator/storage.go
@@ -14,6 +14,11 @@ type StorageApi interface {
 	VMDKCapable
 }
 
+// VolumeTaggingSupport is an optional interface for storage vendors that support volume tagging
+type VolumeTaggingSupport interface {
+	TagVolume(volumeName string, volumeHandle string, migrationType string, copyMethod string, migrationUID string, vmID string) error
+}
+
 // StorageResolver resolves a PersistentVolume to LUN details
 // This interface is embedded by VVolCapable, RDMCapable, and VMDKCapable
 type StorageResolver interface {
--- a/cmd/vsphere-xcopy-volume-populator/internal/pure/flashArray.go
+++ b/cmd/vsphere-xcopy-volume-populator/internal/pure/flashArray.go
@@ -20,6 +20,20 @@ import (
 
 const FlashProviderID = "624a9370"
 
+// Volume tagging constants for Forklift migrations
+const (
+	TagNamespace         = "openshift-forklift"
+	TagKeyCopyMethod     = "copyMethod"
+	TagKeySourceProvider = "sourceProvider"
+	TagKeyTargetProvider = "targetProvider"
+	TagKeyMigrationType  = "migrationType"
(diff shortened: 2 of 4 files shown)


rgolangh:
That kind of functionality is not in the responsibility of the offload mechanism. I understand that it is compelling to tag volume 'while we are there' but this opens the door for unrelated chnages that this mechanism isn't designed for.

A more suitable component is the CSI driver. We help by placing a related label on the PVC, and with that that the CSI could tag the pure volumes with more data.

An admission-controller approach would be more suitable here, although we're not keen on using hooks.

waliu-ps:
Hi Roy @rgolangh Thanks for the feedback! I agree that the CSI driver approach is a cleaner architecture.
I dug into how we can wire this up with PVC labels and here's a proposal:

From the Forklift side (PVC labels):
1. In  pkg/controller/plan/adapter/vsphere/builder.go PopulatorVolumes(), we'd add these labels to the PVC:
```
migration.forklift.konveyor.io/source-provider: vsphere
migration.forklift.konveyor.io/target-provider: openshift
migration.forklift.konveyor.io/migration-type: cold|warm
migration.forklift.konveyor.io/datastore-type: VMFS|NFS|VVOL|vsan
migration: <migration-uid> (already exists)
vmID: <vm-id> (already exists)
```
2. In vsphere-xcopy-volume-populator.go (after population completes), we'd add:
`migration.forklift.konveyor.io/copy-method: xcopy|volumecopy`

3. In pkg/lib-volume-populator/populator-machinery/controller.go, we need to copy labels from the original PVC to PVC' (prime):
```
pvcPrime = &corev1.PersistentVolumeClaim{
    ObjectMeta: metav1.ObjectMeta{
        Name:      pvcPrimeName,
        Namespace: populatorNamespace,
        Labels:    pvc.Labels,  // Copy labels from original PVC
    },
    // ... rest unchanged
}
```


From the Pure CSI driver side (tag volume):
1. CreateVolume() has access to PVC' (prime) and can read the Builder's labels (source-provider, target-provider, migration-type, datastore-type, migration-uid and vm-id)
2. ControllerPublishVolume() has access to the original PVC (after rebind) and can read the Populator's label (copy-method)

How does this proposal sound to you?

mnecas:
closing as stale, feel free to reopen as needed

PR #4490, opened 2026-01-28
MTV-4574 | Hitachi Vantara VSP vol-to-vol storage offload feature

tatsumir:
Support volume-to-volume clone storage offload capability with Hitachi Vantara VSP.
This feature is for VM migration with vVol or RDM as a vDisk.
The target StorageClass must be located on the source Datastore.


rgolangh:
@tatsumir please rebase and update the commit message. If that is verified I'll merge it
Please also provide a list of model this would work with

rgolangh:
this backport will fail till this will be merged #4878

PR #4494, opened 2026-01-28
fix(deps): use forked infinibox-csi-driver with pagination fix   

amitosw15:
Update go.mod to use forked version of infinibox-csi-driver that includes pagination support for GetAllHosts function. This is a temporary change until the fix is merged upstream.


mnecas:
closing as stale

PR #4496, opened 2026-01-28
MTV-3613 | fix transient ConnectionTestFailed error

jonner:
There was a previous attempt to fix this bug in commit 2aafff4879c0be106ffe502955fa295432413ba2, but it wasn't a complete fix. This patch series should fully fix the issue. The first commit is the most important:
- **Don't indicate a transient connection error when creating an ova provider**

The other two minor commits fix some small related corner cases or race conditions:
- **Move endStagingConditions to defer func**
- **Add a Readiness probe to the ova-provider-server image**

### DIFF
--- a/pkg/controller/ova/builder.go
+++ b/pkg/controller/ova/builder.go
@@ -191,6 +191,19 @@ func (r *Builder) PodSpec(provider *api.Provider, pvc *core.PersistentVolumeClai
 					},
 				},
 				SecurityContext: r.securityContext(),
+				ReadinessProbe: &core.Probe{
+					ProbeHandler: core.ProbeHandler{
+						HTTPGet: &core.HTTPGetAction{
+							Path: "/test_connection",
+							Port: intstr.FromInt32(8080),
+						},
+					},
+					InitialDelaySeconds: 5,
+					PeriodSeconds:       3,
+					TimeoutSeconds:      2,
+					SuccessThreshold:    1,
+					FailureThreshold:    3,
+				},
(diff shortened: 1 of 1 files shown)


pkg/controller/provider/validation.go
```diff
@@ -420,6 +433,139 @@ func (r *Reconciler) testConnection(provider *api.Provider, secret *core.Secret)
...
+		deploymentList,
+		&client.ListOptions{
+			LabelSelector: labels.SelectorFromSet(labeler.ProviderLabels(provider)),
+			Namespace:     Settings.Namespace,
+		})
+
+	if err != nil {
+		// If we can't list deployments, log but don't fail - allow connection test to proceed
+		log.Error(err, "Failed to list deployments for OVA provider")
+		return true
```
yaacov, line 464: Q: why do we return true ( e.g. service available ) on fail ?

yaacov (approved):
lgtm, nitpick: i have concerns about the "fail to get deployment" case

pkg/controller/provider/validation.go
jonner: My reasoning was because if we get an error from trying to list deployments, it may just be a kubernetes api failure, but the service could still potentially exist. So I thought it might be worth just trying to do the connection test. But I can certainly understand the opposite viewpoint. Would you like me to change this to false?

yaacov: no, no, need to change, just a question, thanks for the answer

PR #4502, opened 2026-01-30
MTV-4516 | Add User-Agent header to Pure Storage REST client

waliu-ps:
Resolves: MTV-4516
set the user-agent of Pure Storage REST client 
```
     root@mm630c-4-ct0:/var/log/purity# less rest_auth.log | grep forklift
2026-02-13 14:29:29,965 REST interface access event v1: USER="waliu" FROM="10.102.29.153" USING="pure/forklift/v2.7.0-1953-g8a9dff2bf" ACTION="user authentication" SUCCESS=True METHOD="password"
```

### DIFF
--- a/cmd/vsphere-xcopy-volume-populator/internal/pure/rest_client.go
+++ b/cmd/vsphere-xcopy-volume-populator/internal/pure/rest_client.go
@@ -13,8 +13,24 @@ import (
 	"strings"
 	"time"
 
+	"github.com/kubev2v/forklift/cmd/vsphere-xcopy-volume-populator/internal/version"
 	"k8s.io/klog/v2"
 )
+
+// userAgent is the User-Agent string sent with all HTTP requests
+var userAgent = "pure/forklift/" + version.Version
+
+// userAgentTransport wraps an http.RoundTripper and adds User-Agent header to all requests
+type userAgentTransport struct {
+	base http.RoundTripper
+}
+
+func (t *userAgentTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	// Clone the request to avoid modifying the original per RoundTripper contract
+	req = req.Clone(req.Context())
+	req.Header.Set("User-Agent", userAgent)
+	return t.base.RoundTrip(req)
+}
 
 // RestClient provides REST API access to Pure FlashArray
 type RestClient struct {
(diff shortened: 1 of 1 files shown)


cmd/vsphere-xcopy-volume-populator/internal/pure/rest_client.go
```diff
@@ -103,16 +118,26 @@ type HostConnectionRequest struct {
 // NewRestClient creates a new REST client for Pure FlashArray
 // If apiToken is provided (non-empty), it will be used directly, skipping username/password authentication
 // If apiToken is empty, username and password will be used to obtain an API token
-func NewRestClient(hostname, username, password, apiToken string, skipSSLVerify bool) (*RestClient, error) {
+// userAgent: User-Agent header to identify the client
+func NewRestClient(hostname, username, password, apiToken, userAgent string, skipSSLVerify bool) (*RestClient, error) {
+	// Create base transport with TLS config
```
rgolangh, line 123: Please remove the 'userAgent' argument - it is anyway a constant and it doesn't look like it needs mutability.

rgolangh:
@waliu-ps are you absolutely sure the version is something you need? I understand this is something you usually see in user agent but bare in mind it is not a must. As far as I understand your use-case, this would be more for logs and audits. 
If this is the case I would prevent this usage altogether and stick with plain MTV or forklift string

waliu-ps:
@mnecas Hi Martin thank you for the comment. I've updated the pull request to using injection.

@rgolangh Hi Roy thanks for the feedback. We want to set the user agent to track the adoption rate of forklift, and it would be better if the forklift version is included. This allows us to correlate problems with specific versions when customers report issues, and proactively notify customers if they're running a version with known bugs that have been fixed in newer releases.

solenoci:
lgtm

rgolangh:
I rather this be more 'welded' into the binary, instead of an ENV variable. 
Since this is the 2nd request for such a thing, the first being from PowerStore team, I'm going to help with a `ldflag` to set a version variable, just like the `VibVersion` is set in build time.
See :
https://github.com/kubev2v/forklift/blob/4d80df334b8fcdb19cfcaf477cb3ca119e11c52f/cmd/vsphere-xcopy-volume-populator/internal/populator/vib.go#L21-L22

And:
https://github.com/kubev2v/forklift/blob/4d80df334b8fcdb19cfcaf477cb3ca119e11c52f/cmd/vsphere-xcopy-volume-populator/Makefile#L37-L43

waliu-ps:
@rgolangh Thank you, Roy! Really appreciate you taking this on. That will help us better track forklift usage.

waliu-ps:
Hi Roy @rgolangh is there a MTV ticket I can use to include in the PR title and commit message?

rgolangh:
> Hi Roy @rgolangh is there a MTV ticket I can use to include in the PR title and commit message?

Here you go https://issues.redhat.com/browse/MTV-4516
Please add this as the first line of the commit message after the subject `Resolves: MTV-4516`
I'll take care of the PR subject

PR #4517, opened 2026-02-02
MTV-2838 | Preserve source PVC/DV access mode and volume mode when storage mapping omits them

gwencasey96:
Resolves: MTV-2838

@https://issues.redhat.com/browse/MTV-2838

OCP live migration: when the plan storage mapping does not set AccessMode or VolumeMode, preserve the source PVC's (or DataVolume's) values instead of leaving them unset. This prevents disk PVCs from flipping RWO to RWX (or vice versa) when the target storage class default differs from the source.

Step 1 of TPM persistent-state PVC fix; CNV team to address reuse of original TPM state PVC (step 2).


mnecas (changes requested):
I don't think this issue is about live migration but cold.
It's possible that the issue is there aswell but know chenli didn't test live.
cc @mansam

mnecas:
as mentioned above ^ the issue is not releated to live migration, closing

PR #4592, opened 2026-02-04
chore(pr-4509-typo): fix error variable and remove debug print

yaacov:
- Fix wrong error variable in migration.go: return jsonErr instead of err
- Remove debug fmt.Printf statement from variables.go

Ref: https://issues.redhat.com/browse/MTV-4454

Resolves: MTV-4454


yaacov:
@mnecas please review

solenoci:
virt-v2v failure is expected as the job only runs on internal konflux cluster not here/on public cluster

mnecas (approved):

PR #4193, opened 2026-01-09
MTV-4155 | add application type in header for powerstore api calls

sid-dell:
Resolves: MTV-4155

PR #4595, opened 2026-02-04
MTV-4517 | standardise version embedding using git-describe

rgolangh:
Resoles: MTV-4517

By using git describe the version of the binary becomes dependant on git
as source-of-truth. When combined with tags this is how the version
string looks like when the HEAD has the tag 'v2.10.4'

    I0208 19:25:09.323603 3637005 vsphere-xcopy-volume-populator.go:313] binary=vsphere-xcopy-volume-populator version=v2.10.4 vib_version=0.3.0 vmkfstools_wrapper_version=0.3.0

If the HEAD is 2 commits above the tag v2.10.4 then version will be:
v2.10.4-2-g{GIT_COMMIT_SHA}

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,5 +1,4 @@
 # Ignore Git files
-.git
 .github/
 
 # Ignore all test-related files
--- a/cmd/vsphere-xcopy-volume-populator/Makefile
+++ b/cmd/vsphere-xcopy-volume-populator/Makefile
@@ -3,6 +3,10 @@
 SHELL = /usr/bin/env bash -o pipefail
 .SHELLFLAGS = -ec
 
+# The build version which relies on release engineering setting tags that starts with 'v' like v2.10.4
+VERSION := $(shell git describe --always --tags --match 'v*' 2>/dev/null || echo "v0.0.0-dev")
+SCRIPT_VERSION := $(shell sh vmkfstools-wrapper/vmkfstools_wrapper.sh --version --output simple)
+
 # Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
 ifeq (,$(shell go env GOBIN))
 GOBIN=$(shell go env GOPATH)/bin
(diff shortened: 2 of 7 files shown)


rgolangh:
This should help with #4502 and #4193

rgolangh:
@solenoci @mnecas PTAL

cmd/vsphere-xcopy-volume-populator/Makefile
```diff
@@ -3,6 +3,10 @@
 SHELL = /usr/bin/env bash -o pipefail
 .SHELLFLAGS = -ec
 
+# The build version which relies on release engineering setting tags that starts with 'v' like v2.10.4
+VERSION := $(shell git describe --always --tags --match 'v*')
```
yaacov: lgtm,

nit pick - in the MCP i use similar approach, and added a fallback for cases the code is compiled on a git-less env [1] , not sure it's needed here.

`VERSION_GIT := $(shell git describe --tags 2>/dev/null || echo "0.0.0-dev")`

[1] https://github.com/yaacov/kubectl-mtv/blob/main/Makefile#L22

rgolangh: done

cmd/vsphere-xcopy-volume-populator/Makefile
```diff
@@ -3,6 +3,10 @@
 SHELL = /usr/bin/env bash -o pipefail
 .SHELLFLAGS = -ec
 
+# The build version which relies on release engineering setting tags that starts with 'v' like v2.10.4
+VERSION := $(shell git describe --always --tags --match 'v*' || echo "0.0.0-dev")
```
yaacov:
```suggestion
VERSION := $(shell git describe --always --tags --match 'v*' 2>/dev/null || echo "v0.0.0-dev")
```

yaacov: this is a nitpick feel ok to ignore

rgolangh: if there's an error then the `|| echo "0.0.0-dev"` will be used so I don't see why  `/dev/null` is needed. Also, the usage of `--always` is to not fail in case there is no tag.

yaacov: what do you think about the v0.0.0 ?

yaacov: Note:
the redirect is for the logging, see this example:

The log has an error line 
```bash
❯ test=$(git describe --always --tags --match 'v*' || echo "hi")
fatal: not a git repository (or any of the parent directories): .git
❯ echo $test
hi
```

The log don't have an error line
```bash
❯ test=$(git describe --always --tags --match 'v*' 2>/dev/null || echo "hi")
❯ echo $test
hi
```

.dockerignore
```diff
@@ -1,5 +1,4 @@
 # Ignore Git files
-.git
```
rgolangh, line 2: Had to remove this so the git info be used inside the container build

cmd/vsphere-xcopy-volume-populator/Makefile
rgolangh: you're obviously try to force the default if git info doesn't exists right?  I don't understand what is the use-case - please tell me if I'm missing some build stages I don't see

I also removed the .git directory from .dockerignore so that the info will be available during build time.

yaacov: this is a super nit pick thing, not important at all, **please ignore if it is waisting your time**

> what is the use-case

one potential use case is build automation on a clean env without git
a. git call will fail for no git command
b. calling --version should work + indicate the build was one dev env 
c. automation logs should be clean

again, this is my use case, feel free to ignore this nit pick

rgolangh:
commit message should say MTV-4517. Will fix on downport

PR #4969, opened 2026-02-25
MTV-2631 | Fix VM migration hanging in Initialize phase.

Hazanel:
Issue 1: 
Stale Glance images cause infinite reconciliation loop.
Image names are deterministic (no migration-specific component), so
images from a previous failed migration persist in Glance as Active
without the forklift_original_volume_id property. The controller
skips volume creation (image exists), but updateImageProperty cannot
set the property (backing volume is gone) and silently returns nil.
The reconcile loop repeats forever.

Fix 1:
- Detect and delete stale images (Active, missing property) in
  ensureVolumeFromSnapshot and ensureImageFromVolumeReady.
- Set the tracking property at image creation time in createImageFromVolume.
- Make updateImageProperty return an error when the volume is not found
  so callers can act on it.
- Defer cleanup goroutines in ensureImagesFromVolumesReady until all images are confirmed ready
  to prevent a race between cleanup and property lookup.

Issue 2:
GetVM/getImage/getVolume resolve the wrong OpenStack resource.
When both ID and Name are set, these functions first fetch by ID
correctly, then fall through to a name-based search that overwrites
the result.
OpenStack APIs use regex name matching, so "mtv-rhel8-sanity"
matches "mtv-rhel8-sanity-2ip-1nic".
PreTransferActions operates on the wrong VM, returns ready=true,
and the real VM transitions to CopyDisks with no PVCs created.

Fix 2: 
Return early after a successful ID-based lookup in getVM,
getImage, and getVolume. 
The name-based search is now only used as a fallback when no ID is provided.
    
 Ref: https://issues.redhat.com/browse/MTV-2631
 Resolves: MTV-2631

### DIFF
--- a/pkg/controller/plan/adapter/openstack/client.go
+++ b/pkg/controller/plan/adapter/openstack/client.go
@@ -230,8 +230,8 @@ func (r *Client) getVM(vmRef ref.Ref) (vm *libclient.VM, err error) {
 			if r.IsNotFound(err) {
 				err = ResourceNotFoundError
 			}
-			return
 		}
+		return
 	}
 	if vmRef.Name != "" {
 		vms := []libclient.VM{}
@@ -264,8 +264,8 @@ func (r *Client) getImage(imageRef ref.Ref) (image *libclient.Image, err error) 
 			if r.IsNotFound(err) {
 				err = ResourceNotFoundError
 			}
-			return
 		}
+		return
 	}
 	if imageRef.Name != "" {
 		images := []libclient.Image{}
@@ -301,8 +301,8 @@ func (r *Client) getVolume(volumeRef ref.Ref) (volume *libclient.Volume, err err
 			if r.IsNotFound(err) {
 				err = ResourceNotFoundError
 			}
-			return
 		}
+		return
 	}
 	if volumeRef.Name != "" {
 		volumes := []libclient.Volume{}
@@ -420,7 +420,9 @@ func (r *Client) updateImageProperty(vm *libclient.VM, image *libclient.Image) (
 		}
 	}
 	if !found {
-		r.Log.Info("cannot find the original volume id within the metadata", "vm", vm.Name, "image", image.Name)
+		err = liberr.New("cannot find the original volume ID for image",
+			"vm", vm.Name, "image", image.Name)
+		return
 	}
 	return
 }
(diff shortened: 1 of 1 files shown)


Hazanel:
Hey @coderabbitai , can you please review and tell me if the PR actually fixes/ prevent the hang issue observed during the migration ?

Hazanel:
Hey @coderabbitai  , can you please review and tell me if the PR actually fixes/ prevent the hang issue observed during the migration ?

Hazanel:
Thanks @coderabbitai for the review.
1. The hang seems to occur at the provider side
2. The issue occurs in the init phase
3. Timeout is indeed configurable 
4. The migration should fail in such case

Hazanel:
Hey @coderabbitai , is adding :
`step.MarkStarted()
step.Phase = api.StepRunning`

(in both phased as I did) necessary ?

Hazanel:
Thanks @coderabbitai, for example :
PhaseWaitForPowerOff is part of the same Initialize step. When vm.FindStep(r.migrator.Step(vm)) is called at line 1206, it returns the same "Initialize" step that was already marked started back in PhaseStarted. So the step already has Started set and Phase == StepRunning by the time we reach PhaseWaitForPowerOff.

what do you think? 
I'm sure it's harmless , I wonder if it's really required

Hazanel:
@coderabbitai, and is it also necessary in PhaseCreateDataVolumes ?

operator/config/crd/bases/forklift.konveyor.io_forkliftcontrollers.yaml
```diff
@@ -303,6 +303,9 @@ spec:
               controller_snapshot_removal_timeout_minuts:
                 x-kubernetes-int-or-string: true
                 description: "Snapshot removal timeout in minutes (default: 120)"
+              controller_initialize_phase_timeout_minutes:
+                x-kubernetes-int-or-string: true
+                description: "Initialize phase timeout in minutes (default: 60)"
```
yaacov: @coderabbitai hi,
I have a concern about the naming of this controller field, specifically, it sounds like a general controller init phase, but it looks like a CDI setting, what am i missing ?

mnecas (changes requested):
According to the bug logs the issue is with the openstack logic, not the migration process itself. 
```
{"level":"debug","ts":"2025-05-30 10:21:04.986","logger":"plan|2nrt7","msg":"Secret updated.","plan":{"name":"osp-test","namespace":"openshift-mtv"},"migration":"openshift-mtv/osp-test-cccl8","secret":"mtv-test/osp-test-39093ebb-b524-4aae-a2ea-69445333d842-8q9cg","vm":" id:39093ebb-b524-4aae-a2ea-69445333d842 name:'mtv-rhel8-sanity-nfs' "}
{"level":"info","ts":"2025-05-30 10:21:05.001","logger":"plan|2nrt7","msg":"the image properties have not been updated yet","plan":{"name":"osp-test","namespace":"openshift-mtv"},"migration":"openshift-mtv/osp-test-cccl8","image":"forklift-migration-vm-39093ebb-b524-4aae-a2ea-69445333d842-volume-edf7520c-ea5e-4370-b586-3a364d2e08d2"}
```
We should take a look at the openstack and fix the problem there.

pkg/controller/plan/migration.go
```diff
@@ -826,6 +828,11 @@ func (r *Migration) execute(vm *plan.VMStatus) (err error) {
 			}
 
 			if !ready {
+				if checkStepTimeout(step, "Timed out waiting for PreTransferActions") {
+					r.Log.Info("Initialize phase timeout exceeded waiting for PreTransferActions.",
+						"vm", vm.String(), "timeout", settings.Settings.Migration.InitializePhaseTimeout)
+					break
```
yaacov: @coderabbitai hi, k8s paradigm is to try to achieve a state, it does not fail, it just retry forever, can we should raise a condition instead of failing, WDYT ?

Hazanel:
Hey @coderabbitai, I've listen to the comments and dig deep into the openstack code issue.
please review and also if you can create a diagram that illustrates  the issue state machine

mnecas (approved):
lgtm

PR #5653, opened 2026-03-29
[release-2.11] MTV-4825 | add Prometheus metrics labels for storage offload

borod108:
Add labeled Prometheus metrics to the vSphere XCOPY volume populator to enable telemetry on storage array usage, copy performance, and source disk sizes.

New metrics:
- storage_array_info gauge (info pattern, constant=1) with vendor/product/model/version labels
- source_disk_bytes gauge with type label (datastore_allocated, provisioned, pvc_request)
- copy_duration_seconds histogram with storage labels
- copy_throughput_mb_per_second histogram with storage labels

Enhanced existing metrics (progress, xcopy_used) with storage_vendor, clone_method, storage_protocol labels.

Each storage backend now implements StorageArrayInfoProvider to report array metadata. The CopyContext carries source disk capacity and datastore-allocated bytes from the vmware client (via layoutEx extents).

Ref: https://redhat.atlassian.net/browse/MTV-4825
Resolves: MTV-4825

### DIFF
--- a/cmd/vsphere-copy-offload-populator/internal/metrics/metrics.go
+++ b/cmd/vsphere-copy-offload-populator/internal/metrics/metrics.go
@@ -1,0 +1,148 @@
+package metrics
+
+import (
+	"strconv"
+	"time"
+
+	"github.com/kubev2v/forklift/cmd/vsphere-copy-offload-populator/internal/populator"
+	"github.com/prometheus/client_golang/prometheus"
+	dto "github.com/prometheus/client_model/go"
+	"k8s.io/klog/v2"
+)
+
+// Label names used for copy metrics.
+const (
+	labelOwnerUID        = "owner_uid"
+	labelResult          = "result"
+	labelStorageVendor   = "storage_vendor"
+	labelCloneMethod     = "clone_method"
+	labelStorageProtocol = "storage_protocol"
+	labelXcopyUsed       = "xcopy_used"
+	labelArrayVendor     = "storage_array_vendor"
+	labelArrayProduct    = "storage_array_product"
+	labelArrayModel      = "storage_array_model"
+	labelArrayVersion    = "storage_array_version"
+	labelDiskSizeType    = "type"
+)
+
+// CopyMetrics holds Prometheus metrics for copy operations and provides labeled recording.
+type CopyMetrics struct {
+	progressCounter       *prometheus.CounterVec
+	xcopyUsedGauge        *prometheus.GaugeVec
+	storageArrayInfoGauge *prometheus.GaugeVec
+	sourceDiskBytesGauge  *prometheus.GaugeVec
+	copyDurationGauge     *prometheus.GaugeVec
+}
+
+func (m *CopyMetrics) RecordProgress(ownerUID string, progress uint64) {
+	metric := dto.Metric{}
+	c := m.progressCounter.WithLabelValues(ownerUID)
+	if err := c.Write(&metric); err != nil {
+		klog.Error(err)
+		return
+	}
+	if float64(progress) > metric.Counter.GetValue() {
+		c.Add(float64(progress) - metric.Counter.GetValue())
+	}
+}
+
+func (m *CopyMetrics) RecordXcopyUsed(ownerUID, storageVendor, cloneMethod string, xcopyUsed int) {
+	m.xcopyUsedGauge.WithLabelValues(ownerUID, storageVendor, cloneMethod).Set(float64(xcopyUsed))
+}
+
+// RecordStorageArrayInfo sets the info metric once with storage array metadata (constant value 1).
+func (m *CopyMetrics) RecordStorageArrayInfo(storageVendor string, arrayInfo populator.StorageArrayInfo) {
+	model := arrayInfo.Model
+	if model == "" {
+		model = "n/a"
+	}
+	ver := arrayInfo.Version
+	if ver == "" {
+		ver = "n/a"
+	}
+	m.storageArrayInfoGauge.WithLabelValues(storageVendor, arrayInfo.Vendor, arrayInfo.Product, model, ver).Set(1)
+}
+
+// RecordSourceDiskBytes records source disk size metrics for each available measurement.
+func (m *CopyMetrics) RecordSourceDiskBytes(ownerUID, storageVendor, cloneMethod, storageProtocol string, sourceDiskCapacityBytes, sourceDatastoreAllocatedBytes int64) {
+	if sourceDatastoreAllocatedBytes > 0 {
+		m.sourceDiskBytesGauge.WithLabelValues(ownerUID, storageVendor, cloneMethod, storageProtocol, "datastore_allocated").Set(float64(sourceDatastoreAllocatedBytes))
+	}
+	if sourceDiskCapacityBytes > 0 {
+		m.sourceDiskBytesGauge.WithLabelValues(ownerUID, storageVendor, cloneMethod, storageProtocol, "provisioned").Set(float64(sourceDiskCapacityBytes))
+	}
+}
+
+// RecordCompletion records copy duration with a result label ("success" or "failure").
+func (m *CopyMetrics) RecordCompletion(result, ownerUID, storageVendor, cloneMethod, storageProtocol string, xcopyUsed int, duration time.Duration) {
+	xcopyUsedStr := strconv.Itoa(xcopyUsed)
+	if xcopyUsedStr != "0" && xcopyUsedStr != "1" {
+		xcopyUsedStr = "0"
+	}
+	labels := []string{ownerUID, result, storageVendor, cloneMethod, xcopyUsedStr, storageProtocol}
+	m.copyDurationGauge.WithLabelValues(labels...).Set(duration.Seconds())
+}
+
+// NewCopyMetrics creates and registers all Prometheus metrics for copy operations.
+func NewCopyMetrics() (*CopyMetrics, error) {
+	// Progress is emitted during the copy and scraped by the controller to update CR status.
+	// Only owner_uid is needed — the controller matches on it to find the right metric line.
+	progressCounter := prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "vsphere_xcopy_volume_populator_progress",
+			Help: "Progress of vsphere XCOPY volume population (percentage 0-100).",
+		},
+		[]string{labelOwnerUID},
+	)
+	// xcopy_used is emitted during the copy before storageProtocol is known.
+	baseLabels := []string{labelOwnerUID, labelStorageVendor, labelCloneMethod}
+	xcopyUsedGauge := prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "vsphere_xcopy_volume_populator_xcopy_used",
+			Help: "Indicates whether XCOPY was used for cloning (0=no, 1=yes). Labels describe what was copied.",
+		},
+		baseLabels,
+	)
+
+	// Info metric (constant value 1) for storage array metadata. Correlate with other metrics via storage_vendor.
+	storageArrayInfoGauge := prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "vsphere_xcopy_volume_populator_storage_array_info",
+			Help: "Storage array metadata (constant 1). Join with other metrics on storage_vendor to get array details.",
+		},
+		[]string{labelStorageVendor, labelArrayVendor, labelArrayProduct, labelArrayModel, labelArrayVersion},
+	)
+
+	// Source disk size in bytes, with type label for each measurement source.
+	sourceDiskBytesGauge := prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "vsphere_xcopy_volume_populator_source_disk_bytes",
+			Help: "Source disk size in bytes. type=datastore_allocated is actual data on datastore (layoutEx extents); provisioned is guest-visible disk size.",
+		},
+		[]string{labelOwnerUID, labelStorageVendor, labelCloneMethod, labelStorageProtocol, labelDiskSizeType},
+	)
+
+	// Duration of copy in seconds. Its presence signals completion (success or failure).
+	completionLabels := []string{labelOwnerUID, labelResult, labelStorageVendor, labelCloneMethod, labelXcopyUsed, labelStorageProtocol}
+	copyDurationGauge := prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "vsphere_xcopy_volume_populator_copy_duration_seconds",
+			Help: "Duration of copy operation in seconds. result label indicates success or failure.",
+		},
+		completionLabels,
+	)
+
+	for _, collector := range []prometheus.Collector{progressCounter, xcopyUsedGauge, storageArrayInfoGauge, sourceDiskBytesGauge, copyDurationGauge} {
+		if err := prometheus.Register(collector); err != nil {
+			return nil, err
+		}
+	}
+
+	return &CopyMetrics{
+		progressCounter:       progressCounter,
+		xcopyUsedGauge:        xcopyUsedGauge,
+		storageArrayInfoGauge: storageArrayInfoGauge,
+		sourceDiskBytesGauge:  sourceDiskBytesGauge,
+		copyDurationGauge:     copyDurationGauge,
+	}, nil
+}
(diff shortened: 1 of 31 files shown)


cmd/vsphere-copy-offload-populator/internal/ontap/ontap.go
```diff
@@ -118,7 +125,23 @@ func NewNetappClonner(hostname, username, password string) (NetappClonner, error
...
+		},
+	}
+
+	// Fetch ONTAP API version
+	ontapVersion, err := client.APIVersion(context.TODO())
+	if err != nil {
+		klog.Warningf("Failed to get ONTAP version for metrics: %v", err)
+	} else {
+		nc.arrayInfo.Version = ontapVersion
+		klog.Infof("ONTAP: version=%s", ontapVersion)
```
rgolangh: Instead of repeating calling those a log statement per vendor  I suggest we make a single call GetStoraeArrayInfo and log it in the main method in vsphere_copy_offload_poppulator.go . With that every array info would be logged.

Also writing "ONTAP: " here is also smelly because in line 132 it alrady sets ONTAP as a vendor

cmd/vsphere-copy-offload-populator/internal/vantara/vantara.go
```diff
@@ -50,11 +59,27 @@ func NewVantaraClonner(hostname, username, password string) (VantaraCloner, erro
...
+	}
+
+	// Fetch model and version from the API
+	storageInfo, err := client.GetStorageInfo()
+	if err != nil {
+		klog.Warningf("Failed to get Vantara storage info for metrics: %v", err)
+	} else {
+		cloner.arrayInfo.Model = storageInfo.Model
+		cloner.arrayInfo.Version = storageInfo.DkcMicroVersion
+		klog.Infof("Vantara: model=%s, version=%s", storageInfo.Model, storageInfo.DkcMicroVersion)
```
rgolangh: Already logged by the calling code. I think this is redundant.

cmd/vsphere-copy-offload-populator/internal/vmware/client.go
```diff
@@ -478,6 +486,179 @@ func (c *VSphereClient) GetVMDiskBacking(ctx context.Context, vmId string, vmdkP
...
+			return virtualDiskProvisionedBytes(disk), nil
+		}
+	}
+	return 0, nil
+}
+
+// GetVirtualDiskDatastoreAllocatedBytes implements Client.
+func (c *VSphereClient) GetVirtualDiskDatastoreAllocatedBytes(ctx context.Context, vmId, vmdkPath string) (int64, error) {
+	finder := find.NewFinder(c.Client.Client, true)
+	datacenters, err := finder.DatacenterList(ctx, "*")
```
rgolangh: this would break with multiple datacenters. we can't rely on 'Default' Data center.

cmd/vsphere-copy-offload-populator/vsphere-copy-offload-populator.go
```diff
@@ -206,34 +218,43 @@ func main() {
 	cloneLog := log.WithName("xcopy").WithName("clone")
 	log.Info("copy-offload started")
 
+	cloneMethod := copyCtx.CloneMethod
+	if cloneMethod == "" {
+		cloneMethod = "unknown"
```
rgolangh: how can we have unknown?

cmd/vsphere-copy-offload-populator/vsphere-copy-offload-populator.go
```diff
@@ -206,34 +218,43 @@ func main() {
...
+	}
+	storageProtocol := copyCtx.StorageProtocol
+	if storageProtocol == "" {
+		storageProtocol = "unknown"
```
rgolangh: what is the meaning of unknown here and how could this be unknown?

cmd/vsphere-copy-offload-populator/internal/pure/rest_client.go
```diff
@@ -671,6 +671,57 @@ func (c *RestClient) FindVolumeBySerial(serial string) (*Volume, error) {
 	return &volumesResponse.Items[0], nil
 }
 
+// ArrayInfo represents the response from GET /api/2.x/arrays
+type ArrayInfo struct {
+	ID      string `json:"id"`
+	Name    string `json:"name"`
+	Model   string `json:"model"`
```
rgolangh: I don't see model in the ArrayInfo .  Perhaps 'os' and its version will be enough for the version model. Its values are 'Purity' or 'Purity/FA' . So it may be enough combinging with 'version.

You can download the openapi from the 'help' menu on the array admin page

rgolangh (commented):
Didn't finish goingn through all of the code. 

Did  you consider also encapsulating the metrics code under internal/metrics?

mnecas:
@solenoci could you ptal?

cmd/vsphere-copy-offload-populator/internal/vmware/client.go
borod108: right! fixed

cmd/vsphere-copy-offload-populator/vsphere-copy-offload-populator.go
borod108: fixed

cmd/vsphere-copy-offload-populator/vsphere-copy-offload-populator.go
borod108: it is fc/iscsi and in the future maybe nvm but if we do rdm or vvol we do not know how it is connected, but unknown is not good - now it is n/a

cmd/vsphere-copy-offload-populator/internal/pure/rest_client.go
borod108: I am getting FA-X50R4 for example now, is that relevant info?

rgolangh: yes I guess so

cmd/vsphere-copy-offload-populator/internal/populator/factory.go
```diff
@@ -53,62 +58,91 @@ func NewPopulator(
...
 	}
 
 	log.Info("using VMDK/Xcopy populator")
-	return createVMDKPopulator(storageApi, vsphereClient, sshConfig)
+	return createVMDKPopulator(storageApi, vsphereClient, sshConfig, sourceDiskCap, sourceDatastoreAlloc)
+}
+
+// fetchSourceDiskSizes retrieves provisioned capacity and datastore-allocated bytes
+// from vSphere for metric labels. Best-effort: returns zeros on failure.
+func fetchSourceDiskSizes(ctx context.Context, log klog.Logger, vsphereClient vmware.Client, vmId, vmdkPath string) (sourceDiskCap, sourceDatastoreAlloc int64) {
```
rgolangh: please use explicit return rather than named returns

cmd/vsphere-copy-offload-populator/internal/metrics/metrics.go
```diff
@@ -0,0 +1,216 @@
...
+	dto "github.com/prometheus/client_model/go"
+	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/klog/v2"
+)
+
+// Label names used for copy metrics.
+const (
+	labelOwnerUID        = "owner_uid"
+	labelStorageVendor   = "storage_vendor"
+	labelVolumeSizeGB    = "volume_size_gb"
```
rgolangh: We can end up with lots of time series if we have a lot of variation in disk sizes. Lets try to have that as a range, say 1-10, 10-100, 100-500, 500+ ranges to reduce that series number.

cmd/vsphere-copy-offload-populator/vsphere-copy-offload-populator.go
```diff
@@ -192,8 +192,18 @@ func main() {
...
+	var arrayInfo populator.StorageArrayInfo
+	if infoProvider, ok := storageApi.(populator.StorageArrayInfoProvider); ok {
+		arrayInfo = infoProvider.GetStorageArrayInfo()
+		klog.Infof("Storage array info: vendor=%s, product=%s, model=%s, version=%s",
+			arrayInfo.Vendor, arrayInfo.Product, arrayInfo.Model, arrayInfo.Version)
+	}
+
 	klog.InfoS("populator", "stage", "setupTracing")
-	progressCounter, xcopyUsedGauge, err := setupTracingMetrics()
+	copyCtx := p.GetCopyContext()
```
rgolangh, line 204: It says in the comment in factory.go that storageProtocol is set during the Populate, and here we are running before the populate, so for the same populateion we would have multiple metrics, one with storageProtocol label "n/a" and one with "iscsi" or "fc" .   that would mess up dashbords and would require more prom query effort

cmd/vsphere-copy-offload-populator/internal/vmware/client.go
```diff
@@ -478,6 +486,165 @@ func (c *VSphereClient) GetVMDiskBacking(ctx context.Context, vmId string, vmdkP
 	}, nil
 }
 
+// getVMWithConfig looks up a VM across all datacenters and fetches its hardware device config.
+// Falls back to a direct moref lookup if the VM is not found by name/path.
+func (c *VSphereClient) getVMWithConfig(ctx context.Context, vmId string) (*object.VirtualMachine, *mo.VirtualMachine, error) {
```
rgolangh, line 512: I think this is mostly of what GetVMDiskBacking is doing no? try to see if you can refactor for better resuse

rgolangh: I think we can make a single roundtrip and retusn the vm and disk object with all the needed info, instead of 3 calls in factory.go

cmd/vsphere-copy-offload-populator/internal/populator/remote_esxcli.go
```diff
@@ -138,6 +145,13 @@ func (p *RemoteEsxcliPopulator) Populate(vmId string, sourceVMDKFile string, pv
 	setupLog.Info("Datastore active adapters", "datastore",vmDisk.Datastore, "activeAdapters", dsActiveAdapters)
 	for _, a := range dsActiveAdapters {
 		initiators = append(initiators, a.Id)
+		if p.copyCtx.StorageProtocol == "" {
+			if strings.HasPrefix(a.Id, "iqn.") || strings.HasPrefix(a.Id, "eui.") || strings.HasPrefix(a.Id, "nqn.") {
```
rgolangh: nqn is NVMe over fabric, not iscsi

cmd/vsphere-copy-offload-populator/internal/metrics/metrics.go
borod108: didn't we want this to calculate throughput in the future?

cmd/vsphere-copy-offload-populator/internal/metrics/metrics.go
```diff
@@ -0,0 +1,216 @@
...
+	q, err := resource.ParseQuantity(pvcSize)
+	if err != nil {
+		return 0, "unknown"
+	}
+	b := q.Value()
+	return b, BytesToVolumeGiLabel(b)
+}
+
+// NewCopyMetrics creates and registers all Prometheus metrics for copy operations.
+func NewCopyMetrics() (*CopyMetrics, error) {
```
solenoci, line 87: MTV exposed metrics start with `mtv_` to uniquely differentiate them from any other metrics on the cluster. Would it make sense to do the same for storage offload metrics? e.g. `mtv_vsphere_xcopy_volume_populator_progress`...

cmd/vsphere-copy-offload-populator/internal/metrics/metrics.go
solenoci: I would advise to have a separate metric instead of a label that would be used for throughput calculation.

solenoci (commented):
Are you planning to add service monitor so the metrics are being scraped in this PR?

cmd/vsphere-copy-offload-populator/internal/metrics/metrics.go
borod108: Good point. I added mtv_ to the metrics now exposed by the controller, since the ones from the pod are not exposed to Prometheus I left them as is.

cmd/vsphere-copy-offload-populator/internal/ontap/ontap.go
borod108: fixed

cmd/vsphere-copy-offload-populator/internal/vantara/vantara.go
borod108: fixed

cmd/vsphere-copy-offload-populator/internal/populator/factory.go
borod108: done

cmd/vsphere-copy-offload-populator/internal/metrics/metrics.go
borod108: done, and as discussed we do collect the exact size but in telemetry we will share aggregate only

cmd/vsphere-copy-offload-populator/vsphere-copy-offload-populator.go
borod108: fixed. we only expose the metrics now when it is fully populated with the fields

cmd/vsphere-copy-offload-populator/internal/vmware/client.go
borod108: refactored

borod108:
@solenoci kindly take another look when you can :)

cmd/vsphere-copy-offload-populator/internal/powerflex/powerflex.go
```diff
@@ -221,7 +230,14 @@ func NewPowerflexClonner(hostname, username, password string, sslSkipVerify bool
...
 	}
 
-	klog.Infof("successfuly logged in to ScaleIO Gateway at %s version %s", client.GetConfigConnect().Endpoint, client.GetConfigConnect().Version)
+	klog.Infof("successfuly logged in to ScaleIO Gateway at %s", client.GetConfigConnect().Endpoint)
 
-	return PowerflexClonner{Client: client, systemId: systemId}, nil
+	return PowerflexClonner{
+		Client:   client,
+		systemId: systemId,
+		arrayInfo: populator.StorageArrayInfo{
```
rgolangh, line 238: Please create a followup PR to handle that missing storage info (either us or Dell can handle this)

cmd/vsphere-copy-offload-populator/vsphere-copy-offload-populator.go
```diff
@@ -376,12 +428,24 @@ func validateStorageAuthentication(token, username, password string) error {
 	return nil
 }
 
-func startMetricsServer(certFile, keyFile string) {
+func startMetricsServer(certFile, keyFile string, completionReady *atomic.Bool, scrapedCh chan<- struct{}) {
+	mux := http.NewServeMux()
+	mux.HandleFunc("/metrics", func(w http.ResponseWriter, r *http.Request) {
+		promhttp.Handler().ServeHTTP(w, r)
+		// After serving, if completion is ready, notify that we've been scraped.
+		if completionReady.Load() {
```
rgolangh, line 436: major issue - We do have to timeout here to not wait forever in case nothing is incoming

cmd/vsphere-copy-offload-populator/internal/metrics/metrics.go
```diff
@@ -0,0 +1,148 @@
...
+
+// Label names used for copy metrics.
+const (
+	labelOwnerUID        = "owner_uid"
+	labelResult          = "result"
+	labelStorageVendor   = "storage_vendor"
+	labelCloneMethod     = "clone_method"
+	labelStorageProtocol = "storage_protocol"
+	labelXcopyUsed       = "xcopy_used"
+	labelArrayVendor     = "storage_array_vendor"
```
rgolangh, line 21: minor  - not sure if I wrote it before, i think the 'storage_' prefix for the 4 labels here is redundant since the metric name already has storage_array_info in its name, don't you think?
also it would be equal to the constant name: labelArrayVendor = "array_vendor"

cmd/vsphere-copy-offload-populator/internal/metrics/metrics.go
rgolangh: @solenoci  `forklift_` is more suited than mtv_ - project name rather than product name. We should be more aware of that throughout the project.

rgolangh: Anyhow @borod108 I see you didn't insert those prefixes yet, as long as we don't create new dashboards we can defer that change to later.

cmd/vsphere-copy-offload-populator/vsphere-copy-offload-populator.go
rgolangh: ok, scratch that, thanks for pointing into me on line 279

PR #5655, opened 2026-03-30
MTV-4812 | Add per-vNIC static IP preservation control

Hazanel:
Add VM-level `PreserveStaticIPs` (`*bool`) override and `ExcludeNICsFromStaticIPs` (`[]string`) MAC-based exclusion list to `plan.VM`, so users can control which NICs get static IP preservation during migration.

When `preserveStaticIPs` is enabled, the effective value is resolved per-VM (`nil` inherits the plan default). NICs listed in `ExcludeNICsFromStaticIPs` are skipped during static IP mapping and validation.

Ref: https://issues.redhat.com/browse/MTV-4812
Resolves: MTV-4812

---

**Setup:**

```bash
# Update CRDs (new per-VM fields)
oc apply -f operator/config/crd/bases/forklift.konveyor.io_plans.yaml
oc apply -f operator/config/crd/bases/forklift.konveyor.io_migrations.yaml

# Update the controller image
oc edit forkliftcontroller forklift-controller -n openshift-mtv
# set controller_image_fqin to your built image
```

**Environment:** MTV v2.11.0, vSphere provider, Windows Server 2019 VMs with multiple NICs and static IPs reported by VMware Tools.

**Test 1 — Per-VM opt-in** (plan default OFF, VM override ON):

```yaml
spec:
  preserveStaticIPs: false
  vms:
    - id: vm-188
      name: mtv-feature-win2019-staticips
      preserveStaticIPs: true         # per-VM override
```

Result: Succeeded. Static IP `10.232.18.19/24` preserved. Migration status shows `preserveStaticIPs: true` on the VM entry.

**Test 2 — MAC exclusion** (plan default ON, exclude one NIC):

```yaml
spec:
  preserveStaticIPs: true
  vms:
    - id: vm-151
      name: mtv-func-WIN2019-test
      excludeNICsFromStaticIPs:
        - "00:50:56:be:cd:4d"         # skip Mgmt Network NIC
```

Result: Succeeded. NIC0 static IPs preserved, NIC1 excluded. Migration status shows `excludeNICsFromStaticIPs: ["00:50:56:be:cd:4d"]`.


Hazanel:
Hey @coderabbitai, please review, if you think perhaps there is a better way, or you think I missed something in my approach and if any critical BUGS are found

Hazanel:
Thanks @coderabbitai for the thorough review. Addressing each point:

1. Pod Network warning false-positive when all NICs are excluded

Acknowledged. validateNetworkMap doesn't have inventory access, so it can't determine the actual NIC count to compare against the exclusion list. The worst case is a harmless extra warning for a very unlikely user configuration (enabling preserve + excluding every single NIC).

2. vsphere StaticIPs returns false for empty GuestNetworks

Intentional. The old behavior  was inconsistent with the HyperV adapter and masked a real problem:
if preserveStaticIPs is on but VMware Tools isn't reporting guest networks, the user should know.

3. ExcludeNICsFromStaticIPs silently ignored for oVirt / OVF-base / OpenStack

By design — those providers don't implement static IP preservation at all. The field lives on the shared plan.VM struct. Today we don't set a warning if a user preservestaticip's for those providers on a plan level, aren't we?

4. UdnStaticIPs double-checks StaticIPs

StaticIPs is cheap and the redundancy acts as a safety net. No change needed.

5. GetVM returns pointer to slice element

Intentional improvement. All current call sites are read-only. The old FindVM took a copy and returned a pointer to the copy, which was  misleading -  mutations would silently not propagate. The new signature is more honest.

mnecas:
i have proposed alternative at https://redhat.atlassian.net/browse/MTV-1528?focusedCommentId=16678497

mnecas:
closing this PR as there will be another direction

PR #5690, opened 2026-04-02
chore: Update copy offload readme

mnecas:
The `STORAGE_HOSTNAME` in the docs has `https://` prefix

### DIFF
--- a/cmd/vsphere-copy-offload-populator/README.md
+++ b/cmd/vsphere-copy-offload-populator/README.md
@@ -191,7 +191,7 @@ metadata:
   namespace: openshift-mtv
 type: Opaque
 stringData:
-  STORAGE_HOSTNAME: "https://192.168.1.1:8080"
+  STORAGE_HOSTNAME: "192.168.1.1:8080"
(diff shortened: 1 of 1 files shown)


cmd/vsphere-copy-offload-populator/README.md
```diff
@@ -191,7 +191,7 @@ metadata:
   namespace: openshift-mtv
 type: Opaque
 stringData:
-  STORAGE_HOSTNAME: "https://192.168.1.1:8080"
+  STORAGE_HOSTNAME: "192.168.1.1:8080"
```
rgolangh, line 194: that is known to work with 3par/primera. did you had a different experience?

mnecas: ah, in that case we need to add new section to ONTAPP, when i was testing it with Nimo we had issue with this

rgolangh: so there is a snippet section for netapp, or we need to add one

PR #5731, opened 2026-04-07
MTV-4436 | Configure snapshot consolidation for VMware warm migration

mrnold:
Issue: [MTV-4436](https://redhat.atlassian.net/browse/MTV-4436) Warm migration with heavy write stress can cause extremely long cutover delays due to snapshot consolidations, because forklift currently allows VMware to perform a consolidation when calling RemoveSnapshot.

Fix: Move the final snapshot removal step (including consolidation) to the end of the warm migration itinerary, after VM creation. Add a wait_for_final_snapshot_consolidation flag to the forklift controller to avoid waiting for the removal/consolidation to complete before moving on with plan completion.

~~Fix: Add an "allowSnapshotConsolidation" plan setting to allow different consolidation behaviors:~~
~~- Always: consolidate snapshots with every removal~~
~~- Never: never consolidate snapshots~~
~~- AfterFinalSnapshot: only consolidate when removing the very last snapshot, and specifically do not wait for that removal to complete. With this flag, forklift assumes it has completely transferred the disk and does not need to wait for any further operations from the source provider, which may speed up some test scenarios.~~
~~- BeforePenultimateSnapshot: consolidate on every snapshot removal except for the final two, which are taken after the cutover is triggered. This avoids waiting for consolidation after that point to reduce total cutover time.~~

~~These are mostly debug levers to help with performance testing. In my own testing I have found it is possible to get data corruption with enough abuse. I have also observed that allowing snapshot consolidation on snapshot removal does not necessarily consolidate every VMDK, so the "consolidation needed" condition may still stick around without manual intervention.~~

Resolves: MTV-4436

[MTV-4436]: https://redhat.atlassian.net/browse/MTV-4436?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/validation/policies/io/konveyor/forklift/vmware/consolidation_needed.rego
+++ b/validation/policies/io/konveyor/forklift/vmware/consolidation_needed.rego
@@ -1,0 +1,17 @@
+package io.konveyor.forklift.vmware
+
+import rego.v1
+
+consolidation_needed if {
+	input.consolidationNeeded == true
+}
+
+concerns contains flag if {
+	consolidation_needed
+	flag := {
+		"id": "vmware.consolidation_needed",
+		"category": "Information",
+		"label": "Snapshot consolidation required",
+		"assessment": "VM has snapshots that require consolidation. This may cause delays between precopies."
+	}
+}
--- a/validation/policies/io/konveyor/forklift/vmware/consolidation_needed.rego
+++ b/validation/policies/io/konveyor/forklift/vmware/consolidation_needed.rego
@@ -10,7 +10,7 @@ concerns contains flag if {
 	consolidation_needed
 	flag := {
 		"id": "vmware.consolidation_needed",
-		"category": "Information",
+		"category": "Warning",
 		"label": "Snapshot consolidation required",
 		"assessment": "VM has snapshots that require consolidation. This may cause delays between precopies."
 	}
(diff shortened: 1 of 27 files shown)


dgur:
What would be the default for - "allowSnapshotConsolidation" plan setting ?

mnecas:
> What would be the default for - "allowSnapshotConsolidation" plan setting ?

@dgur  `"Always" (default): always consolidate snapshots after removal`

mnecas:
I want to hear more feedback from QE about this PR before merge.
If possible i don't want to maintain additional code if not needed especially if it could be causing problems.

validation/policies/io/konveyor/forklift/vmware/consolidation_needed.rego
```diff
@@ -0,0 +1,17 @@
...
+
+consolidation_needed if {
+	input.consolidationNeeded == true
+}
+
+concerns contains flag if {
+	consolidation_needed
+	flag := {
+		"id": "vmware.consolidation_needed",
+		"category": "Information",
```
mnecas: Lets move this to warn to keep it consistant with the paln validation

mnecas (approved):

PR #5775, opened 2026-04-10
MTV-4913 | Set disk type to 'raw' for in-place domain

mnecas:
Resolves: MTV-4913


mnecas:
ups accidentally closed it

mrnold (approved):

PR #5824, opened 2026-04-13
MTV-5012 | Add enhancement doc for bridging VM concerns to plan readiness

yaacov:
Add enhancement proposal to bridge inventory VM concerns (evaluated by Rego policies) into plan validation so that Critical concerns block plan readiness. Currently these two systems are disconnected: a VM can have Critical inventory concerns while the plan is Ready and executes, only to fail at runtime.

The enhancement covers two parts:
- Bridge Critical inventory concerns into plan readiness checks
- Reclassify non-blocking concerns (RDM, oVirt/OpenStack VM status) from Critical to Warning

Ref: https://issues.redhat.com/browse/MTV-5012
Resolves: MTV-5012

### DIFF
--- a/docs/enhancements/bridge-vm-concerns-to-plan-readiness.md
+++ b/docs/enhancements/bridge-vm-concerns-to-plan-readiness.md
@@ -1,0 +1,254 @@
+---
+title: bridge-vm-concerns-to-plan-readiness
+authors:
+  - "@yaacov"
+reviewers:
+  - TBD
+approvers:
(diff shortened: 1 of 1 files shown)


yaacov:
This is a design doc,
Offline input from @mnecas 
a. don't remove hardcoded go validations that overlap .rego validations, because the hardcoded validations may have more information about the issue.
b. make sure all providers have the needed .rego files, and new vm-concerns-validation test

PR #5828, opened 2026-04-13
MTV-5016 | Hide HyperV params from ForkliftController create form

Hazanel:
HyperV fields were missing specDescriptors with the hidden annotation in the CSV, causing OLM to render them in the create form. Added hidden descriptors for all HyperV fields in all three CSV files.

Ref: https://issues.redhat.com/browse/MTV-5016
Resolves: MTV-5016

### DIFF
--- a/operator/.downstream_manifests
+++ b/operator/.downstream_manifests
@@ -8321,6 +8321,41 @@ spec:
         path: ova_container_requests_memory
         x-descriptors:
         - urn:alm:descriptor:com.tectonic.ui:hidden
+      - description: HyperV CPU limit (default 1000m)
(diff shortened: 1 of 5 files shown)


Hazanel:
Hey @coderabbitai, did I miss anything or now the hyperv params would be all hidden?

mnecas (approved):

mnecas:
lgtm, but the konflux should not be failing 
operator is quite stable @Hazanel ptal and the CI

Hazanel:
@mnecas  - it seems those 2 jobs fail broadly , not just on my PR

PR #5855, opened 2026-04-14
chore(hyperv): improve SMB CSI validation message

Hazanel:
Improve the validation error message when the SMB CSI driver is not
found. The new message tells users exactly what to check: that the
CIFS/SMB CSI Driver Operator is installed and that the ClusterCSIDriver
CR for smb.csi.k8s.io exists. This is a read-only, validation-only
change with no new RBAC requirements.


mnecas (changes requested):
we should not touch other operators, this should be documented from our side we should start screaming about it once the user tries to create the provider with SMB

mnecas (approved):
lgtm

PR #5864, opened 2026-04-14
MTV-5033 | Add AAP job-templates proxy to forklift-api services

gwencasey96:
Expose GET /aap/job-templates?url=... on the services HTTPS server (port 8444), proxying to the AAP Controller API job_templates list with X-AAP-Token sent as Bearer authorization, so the console can list templates without browser CORS blocking direct calls to AAP. Follows the existing tls-certificate service pattern.

Ref: https://redhat.atlassian.net/browse/MTV-5033
Resolves: MTV-5033
Made-with: Cursor

### DIFF
--- a/pkg/forklift-api/services/aap-job-templates.go
+++ b/pkg/forklift-api/services/aap-job-templates.go
@@ -131,7 +131,13 @@ func isBlockedAAPHostname(host string) bool {
 }
 
 func isBlockedAAPIP(ip net.IP) bool {
-	if ip.IsLoopback() || ip.IsPrivate() || ip.IsUnspecified() || ip.IsLinkLocalUnicast() || ip.IsLinkLocalMulticast() {
+	// Block loopback, unspecified, and multicast. Do not block RFC1918: on-prem AAP
+	// commonly uses private IPs reachable only from the cluster.
+	if ip.IsLoopback() || ip.IsUnspecified() || ip.IsLinkLocalMulticast() {
+		return true
+	}
+	// Link-local unicast (e.g. fe80::/10); allow RFC1918 private networks for customer AAP.
+	if ip.IsLinkLocalUnicast() {
 		return true
 	}
 	// Cloud instance metadata
--- a/pkg/forklift-api/services/aap-job-templates.go
+++ b/pkg/forklift-api/services/aap-job-templates.go
@@ -1,0 +1,88 @@
+package services
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"strings"
+	"time"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	// AAPJobTemplatesPath is the forklift-services path that proxies to the AAP Controller API job_templates list.
+	AAPJobTemplatesPath = "/aap/job-templates"
+	headerAAPToken      = "X-AAP-Token"
+	aapProxyTimeout     = 60 * time.Second
+)
+
+func serveAAPJobTemplates(resp http.ResponseWriter, req *http.Request, _ client.Client) {
+	if req.Method != http.MethodGet {
+		http.Error(resp, "method not allowed", http.StatusMethodNotAllowed)
+		return
+	}
+
+	rawBase := req.URL.Query().Get("url")
+	if rawBase == "" {
+		http.Error(resp, "required query parameter: url", http.StatusBadRequest)
+		return
+	}
+
+	aapBase, err := url.Parse(rawBase)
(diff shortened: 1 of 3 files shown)


pkg/forklift-api/services/aap-job-templates.go
```diff
@@ -0,0 +1,166 @@
...
+	h := strings.ToLower(strings.TrimSpace(host))
+	switch h {
+	case "localhost":
+		return true
+	}
+	if strings.HasSuffix(h, ".svc.cluster.local") || strings.HasSuffix(h, ".svc") {
+		return true
+	}
+	switch h {
+	case "metadata.google.internal", "kubernetes.default.svc.cluster.local":
```
mnecas, line 127: @coderabbitai what are these?

pkg/forklift-api/services/aap-job-templates.go
```diff
@@ -0,0 +1,166 @@
...
+	// commonly uses private IPs reachable only from the cluster.
+	if ip.IsLoopback() || ip.IsUnspecified() || ip.IsLinkLocalMulticast() {
+		return true
+	}
+	// Link-local unicast (e.g. fe80::/10); allow RFC1918 private networks for customer AAP.
+	if ip.IsLinkLocalUnicast() {
+		return true
+	}
+	// Cloud instance metadata
+	if ip.Equal(net.IPv4(169, 254, 169, 254)) {
```
mnecas, line 144: @coderabbitai what's this IP? and why do we ned to block it in the openshift?

yaacov (changes requested):
why do we proxy aap , how is this related to a migration flow ?

pkg/forklift-api/services/aap-job-templates.go
```diff
@@ -0,0 +1,166 @@
...
+// aapProxyHTTPClient is the client used for upstream AAP requests (overridable in tests).
+var aapProxyHTTPClient = &http.Client{Timeout: aapProxyTimeout}
+
+func serveAAPJobTemplates(resp http.ResponseWriter, req *http.Request, _ client.Client) {
+	if req.Method != http.MethodGet {
+		http.Error(resp, "method not allowed", http.StatusMethodNotAllowed)
+		return
+	}
+
+	rawBase := req.URL.Query().Get("url")
```
yaacov, line 33: @coderabbitai, hi, I have a concern about passing the url as a query, does that create a general purpose proxy, users can use for any url, even if it's not AAP ?

pkg/forklift-api/services/aap-job-templates.go
```diff
@@ -0,0 +1,166 @@
...
+	ctx, cancel := context.WithTimeout(req.Context(), aapProxyTimeout)
+	defer cancel()
+
+	httpReq, err := http.NewRequestWithContext(ctx, http.MethodGet, target, nil)
+	if err != nil {
+		log.Error(err, "failed to build AAP request", "target", target)
+		http.Error(resp, err.Error(), http.StatusInternalServerError)
+		return
+	}
+	httpReq.Header.Set("Authorization", "Bearer "+token)
```
yaacov, line 78: @coderabbitai, hi, this is a permanent token (e.g. not time limited) does it make sense to send it for each request? should we  negotiate some temporary session token?

yaacov (changes requested):
The approach suggested here present high security issues.

IMHO the original design where the UI handle the AAP comunication and fill in the hook as text is best.

In the original design we kept the Hook CRD without change, and used the known to work text hook field.

IMHO we should revert the changes adding AAP url, token and references, and return to use regular text fields the UI can fill in from any source they need to.

mnecas:
> IMHO the original design where the UI handle the AAP comunication and fill in the hook as text is best.

the UI can't get the playbook as text from AAP, we need to call AAP to get it started

yaacov:
> > IMHO the original design where the UI handle the AAP comunication and fill in the hook as text is best.
> 
> the UI can't get the playbook as text from AAP, we need to call AAP to get it started

My 2 cents, we need to re-design, this solution is too far away from the original design and introduce a lot of security risks.

mnecas:
@yaacov understand but as far as I see there are no alternatives compared to the current solution

yaacov:
> there are no alternatives compared to the current solution

we have many other choices, we need to discuss in the JIRA ticket or mtgs, the options are managing the AAP hooks using MTV instead of letting the UI send URL and Token for each request, we have many ways to implement this securely and we need to discuss them, so we come up with best design.

mnecas:
replaced by https://github.com/kubev2v/forklift/pull/5950
closing

PR #5872, opened 2026-04-15
chore(ci): fix redundant or-conditions failing go vet

yaacov:
The watch test polling loop duplicated `len(handler.created)` checks instead of checking `len(handler.deleted)`, making three conditions no-ops. go vet correctly flags these as "redundant or" expressions, breaking `make test`. Replace with the intended `.deleted` checks.

Resolves: none

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -76,3 +76,6 @@ __pycache__
 
 .gemini/
 gha-creds-*.json
+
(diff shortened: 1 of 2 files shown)


yaacov:
@coderabbitai, hi, can you verify this is just fixing an old copy / paste issue ?

PR #5875, opened 2026-04-15
MTV-5031 | Validate hook ServiceAccount exists in plan namespace

Hazanel:
When a Hook CR specifies a serviceAccount that does not exist or exists only in a different namespace, the plan passes validation and the migration hangs at the PreHook/PostHook phase with no error.

Fix:
Add validation in validateHooks() to check that each hook's serviceAccount exists in the plan namespace where hook Jobs run.

Ref: https://issues.redhat.com/browse/MTV-5031
Resolves: MTV-5031

### DIFF
--- a/pkg/controller/plan/validation.go
+++ b/pkg/controller/plan/validation.go
@@ -99,6 +99,7 @@ const (
 	GuestToolsIssue                 = "GuestToolsIssue"
 	VDDKAndOffloadMixedUsage        = "VDDKAndOffloadMixedUsage"
 	ServiceAccountNotValid          = "ServiceAccountNotValid"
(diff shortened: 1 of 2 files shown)


pkg/controller/plan/validation.go
```diff
@@ -1590,9 +1599,25 @@ func (r *Reconciler) validateHooks(plan *api.Plan) (err error) {
 			if !hook.Status.HasCondition(libcnd.Ready) {
```
yaacov, line 1599: @coderabbitai, hi, how does the hook have ready condition if the sa is not valid ?
do we check sa in the hook validation ? should we add that check ?

yaacov (approved):

PR #5882, opened 2026-04-15
MTV-3666 | Add Conversion CR to handle virt-v2v and deep inspection

solenoci:
In order to keep backwards compatibility and a "fallback" option, the migration flow is designed to work in any of the 3 modes:
- plan + conversion = plan creates the conversion CR and conversion reconciler then creates pods
- only plan = plan doesn't create conversion CR but directly the conversion pods (this mimics the old flow)
- only CR = if users supplies all the required args to the conversion spec, then the conversion can happen even without the plan   

Changes:
- Move pod creation logic from kubevirt into separate pkg
- Add new CR controller
- Add flag `UseConversionCR` to enable this feature
- Add deep inspection container image
- Add data extraction and reporting from inspection

Please see `docs/conversion-cr.md` section for documentation

Resolves: MTV-3666


operator/.upstream_manifests
```diff
@@ -5021,6 +5021,12 @@ spec:
                   - true (default): Use compatibility devices (SATA bus, E1000E NIC) to ensure bootability
                   - false: Use high-performance VirtIO devices (requires VirtIO drivers already installed in source VM)
                 type: boolean
+              useConversionCR:
+                default: false
+                description: |-
+                  UseConversionCR controls whether the plan creates Conversion custom resources
+                  to drive the VM conversion process instead of managing it directly.
+                type: boolean
```
mnecas: This is internal thingy, this should be in the ForkliftController

operator/config/crd/bases/forklift.konveyor.io_plans.yaml
```diff
@@ -2467,6 +2467,12 @@ spec:
                   - true (default): Use compatibility devices (SATA bus, E1000E NIC) to ensure bootability
                   - false: Use high-performance VirtIO devices (requires VirtIO drivers already installed in source VM)
                 type: boolean
+              useConversionCR:
+                default: false
```
mnecas: by default we should use the CR

pkg/controller/conversion/controller.go
```diff
@@ -0,0 +1,220 @@
...
+			err = nil
+		}
+		return
+	}
+	defer func() {
+		r.Log.V(2).Info("Conditions.", "all", conversion.Status.Conditions)
+	}()
+
+	// only reconcile if the conversion pod is not finished
+	if conversion.Status.Phase == api.PhaseSucceeded || conversion.Status.Phase == api.PhaseFailed {
```
mnecas, line 99: NP: succeded or failed should not have Requeue so it will not get started again. We should have similar logic as in the plan controller where is `NoReQ   = time.Duration(0)` if it finishes

pkg/controller/conversion/controller.go
```diff
@@ -0,0 +1,220 @@
...
+
+	if conversion.Status.HasBlockerCondition() {
+		conversion.Status.EndStagingConditions()
+		r.Record(conversion, conversion.Status.Conditions)
+		conversion.Status.ObservedGeneration = conversion.Generation
+		err = r.Status().Update(ctx, conversion)
+		return
+	}
+
+	conversion.Status.Phase = api.PhaseCreating
```
mnecas: In first run we set the phase to `api.PhasePending` and then override it to `api.PhaseCreating` without updating the resource. So the pending is never used.

pkg/controller/conversion/controller.go
```diff
@@ -0,0 +1,220 @@
...
+
+	// Ensure the conversion CR pod exists and track its state.
+	err = r.ensurePod(conversion)
```
mnecas: Think the `ensurePod` should return the pod yaml so we can check the pod status.

pkg/controller/conversion/ensurer.go
```diff
@@ -0,0 +1,414 @@
...
+func (e *Ensurer) GetPod(conversion *api.Conversion, labels map[string]string) (*core.Pod, error) {
+	list := &core.PodList{}
+	err := e.DestinationClient.List(context.TODO(), list,
+		client.InNamespace(conversion.Spec.TargetNamespace),
+		client.MatchingLabels(labels),
+	)
+	if err != nil {
+		return nil, liberr.Wrap(err)
+	}
+	if len(list.Items) > 0 {
```
mnecas: if len(list.Items) == 1 {...} else { return err{"ambiguous"}}}

pkg/apis/forklift/v1beta1/conversion.go
```diff
@@ -0,0 +1,195 @@
...
+}
+
+// PodSettings groups pod-level overrides for the conversion pod.
+type PodSettings struct {
+	// Pre-resolved transfer network annotations.
+	// +optional
+	TransferNetworkAnnotations map[string]string `json:"transferNetworkAnnotations,omitempty"`
+	// Labels to add to the conversion pod.
+	// +optional
+	Labels map[string]string `json:"labels,omitempty"`
```
mnecas: The K8S CRs already have Labels by default, we don't need to specify it on the spec. alternative could be propagating the labels from CR to the Pods.

pkg/apis/forklift/v1beta1/conversion.go
```diff
@@ -0,0 +1,195 @@
...
+	// Annotations to add to the conversion pod.
+	// +optional
+	Annotations map[string]string `json:"annotations,omitempty"`
```
mnecas: Same applies to the annotations, for example CDI is doing this to pass the networking stuff.
https://github.com/kubevirt/containerized-data-importer/blob/a0f9566d61db5a8d578c19bc7cf5173f5f4429b3/pkg/controller/common/util.go#L399-L408

pkg/apis/forklift/v1beta1/conversion.go
```diff
@@ -0,0 +1,195 @@
...
+type Connection struct {
+	// Secret containing virt-v2v credentials and connection parameters.
+	Secret core.ObjectReference `json:"secret" ref:"Secret"`
+}
+
+// PodSettings groups pod-level overrides for the conversion pod.
+type PodSettings struct {
+	// Pre-resolved transfer network annotations.
+	// +optional
+	TransferNetworkAnnotations map[string]string `json:"transferNetworkAnnotations,omitempty"`
```
mnecas, line 123: Also don't tihnk we need this as we can set the annotation on the resource itself

pkg/apis/forklift/v1beta1/conversion.go
```diff
@@ -0,0 +1,195 @@
...
+	DestinationProvider core.ObjectReference `json:"destinationProvider,omitempty" ref:"Provider"`
+	// Reference to the source VM.
+	VM ref.Ref `json:"vm"`
+	// Disks to be converted or inspected.
+	// For InPlace/Remote: populated from PVCs (namespaced name + volume mode).
+	// For Inspection: populated with disk paths from the source inventory.
+	// +optional
+	Disks []DiskRef `json:"disks,omitempty"`
+	// Source connection details including the virt-v2v credentials secret.
+	Connection Connection `json:"connection"`
```
mnecas: We discussed this offline, we should not need the connection but only run on local clusters

pkg/apis/forklift/v1beta1/conversion.go
```diff
@@ -0,0 +1,195 @@
...
+	// Disk decryption LUKS keys.
+	// +optional
+	LUKS core.ObjectReference `json:"luks,omitempty" ref:"Secret"`
```
mnecas: As LUKS require additional configs please make it a struct so it's more nested, plus makes it easily configurable in future.
core.ObjectReference -> secret for normal LUKS
clavis -> bool?

pkg/apis/forklift/v1beta1/conversion.go
```diff
@@ -0,0 +1,195 @@
...
+	// Container image for the virt-v2v pod. When empty the controller
+	// falls back to the global default from settings.
+	// +optional
+	Image string `json:"image,omitempty"`
```
mnecas, line 166: Note: With having MTV on two clusters we will need to be careful with the images as if we would provide image that would not be available on the destination cluster it could fail.

pkg/apis/forklift/v1beta1/conversion.go
```diff
@@ -0,0 +1,195 @@
...
+	// Whether the pod needs a KVM device and kubevirt.io/schedulable node selector.
+	// +optional
+	// +kubebuilder:default:=true
+	RequestKVM bool `json:"requestKVM,omitempty"`
+	// Sets LOCAL_MIGRATION env var in the conversion pod.
+	// +optional
+	LocalMigration bool `json:"localMigration,omitempty"`
+	// Whether to add UDN open-default-ports annotation.
+	// +optional
+	UDN bool `json:"udn,omitempty"`
```
mnecas: Hmmm we have two options:
1. ommit the UDN information and automatically get the information from the target namespace inside the conversion controller
2. leave it up to the forklift-controller to create the annotation on the Conversion resource

pkg/apis/forklift/v1beta1/conversion.go
```diff
@@ -0,0 +1,195 @@
...
+	// +kubebuilder:validation:Enum=DeepInspection;Inspection;InPlace;Remote
+	Type ConversionType `json:"type"`
+	// Reference to the source provider.
+	Provider core.ObjectReference `json:"provider" ref:"Provider"`
+	// Reference to the destination provider where pods and PVCs live.
+	// When empty or pointing to the host provider the local client is used;
+	// otherwise a remote k8s client is constructed from the provider URL
+	// and its secret.
+	// +optional
+	DestinationProvider core.ObjectReference `json:"destinationProvider,omitempty" ref:"Provider"`
```
mnecas: We discussed this offline, we should not need the connection but only run on local clusters

pkg/apis/forklift/v1beta1/conversion.go
```diff
@@ -0,0 +1,195 @@
...
+	GenerateName string `json:"generateName,omitempty"`
+}
+
+// ConversionSpec defines the desired state of Conversion.
+type ConversionSpec struct {
+	// Type of conversion.
+	// +kubebuilder:validation:Enum=DeepInspection;Inspection;InPlace;Remote
+	Type ConversionType `json:"type"`
+	// Reference to the source provider.
+	Provider core.ObjectReference `json:"provider" ref:"Provider"`
```
mnecas: What is the source provider used for?

pkg/apis/forklift/v1beta1/conversion.go
```diff
@@ -0,0 +1,195 @@
...
+	// Reference to the destination provider where pods and PVCs live.
+	// When empty or pointing to the host provider the local client is used;
+	// otherwise a remote k8s client is constructed from the provider URL
+	// and its secret.
+	// +optional
+	DestinationProvider core.ObjectReference `json:"destinationProvider,omitempty" ref:"Provider"`
+	// Reference to the source VM.
+	VM ref.Ref `json:"vm"`
```
mnecas, line 150: This is only for info on which VM this is running on? Think this should be annotation. Spec should contain only input that is activity used for something

pkg/apis/forklift/v1beta1/conversion.go
```diff
@@ -0,0 +1,195 @@
...
+	// +optional
+	Settings map[string]string `json:"settings,omitempty"`
+	// VDDK init container image. Required when type is DeepInspection.
+	// For other types, empty means no VDDK sidecar.
+	// +optional
+	VDDKImage string `json:"vddkImage,omitempty"`
+	// Whether the pod needs a KVM device and kubevirt.io/schedulable node selector.
+	// +optional
+	// +kubebuilder:default:=true
+	RequestKVM bool `json:"requestKVM,omitempty"`
```
mnecas: ahhh, yeah we don't need this one, it was only for the upstream github CI

pkg/apis/forklift/v1beta1/conversion.go
```diff
@@ -0,0 +1,195 @@
...
+	// +optional
+	Image string `json:"image,omitempty"`
+	// Namespace where conversion pods will be created.
+	// Defaults to the Conversion CR's own namespace.
+	// +optional
+	TargetNamespace string `json:"targetNamespace,omitempty"`
+	// XfsCompatibility selects the XFS-compatible virt-v2v image variant.
+	// +optional
+	// +kubebuilder:default:=false
+	XfsCompatibility bool `json:"xfsCompatibility,omitempty"`
```
mnecas, line 174: Note: I was thinking of removing this and have only the `Image` but that could be problematic if the other image would not be accessible in disconnected envs or in different versions of the MTV

pkg/apis/forklift/v1beta1/conversion.go
```diff
@@ -0,0 +1,195 @@
...
+	// Freeform settings passed to the conversion process as environment variables.
+	// +optional
+	Settings map[string]string `json:"settings,omitempty"`
```
mnecas, line 177: What other settings are we passing?

pkg/apis/forklift/v1beta1/conversion.go
```diff
@@ -0,0 +1,195 @@
...
+	// Namespace of the PVC.
+	// +optional
+	Namespace string `json:"namespace,omitempty"`
+	// VolumeMode indicates whether the disk is block or filesystem.
+	// +optional
+	VolumeMode *core.PersistentVolumeMode `json:"volumeMode,omitempty"`
+	// Filesystem mount path inside the conversion pod (e.g. /mnt/disks/disk0).
+	// Mutually exclusive with DevicePath.
+	// +optional
+	MountPath string `json:"mountPath,omitempty"`
```
mnecas, line 78: Do we need to have the mountPath configurable?
it's quite internal thing how the volume is mounted and how the v2v-wrapper reads it

pkg/controller/conversion/builder.go
```diff
@@ -0,0 +1,483 @@
...
+			Affinity:       cfg.Affinity,
+			RestartPolicy:  core.RestartPolicyNever,
+			InitContainers: initContainers,
+			Containers: []core.Container{
+				{
+					Env:             nil, // set by type-specific builder
+					ImagePullPolicy: core.PullAlways,
+					Resources: core.ResourceRequirements{
+						Requests: core.ResourceList{
+							core.ResourceCPU:    resource.MustParse(Settings.Migration.VirtV2vContainerRequestsCpu),
```
mnecas, line 188: Note: If we will go ahead with the 2 MTV design, then we will need to pass the CPU config via spec and the settings have as default. This would apply to everything using Settings. So the users still could configure the "main" cluster and the pods on the remote cluster would be correctly run.

pkg/controller/conversion/controller.go
```diff
@@ -0,0 +1,220 @@
...
+	if pod == nil {
+		return nil
+	}
+
+	conversion.Status.Pod = core.ObjectReference{
+		Namespace: pod.Namespace,
+		Name:      pod.Name,
+	}
+
+	switch pod.Status.Phase {
```
mnecas: Ahh here it is! 
I think ensurer should just ensure the conversion, because now we are also updating the conversion CR when calling it, when I was reading it I was bit confused where the status is getting updated from

pkg/controller/conversion/ensurer.go
```diff
@@ -0,0 +1,414 @@
...
+	case api.DeepInspection:
+		return e.ensureDeepInspectionPodFromSpec(conversion, cfg)
+	}
+	return nil
+}
+
+// ensureDeepInspectionPodFromSpec creates the deep inspection pod for a Conversion
+// CR if one does not already exist.
+func (e *Ensurer) ensureDeepInspectionPodFromSpec(conversion *api.Conversion, cfg convctx.PodConfig) error {
+	existing, err := e.GetPod(conversion, cfg.PodLabels)
```
mnecas: Is the `cfg.PodLabels` getting updated according to the type?
If one plan would run multiple types would it return the Pod or two different.

pkg/controller/conversion/ensurer.go
```diff
@@ -0,0 +1,414 @@
...
+
+		volumeMode := disk.VolumeMode
+		if volumeMode == nil && pvc.Spec.VolumeMode != nil {
+			volumeMode = pvc.Spec.VolumeMode
+		}
+
+		if volumeMode != nil && *volumeMode == core.PersistentVolumeBlock {
+			devPath := disk.DevicePath
+			if devPath == "" {
+				devPath = "/dev/block" + disk.Name
```
mnecas: Watchout this changes behaviour from index to disk name. 
`DevicePath: fmt.Sprintf("/dev/block%v", i),`
Same for `MountPath`

pkg/controller/conversion/controller.go
```diff
@@ -0,0 +1,220 @@
...
+			Message:  "The conversion pod is pending.",
+		})
+	}
+
+	conversion.Status.EndStagingConditions()
+
+	r.Record(conversion, conversion.Status.Conditions)
+
+	conversion.Status.ObservedGeneration = conversion.Generation
+	err = r.Status().Update(ctx, conversion)
```
mnecas: We are missing the request to the v2v API to get information such as OS or we are missing metrics about the disk copy. The conversion CR should have this information in the status including the progress.

pkg/controller/plan/migration.go
```diff
@@ -1239,10 +1239,23 @@ func (r *Migration) execute(vm *plan.VMStatus) (err error) {
...
+					step.AddError(resolveErr.Error())
+					break
+				}
+				if ready, err = r.kubevirt.EnsureConversion(vm, convType, r.Plan.Name, r.Plan.Namespace, string(r.Plan.UID), r.Migration, step); err != nil {
+					step.AddError(err.Error())
+					err = nil
+					break
+				}
+			} else {
+				if ready, err = r.kubevirt.EnsureGuestConversionPod(vm, step); err != nil {
```
mnecas, line 1289: I was thinking that we would keep the old path wihtout any changes, but now the `EnsureGuestConversionPod` is makde from the new builder which is used by both paths. In this case we can just remove the `ShouldUseConversionCR` and do it always.

pkg/controller/conversion/controller.go
```diff
@@ -0,0 +1,220 @@
...
+
+// ensurePod creates the pod for the Conversion CR if it does
+// not already exist and updates the status phase from the pod state.
+func (r *Reconciler) ensurePod(conversion *api.Conversion) (err error) {
+	ensurer, err := NewEnsurer(r.Client, r.Log, conversion.Spec)
+	if err != nil {
+		return
+	}
+
+	err = ensurer.EnsurePod(conversion)
```
mnecas: There should be extra builder method which will return Pod spec and this will be passed to the esurer

pkg/apis/forklift/v1beta1/conversion.go
```diff
@@ -0,0 +1,195 @@
...
+	// +optional
+	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
+	// Phase of the conversion lifecycle.
+	// +optional
+	// +kubebuilder:validation:Enum=Pending;CreatingPod;Running;Succeeded;Failed
+	Phase ConversionPhase `json:"phase,omitempty"`
+	// Reference to the managed virt-v2v pod.
+	// +optional
+	Pod core.ObjectReference `json:"pod,omitempty"`
+}
```
mnecas, line 311: missing v2v info

pkg/apis/forklift/v1beta1/conversion.go
mnecas: lets go with 2: leave it up to the forklift-controller to create the annotation on the Conversion resource

pkg/apis/forklift/v1beta1/conversion.go
mnecas: NP: DestinationProvider -> Destination? idk

operator/.upstream_manifests
solenoci: moved to the FC resource

operator/config/crd/bases/forklift.konveyor.io_plans.yaml
solenoci: change default to true

pkg/controller/conversion/controller.go
solenoci: added

pkg/controller/conversion/controller.go
solenoci: leftover from prototyping, removed the phase

pkg/controller/conversion/ensurer.go
solenoci: fixed

pkg/apis/forklift/v1beta1/conversion.go
solenoci: fixed: labels and annotations are now propagated from conversion CR to pods

pkg/apis/forklift/v1beta1/conversion.go
solenoci: fixed: labels and annotations are now propagated from conversion CR to pods

pkg/apis/forklift/v1beta1/conversion.go
solenoci: yep, same story with old workflow, really depends on env preparation by cluster admins

pkg/apis/forklift/v1beta1/conversion.go
solenoci: UDN removed, now it's an annotation on conversion CR which is then propagated down to pods

pkg/apis/forklift/v1beta1/conversion.go
solenoci: changed to destination (makes more sense)

pkg/apis/forklift/v1beta1/conversion.go
solenoci: removed

pkg/apis/forklift/v1beta1/conversion.go
solenoci: these settings are for ENV vars which are passed to pods, but we also use it for snapshot_moref

pkg/apis/forklift/v1beta1/conversion.go
solenoci: in order to make conversion CR standalone and to not do internal reconstruction (which could differ from the original), I chose to go this way. Everything is clear where it should be mounted so there are no miss-understandings between plan controller, conversion controller and pods

pkg/controller/conversion/ensurer.go
solenoci: yes, pod label is `deep-inspection`, it is passed to the pods in conversion pod builder based on the conversion cr type. the full label is `forklift.app=deep-inspection`. similar story with virt-v2v pods, the get `forklift.app=virt-v2v`

pkg/controller/conversion/ensurer.go
solenoci: fixed, thanks!

pkg/controller/plan/migration.go
solenoci: there is still the legacy virt-v2v-inspector pod, so I kept it like this. by default the conversion should be used either way, but this more of a fallback settings so we don't break everything at once

pkg/controller/plan/migration.go
```diff
@@ -1309,9 +1328,81 @@ func (r *Migration) execute(vm *plan.VMStatus) (err error) {
...
+					step.AddError(err.Error())
+					err = nil
+					break
+				}
+
+				if cr == nil {
+					// No CR yet, first attempt
+					// retryAllowed=false means if this one fails it will not be retried automatically.
+					_, err = r.kubevirt.CreateDeepInspectionConversion(
+						vm, snapshotMoref, r.Plan.Name, string(r.Plan.UID), false)
```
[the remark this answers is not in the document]
solenoci: This is intended behavior. When the conversion is created from plan, it should always have false, only if the conversion was created outside of plan it can be retried.

pkg/settings/features.go
```diff
@@ -62,6 +63,8 @@ type Features struct {
 	InsecureSkipVerifySupported bool
 	// Whether to run VMware driver removal scripts during Windows vSphere conversion (virt-customize).
 	VsphereVmwareDriverRemoval bool
+	// Whether to delegate VM conversion to Conversion CRs instead of managing it directly.
+	UseConversionCR bool
```
mnecas, line 70: We landed on keeping this flag?

operator/streams/downstream/mtv-operator.clusterserviceversion.yaml
```diff
@@ -160,6 +160,12 @@ spec:
           x-descriptors:
           - urn:alm:descriptor:com.tectonic.ui:radio:true
           - urn:alm:descriptor:com.tectonic.ui:radio:false
+        - description: Delegate VM conversion to Conversion CRs instead of managing it directly (default true)
```
mnecas, line 163: please hide this option from the UI

mnecas (approved):
i have some small notes, but think we can merge it so we can unblock the UI

PR #5885, opened 2026-04-15
MTV-4674 | copy-offload allow ssh command run without a timeout

rgolangh:
Resolves: MTV-4674

Commit 254555f6b enforced a default 30 sec timeout on the ssh internal
client. This prevents longer copy from working with ssh.

Bringing back the indefinite timeout behaviour on ssh commands.


mnecas (approved):
100%

PR #5902, opened 2026-04-16
MTV-3866 | Warm migration snapshot cleanup

Kyl-Bempah:


Kyl-Bempah:
@coderabbitai review

pkg/controller/plan/migration.go
```diff
@@ -465,17 +465,51 @@ func (r *Migration) removeLastWarmSnapshot(vm *plan.VMStatus) {
...
-			"Failed to clean up warm migration snapshots.",
-			"vm", vm)
+
+	// RemoveSnapshot is asynchronous on vSphere. Poll until the task completes so
+	// we don't return with an orphaned snapshot still being removed in the background.
+	precopy := plan.Precopy{RemoveTaskId: taskId}
+	retries := settings.Settings.Migration.SnapshotRemovalCheckRetries
+	checkRate := time.Duration(settings.Settings.Migration.SnapshotStatusCheckRate) * time.Second
+	for i := 0; i < retries; i++ {
+		time.Sleep(checkRate)
```
mrnold: I am going to agree with code rabbit, I don't think this is a good place to sleep. Also I have done a bunch of stuff in this area recently and SnapshotRemovalCheckRetries * SnapshotStatusCheckRate is usually not the solution. Sometimes it really does take VMware longer than this to remove a snapshot,

mrnold (changes requested):
I didn't see much to go on in MTV-3866, does this happen every time? The root cause called out in the code comments is plausible but I don't see a way to tell if it was this vs. removeLastWarmSnapshot not getting called at all.

pkg/controller/plan/migration.go
Kyl-Bempah: Yeah, will be updating this, mostly just using the AI to identify where in the codebase this would be happening.

Kyl-Bempah:
@mrnold The way Chenli described it as always reproducible makes me assume it happens every time. Is there a way to trace it to identify if `removeLastWarmSnapshot` was not called at all, failed silently, or was just running in the background?

pkg/controller/plan/migration.go
mrnold: It might be the right place to find and delete the forklift-named snapshot, it's just that it's already in a loop from the reconcile. The usual pattern is to try the delete once, and set some state so you can check on the result when it is called later. Then when it is successful, meaning the snapshot is cleanly removed, set the state to move along to the next part of the migration sequence.

You don't need to worry about polling anything in a loop like this, because as long as you make it correctly idempotent the reconcile loop will take care of it.

mrnold:
> @mrnold The way Chenli described it as always reproducible makes me assume it happens every time. Is there a way to trace it to identify if `removeLastWarmSnapshot` was not called at all, failed silently, or was just running in the background?

I think you will have to add your own temporary log messages and try it out, looking at the logging in removeLastWarmSnapshot I don't think you will get a really complete trace even if you had a full forklift controller log.

Kyl-Bempah:
@coderabbitai Do the latest changes sufficiently address all of the previous comments/concerns?

Kyl-Bempah:
@coderabbitai The latest commit should be `00638e18fec8db589a4f670e722f479386f38ab9` after the PR has been rebased, are you sure that analysis is correct?

mnecas (changes requested):
This is just workaround.

The issue with not cleaning up snapshots is not that we don't delete the snapshot.
But that we create 2 snapshots instead of 1. This is commonly caused by reconcile getting started imidiatly after previous run (usually caused by update to the Plan). 
We need forklift-controller logs to see the whole picture.

Kyl-Bempah:
Agreed, like I was telling Matthew, initially I was just looking for direction on where to point to, since the Jira ticket only includes the steps they took and some screenshots. I've left a [comment](https://redhat.atlassian.net/browse/MTV-3866?focusedCommentId=16771022), but at this point, I don't know if any further progress can be made on this.

mrnold:
I totally forgot about the duplicate snapshot operations, that's probably what it is. If that's the case there's a good chance this is already fixed in 2.10 and up, I had some pull requests in this area that I don't think got backported to 2.9.6 (the "affects version" in the bug). Maybe all we need is a re-test to verify this is no longer happening.

mnecas:
Personally I'm looking into larger redesign of the Forklift resource reconcile for next major version so we would not have any random reconcile triggeres and we could easily predict the flow.
agree with retesting as we have done lot of changes in this area, closing this PR

PR #5950, opened 2026-04-19
MTV-3663 | Centralize AAP config on ForkliftController

gwencasey96:
Add ForkliftController fields aap_url, aap_token_secret_name, aap_timeout and inject AAP_* env vars on main and inventory containers. Hooks select only job template id; legacy per-hook URL/token remain supported.

Add inventory GET /aap/job-templates. Move AAP HTTP client and token helpers to pkg/lib/aap.

Ref: https://issues.redhat.com/browse/MTV-3663
Resolves: MTV-3663
Made-with: Cursor


operator/config/crd/bases/forklift.konveyor.io_hooks.yaml
```diff
@@ -61,19 +61,23 @@ spec:
...
                       Any negative value: wait without a wall-clock limit until the job finishes or fails.
                     format: int64
                     type: integer
                   tokenSecret:
                     description: |-
-                      Reference to the Secret containing the AAP API token.
-                      The Secret must contain a key named "token" with the Bearer token value.
-                      The controller reads the Secret only from the migration plan namespace.
-                      Namespace must be empty or equal to that plan namespace.
+                      Deprecated: use ForkliftController `aap_token_secret_name` (Secret in the controller namespace) instead.
```
yaacov, line 78: AFAIR It was actually never introduced in a released API , so it can't be deprecated, IMHO we can just omit it, @gwencasey96 hi, was this API released in some version ? if not we can just remove it

operator/config/crd/bases/forklift.konveyor.io_hooks.yaml
```diff
@@ -61,19 +61,23 @@ spec:
...
                     type: integer
+                  jobTemplateName:
+                    description: Optional display name for UI (not used when launching
+                      jobs).
+                    type: string
                   timeout:
                     description: |-
                       Timeout in seconds to wait for the AAP job after launch (wall-clock polling limit).
-                      Default when omitted or 0: first use HookSpec.deadline if it is > 0; otherwise 3600 seconds (1 hour).
+                      Default when omitted or 0: first use HookSpec.deadline if it is > 0; otherwise ForkliftController
```
yaacov, line 71: while timout, token and url are nice to have here, they complicate the logic and make it less clear to users, it also complicate the code, my two cents is to omit these fields from this API and only use the centralized source, @coderabbitai WDYT ?

operator/.downstream_manifests
```diff
@@ -40,6 +40,19 @@ spec:
...
+                x-kubernetes-int-or-string: true
+              aap_token_secret_name:
+                description: 'Name of the Secret in the operator namespace containing
+                  the AAP API Bearer token (data key: token). Optional unless using
+                  AAP hooks.'
+                type: string
+              aap_url:
+                description: Ansible Automation Platform base URL (e.g. https://aap.example.com).
+                  Optional unless using AAP hooks.
+                type: string
```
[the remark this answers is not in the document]
yaacov: @coderabbitai hi, in https://github.com/kubev2v/forklift/pull/5950/changes#r3106433469 i suggest to remove the settings in the hook, and use only the centralized settings, it will solve this issue too, WDYT ?

operator/config/crd/bases/forklift.konveyor.io_hooks.yaml
yaacov: @gwencasey96 WDYT ^ ?

I understand why @coderabbitai suggest to leave the timeout override in this API, IMHO it's an overkill for here and we can do find without it, so it's up to you :-)

pkg/controller/provider/web/aap/handler.go
```diff
@@ -0,0 +1,75 @@
...
+			pageSize = n
+		}
+	}
+
+	httpTimeout := 30 * time.Second
+	if m.AAPTimeoutSeconds > 0 {
+		httpTimeout = time.Duration(m.AAPTimeoutSeconds) * time.Second
+	}
+	cl := aap.NewClient(m.AAPURL, token, httpTimeout)
+	list, err := cl.ListJobTemplates(ctx.Request.Context(), page, pageSize)
```
yaacov, line 68: @coderabbitai hi, IMHO the list job template and the pagination handling to aggregate the jobs to one flat list should be handled by our aap client, WDYT?

yaacov: @gwencasey96 WDYT ?

pkg/lib/aap/client.go
```diff
@@ -132,9 +147,52 @@ func (c *AAPClient) GetJobStatus(ctx context.Context, jobID int) (*JobStatusResp
...
+	}
+	rel, err := url.Parse("api/controller/v2/job_templates/")
+	if err != nil {
+		return nil, liberr.Wrap(err, "invalid AAP API path")
+	}
+	u := base.ResolveReference(rel)
+	q := u.Query()
+	q.Set("page", strconv.Itoa(page))
+	q.Set("page_size", strconv.Itoa(pageSize))
+	u.RawQuery = q.Encode()
```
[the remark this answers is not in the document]
yaacov: @coderabbitai hi, in https://github.com/kubev2v/forklift/pull/5950/changes#r3106445842 i suggested to move the page handling logic to this method, this will solve this issue, you also suggested adding a simple max jobs var to handle bit lists.

pkg/lib/aap/token.go
```diff
@@ -0,0 +1,72 @@
...
+	tokenBytes, ok := secret.Data["token"]
+	if !ok {
+		return "", fmt.Errorf("secret %s/%s does not contain 'token' key", ns, ref.Name)
+	}
+
+	return string(tokenBytes), nil
+}
+
+// GetTokenFromSecretName reads the token from a Secret by name in the given namespace.
+func GetTokenFromSecretName(ctx context.Context, k8sClient client.Client, namespace, name string) (string, error) {
```
yaacov, line 49: @coderabbitai, hi, I see we have getting token by secret name and by secret ref, can we refactor the code so we will not require two so similar methods ?

pkg/lib/aap/client.go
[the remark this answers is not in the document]
yaacov: @gwencasey96 WDYT ?

operator/.downstream_manifests
```diff
@@ -657,12 +674,11 @@ spec:
                     type: object
                     x-kubernetes-map-type: atomic
                   url:
-                    description: URL of the AAP instance (e.g., "https://aap.example.com").
+                    description: 'Deprecated: use ForkliftController `aap_url` instead.
+                      If set, takes effect only when cluster-wide AAP URL is not configured.'
```
mnecas, line 678: I think the logic should be opossite, if set explicity then use the url, otherwise use the `ForkliftController`. This applies to other parameters. Additionally it would not be "deprecated" if you still could use it and there would be some behaviour.

operator/config/crd/bases/forklift.konveyor.io_forkliftcontrollers.yaml
```diff
@@ -336,6 +336,15 @@ spec:
...
                 description: "Max retries when getting parent disks (default: 10)"
+              aap_url:
+                type: string
+                description: "Ansible Automation Platform base URL (e.g. https://aap.example.com). Optional unless using AAP hooks."
+              aap_token_secret_name:
+                type: string
+                description: "Name of the Secret in the operator namespace containing the AAP API Bearer token (data key: token). Optional unless using AAP hooks."
+              aap_timeout:
+                x-kubernetes-int-or-string: true
+                description: "Default timeout in seconds for AAP HTTP calls and job polling when not set on the Hook (optional)."
```
mnecas, line 347: Thiink we should add these to the hidden params on the forklift controller. `urn:alm:descriptor:com.tectonic.ui:hidden`

This should be configurable througth the console-plugion "settings" section

operator/.upstream_manifests
```diff
@@ -602,19 +615,23 @@ spec:
                   jobTemplateId:
                     description: ID of the AAP job template to execute.
                     type: integer
+                  jobTemplateName:
+                    description: Optional display name for UI (not used when launching
+                      jobs).
+                    type: string
```
mnecas, line 621: `jobTemplateName` for which UI? will this be visible in the AAP?

pkg/apis/forklift/v1beta1/hook_exec.go
```diff
@@ -17,7 +17,26 @@ func HookExecutionConfigValid(hook *Hook) bool {
 			return false
 		}
 		a := hook.Spec.AAP
-		return strings.TrimSpace(a.URL) != "" && a.JobTemplateID > 0 && strings.TrimSpace(a.TokenSecret.Name) != ""
+		if a.JobTemplateID <= 0 {
+			return false
```
mnecas, line 21: @gwencasey96 ptal comment above ^^

operator/config/crd/bases/forklift.konveyor.io_hooks.yaml
mnecas: I think we could keep the parameters wtihout deprecating.
1. by default lookup for the controller settings
2. if hook has tokenSecret url etc. use that instead

The problems we had were with the console-plugin integration. 
But the API itself I think is still valid.

yaacov: @mnecas , IMHO we should remove all unneeded settings,
One place to set the credentials is enough, we don't need to support a local secret we will not use, e.g. the only users for now is the UI, and they will use the centralized secret, removing this settings will simplify the code and simplify usage, WDYT ?

operator/.upstream_manifests
yaacov: It is for human usability only

from AAP we can get infomation about the project, we get:
**machine ID**, e.g. "6" - this is the only information we need to run the script
**name** - human readable name - we don't need it, but it's nice for humans
**information about how to get the script text** - the URL to get the script and the the credentials type ( nill if no authentication needed ) - require a 2nd fetch command with optional auth


If we want to create a CLI flow ( UI should be similar )
a. **get the projects list** - 
name | ID | optional the script text (we don't fetch it, it require a 2'nd fetch command)
b. **select the ID by the human readable name**
c. **run the AAP project using it's ID**

So the name is important for usability, without it we will see a list of IDs, ( "3,4,5,7" ) :-(

pkg/controller/provider/web/aap/handler.go
```diff
@@ -0,0 +1,75 @@
...
+	if ns == "" {
+		log.Error(nil, "inventory namespace not set for AAP handler")
+		ctx.Status(http.StatusInternalServerError)
+		return
+	}
+	token, err := aap.GetTokenFromSecretName(ctx.Request.Context(), h.Client, ns, m.AAPTokenSecretName)
+	if err != nil {
+		log.Error(err, "failed to read AAP token Secret")
+		ctx.JSON(http.StatusBadGateway, gin.H{"message": err.Error()})
+		return
```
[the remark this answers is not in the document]
yaacov: @gwencasey96 hi, I agree with @coderabbitai ^^

yaacov:
@gwencasey96 hi, don't forget to squash the commits

operator/.upstream_manifests
mnecas: in that case it should go to the status not to the input information if we are not going to use it or some annotation

operator/config/crd/bases/forklift.konveyor.io_hooks.yaml
mnecas: I'm thinking about users who would want to automate this, and having this expicitly on the Hook CR itself gives you better context.

mnecas: @yaacov Yeah for the UI everything done via ForkliftController makes sense, thats was one of the main problems that we were trying to solve with this PR. 
I'm more thinking about the users who want to automate it, and I think setting everyhing at one place (in Hook)  is more user friendly as you see the whole context at once and don't need to jump between ForkliftController CR.

I'm not biggest fan of the ForkliftController CR changes as these changes cause forklift-controller pod restart as it gets restarted with new env variable. However i don't see any other better way around this.

pkg/lib/aap/client.go
```diff
@@ -0,0 +1,289 @@
...
+	defaultJobTemplatePageSize = 20
+	maxJobTemplatePageSize     = 100
+	// defaultMaxJobTemplatesList caps how many templates ListAllJobTemplates will aggregate from AAP.
+	defaultMaxJobTemplatesList = 500
+	listAllInternalPageSize    = 100
+)
+
+// LaunchJob launches an AAP job template with the given extra variables.
+func (c *Client) LaunchJob(ctx context.Context, jobTemplateID int, extraVars map[string]string) (int, error) {
+	u := fmt.Sprintf("%s/api/controller/v2/job_templates/%d/launch/", c.URL, jobTemplateID)
```
yaacov: @coderabbitai hi, we use the api version prefix in several places

a. should we refactor the api version prefix to one place ?
b. different AAP/Tower/AWX install may have different versions names, should we add a "get api prefix" method to fetch the current version prefix for the server ?

for example:
``` go
  // detectAPIPrefix probes the /api/ endpoint to determine the API path                            
  func (c *Client) detectAPIPrefix(ctx context.Context) error {
      u := fmt.Sprintf("%s/api/", c.URL)                                                            
      ...                                          
  } 
```
WDYT ?

yaacov: @gwencasey96 WDYT ^^ ?

operator/.upstream_manifests
yaacov: I agree with @mnecas , we can detect the name using the id in the hook reconcile loop, and update the hook status to add a human readable name to go with the machine readable id in the spec.

operator/config/crd/bases/forklift.konveyor.io_hooks.yaml
yaacov: after offline discussion with @mnecas I agree we can re-add url/token secret/timout in the hook itself in addition to the central settings as overrides of the central url/token secret/timout in the settings.

logic: first look for url/token/timout in hook, if missing use central information, if missing err

pros of allowing overides:
users can link a project id to an AAP url in 1:1 connection.

cons of allowing overrides:
complicated and may be confusing to follow the logic.

pkg/controller/plan/hook.go
```diff
@@ -405,29 +407,49 @@ func (r *HookRunner) aapJobExtraVars() map[string]string {
...
+		step.MarkCompleted()
+		return
+	}
+	if strings.TrimSpace(m.AAPURL) == "" || strings.TrimSpace(m.AAPTokenSecretName) == "" {
+		step.AddError("AAP is not configured: set ForkliftController aap_url and aap_token_secret_name")
+		step.MarkCompleted()
+		return
+	}
+	aapURL := strings.TrimSpace(m.AAPURL)
+	ns := strings.TrimSpace(os.Getenv("POD_NAMESPACE"))
```
mnecas: @coderabbitai is this POD_NAMESPACE the hook namespace or forklift-controller namespace?

operator/.upstream_manifests
mnecas: I think the status would be best, but lets keep it to annotation. I went over the code and it would add quite large complexity to the PR that is not releated to this fix

mnecas (approved):
lgtm, i'll wait for @yaacov review aswell before merge

pkg/lib/aap/client.go
```diff
@@ -0,0 +1,299 @@
...
+	JobStatusFailed     = "failed"
+	JobStatusError      = "error"
+	JobStatusCanceled   = "canceled"
+	JobStatusPending    = "pending"
+	JobStatusWaiting    = "waiting"
+	JobStatusRunning    = "running"
+)
+
+// aapControllerV2APIPrefix is the URL path prefix for the Ansible Automation Platform controller API.
+// (AWX/Tower historically used /api/v2/; dynamic detection can be added later if needed.)
```
yaacov: @coderabbitai the static prefix will fail on any Ansible setup that is not AAP, can you suggest a more robust dynamic detection method ?

yaacov: @coderabbitai, can you simplify your example to
a. check if prefix value already set, if so, exit
b. fetch the api endpoint struct, and return the current version string, or default aapControllerV2APIPrefix if err or missing/mallformed field ?

yaacov: from ansible docs, response for `GET /api ` is -
```
{
  "available_versions": {
    "v2": "/api/v2/"
  },
  "current_version": "/api/v2/",
  "custom_login_info": "",
  "custom_logo": "",
  "description": "AWX REST API",
  "login_redirect_override": "",
  "oauth2": "/api/o/"
}
```
why do you think current_version may be missing the full path ? what am i missing ?

pkg/controller/provider/web/aap/handler.go
```diff
@@ -0,0 +1,84 @@
...
+	Client client.Client
+}
+
+// AddRoutes implements libweb.RequestHandler.
+func (h *Handler) AddRoutes(e *gin.Engine) {
+	e.GET(jobTemplatesPath, h.ListJobTemplates)
+}
+
+// ListJobTemplates returns all configured AAP job templates (flat list). Pagination is handled inside pkg/lib/aap.
+func (h *Handler) ListJobTemplates(ctx *gin.Context) {
```
yaacov, line 38: @coderabbitai hi, in all our inventory endpoint we check that the user have get/list permissions for the providers CRD, I am missing this check here, what am i missing, do we handle RBAC for this endpoint ?

yaacov: we need the fix in this PR, can you suggest a fix?

pkg/lib/aap/client.go
```diff
@@ -30,20 +31,134 @@ const (
...
+	if v == "" {
+		return ""
+	}
+	if !strings.HasPrefix(v, "/") {
+		v = "/" + v
+	}
+	return v
+}
+
+func (c *Client) apiV2ResourceURL(ctx context.Context, path string) (string, error) {
```
yaacov: @coderabbitai, hi, the ansible api ensure we get the correct path from the API call, why do we need this helper, what am i missing ?

pkg/lib/aap/client.go
```diff
@@ -98,7 +213,10 @@ const (
 
 // LaunchJob launches an AAP job template with the given extra variables.
 func (c *Client) LaunchJob(ctx context.Context, jobTemplateID int, extraVars map[string]string) (int, error) {
-	u := c.apiControllerV2URL(fmt.Sprintf("/job_templates/%d/launch/", jobTemplateID))
+	u, err := c.apiV2ResourceURL(ctx, fmt.Sprintf("/job_templates/%d/launch/", jobTemplateID))
```
yaacov: we already build the string using sprintf, why do we need the helper apiV2ResourceURL method, what am i missing?

pkg/lib/aap/client.go
```diff
@@ -30,20 +31,134 @@ const (
...
+// automatic discovery (GET /api) fails or is unavailable.
 const aapControllerV2APIPrefix = "/api/controller/v2"
 
 // Client handles communication with Ansible Automation Platform API.
 type Client struct {
 	URL    string
 	Token  string
 	client *http.Client
+	// manualAPIPrefix, if set via SetAPIPrefix, is used and GET /api is not called.
+	manualAPIPrefix string
```
yaacov: we can set the client api prefix how ever we want, automatically, use a default  value, or manually, but once set, we should only have one source of truth in the client struct, what am i missing?

pkg/lib/aap/client.go
```diff
@@ -30,20 +31,134 @@ const (
...
+	apiPrefixReady bool
+	apiPrefixMu    sync.Mutex
 }
 
-// apiControllerV2URL returns base AAP URL with trailing segment path under aapControllerV2APIPrefix (path must start with /).
-func (c *Client) apiControllerV2URL(path string) string {
-	return strings.TrimRight(c.URL, "/") + aapControllerV2APIPrefix + path
+// SetAPIPrefix sets the API path prefix (e.g. "/api/v2") for this client, skipping automatic
+// discovery on first use. Pass empty to use GET /api and current_version.
+func (c *Client) SetAPIPrefix(prefix string) {
```
yaacov: this is a whole method to basically do `client.apiPrefix = ...` what am i missing, what does this method do?

pkg/lib/aap/client.go
yaacov: @coderabbitai ^^ ?

pkg/lib/aap/client.go
yaacov: @coderabbitai ^^ ?

yaacov: the question is:
if we no the input to this method is already normalized because it's the output of API that give the normalized version, why do we normalize it again ?

pkg/lib/aap/client.go
yaacov: sorry, the question was why we normalize the responce?

yaacov: where do we add the doubld '/' why can't we just not do that ?

pkg/lib/aap/client.go
```diff
@@ -0,0 +1,421 @@
...
+func WithPathPrefix(prefix string) ClientOption {
+	return func(c *Client) {
+		c.staticPathPrefix = prefix
+	}
+}
+
+// resolvePathPrefix sets apiPathPrefix once: optional static ([WithPathPrefix]), else GET {baseURL}/api
+// (Ansible/AWX root doc) and current_version, else the package default aapControllerV2APIPrefix.
+func (c *Client) resolvePathPrefix(ctx context.Context) {
+	c.prefixMu.Lock()
```
yaacov: hi, why do you need mutex for this call?

pkg/lib/aap/client.go
```diff
@@ -0,0 +1,421 @@
...
+	if c.apiPathPrefix != "" {
+		c.prefixMu.Unlock()
+		return
+	}
+	if c.staticPathPrefix != "" {
+		c.apiPathPrefix = normalizePathPrefix(c.staticPathPrefix)
```
yaacov, line 109: the static path is human source of truth, why do we need to normalize it ?

pkg/lib/aap/client.go
```diff
@@ -0,0 +1,421 @@
...
+	if err == nil && prefix != "" {
+		chosen = prefix
+	} else {
+		if err != nil {
+			log.Error(err, "AAP/AWX API prefix: using default", "default", aapControllerV2APIPrefix)
+		}
+	}
+
+	c.prefixMu.Lock()
+	if c.apiPathPrefix == "" {
```
yaacov: i am missing the point of this if, what am i missing?

pkg/lib/aap/client.go
gwencasey96: The mutex is not needed in the current code-I removed it now since you prefer to keep the code smaller

I put it there for future possible changes to keep the lazy apiPathPrefix safe. For example-if we want to fetch job templates in parallel as a future story.

pkg/lib/aap/client.go
gwencasey96: The tests don’t cover everything, I think they would not catch for example missing leading slash, trailing slash, spaces. We could* add more tests but I think this is easier. When the user provides the URL we get current_version from Ansible API we don’t control the formatting

pkg/lib/aap/client.go
gwencasey96: Same situation as the mutex, not needed now, removed this-it was there for future possible change

pkg/lib/aap/client.go
```diff
@@ -0,0 +1,409 @@
...
+		},
+	}
+	for _, o := range opts {
+		o(c)
+	}
+	return c
+}
+
+// LaunchJobResponse represents the response from AAP job launch.
+type LaunchJobResponse struct {
```
yaacov: i have a concern about the placement of this type and consts block, should it not be on top of the file with the other types and consts, what am i missing?

pkg/apis/forklift/v1beta1/hook_exec.go
```diff
@@ -17,7 +17,23 @@ func HookExecutionConfigValid(hook *Hook) bool {
...
-		return strings.TrimSpace(a.URL) != "" && a.JobTemplateID > 0 && strings.TrimSpace(a.TokenSecret.Name) != ""
+		return a.JobTemplateID > 0
 	}
 	return strings.TrimSpace(hook.Spec.Image) != ""
 }
+
+// HookAAPRunnable returns true when an AAP hook has a job template and either:
+//   - per-hook connection: non-empty spec.aap.url and spec.aap.tokenSecret.name, or
+//   - cluster connection: non-empty ForkliftController aap_url and aap_token_secret_name.
+func HookAAPRunnable(hook *Hook, clusterAAPURL, clusterAAPTokenSecretName string) bool {
```
yaacov, line 28: in all calls to this methos the url and token are:
settings.Settings.Migration.AAPURL,
settings.Settings.Migration.AAPTokenSecretName,

can we simplify the method signature, by removing this constant/78known args?

pkg/lib/aap/client.go
gwencasey96: moved to top and added a rule in my local environment to automatically fix this formatting for me in the future

pkg/apis/forklift/v1beta1/hook_exec.go
gwencasey96: um-added a new file with small wrapper and updated validation files, not sure if you had a simpler approach in mind

yaacov: I was thinking of importing settings, here and replacing the args with consts from the settings pkg

offline discussion, @gwencasey96 raised a concern about cyclic imports,

@gwencasey96 is it possible to move this method to the hook client pkg ?
will it help with the cyclic import ?

.cursor/rules/go-declarations-at-top.mdc
```diff
@@ -0,0 +1,17 @@
+---
+description: Keep package-level const, var, and type declarations at the top of Go files (before functions).
```
yaacov: coding syle conventions should be documented as the project coding style, and everyone should strive to follow them if they are not strict rules that are enforced by the linter.

please move to a different PR that concern with coding style.

same with the other file rule, once we have project coding style, we can create a rule to follow project coding style.

pkg/apis/forklift/v1beta1/hook_exec.go
yaacov: @gwencasey96 hi, the cyclic impoerts issue makes sense to me, closing this comment

PR #5952, opened 2026-04-19
MTV-5068 | Always use the provider url and not the secret url

rgolangh:
The vsphere provider url is not guaranteed to be in the secret or not
guaranteed to be even correct.
The source of truth of the provider url is from the provider .Spec.Url.
That Provider READY state depends on the Spec.Url to be correct
otherwise the connection test fails and the provider is unusable.
Meaning that if a plan is executing then the provider spec.url is the
only correct url to use.

The fix is specific to storage offload secret handling.

Resolves: MTV-5068


mnecas (approved):

mnecas (commented):
@rgolangh please take a look at the build CI

rgolangh:
@mnecas ci is green. Merge?

rgolangh:
@mnecas ping @borod108

PR #5973, opened 2026-04-23
MTV-5091 | Switch VMware search from UUID to moref ID

mnecas:
Issue:
FindByUuid on config.uuid is ambiguous when multiple VMs share a BIOS UUID.

Fix:
Use the inventory managed object id so cold migration uses VM ID from plan spec.

Resolves: MTV-5091

Important note:
Warm migraiton will not work as CDI uses UUIDs, this woudl require additional changes.


mnecas:
@coderabbitai ptal

yaacov (approved):
need squashing

PR #6014, opened 2026-04-26
MTV-5072 | Add iSCSI disk transfer for Hyper-V migrations

Hazanel:
Add iSCSI as an alternative to SMB for transferring Hyper-V VM disks. Key changes:

- hyperv-populator: new binary that logs in to an iSCSI target on the Hyper-V host, discovers the LUN, and copies it to the target PVC
- iSCSI target lifecycle: create/teardown iSCSI virtual disks and targets on the Hyper-V host via WinRM PowerShell
- Volume populator integration: register HyperVVolumePopulator CRD, wire it into the populator-controller with progress reporting
- Provider setting: transfer method controlled by hyperVTransferMethod in provider.Spec.Settings (values: "smb" (default) or "iscsi")

Ref: https://issues.redhat.com/browse/MTV-5072

Resolves: MTV-5072

Please use the an interactive design flow - [hyperv-iscsi-flow-interactive.html](https://github.com/user-attachments/files/27122787/hyperv-iscsi-flow-interactive.html)

image


build/forklift-operator-bundle/Containerfile-downstream
```diff
@@ -42,6 +42,9 @@ ARG VIRT_V2V_IMAGE="registry.redhat.io/mtv-candidate/mtv-virt-v2v-rhel10@sha256:
 ARG VIRT_V2V_IMAGE_RHEL9="registry.redhat.io/mtv-candidate/mtv-virt-v2v-rhel9@sha256:f9f62a6716d9ee1ae156d15f54f5a1d8887ff379eac94912a7db5596ba46b1ab"
 
 ARG VSPHERE_COPY_OFFLOAD_POPULATOR_IMAGE="registry.redhat.io/mtv-candidate/mtv-vsphere-copy-offload-populator-rhel9@sha256:a1f8308e6fa235654a4d35fe9c40e4657777814968879bde3f6c3713d6728150"
+# TODO: Replace with registry.redhat.io/mtv-candidate/mtv-hyperv-populator-rhel9@sha256:...
+# once the downstream image is published.
+ARG HYPERV_POPULATOR_IMAGE="quay.io/kubev2v/hyperv-populator:latest"
```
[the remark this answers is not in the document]
Hazanel: Already tracked by the inline TODO. This will be replaced with the downstream digest before GA, same as the other ARG's

cmd/hyperv-populator/main.go
```diff
@@ -0,0 +1,702 @@
...
+		if b != 0 {
+			allZero = false
+			break
+		}
+	}
+	if allZero {
+		return fmt.Errorf("post-copy verify FAILED for LUN %d: first %d bytes of %s are all zeros. "+
+			"The iSCSI source LUN likely served empty data (check differencing disk setup on the Hyper-V host)",
+			lunID, n, path)
+	}
```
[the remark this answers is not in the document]
Hazanel: These are Windows VMs. An all-zero first 1KiB is not a valid Windows disk ,it means the differencing disk served empty data. The ext4 superblock-at-1024 scenario doesn't apply here. The check also looks for MBR/GPT signatures after the zero check and only warns if those are missing

pkg/controller/provider/container/hyperv/client.go
```diff
@@ -235,8 +247,20 @@ func (r *Client) ListNetworks() ([]types.Network, error) {
...
+// returns a synthetic entry so the UI can build a valid StorageMap.
 func (r *Client) ListStorages() ([]types.Storage, error) {
+	if r.provider != nil && r.provider.GetHyperVTransferMethod() == api.HyperVTransferMethodISCSI {
+		storage := types.Storage{
+			ID:   hvutil.StorageIDDefault,
+			Name: StorageNameISCSI,
+			Type: StorageTypeISCSI,
+		}
+		r.Log.Info("Returning synthetic iSCSI storage entry", "name", storage.Name)
+		return []types.Storage{storage}, nil
```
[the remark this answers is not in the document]
Hazanel: Intentional. The Path field represents the SMB share path and has no meaning for iSCSI transfers. Setting it to a fake value would be misleading. The synthetic entry provides just enough for the UI to build a valid StorageMap — ID, Name, and Type are sufficient.

pkg/lib/hyperv/powershell/scripts.go
```diff
@@ -128,3 +140,139 @@ const (
...
+	// Parameters: windowsPath
+	TestPath = `Test-Path -Path '%s' -PathType Leaf`
+
+	// TestPaths checks multiple file paths in a single WinRM call.
+	// Returns JSON: {"Missing": ["path1", "path3"]} for paths that don't exist.
+	// Parameter: comma-separated quoted strings, e.g. "'C:\\path1','C:\\path2'"
+	TestPaths = `$paths=@(%s)
+$missing=@()
+foreach($p in $paths){if(-not (Test-Path -Path $p -PathType Leaf)){$missing+=$p}}
+[PSCustomObject]@{Missing=$missing}|ConvertTo-Json -Compress`
```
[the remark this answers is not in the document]
Hazanel: TestPaths is called via fmt.Sprintf (not BuildCommand), with pre-quoted paths from buildPathList() which handles escaping. The test TestTestPaths_Template covers this.
Add-IscsiVirtualDiskTargetMapping on Windows Server does succeed silently if the exact same path+LUN mapping already exists

pkg/lib-volume-populator/populator-machinery/controller.go
```diff
@@ -645,6 +661,67 @@ func (c *controller) syncPvc(ctx context.Context, key, pvcNamespace, pvcName str
...
+				// Clear FSGroup/RunAsUser from the template — they conflict with
+				// privileged hostPath mounts to /dev and /etc/iscsi.
+				pod.Spec.SecurityContext = &corev1.PodSecurityContext{}
+				// NETLINK_ISCSI is scoped to the network namespace, hostNetwork
+				// lets the pod use the host kernel's iSCSI transport and the
+				// host's iscsid daemon.
+				pod.Spec.HostNetwork = true
+				// HostPID allows nsenter -t 1 -m to run iscsiadm in the host's
+				// mount namespace, ensuring version compatibility with iscsid.
+				pod.Spec.HostPID = true
```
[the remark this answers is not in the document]
Hazanel: Intentional. The pod runs Privileged: true with hostPID, hostNetwork, and hostPath mounts to /dev and /etc/iscsi. In this context, SeccompProfile, RunAsNonRoot, AllowPrivilegeEscalation=false, and Capabilities.Drop: ALL are all overridden by the privileged flag and would have no effect. Clearing them avoids confusing the security context with settings that look restrictive but aren't enforced.

Hazanel:
@coderabbitai - please check whether I addressed all of your concerns

Hazanel:
@coderabbitai - I fixed them on the code please check

build/forklift-operator-bundle/Containerfile-downstream
solenoci: this will fail the build

build/hyperv-populator/Containerfile
```diff
@@ -0,0 +1,34 @@
+FROM registry.access.redhat.com/ubi9/go-toolset:1.25.7-1772728670 AS builder
+USER 0
+WORKDIR /app
+COPY --chown=1001:0 ./ ./
+ENV GOFLAGS="-mod=vendor -tags=strictfipsruntime"
+ENV GOEXPERIMENT=strictfipsruntime
+ENV GOCACHE=/go-build/cache
+RUN --mount=type=cache,target=${GOCACHE},uid=1001 go build -buildvcs=false -ldflags="-w -s" -o hyperv-populator github.com/kubev2v/forklift/cmd/hyperv-populator
+
+FROM quay.io/centos/centos:stream9
```
solenoci: please, use ubi image if possible

Kyl-Bempah: No I think this was fine, what was needed was the a `Containerfile-downstream` for this component built on UBI with the predefined enterprise requirements.

Kyl-Bempah:
@Hazanel is there an ETA on this being merged, won't be able to onboard it to CI until the actual Containerfile-downstream exists for RPMs

build/hyperv-populator/Containerfile
Hazanel: @Kyl-Bempah , @solenoci  - So what should I do, revert to "`FROM quay.io/centos/centos:stream9`" or use ubi image?

Kyl-Bempah: From my perspective, in order to match parity with the other components, you can revert it to centos in this file and then create another file in this directory called `Containerfile-downstream` that would use the ubi image, @solenoci do you agree?

Kyl-Bempah: @Hazanel Sorry for the confusion regarding this, Stefan & I discussed this offline, this file can be left as is. What is desired is a second file named `Containerfile-downstream` that would use `registry.redhat.io` as its base image, similar to how the other directories have it https://github.com/kubev2v/forklift/tree/main/build/hyperv-provider-server

Hazanel:
@mnecas @yaacov  - I did my best please take a look.
Personally I still don't like it but up to you

mnecas:
This does lot of things that should be done inside the CSI, i don't like to have privileged container that does host operations. I can't find any ways to work around this inside the kubernetes. I think we should go in direction of improving the SMB connections like multipath.
closign

PR #6027, opened 2026-04-28
MTV-4365 | Standardize logging across all storage providers

amitosw15:
Storage providers used flat `klog.Infof()`/`klog.Warningf()` calls with no
consistent prefix or structure. In production, this made it difficult to filter
logs by provider or correlate a log line with the provider that emitted it —
especially when multiple providers run concurrently during a migration plan.

This PR switches all providers to structured logging via `logr`, with each
provider carrying its own named logger as a struct field. Every log line now
includes the provider name as a key-value pair, making it possible to filter
by provider in any log aggregation tool.

`logr` has no Warning level, so messages that were previously `klog.Warningf()`
are emitted via `log.Error(nil, ...)` — this is the idiomatic `logr` approach
and ensures they surface at a severity above Info.

Resolves: MTV-4365

### DIFF
--- a/cmd/vsphere-copy-offload-populator/internal/flashsystem/flashsystem.go
+++ b/cmd/vsphere-copy-offload-populator/internal/flashsystem/flashsystem.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/google/uuid"
 	"github.com/kubev2v/forklift/cmd/vsphere-copy-offload-populator/internal/fcutil"
+	"github.com/kubev2v/forklift/cmd/vsphere-copy-offload-populator/internal/logutil"
 	"github.com/kubev2v/forklift/cmd/vsphere-copy-offload-populator/internal/populator"
 	"github.com/kubev2v/forklift/cmd/vsphere-copy-offload-populator/internal/vmware"
 	"k8s.io/klog/v2"
@@ -39,17 +40,17 @@ type ExtractedMapping struct {
 
 // extractWWPNsFromFCFormat extracts individual WWPNs from fc.WWNN:WWPN format
 // Uses the second part (after colon) as the real WWPN
-func extractWWPNsFromFCFormat(fcStrings []string) []string {
+func extractWWPNsFromFCFormat(fcStrings []string, log klog.Logger) []string {
 	var wwpns []string
 	for _, fcStr := range fcStrings {
 		if strings.HasPrefix(fcStr, "fc.") {
 			wwpn, err := fcutil.ExtractWWPN(fcStr)
 			if err != nil {
-				klog.Warningf("Failed to extract WWPN from %s: %v", fcStr, err)
+				log.Info("failed to extract WWPN", "fc_string", fcStr, "err", err)
 				continue
 			}
 			wwpns = append(wwpns, wwpn)
-			klog.Infof("Extracted WWPN: %s from %s", wwpn, fcStr)
+			log.V(2).Info("extracted WWPN", "wwpn", wwpn, "fc_string", fcStr)
 		}
 	}
 	return wwpns
@@ -101,6 +102,7 @@ type FlashSystemAPIClient struct {
 	authToken    string // Session token from /auth
 	username     string // Store for re-authentication
 	password     string // Store for re-authentication
+	log          klog.Logger
 }
 
 // NewFlashSystemAPIClient creates and authenticates a new API client.
@@ -115,6 +117,7 @@ func NewFlashSystemAPIClient(managementIP, username, password string, sslSkipVer
 		httpClient:   httpClient,
 		username:     username,
 		password:     password,
+		log:          logutil.NewLogger("flashsystem"),
 	}
 
 	// Initial authentication
@@ -140,7 +143,7 @@ func (c *FlashSystemAPIClient) authenticate() error {
 	req.Header.Set("X-Auth-Username", c.username)
 	req.Header.Set("X-Auth-Password", c.password)
 
-	klog.Infof("Attempting to authenticate with FlashSystem at %s for user %s", c.ManagementIP, c.username)
+	c.log.V(2).Info("authenticating with FlashSystem", "host", c.ManagementIP)
 	resp, err := c.httpClient.Do(req)
 	if err != nil {
 		return fmt.Errorf("failed to send auth request to FlashSystem: %w", err)
@@ -167,7 +170,7 @@ func (c *FlashSystemAPIClient) authenticate() error {
 
 	c.authToken = authResp.Token
 
-	klog.Infof("Successfully authenticated with FlashSystem and obtained session token.")
+	c.log.V(2).Info("Successfully authenticated with FlashSystem and obtained session token.")
 	return nil
 }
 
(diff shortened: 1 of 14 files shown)


cmd/vsphere-copy-offload-populator/internal/flashsystem/flashsystem.go
```diff
@@ -167,7 +169,7 @@ func (c *FlashSystemAPIClient) authenticate() error {
 
 	c.authToken = authResp.Token
 
-	klog.Infof("Successfully authenticated with FlashSystem and obtained session token.")
+	log.V(2).Info("authentication successful, session token obtained")
```
rgolangh: keep the original message that says flashsystem

cmd/vsphere-copy-offload-populator/internal/flashsystem/flashsystem.go
```diff
@@ -140,7 +141,8 @@ func (c *FlashSystemAPIClient) authenticate() error {
 	req.Header.Set("X-Auth-Username", c.username)
 	req.Header.Set("X-Auth-Password", c.password)
 
-	klog.Infof("Attempting to authenticate with FlashSystem at %s for user %s", c.ManagementIP, c.username)
+	log := klog.Background().WithName(loggerName).WithName("setup")
```
rgolangh: why not store the log in the client Struct?

cmd/vsphere-copy-offload-populator/internal/flashsystem/flashsystem.go
```diff
@@ -445,8 +447,10 @@ func (c *FlashSystemClonner) UnmapTarget(targetLUN populator.LUN, context popula
 
 // EnsureSingleHost creates or finds a single host with the given identifiers.
 func (c *FlashSystemClonner) EnsureClonnerIgroup(hostName string, clonnerIdentifiers []string) (populator.MappingContext, error) {
+	log := klog.Background().WithName(loggerName).WithName("map").WithName("ensure-igroup")
```
rgolangh: So are we going to initialize a logger with the name of every method? 
We should see how this fits in with the logger from the calling layer:
We have few loggers the Populate phase see remote_esxcli.go:83
```
   setupLog := log.WithName("setup")
	mapLog := log.WithName("map-volume")
	rescanLog := log.WithName("rescan")
	cloneLog := log.WithName("clone")
	cleanupLog := log.WithName("cleanup")

```

The call to concrete implementation is mixed:
setupLog
  ensureIgroup    
  currentMappedGroup
MapLog
   mapTarget
RescanLog
   rescan
CloneLog
   clone

I don't think we need a unique logger name for every method , like Ensure and MapTarget and so on. We can go with a single log name for all clonner interaction, and preferably set them as a struct field

cmd/vsphere-copy-offload-populator/internal/flashsystem/flashsystem.go
```diff
@@ -167,7 +171,7 @@ func (c *FlashSystemAPIClient) authenticate() error {
 
 	c.authToken = authResp.Token
 
-	klog.Infof("Successfully authenticated with FlashSystem and obtained session token.")
+	log.V(2).Info("FlashSystem authentication successful, session token obtained")
```
rgolangh: please keep the old log message

cmd/vsphere-copy-offload-populator/internal/infinibox/infinibox.go
```diff
@@ -173,27 +192,35 @@ func NewInfiniboxClonner(hostname, username, password string, insecure bool) (In
...
 		clonner.arrayInfo.Model = sysInfo.Model
 		clonner.arrayInfo.Version = sysInfo.Version
+		log.V(2).Info("InfiniBox array info", "vendor", clonner.arrayInfo.Vendor, "product", clonner.arrayInfo.Product, "model", clonner.arrayInfo.Model, "version", clonner.arrayInfo.Version)
 	}
 
 	return clonner, nil
 }
 
 func (c *InfiniboxClonner) ResolvePVToLUN(pv populator.PersistentVolume) (populator.LUN, error) {
+	log := c.log
```
rgolangh: do we need a local instance?

rgolangh: just to make is clear, why isn't c.log enough?

amitosw15: sounds cleaner ill change

cmd/vsphere-copy-offload-populator/internal/flashsystem/flashsystem.go
```diff
@@ -478,21 +483,21 @@ func (c *FlashSystemClonner) EnsureClonnerIgroup(hostName string, clonnerIdentif
 	if len(fcWWPNs) > 0 {
 		existingFCHosts, err := c.findAllHostsByWWPNs(fcWWPNs)
 		if err != nil {
-			klog.Warningf("Error searching for existing FC hosts: %v", err)
+			c.log.Error(nil, "error searching for existing FC hosts", "err", err)
```
rgolangh: See the logger interface, the first argument for Error method

cmd/vsphere-copy-offload-populator/internal/flashsystem/flashsystem.go
```diff
@@ -565,12 +572,16 @@ func (c *FlashSystemClonner) createNewHost(hostName string, identifiers []string
...
+			c.log.Info("host name conflict, using alternative name", "host", hostName)
 		}
 	}
 
 	// Create new host with unique WWPNs and name
-	klog.Infof("Creating NEW host '%s' with identifiers: %v (FC: %t)", hostName, identifiers, isFibreChannel)
+	protocol := "iSCSI"
+	if isFibreChannel {
+		protocol = "FC"
+	}
```
rgolangh: I'm fine with logging only the identifiers,  I rather not add specific log for the protocol

cmd/vsphere-copy-offload-populator/internal/flashsystem/flashsystem.go
```diff
@@ -784,43 +795,43 @@ func (c *FlashSystemClonner) cleanupEmptyHost(hostID string) {
 
 	mappingsBytes, status, err := c.api.makeRequest("POST", "/lshostvdiskmap", filterPayload)
 	if err != nil {
-		klog.Warningf("Failed to check host mappings before cleanup: %v", err)
+		c.log.Error(nil, "failed to check host mappings before cleanup", "host_id", hostID, "err", err)
```
rgolangh: same here

cmd/vsphere-copy-offload-populator/internal/logutil/logutil.go
```diff
@@ -0,0 +1,9 @@
+package logutil
```
rgolangh: I take the blame of creating fcutil with a bad name. It should have been just 'fc'
same here, lets make the package name 'logger' and then the factory function to 'New' so a new logger would just be called `logger.New(....)`

PR #6077, opened 2026-05-01
MTV-5244 | Add warning for missing IP addresses when preserveStaticIps

mnecas:
Issue: when migrating VM that MTV does not see any IP address we are not showing any warning in the status.

Resolves: MTV-5244


mnecas:
I was concidering to make the status critical but we have actully got feedback about it and we lowered the severity before so keeping as is.
Ref https://github.com/kubev2v/forklift/pull/1161

PR #6086, opened 2026-05-01
MTV-5251 | Fix warm migration task lookup to use vCenter client

mnecas:
Issue:
MTV-5091 switched getVM to use the vCenter client, but task operations still resolved to an ESXi host client, causing tasks created on vCenter to not be found on the ESXi host.

Fix:
Use vCenter client for all task and snapshot operations. Remove unused ESXi host client infrastructure from the vsphere Client.

Ref: https://issues.redhat.com/browse/MTV-5251
Resolves: MTV-5251


mnecas:
i was not able to reproduce it, waiting for QE confirmation

PR #6088, opened 2026-05-02
MTV-5251 | Use instance UUID instead of moref for VM lookup

mnecas:
Issue:
PR #5973 switched some parts to use vcenter api instead of esxi this
broke warm migraiton tasks, so we need to revert the changes. And to
properly fix MTV-5091 we will swtich search from bios uuid to instance
uuid.

Fix:
Use FindByUuid with instance UUID (config.instanceUuid) which is
globally unique and valid on both vCenter and ESXi. Add InstanceUUID
field to the VM model and collector.

Ref: https://issues.redhat.com/browse/MTV-5251
Resolves: MTV-5251


mnecas:
@coderabbitai ptal

yaacov (approved):
lgtm

PR #6140, opened 2026-05-04
MTV-5245 | Add registry-based Windows static IP firstboot scripts

mnecas:
Issue:
With the DHCP Client service disabled, applying static IPs via New-NetIPAddress in the firstboot network script can fail with "The RPC server is unavailable".

Fix:
Add ForkliftController feature_windows_registry_network_config into conversion pods. When enabled virt-v2v uses registry-based *-registry firstboot sources for static IP, duplicate route cleanup, and complementary multi-IP scripts; otherwise the existing PowerShell path is used.

Ref: https://issues.redhat.com/browse/MTV-5245

Resolves: MTV-5245

### DIFF
--- a/operator/.downstream_manifests
+++ b/operator/.downstream_manifests
@@ -271,6 +271,13 @@ spec:
                 - "true"
                 - "false"
                 type: string
+              feature_windows_registry_network_config:
+                description: 'Use registry-based network configuration scripts for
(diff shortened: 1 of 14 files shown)


mnecas:
@coderobbitai ptal again

yaacov (approved):

PR #6157, opened 2026-05-05
MTV-5068 | Derive GOVMOMI_HOSTNAME from provider URL

amitosw15:
## Summary
- Fix GOVMOMI_HOSTNAME being unset when the provider
  secret has no `url` key (the common case), causing
  copy-offload populator pods to fail
- Remove nil-pointer bug in the `url` parse error path
- Always derive hostname from `Provider.Spec.URL`

## Test plan
- [ ] Existing `mergeSecrets` test updated to assert
      GOVMOMI_HOSTNAME is always present
- [ ] New test: secret without `url` key still produces
      correct GOVMOMI_HOSTNAME, GOVMOMI_USERNAME,
      GOVMOMI_PASSWORD
- [ ] `go test ./pkg/controller/plan/adapter/vsphere/...`
      passes (104/104)

### DIFF
--- a/pkg/controller/plan/adapter/vsphere/builder.go
+++ b/pkg/controller/plan/adapter/vsphere/builder.go
@@ -1975,13 +1975,6 @@ func (r *Builder) mergeSecrets(migrationSecret, migrationSecretNS, storageVendor
 
 	for key, value := range dst.Data {
 		switch key {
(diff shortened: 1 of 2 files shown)


amitosw15:
duplicate roy's pr, closing

PR #6158, opened 2026-05-05
MTV-5270 | fix: Powerstore Map volumes to host group when ESXi host is a member

amitosw15:
## Summary

- PowerStore rejects volume attach/detach on individual hosts that belong to a host group. The populator now checks `HostGroupID` and uses the host group API when the host is a member.
- `UnMap` re-fetches the host via `GetHostByName` to pick up current group membership instead of relying on a stale mapping context value.
- Added unit tests covering both standalone and host-group paths for `Map` and `UnMap`, including idempotency and error cases.

## Test plan

- [ ] `go test ./internal/powerstore/... -v` — all 8 new tests pass
- [ ] Run a copy-offload migration against a PowerStore cluster where the ESXi host belongs to a host group
- [ ] Verify standalone host migrations still work (no host group)

### DIFF
--- a/cmd/vsphere-copy-offload-populator/internal/powerstore/powerstore.go
+++ b/cmd/vsphere-copy-offload-populator/internal/powerstore/powerstore.go
@@ -235,36 +235,57 @@ func (p *PowerstoreClonner) Map(initiatorGroup string, targetLUN populator.LUN, 
 		hostName = mappingContext[esxRealHostNameKey].(string)
 	}
 
-	// Get the host by the real PowerStore host name
 	host, err := p.Client.GetHostByName(ctx, hostName)
 	if err != nil {
 		return targetLUN, fmt.Errorf("failed to find host for host name %s: %w", hostName, err)
-	}
-
-	hostID := host.ID
-
-	// idempotency: skip attach if already mapped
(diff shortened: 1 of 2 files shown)


borod108:
In some other providers we create host groups and put individual hosts in them, then connect to said host group, what is there a reason we do not do it for powerstore?

Also in EnsureClonnerIgroup - it looks like a bug to me - we search for a host with hba/wwpn, then if we do not find it we search for a host with a name matching the initiatorGroup and select it if it exists. Which would not work because it does not have the right iqn/wwpn  - am I getting this right? I do not want to fix it here since that would be a scope creep but I do want to make sure I understand what is going on with groups.

amitosw15:
@borod108  Thanks for the review!
On the host group question: PowerStore's model is different from ontap. A host can be mapped to  volumes either directly, or as a member of a host group - but these are mutually exclusive. 
So when we need to map a volume to the ESXi host, we have to use the same method that host is already using for its existing volume mappings. If it's part of a host group, we must map to the group; if it's standalone, we map directly to the host. 
This PR detects which mode the host is in and maps accordingly.
 
On the second point - you're right, that looks like a bug. I'll open a separate ticket for it, thanks for catching it.

PR #6159, opened 2026-05-05
MTV-3736 | RDM and independent disk concerns should not block migration

borod108:
Change RDM and independent disk concerns from Critical to Warning so they no longer block migration plans. When copy-offload (XCOPY) is active in the storage mapping, suppress these concerns entirely since XCOPY can handle both disk types.

- Rego policies: change category from "Critical" to "Warning"
- Plan validation: add RDMAndIndependentDiskConcerns() to Validator interface
- vSphere validator: detect RDM/independent disks, suppress when copy-offload active
- validation.go: surface RDMDiskWarning and IndependentDiskWarning as Warn conditions
- Unit tests for all new validation logic

Resolves: MTV-3736

### DIFF
--- a/pkg/controller/plan/adapter/vsphere/validator_test.go
+++ b/pkg/controller/plan/adapter/vsphere/validator_test.go
@@ -112,8 +112,6 @@ func (m *mockInventory) Find(resource interface{}, ref ref.Ref) error {
 			res.ToolsStatus = ToolsNotInstalled // Should be ignored when powered off
 		case "tools_unmanaged":
 			res.ToolsVersionStatus = GuestToolsUnmanaged
-		case "missing_from_inventory":
-			return base.NotFoundError{}
 		}
 	}
 	return nil
@@ -620,6 +618,7 @@ var _ = Describe("vsphere validation tests", func() {
 		Entry("should warn on consolidation needed", true),
 		Entry("should not warn when consolidation is not needed", false),
 	)
+
 })
 
 func createPlan() *v1beta1.Plan {
--- a/pkg/controller/plan/validation.go
+++ b/pkg/controller/plan/validation.go
@@ -108,6 +108,8 @@ const (
 	RestrictedPodSecurity           = "RestrictedPodSecurity"
 	NetMapDestinationNADNotValid    = "NetMapDestinationNADNotValid"
 	VMCriticalConcerns              = "VMCriticalConcerns"
+	RDMDiskWarning                  = "RDMDiskWarning"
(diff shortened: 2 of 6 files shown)


pkg/controller/plan/adapter/base/doc.go
```diff
@@ -263,6 +263,12 @@ type Validator interface {
 	PVCNameTemplate(vmRef ref.Ref, pvcNameTemplate string) (bool, error)
 	// Validate guest tools installation and status (e.g., VMware Tools, VirtIO drivers).
 	GuestToolsInstalled(vmRef ref.Ref) (ok bool, err error)
+	// Check for RDM and independent disks.
+	// Returns whether these disk types are present AND copy-offload is NOT active.
+	// When copy-offload is active, both return false (no concern needed).
+	// For warm migrations, only physical compatibility mode RDM disks are flagged;
+	// virtual RDM is not a concern for warm migration.
+	RDMAndIndependentDiskConcerns(vmRef ref.Ref) (hasRDM bool, hasIndependent bool, err error)
```
mnecas: @coderabbitai instead of adding new method could we use typecast to get all informaiton from VM similar to https://github.com/kubev2v/forklift/commit/a49eb4d5265416e542f20697fbe56318eebe037e
```
			if vsphereVM, ok := v.(*vsphere.VM); ok {
```

pkg/controller/plan/adapter/vsphere/validator_test.go
```diff
@@ -112,8 +112,6 @@ func (m *mockInventory) Find(resource interface{}, ref ref.Ref) error {
 			res.ToolsStatus = ToolsNotInstalled // Should be ignored when powered off
 		case "tools_unmanaged":
 			res.ToolsVersionStatus = GuestToolsUnmanaged
-		case "missing_from_inventory":
-			return base.NotFoundError{}
```
mnecas, line 116: @borod108 Why to remove this?

mnecas (approved):

borod108:
RE: [the question about removing `missing_from_inventory`](https://github.com/kubev2v/forklift/pull/6159#discussion_r3256094175):

The `missing_from_inventory` case inside the `switch ref.Name` was unreachable dead code — it's already handled by the `if ref.Name == "missing_from_inventory"` check at [line 85](https://github.com/kubev2v/forklift/blob/f4a59505c9481bca44be57efeba0317b82d5ca8c/pkg/controller/plan/adapter/vsphere/validator_test.go#L85) which returns `base.NotFoundError{}` before the VM is populated and before the switch is reached. So the case could never execute.

PR #6173, opened 2026-05-06
MTV-5260 | OSP fix race between image upload and volume snapshot

mnecas:
Move volume snapshot creation from createVmSnapshotImage into ensureSnapshotsFromVolumes so snapshots are only created after the VM task_state clears. Extract task_state from Nova server response and gate snapshot creation on it being empty. Handle error status snapshots by deleting them for retry. Reorder PopulatorVolumes after PreTransferActions readiness check to prevent populator pods from downloading images still uploading.

Ref: https://issues.redhat.com/browse/MTV-5260
Resolves: MTV-5260

### DIFF
--- a/pkg/controller/plan/migration.go
+++ b/pkg/controller/plan/migration.go
@@ -807,6 +807,11 @@ func (r *Migration) execute(vm *plan.VMStatus) (err error) {
 				}
 			}
 
+			if !ready {
+				r.Log.Info("PreTransferActions hook isn't ready yet")
+				return
+			}
+
 			if r.builder.SupportsVolumePopulators() {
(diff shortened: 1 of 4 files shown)


pkg/controller/plan/migration.go
```diff
@@ -817,6 +807,11 @@ func (r *Migration) execute(vm *plan.VMStatus) (err error) {
 				}
 			}
 
+			if !ready {
+				r.Log.Info("PreTransferActions hook isn't ready yet")
```
gwencasey96, line 811: can you add VM information for easier debugging?

mnecas:
closing as stale

PR #6179, opened 2026-05-06
MTV-5245 | Fix static IP scripts not injected

mnecas:
Fix static IP firstboot script injection when neither legacy drivers nor registry config was set.

Ref: https://issues.redhat.com/browse/MTV-5245
Resolves: MTV-5245

### DIFF
--- a/pkg/virt-v2v/customize/customize.go
+++ b/pkg/virt-v2v/customize/customize.go
@@ -321,8 +321,7 @@ func (c *Customize) addWinFirstbootScripts(cmdBuilder utils.CommandBuilder) erro
 		initPath = filepath.Join(windowsScriptsPath, "9999-run-mtv-ps-scripts-legacy.bat")
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/virt-v2v/customize/customize.go
```diff
@@ -321,8 +321,7 @@ func (c *Customize) addWinFirstbootScripts(cmdBuilder utils.CommandBuilder) erro
 		initPath = filepath.Join(windowsScriptsPath, "9999-run-mtv-ps-scripts-legacy.bat")
 	}
 
-	staticIPFirstbootScripts := c.appConfig.VirtIoWinLegacyDrivers != "" || c.appConfig.WindowsRegistryNetworkConfig
```
solenoci, line 324: Why was this condition required in the first place?

mnecas: it was not, brain not working at that time

solenoci (approved):

mnecas:
update branch name and commit so swtiching to other PR https://github.com/kubev2v/forklift/pull/6179

PR #6180, opened 2026-05-06
MTV-5316 | Fix static IP scripts not injected

mnecas:
Fix static IP firstboot script injection when neither legacy drivers nor registry config was set.

Ref: https://issues.redhat.com/browse/MTV-5316
Resolves: MTV-5316


yaacov (approved):
lgtm

PR #6198, opened 2026-05-07
MTV-5316 | Fix secondary static IP loss on Windows VM migration

Hazanel:
The complementary IP script now skips any IP
already configured, and adds secondaries with -SkipAsSource.
The primary-IP setter is gated to legacy/registry mode only.

Ref: https://redhat.atlassian.net/browse/MTV-5316
Resolves: MTV-5316


Hazanel:

Hazanel:
@coderabbitai, ptal

PR #6205, opened 2026-05-07
MTV-5327 | Fix broken Jinja2 template and exclude missing deep-inspection image

yaacov:
Restore missing `value: "true"` and two `{% endif %}` closing tags in deployment-controller.yml.j2. Commit 12adee73b (MTV-3666) accidentally deleted these lines from the FEATURE_WINDOWS_REGISTRY_NETWORK_CONFIG block when adding FEATURE_USE_CONVERSION_CR, leaving two unclosed {% if %} blocks that caused the operator Ansible reconcile to fail with a Jinja2 "Unexpected end of template" error.

Also exclude deep-inspection-image from build-all-images and push-all-images until its Containerfile exists, preventing build failures.

Ref: https://redhat.atlassian.net/browse/MTV-5327

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -480,6 +480,7 @@ push-deep-inspection-image: build-deep-inspection-image ## Push forklift-deep-in
 	fi
 
 build-all-images: ## Build all container images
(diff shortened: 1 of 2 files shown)


solenoci (approved):
lgtm

PR #6217, opened 2026-05-08
Add cursor skill grill-me

mnecas:
ref: https://www.aihero.dev/my-grill-me-skill-has-gone-viral


mrnold:
It looks like there's a grill-with-docs now that can pick up more context automatically: https://www.aihero.dev/grill-with-docs

PR #6262, opened 2026-05-11
MTV-5359 | Restore KVM device request on conversion pods

yaacov:
The MTV-3666 refactor moved pod construction into
pkg/controller/conversion/builder.go but dropped the setKvmOnPodSpec call that added devices.kubevirt.io/kvm resource requests/limits and the kubevirt.io/schedulable node selector. Without /dev/kvm the virt-v2v appliance falls back to QEMU emulation, which is significantly slower.

Add a RequestKVM flag to PodConfig, set it based on source provider type (vSphere, OVA, HyperV) and the VIRT_V2V_DONT_REQUEST_KVM setting, and apply it in the builder. The flag is also propagated through the
Conversion CR PodSettings so both the direct-pod and CR-based paths get KVM.

Ref: https://redhat.atlassian.net/browse/MTV-5359
Resolves: MTV-5359

### DIFF
--- a/pkg/apis/forklift/v1beta1/conversion.go
+++ b/pkg/apis/forklift/v1beta1/conversion.go
@@ -133,6 +133,10 @@ type PodSettings struct {
 	// GenerateName prefix for the managed pod.
 	// +optional
 	GenerateName string `json:"generateName,omitempty"`
+	// Whether the pod needs a KVM device and kubevirt.io/schedulable node selector.
+	// +optional
+	// +kubebuilder:default:=true
+	RequestKVM bool `json:"requestKVM,omitempty"`
 }
 
 // ConversionSpec defines the desired state of Conversion.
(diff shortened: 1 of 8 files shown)


solenoci:
https://github.com/kubev2v/forklift/pull/5882/changes/184addd0612157bbcc8c24a6907105d3baef2036 please follow this commit to revert changes

pkg/controller/conversion/builder.go
```diff
@@ -157,6 +157,12 @@ func (b *Builder) GetVirtV2vPodSpec(vm *plan.VMStatus, volumes []core.Volume, vo
 		podNodeSelector = make(map[string]string)
 		maps.Copy(podNodeSelector, cfg.PodNodeSelector)
 	}
+	if cfg.RequestKVM {
```
solenoci: let's extract the kvm request to one separate method

yaacov: @solenoci one method now, please re-review

pkg/apis/forklift/v1beta1/conversion.go
```diff
@@ -133,6 +133,10 @@ type PodSettings struct {
 	// GenerateName prefix for the managed pod.
 	// +optional
 	GenerateName string `json:"generateName,omitempty"`
+	// RequestKVM adds devices.kubevirt.io/kvm to the conversion pod
+	// so that virt-v2v uses hardware virtualisation instead of emulation.
+	// +optional
```
solenoci, line 137: Set the default value to true

yaacov: upd at ed to default true

solenoci (approved):

PR #6283, opened 2026-05-12
MTV-4593 | Add missing StorePowerState phase to EC2 migration itinerary

yaacov:
The EC2 cold migration itinerary was missing the PhaseStorePowerState phase that all other providers include before PhasePowerOffSource. Without it, RestorePowerState is never recorded on the VMStatus, so determineRunStrategy sees an empty value and defaults to RunStrategyHalted regardless of the source instance's actual power state.

Add PhaseStorePowerState between PhasePreHook and PhasePowerOffSource in the EC2 itinerary, consistent with the base cold/warm itineraries.

Ref: https://redhat.atlassian.net/browse/MTV-4593
Resolves: MTV-4593


yaacov:

Hazanel (approved):
lgtm

PR #6284, opened 2026-05-12
chore(upstream): skip --no-fstrim on upstream virt-v2v images

yaacov:
The --no-fstrim flag is a RHEL-only downstream patch that does not exist in upstream virt-v2v (CentOS Stream or Fedora). The wrapper was unconditionally passing this flag, causing migrations to fail with "unrecognized option" on upstream images.

Detect the OS at startup by reading /etc/os-release and only pass --no-fstrim when running on RHEL (ID=rhel). On upstream images the flag is silently omitted; fstrim runs by default and logs a non-fatal warning if it fails.

Resolves: none

### DIFF
--- a/pkg/virt-v2v/config/variables.go
+++ b/pkg/virt-v2v/config/variables.go
@@ -1,6 +1,7 @@
 package config
 
 import (
+	"bufio"
 	"encoding/json"
 	"flag"
 	"fmt"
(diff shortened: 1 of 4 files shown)


yaacov:

Hazanel (approved):
lgtm

PR #6279, opened 2026-05-12
MTV-5390 | Use forked infinibox-csi-driver with pagination fix

amitosw15:
## Summary

- Switch the `infinibox-csi-driver` dependency to the forked version (`amitosw15/infinibox-csi-driver`) that includes pagination support for `GetAllHosts`. The upstream driver fails when the Infinidat array returns paginated results, causing host lookups to silently return incomplete data.
- This is a temporary `replace` directive until the fix is merged upstream.

## Test plan

- [ ] Verify Infinidat storage provider discovery works with arrays that have more hosts than a single page
- [ ] Confirm no regression in other storage provider flows (Dell, NetApp)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #6299, opened 2026-05-13
[release-2.11] MTV-5390 | Use forked infinibox-csi-driver with pagination fix

amitosw15:
**Backport:** https://github.com/kubev2v/forklift/pull/6279

## Problem

The upstream infinibox-csi-driver fails when the Infinidat array returns paginated results from GetAllHosts, causing host lookups to return incomplete data.

## What it achieves

Switches the `infinibox-csi-driver` dependency to a forked version (`amitosw15/infinibox-csi-driver`) that adds pagination support for `GetAllHosts`. This is a temporary `replace` directive until the fix is merged upstream. `go mod tidy` also cleaned up stale unused transitive deps (sonic, gin, etc.).

## Test plan

- [ ] Verify Infinidat storage provider discovery works with arrays that have more hosts than a single page
- [ ] Confirm no regression in other storage provider flows (Dell, NetApp)


amitosw15:
i know it seems a lot of files, it's because of stale dependencies. when i did go mod tidy to the 2.11 branch without any changes i got the same changes - not related to us.

PR #6326, opened 2026-05-13
MTV-5419 MTV-5429 MTV-5430 MTV-5431 | Fix for deep inspection XFS

solenoci:
Issue: deep inspection also needs to use variable image based on XFS compat setting.

Fix: add XFS compatible deep inspection image

Resolves: MTV-5419 MTV-5429 MTV-5430 MTV-5431


solenoci:
I suspect that golangci-lint fails because of cache, the affected code was fixed.

Kyl-Bempah (approved):
LGTM from a Delivery standpoint

PR #6344, opened 2026-05-13
MTV-5237 | Validate image with pattern

solenoci:
Issue: forklift-controller currently call panic() when an invalid
setting is set.

Fix: If user defined value is empty string, use the default.

Resolves: MTV-5237

Signed-off-by: Stefan Olenocin <solenoci@redhat.com>

### DIFF
--- a/operator/roles/forkliftcontroller/templates/api/deployment-forklift-api.yml.j2
+++ b/operator/roles/forkliftcontroller/templates/api/deployment-forklift-api.yml.j2
@@ -28,7 +28,7 @@ spec:
       serviceAccountName: forklift-api
       containers:
         - name: {{ api_container_name }}
-          image: {{ api_image_fqin }}
+          image: "{{ api_image_fqin or lookup('env', 'API_IMAGE') or lookup('env', 'RELATED_IMAGE_API') }}"
           imagePullPolicy: {{ image_pull_policy }}
           ports:
           - name: api
--- a/operator/roles/forkliftcontroller/templates/cli-download/deployment-cli-download.yml.j2
+++ b/operator/roles/forkliftcontroller/templates/cli-download/deployment-cli-download.yml.j2
@@ -21,7 +21,7 @@ spec:
     spec:
       containers:
         - name: "{{ cli_download_container_name }}"
-          image: "{{ cli_download_image_fqin }}"
+          image: "{{ cli_download_image_fqin or lookup('env', 'CLI_DOWNLOAD_IMAGE') or lookup('env', 'RELATED_IMAGE_CLI_DOWNLOAD') }}"
           ports:
             - containerPort: 8443
               protocol: TCP
--- a/operator/roles/forkliftcontroller/templates/controller/deployment-controller.yml.j2
+++ b/operator/roles/forkliftcontroller/templates/controller/deployment-controller.yml.j2
@@ -47,13 +47,13 @@ spec:
         - name: KUBEVIRT_CLIENT_GO_SCHEME_REGISTRATION_VERSION
           value: "v1"
         - name: VIRT_V2V_IMAGE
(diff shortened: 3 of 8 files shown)


solenoci:
don't know if we want to add this to all other images

solenoci:
@yaacov could you PTAL?

yaacov:
> @yaacov could you PTAL?

If you want validations, lgtm

my concern is that we don't need validations for "read-only" "don't touch unless you know what you are doing" settings.
adding validations mean we actually expect users who don't have a very good reason to touch this settings.

yaacov (approved):
lgtm

nit pick - why you want validations for this settings, my concern is that validation are another moving part that will need to be maintained, and may break.

solenoci:
We can:
- either use default value when "" but we would still fail on an invalid image fqin
- or add validations beforehand so that "" or "fail" can't happen
- we can add those validations in the settings pkg or to the FC CR so I thought that having it in FC CR would propagate sooner

yaacov (approved):

yaacov (approved):
nitpick:
if you find a nice way to define the defaults as vars instead of directly accessing the env vars it will be nice, but not a blocking issue for this pull request

operator/roles/forkliftcontroller/templates/api/deployment-forklift-api.yml.j2
```diff
@@ -28,7 +28,7 @@ spec:
       serviceAccountName: forklift-api
       containers:
         - name: {{ api_container_name }}
-          image: {{ api_image_fqin }}
+          image: "{{ api_image_fqin or lookup('env', 'API_IMAGE') or lookup('env', 'RELATED_IMAGE_API') }}"
```
mnecas, line 31: @coderabbitai could we move this to the default so we don't need to think about this check?

yaacov: the edge case mtv-5237 is about is the case of setting the value to "" instead of nil (e.g. not setting the value in the CR)
@coderabbitai , when setting the value to "" in the forkliftcontroller CR what value will be set in the controller, what is the best way to fallback to defaults in that case e.g. how can we make "" behave like nil in this case?

yaacov: does it make sense to create default value variables, and use them in case user set nil or "", how will that look like ?

yaacov: @coderabbitai ^^ ?

yaacov: @coderabbitai 

```
api_image_fqin_default: "{{ lookup('env', 'API_IMAGE') or lookup('env', 'RELATED_IMAGE_API') }}"
api_image_fqin: "{{ api_image_fqin_default }}"   # fallback to default when CR doesn't set it
```
 when using this pattern, do we neet the `api_image_fqin: "{{ api_image_fqin_default }}"` line ?
 will the 
`image: "{{ api_image_fqin or api_image_fqin_default }}"`
work without explicitly setting the api_image_fqin var ?

yaacov: @mnecas @solenoci @coderabbitai, hi

I like the option of using:
`api_image_fqin_default: "{{ lookup('env', 'API_IMAGE') or lookup('env', 'RELATED_IMAGE_API') }}"`
with:
`image: "{{ api_image_fqin | default(api_image_fqin_default, true) }}"
`

My reasoning is that the code is self explanatory, it is clear that we set default ( the var name has `_default` ) and it is clear we override in the case of "" and nil ( we explicitly use the method `default( , true)` ) , WDYT?

PR #6414, opened 2026-05-17
MTV-5323 | Preserve advanced Windows network settings during migration

Hazanel:
Add a REGF registry parser to extract per-adapter network settings
from the Windows SYSTEM hive after virt-v2v conversion.
MAC addresses are resolved directly from the registry, eliminating any dependency on
provider-supplied NIC MAC lists.

Settings preserved: 

- InterfaceMetric
- DNS registration,
- NetBIOS over TCP/IP,
- LanmanServer start type, and File & Printer Sharing bindings.

Ref: https://redhat.atlassian.net/browse/MTV-5323
Resolves: MTV-5323

### DIFF
--- a/cmd/virt-v2v/entrypoint.go
+++ b/cmd/virt-v2v/entrypoint.go
@@ -3,10 +3,12 @@ package main
 import (
 	_ "embed"
 	"fmt"
+	"log/slog"
 	"os"
 
 	"github.com/kubev2v/forklift/pkg/virt-v2v/config"
 	"github.com/kubev2v/forklift/pkg/virt-v2v/conversion"
+	"github.com/kubev2v/forklift/pkg/virt-v2v/customize/advancednet"
 	"github.com/kubev2v/forklift/pkg/virt-v2v/server"
 	utils "github.com/kubev2v/forklift/pkg/virt-v2v/utils"
 )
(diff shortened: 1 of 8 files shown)


Hazanel:
@coderabbitai please take a look and also share your thought whether it's a good approach or you think we can come up with a better way

Hazanel:
Hey @coderabbitai 

Thanks for the thorough review:

🔴 virt-cat --format=raw hardcoded

By the time this code runs, virt-v2v has already completed conversion — all disks are raw at this point regardless of the source format (vSphere, OVA, etc.). The --format=raw flag is correct and intentional.

🔴 Only first disk is read — no fallback

Windows always places C:\Windows\System32\config\SYSTEM on the boot disk, which is Disks[0]. Added a comment documenting this assumption explicitly. Iterating through all disks would add complexity with no practical benefit — if the boot disk isn't Disks[0], we have bigger problems than advanced network settings.

🟡 injectComplementryStaticIPTemplate not refactored to use renderTemplate

Good catch. Refactored — extended renderTemplate with a variadic funcMaps ...template.FuncMap parameter so the custom funcs (lower, add, len, formatIPs, formatDNS) pass through cleanly.

🟡 PowerShell JSON injection — single-quote assumption

The JSON is produced by json.MarshalIndent from our own Go structs (AdvancedNetSettings), which contain only MAC addresses (AA:BB:CC:DD:EE:FF), GUIDs ({...}), and uint32 values. None of these can contain single quotes.

mnecas (changes requested):
no hivex parsing inside the v2v wrapper plust there is https://github.com/libguestfs/hivex

if this is the cost of the RFE, we should not support it

Hazanel:
Can we use huvex-libs in the wrapper or not at all?
Another option is to use hivexml to dump the relevant subtree as XML in a single call, then parse the XML in Go with encoding/xml. One process spawn, structured output, and Go's XML parser is stdlib -- no custom binary parsing at all.

WDYT?

Hazanel:
After discussing our options here on how to reliably;y implement that  with @Martin Necas , We decided it’s not something MTV should be responsible , hence we close as “Won’t Do”

PR #6418, opened 2026-05-17
MTV-5530 |  static IP preservation on RHEL 9/10+ after migration

Hazanel:
On RHEL 9/10+, predictable network interface naming (e.g. enp1s0)
overrides udev NAME= rename rules to e.g eth0.
Since NM keyfiles reference interface-name=eth0, NetworkManager never activates the
connection and the VM loses its static IP.

Add generate_link_files() to network_config_util.sh which writes
systemd .link files into /etc/systemd/network/.  A .link [Match]
by MAC with [Link] Name=<original> tells systemd-udevd to rename
the interface before NetworkManager sees it, so existing NM
connections activate unchanged.  The .link file is harmless on
older distros where the udev rule already works.

Ref: https://redhat.atlassian.net/browse/MTV-5530
Resolves: MTV-5530

image


Hazanel:
@coderabbitai - please check the fixes and in general go over and suggest whether you think there is a better approach

Hazanel:
Thanks @coderabbitai for the thorough review, please check the fixes.

mnecas (changes requested):
@Hazanel editing the source config file can be bit complicated. I would rather lean to creating `.link` file for the NM.  https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/10/html/configuring_and_managing_networking/implementing-consistent-network-interface-naming

wdyt @yaacov ?

yaacov:
> wdyt @yaacov ?

my concern is that we are solving missing support for specific OSs in virt-v2v by adding ad-hock scripts.
IIRC, we were suppose to move the missing support to virt-v2v where is can be tested and QE'd using the infrastructure they have.
when we add support using our scripts we don't have the infrastructure and CI to test out scripts using a known matrix of supported OSs to make sure we didn't miss some configuration we should support.

> editing the source config file can be bit complicated.

yes, specifically for this issue, if a less intrusive fix work ( e.g. adding a separate named file ) my vote is not to edit an existing file, but to add a separate file with the correct priority.

pkg/virt-v2v/customize/scripts/rhel/run/network_config_util.sh
```diff
@@ -523,6 +524,37 @@ udev_from_ifquery() {
...
 }
 
+# Generate systemd .link files to ensure interface renaming works on all distros.
+# On RHEL 10+ predictable naming overrides raw udev NAME= rules, but a .link
+# file with a lower priority number takes precedence over the default policy.
+# On older distros the .link file is redundant with the udev rule but harmless.
+generate_link_files() {
+    mkdir -p "$SYSTEMD_LINK_DIR"
+
+    if [ ! -f "$UDEV_RULES_FILE" ] || [ ! -s "$UDEV_RULES_FILE" ]; then
```
mnecas: please add check if system-connections/$name.nmconnection
1. exists
2. does not contian the MAC

please add validations for this aswell ^^ that we don't create the file when not needed

I'm also thinking that for those cases we could skip the udev rule, but that should be smaller detial

mnecas (changes requested):

mnecas (approved):
lgtm

PR #6465, opened 2026-05-18
MTV-5301 | Fix duplicate default routes in registry network config mode

Hazanel:
When feature_windows_registry_network_config is enabled, the remove_duplicate_persistent_routes-registry script used Remove-NetRoute -PolicyStore PersistentStore to clean up redundant routes. This also cleared the DefaultGateway registry key set by the network-config-registry script, leaving the VM with no gateway after reboot.

Fix: when the gateway already exists in the interface-level DefaultGateway registry key, remove only the PersistentRoutes registry entry directly via Remove-ItemProperty, preserving the DefaultGateway value.

Ref: https://redhat.atlassian.net/browse/MTV-5301
Resolves: MTV-5301


Hazanel:
@coderabbitai, I addressed your comments, thanks for that.
Any other thoughts/concerns?

Hazanel:
Thanks for the feedback @coderabbitai, I certianly addressed also the minor issues.

mnecas (approved):

PR #6477, opened 2026-05-18
MTV-5556 | Solves the issue that warm precopy on Pure FlashArray vVol falls back to slow vmkfstools clone instead of array offload

waliu-ps:
When a precopy snapshot is active, the top-of-chain backing's filename is the child disk (foo-000001.vmdk) while forklift passes the base path (foo.vmdk). The old top-only path comparison missed this and the populator fell through to the VMDK/vmkfstools fallback. vmkfstools still produced correct bytes (it opens foo.vmdk on the host and VASA serves the snap-vVol), but at host-mediated speed.

Walk the Parent chain and return the BackingObjectId of the node whose filename matches the caller's vmdkPath. Warm precopy now matches at the parent, identifies the snap-vVol, and copies it via Pure REST.

## Release Notes

* **Bug Fixes**
  * Enhanced vSphere volume detection to properly traverse disk backing chains and accurately match disk paths in snapshot scenarios
  * Improved VVol identification for more reliable volume population operations

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubev2v/forklift/pull/6477?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


waliu-ps:
@coderabbitai review

mnecas:
@mrnold could you ptal?

mrnold:
I think this looks okay, assuming the Code Rabbit comments can be cleaned up.

borod108:
@waliu-ps can you please cleanup the coderabbit comments and we can proceed with merge? or if you think they are misplaced, please comment that in reply to them.

cmd/vsphere-xcopy-volume-populator/internal/vmware/client.go
```diff
@@ -247,6 +243,25 @@ func (c *VSphereClient) GetVMDiskBacking(ctx context.Context, vmId string, vmdkP
...
+// filename matches vmdkPath, or nil if none does.
+//
+// While a snapshot is active the chain is "child disk (foo-000001.vmdk) ->
+// parent (foo.vmdk)", and VASA reports a different BackingObjectId per level:
+// child -> base vVol (current writes), parent -> snap-vVol (frozen). The
+// caller's vmdkPath picks which level matches.
+func findMatchingFlatBacking(b *types.VirtualDiskFlatVer2BackingInfo, normalizedPath, vmdkPath string) *types.VirtualDiskFlatVer2BackingInfo {
+	for cur := b; cur != nil; cur = cur.Parent {
+		fn := strings.ToLower(cur.FileName)
+		if strings.Contains(fn, normalizedPath) ||
```
snehal-pure: Is "normalizedPath" guaranteed to be in lower case or should it also be processed as lowerCase for comparison?

waliu-ps: normalizedPath was set lowercase before. It's line 183 normalizedPath := strings.ToLower(vmdkPath)

waliu-ps:
@mnecas @borod108 I have addressed the coderabbit comments in the 2nd commit. Could you please review? Let me know if there is anything else I need to do.

rgolangh:
@mnecas @borod108 we are going to have the same resolution problem with other vendors?

mnecas (changes requested):
This is against release-2.10, please change the target branch to main

waliu-ps:
@mnecas I have changed the target to main. Could you please take a look?

mnecas (approved):

PR #6541, opened 2026-05-19
chore(upstream-ci): add CLI release artifacts workflow

yaacov:
Add automation to attach cross-compiled kubectl-mtv binaries to GitHub Releases. On release creation, a new workflow builds archives for all supported platforms (linux, darwin, windows) with SHA-256 checksums and uploads them as release assets.

- .github/workflows/release-cli.yml: release-triggered workflow
- cmd/kubectl-mtv/Makefile: dist-* targets for per-platform archives
- cmd/kubectl-mtv/.gitignore: ignore bin/ and dist/ build outputs
- Makefile: dist-cli target delegating to the hack script
- docs/enhancements/cli-release-artifacts.md: enhancement document

Resolves: none


yaacov:

.github/workflows/release-cli.yml
```diff
@@ -0,0 +1,87 @@
...
+          VERSION="${{ github.event.release.tag_name }}"
+          VERSION="${{ github.event.release.tag_name }}"
+          url="https://github.com/${{ github.repository }}/releases/download/${VERSION}/kubectl-mtv-${VERSION}-linux-amd64.tar.gz"
+          if curl -sSL --fail -I "$url" > /dev/null 2>&1; then
+            echo "Assets already exist, skipping build and upload"
+            echo "assets_exist=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "Assets do not exist, will build and upload"
+            echo "assets_exist=false" >> "$GITHUB_OUTPUT"
+          fi
```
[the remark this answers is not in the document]
yaacov: thank you, removed

cmd/kubectl-mtv/Makefile
```diff
@@ -46,5 +97,5 @@ update-kubectl-mtv:
 
 .PHONY: clean
 clean:
-	rm -rf bin/ vendor/
+	rm -rf bin/ dist/ vendor/
```
solenoci: can we use dist-clean phony instead of directly specifying dist dir?

cmd/kubectl-mtv/Makefile
```diff
@@ -1,3 +1,6 @@
+VERSION ?= $(shell grep '^RELEASE=' ../../build/release.conf 2>/dev/null | cut -d= -f2 || echo "v0.0.0")-dev
```
solenoci: release.conf contains only vX.Y, is that oky for your use case?

.github/workflows/release-cli.yml
```diff
@@ -0,0 +1,86 @@
...
+        if: steps.check_assets.outputs.assets_exist != 'true'
+        uses: softprops/action-gh-release@da05d552573ad5aba039eaac05058a918a7bf631
+        with:
+          files: cmd/kubectl-mtv/dist/*
+
+      - name: Wait for assets to be available
+        if: steps.check_assets.outputs.assets_exist != 'true'
+        run: |
+          echo "Waiting 30 seconds for GitHub to propagate release assets..."
+          sleep 30
```
solenoci, line 56: this seems like an arbitrary sleep time. is it enough?

solenoci (commented):
I have few NPs but overall lgtm

.github/workflows/release-cli.yml
yaacov: hmm, it's totally arbitrary, 30s worked for me in yaacov/kubectl-mtv so I think it may work here too, but IDK.
Since it's running once per release, ~2 weeks, it's ok to err on the longer wait side, we may need to play with it if it fails here.

cmd/kubectl-mtv/Makefile
yaacov: now using the dist-clean as pre requisite

cmd/kubectl-mtv/Makefile
yaacov: Thanks !, now using MTV_VERSION , wow we are in 5.0.0 🎉

cmd/kubectl-mtv/Makefile
```diff
@@ -1,3 +1,6 @@
+VERSION ?= $(shell grep '^MTV_VERSION=' ../../build/release.conf 2>/dev/null | cut -d= -f2 || echo "0.0.0")-dev
```
solenoci: NP: would you like to prepend "v" so we are consistent with the tagged artifacts? e.g. v5.0.0-dev....

docs/enhancements/cli-release-artifacts.md
```diff
@@ -0,0 +1,144 @@
...
+  reproducibility without network access during compilation.
+
+## Design Details
+
+### VERSION Resolution
+
+| Context    | Source                                         |
+|------------|------------------------------------------------|
+| CI         | `github.event.release.tag_name`                |
+| Local      | `RELEASE` from `build/release.conf` + `-dev` suffix |
```
solenoci: after changes to the grep to MTV_VERSION, this is no longer true

cmd/kubectl-mtv/Makefile
yaacov: i agree v is important, also remove the "-dev" the version number is a good indicator it's dev version

docs/enhancements/cli-release-artifacts.md
yaacov: yes it is, fixed

solenoci (approved):

PR #6544, opened 2026-05-19
MTV-5555 | Fix Windows firstboot scripts for Constrained Language Mode

Hazanel:
Replace .NET method calls blocked by PowerShell Constrained Language Mode (CLM) with CLM-compatible alternatives:

- Convert-PrefixToMask: replace [Convert]::ToInt32() binary-to-decimal conversion with a static CIDR-to-mask lookup table (all 5 scripts)
- Replace [math]::Round() with [int] cast (network-config-registry)
- Replace New-Object -ComObject WScript.Shell with cmd /c calls (network-config legacy)
- Replace [System.Collections.ArrayList] with plain arrays (preserve-complementary-ips-registry)

Tested on CLM-enabled Windows Server 2019 VM with 2 NICs and multiple static IPs. All scripts complete with exit code 0 and subnet masks are correctly written to the registry.

Ref: https://redhat.atlassian.net/browse/MTV-5555
Resolves: MTV-5555

### DIFF
--- a/pkg/virt-v2v/customize/scripts/windows/9999-network-config-registry.ps1.tmpl
+++ b/pkg/virt-v2v/customize/scripts/windows/9999-network-config-registry.ps1.tmpl
@@ -6,16 +6,21 @@
 $inputString = "{{.InputString}}"
 $adapterWaitMinutes = 5
 
+# Lookup table avoids [Convert]::ToInt32() which is blocked in Constrained Language Mode
+$PrefixToMaskTable = @(
+    "0.0.0.0",         "128.0.0.0",       "192.0.0.0",       "224.0.0.0",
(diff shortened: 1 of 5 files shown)


Hazanel:
The PrefixToMaskTable and Convert-PrefixToMask are intentionally duplicated across scripts. 
I attempted extracting them into a shared `mtv-network-utils.ps1` file and dot-sourcing it (both from within scripts and via the bat runner using `-Command "& { . '...utils.ps1'; & '%%~f' }")`, but dot-sourcing fails silently under Constrained Language Mode , the functions are never defined in the caller's session, resulting in empty subnet masks. Inlining is the only reliable approach in CLM.

mnecas (approved):
lgtm

PR #6547, opened 2026-05-19
MTV-5328 | ocp: use allocated capacity when creating target volumes for block sources

mansam:
Some block storage backings might create a volume that is larger than the requested size. (for instance, powerflex creating volumes in 8Gi increments)

To accommodate that, use the source capacity instead of the requested size when creating target volumes for block sources.


mnecas (approved):
interesting interesting, lgtm

PR #6548, opened 2026-05-19
MTV-5082 | ova: skip unreadable paths when walking appliances

mansam:
If the file walker hit an error then the entire scan would be aborted, preventing any appliances from being surfaced in the inventory.

Fixes the problem by only failing if the root of the appliance catalog is unreadable.

Resolves: MTV-5082


mnecas (approved):
lgtm

PR #6549, opened 2026-05-20
MTV-5563 | Fix IP prefix matching in static IP udev rule generation

krcmarik:
Add word boundary (\b) assertions to grep patterns in network_config_util.sh that match IP addresses in guest network configuration files. Without these boundaries, an IP like 192.168.1.1 incorrectly matches 192.168.1.10 or 192.168.1.100, potentially assigning the wrong interface name in the generated udev rule.

Fixed patterns:
- udev_from_ifcfg: IPADDR matching (also adds IPADDR0/1 support)
- udev_from_nm: NetworkManager keyfile address matching
- udev_from_netplan: netplan get and systemd-networkd fallback
- udev_from_ifquery: Debian/Ubuntu ifquery matching

Added test fixtures with overlapping IPs (192.168.1.1 vs 192.168.1.10) for ifcfg and NetworkManager formats to prevent regression.

Resolves: MTV-5563

Signed-off-by: Marian Krcmarik <mkrcmari@redhat.com>


krcmarik:
@coderabbitai review

mnecas (approved):
lgtm

PR #6586, opened 2026-05-20
MTV-5437 | DI: orphaned vSphere snapshot after Conversion CR failure

solenoci:
Issue: When a standalone DeepInspection Conversion CR (no plan) is created without SNAPSHOT_MOREF, the controller creates its own snapshot. If the CR fails, the snapshot is not removed from vCenter. This blocks future warm migrations for the same VM with: "VM has pre-existing snapshots which are incompatible with warm migration."

Fix: Add condition for snapshot removal in the DI pipeline state machine

### DIFF
--- a/pkg/controller/conversion/controller.go
+++ b/pkg/controller/conversion/controller.go
@@ -163,6 +163,23 @@ func (r Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (r
 			"stage", conversion.Status.Stage)
 	}
 
+	// Safety net: if the pipeline terminates with Phase=Failed and an owned
+	// snapshot is still present (e.g. the failure happened before the divert
+	// logic could redirect to StageRemoveSnapshot, or the snapshotRemoval stage
+	// itself errored and set Phase=Failed internally), submit a best effort
+	// removal task here. If in the primary failure path, the divert in
+	// runDeepInspection clears Status.Snapshot before StageFinished (successful snapshot removal) then this block is a no-op.
+	if conversion.Status.Phase == api.PhaseFailed {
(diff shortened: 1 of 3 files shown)


pkg/controller/conversion/controller.go
```diff
@@ -163,6 +163,23 @@ func (r Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (r
...
+	if conversion.Status.Phase == api.PhaseFailed {
+		ensurer, ensureErr := NewEnsurer(r.Client, r.Log, conversion.Spec)
+		if ensureErr == nil {
+			if _, snapErr := ensurer.RemoveOwnedSnapshot(ctx, conversion); snapErr != nil {
+				r.Log.Error(snapErr, "Failed to trigger snapshot removal for failed Conversion.")
+			}
+		} else {
+			r.Log.Error(ensureErr, "Failed to build Ensurer for snapshot cleanup on failure.")
+		}
+	}
```
mnecas, line 181: would it make sense to have this in defer that we would cleanup if there would error in any other place?

solenoci: The pipeline is the only place a snapshot can be created. Every early return before it is safe without needing a snapshot cleanup. So a defer firing on those paths would just evaluate conversion.Status.Phase == api.PhaseFailed as a no-op.

Hazanel:
NP:
The renamed function's doc comment uses lowercase podhandler instead of podHandler:

`// podhandler is called by the pipeline runners when StageFinished is reached.`
Should be:

`// podHandler is called by the pipeline runners when StageFinished is reached.`

and please address @coderabbitai major issue, I agree with him

solenoci:
addressed the comments, please review @Hazanel

Hazanel (approved):

PR #6590, opened 2026-05-21
MTV-5584 | xfs_repair ignore RHEL 7 AGFL inconsistency

mnecas:
Relates to [RHEL-178287](https://redhat.atlassian.net/browse/RHEL-178287), there is issue when XFSv5 on RHEL 7 reports corrupted FS even if there is no actuall corrpution.

When feature_xfs_repair_ignore is enabled on the ForkliftController CR,
xfs_repair exit status is suppressed so conversions are not blocked by
minor XFS issues. The flag flows from the operator through the controller
env into LIBGUESTFS_APPEND and is checked by a wrapper script that
replaces xfs_repair inside the appliance.

Resolves: MTV-5584

[RHEL-178287]: https://redhat.atlassian.net/browse/RHEL-178287?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/build/virt-v2v-rhel9/Containerfile
+++ b/build/virt-v2v-rhel9/Containerfile
@@ -75,6 +75,14 @@ COPY --from=builder /app/virt-v2v-wrapper /usr/bin/virt-v2v-wrapper
 
 COPY --from=builder /app/forklift-wait-for-reboot /usr/local/bin/forklift-wait-for-reboot
 
+# Replace xfs_repair with a wrapper that can ignore failures; register the
+# original binary in supermin hostfiles so both get copied into the appliance.
+COPY build/virt-v2v/xfs_repair_wrapper.sh /tmp/xfs_repair_wrapper.sh
+RUN cp /usr/sbin/xfs_repair /usr/sbin/xfs_repair.original && \
+    cp /tmp/xfs_repair_wrapper.sh /usr/sbin/xfs_repair && \
+    chmod 755 /usr/sbin/xfs_repair && \
+    echo '/usr/sbin/xfs_repair.original' >> /usr/lib64/guestfs/supermin.d/hostfiles
+
 ENTRYPOINT ["/usr/libexec/catatonit/catatonit", "/usr/bin/virt-v2v-wrapper"]
 
 LABEL \
--- a/operator/config/manifests/bases/forklift-operator.clusterserviceversion.yaml
+++ b/operator/config/manifests/bases/forklift-operator.clusterserviceversion.yaml
@@ -451,6 +451,11 @@ spec:
         path: controller_retain_populator_pods
         x-descriptors:
         - urn:alm:descriptor:com.tectonic.ui:hidden
+      - description: Ignore xfs_repair exit status during conversion (default false)
+        displayName: XFS Repair Ignore
+        path: feature_xfs_repair_ignore
+        x-descriptors:
+        - urn:alm:descriptor:com.tectonic.ui:hidden
       - description: Enable static udn IP addresses feature (default false)
         displayName: Static UDN IP Addresses
         path: controller_static_udn_ip_addresses
--- a/operator/streams/downstream/mtv-operator.clusterserviceversion.yaml
+++ b/operator/streams/downstream/mtv-operator.clusterserviceversion.yaml
@@ -521,6 +521,11 @@ spec:
           path: controller_retain_populator_pods
           x-descriptors:
           - urn:alm:descriptor:com.tectonic.ui:hidden
+        - description: Ignore xfs_repair exit status during conversion (default false)
+          displayName: XFS Repair Ignore
+          path: feature_xfs_repair_ignore
+          x-descriptors:
+          - urn:alm:descriptor:com.tectonic.ui:hidden
         - description: Enable static udn IP addresses feature (default false)
           displayName: Static UDN IP Addresses
           path: controller_static_udn_ip_addresses
--- a/operator/streams/upstream/forklift-operator.clusterserviceversion.yaml
+++ b/operator/streams/upstream/forklift-operator.clusterserviceversion.yaml
@@ -525,6 +525,11 @@ spec:
           path: controller_retain_populator_pods
           x-descriptors:
           - urn:alm:descriptor:com.tectonic.ui:hidden
+        - description: Ignore xfs_repair exit status during conversion (default false)
+          displayName: XFS Repair Ignore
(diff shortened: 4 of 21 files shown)


operator/config/manifests/bases/forklift-operator.clusterserviceversion.yaml
```diff
@@ -94,6 +94,12 @@ spec:
         x-descriptors:
         - urn:alm:descriptor:com.tectonic.ui:radio:true
         - urn:alm:descriptor:com.tectonic.ui:radio:false
+      - description: Skip xfs_repair in virt-v2v containers (default false)
+        displayName: Skip XFS Repair
+        path: feature_xfs_repair_skip
+        x-descriptors:
+        - urn:alm:descriptor:com.tectonic.ui:radio:true
+        - urn:alm:descriptor:com.tectonic.ui:radio:false
```
solenoci: Is it a good idea to hide this from default view as the setting is possibly dangerous?

operator/streams/downstream/mtv-operator.clusterserviceversion.yaml
```diff
@@ -160,6 +160,12 @@ spec:
           x-descriptors:
           - urn:alm:descriptor:com.tectonic.ui:radio:true
           - urn:alm:descriptor:com.tectonic.ui:radio:false
+        - description: Skip xfs_repair in virt-v2v containers (default false)
+          displayName: Skip XFS Repair
+          path: feature_xfs_repair_skip
+          x-descriptors:
```
solenoci: Same as other comment

operator/streams/upstream/forklift-operator.clusterserviceversion.yaml
```diff
@@ -164,6 +164,12 @@ spec:
           x-descriptors:
           - urn:alm:descriptor:com.tectonic.ui:radio:true
           - urn:alm:descriptor:com.tectonic.ui:radio:false
+        - description: Skip xfs_repair in virt-v2v containers (default false)
+          displayName: Skip XFS Repair
+          path: feature_xfs_repair_skip
+          x-descriptors:
```
solenoci: Same as other comment

solenoci (commented):
NP

solenoci:
Please run: `make update-manifests`

mnecas:
cc @rwmjones

rwmjones:
I don't exactly _like_ it, but it's gated behind a flag so that's fine I guess.
The proposed fix upstream for this is: https://github.com/libguestfs/libguestfs/pull/355

mnecas:
@rwmjones after doing e2e testing my approach, it does not work because the enviroment vars are not propagated to the guestfs appliance so i can't pass the feature flag. I'm thinking to just adapt your suggestion but to the wrapper. Any other way it start to look too much complicated.

solenoci:
@mnecas can we just symlink from the wrapper to xfs_repair from virt-v2v go cmd when there is env var present?

mnecas:
@solenoci we would need to do it during runtime of the contianer and the user does not have permissions to override the binary. Alternative we could do some configmap mount but that gets too complicated

solenoci:
yeah, permission issue would complicate it a bit, we could link the /usr xfs_repair to qemu owned xfs_repair somewhere where it can be accessible (this link would be always the same) and then replace the xfs_repair owned by qemu with one of the binaries, either the real one or the wrapper during runtime. I know this is a big hack but it could maybe work.

rwmjones:
This all sounds like a massive hack. Could we pick up the fixed version of libguestfs instead. We can have an unsigned build of that today if someone will review the PR.

mnecas:
i think we need to fix it on our side so I have duplicated the logic from libguestfs to the wrapper. Once the fix will be avilable we will revert this change. I have create followup task for this [MTV-5588](https://redhat.atlassian.net/browse/MTV-5588).
Hopefully in future we could directly pick up the builds but i don't want to start with it right now. wdyt @rwmjones ?

[MTV-5588]: https://redhat.atlassian.net/browse/MTV-5588?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

mnecas:
I was able to make make it work by adding `LIBGUESTFS_APPEND` this way we can pass information into libugestfs, it's hacky but better than any other alternative. wdyt @solenoci @rwmjones ?

build/virt-v2v-rhel9/Containerfile
```diff
@@ -75,6 +75,12 @@ COPY --from=builder /app/virt-v2v-wrapper /usr/bin/virt-v2v-wrapper
 
 COPY --from=builder /app/forklift-wait-for-reboot /usr/local/bin/forklift-wait-for-reboot
 
+COPY build/virt-v2v/xfs_repair_wrapper.sh /tmp/xfs_repair_wrapper.sh
+RUN cp /usr/sbin/xfs_repair /usr/sbin/xfs_repair.original && \
+    cp /tmp/xfs_repair_wrapper.sh /usr/sbin/xfs_repair && \
+    chmod 755 /usr/sbin/xfs_repair && \
+    echo '/usr/sbin/xfs_repair.original' >> /usr/lib64/guestfs/supermin.d/hostfiles
```
solenoci, line 84: please, document this workflow for the future ref

pkg/virt-v2v/config/variables.go
```diff
@@ -175,6 +176,16 @@ func (s *AppConfig) Load() (err error) {
 
 	s.SupportsNoFstrim = detectNoFstrimSupport("/etc/os-release")
 
+	if s.getEnvBool(EnvXfsRepairIgnoreName, false) {
+		const token = "xfs_repair_ignore=1"
+		if existing := os.Getenv("LIBGUESTFS_APPEND"); existing != "" {
```
solenoci, line 181: please, document this as we discussed offline

solenoci (commented):
NP: Mainly doc for the more "abstract" logic behind this workaround, otherwise we should test that this not break different types of migrations.

Other than those 2 points, LGTM

build/virt-v2v-rhel9/Containerfile
mnecas: done ptal

solenoci (approved):

PR #6599, opened 2026-05-21
ECOPROJECT-4296 | Read the appInfo from the extraConfig VM field

machacekondra:
In assisted-migration-agent we would like to display running applications in VMs.

Jira: https://redhat.atlassian.net/browse/ECOPROJECT-4296


mnecas (approved):
lgtm

PR #6662, opened 2026-05-26
MTV-5604 | Fix slow vSphere tag fetching causing collector to hang

Hazanel:
The vSphere collector's getVMsWithTags function suffered from an N+1 API call problem: it called GetTags (which internally does ListTags + N sequential GetTag calls) and GetAttachedObjectsOnTags (which also does N sequential GetTag calls). With hundreds of tags in vCenter, this caused the collector to hang or become unresponsive.

Replace the inefficient govmomi convenience methods with an optimized approach:
1. ListTags - single API call to get all tag IDs
2. ListAttachedObjectsOnTags - single batch call to get attachments
3. Filter to only tags attached to VirtualMachine objects
4. Fetch tag details concurrently (bounded at 10 workers) only for VM-relevant tags

Tested with 1000 tags in vCenter (872 attached to a single VM). Migration of all 872 tags as labels on the destination VM was verified end-to-end.

Ref: https://redhat.atlassian.net/browse/MTV-5604
Resolves: MTV-5604

### DIFF
--- a/pkg/controller/provider/container/vsphere/collector.go
+++ b/pkg/controller/provider/container/vsphere/collector.go
@@ -8,9 +8,11 @@ import (
 	"path"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/kubev2v/forklift/pkg/lib/util"
+	"github.com/kubev2v/forklift/pkg/settings"
 
 	api "github.com/kubev2v/forklift/pkg/apis/forklift/v1beta1"
 	"github.com/kubev2v/forklift/pkg/controller/base"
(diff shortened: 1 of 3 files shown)


Hazanel:
Hey @coderabbitai  - what do you think of the solution?

Hazanel:
Good points @coderabbitai - thanks!

 here's how I see each:

1. Chunking ListAttachedObjectsOnTags — Agree it's worth considering at extreme scale. In practice, the environments we've validated (up to 1,000 tags) work fine with a single call, and typical MTV deployments won't hit 10k+. I'll keep this as a follow-up if we get reports of payload limits being hit.

2. TTL cache on the tag map — Looking at the call site, getVMsWithTags is already guarded by vmTagsFetched inside apply(), so it runs at most once per WaitForUpdatesEx batch, not per-VM. Across batches, re-fetching is intentional since tags may have changed. No cache needed.

3. No bulk GetTag API — Confirmed. govmomi's tags.Manager only exposes GetTag(ctx, id) individually — there's no GetTagsByIDs([]string). The bounded worker pool is the best available approach.

4. Zero-value results on cancellation — Good catch. Added a defensive if res.tag.ID == "" { continue } in the results loop to guard against zero-value entries from cancelled goroutines.

pkg/controller/provider/container/vsphere/collector.go
```diff
@@ -173,6 +175,23 @@ const (
...
+const (
+	// Default concurrency for parallel GetTag API calls.
+	DefaultTagFetchConcurrency = 10
+	// Env var to override tag fetch concurrency.
+	EnvTagFetchConcurrency = "VSPHERE_TAG_FETCH_CONCURRENCY"
+)
+
+var tagFetchConcurrency = DefaultTagFetchConcurrency
+
+func init() {
```
yaacov: @coderabbitai hi, we usually use the settings pkg convention as a central place for all our env vars settings, does it make sense to use init here ?

yaacov: @Hazanel lets move this to the settings pkg

Hazanel: Hey @coderabbitai and @yaacov , please take a look

yaacov (approved):

PR #6699, opened 2026-05-27
MTV-5621 | Add per-ESXi-host populator throttling with dedicated MAX_POPULATOR_INFLIGHT

mnecas:
Add per-ESXi-host populator throttling via controller_max_populator_inflight

Introduce controller_max_populator_inflight ForkliftController setting
(default 20) to control how many populator pods can run concurrently
per ESXi source host, independent of controller_max_vm_inflight.

The populator controller reads the sourceHost label from the xcopy CR
and propagates it to the populator pod. Before creating a new pod, it
counts active pods for that host and defers if the limit is reached.
This prevents a single busy host from starving transfers on other hosts.

The plan-level scheduler now treats offload-plugin (xcopy) VMs the same
as v2v transfers (cost=1 per VM) since per-disk throttling is handled
by the populator controller.

Example ForkliftController CR:
```
  spec:
    controller_max_populator_inflight: 10
```
Resolves: MTV-5621

### DIFF
--- a/cmd/populator-controller/populator-controller.go
+++ b/cmd/populator-controller/populator-controller.go
@@ -4,6 +4,7 @@ import (
 	"flag"
 	"os"
 	"os/signal"
+	"strconv"
 	"syscall"
 
 	"github.com/kubev2v/forklift/pkg/apis/forklift/v1beta1"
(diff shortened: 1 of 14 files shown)


mnecas:
@coderabbitai ptal

solenoci:
just FYI, konflux will fail because github has API issues

PR #6762, opened 2026-05-28
MTV-5319 | Cache ShouldUseV2vForTransfer in BasePredicate

gwencasey96:
Evaluate() for CDI/virt-v2v flags called ShouldUseV2vForTransfer on every predicate check, repeating destination StorageClass API work per VM during begin() pipeline setup. Cache the result (and errors) once per BasePredicate.


mrnold:
I think this is okay. I think there might be more places that have similar slowdowns, but if this fixes the original bug we can deal with them later.

mrnold (approved):

solenoci (approved):

PR #6768, opened 2026-05-28
MTV-3505 | Fix VMware serial number feature flag not disabling from CR

yaacov:
The Jinja2 template only emitted FEATURE_VMWARE_SYSTEM_SERIAL_NUMBER when the CR value was true, omitting the env var entirely when false. Since the Go code defaults to true when the env var is absent, setting feature_vmware_system_serial_number: "false" in the ForkliftController CR had no effect.

Replace the conditional block with an unconditional emit using |bool|lower, matching the pattern already used for FEATURE_USE_CONVERSION_CR.

Bash test plan added to Jira ticket.

Ref: https://redhat.atlassian.net/browse/MTV-3505
Resolves: MTV-3505


yaacov:
@mnecas please review

PR #6838, opened 2026-05-31
MTV-5644 | Add overlay to the virt-v2v-in-place so we write changes only if conversion passes

Hazanel:
Before running virt-v2v-in-place, create a qcow2 overlay per disk backed by the original. The conversion writes only to the overlay. On success, qemu-img commit merges changes back to the base disk. On failure, the overlay is discarded and the base disk is untouched, enabling safe retry.

The overlay is always active for in-place conversions (no opt-out flag). The domain XML driver type is set to qcow2 to match the overlay format.

Ref: https://redhat.atlassian.net/browse/MTV-5644
Resolves: MTV-5644

### DIFF
--- a/pkg/virt-v2v/conversion/overlay.go
+++ b/pkg/virt-v2v/conversion/overlay.go
@@ -1,8 +1,6 @@
 package conversion
 
 import (
-	"bytes"
-	"encoding/json"
 	"fmt"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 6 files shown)


Hazanel:
Hey @coderabbitai  - please take a look and perhaps suggest if there is a better/enhanced way to implement  this feature, also please create a diagram/state machine of the scenario

Hazanel:
Thanks @coderabbitai for the review!

suggestion #2 (stale overlays from previous failed runs):

Not applicable — the virt-v2v pod is ephemeral. Each migration attempt spawns a fresh pod with its own filesystem. stale overlays from a previous pod don't persist. The defer safety net handles in-pod crashes (panics, SIGTERM). SIGKILL is the only gap, and that's acceptable since the pod won't be reused after a forced kill.

Re: suggestion #3 (pre-flight disk space check):

Overlays are sparse qcow2 — they start at ~200KB regardless of the backing disk size and grow only as virt-v2v writes. A pre-flight check against a fixed threshold (e.g., 100MB) wouldn't predict whether there's enough space for the actual conversion writes. The existing error handling already covers this: if qemu-img create fails due to ENOSPC we clean up, and if the conversion fills the overlay and hits ENOSPC, virt-v2v fails and we discard — base disk is safe either way.

Re: suggestion #4 (recover in deferred cleanup):

A panic inside DiscardOverlays (which calls os.Remove and os.Symlink) would require a fundamentally broken kernel/filesystem. Adding recover() adds code complexity for a near-impossible scenario, and re-panicking after partial cleanup doesn't improve the outcome. The current defer is sufficient.

Re: suggestion #5 (structured logging):

The virt-v2v binary uses fmt.Printf throughout (it's a standalone CLI-style process, not the controller). Our code is consistent with the surrounding codebase. Happy to migrate if the whole binary moves to structured logging in a future PR.

WDYT?

Hazanel:
Hey @coderabbitai 
About rollbackOverlays on partial commit is misleading (rated High)

The concern is valid from a naming/readability perspective — calling "rollback" after irreversible commits is misleading. But functionally, the code is correct: os.Remove on already-deleted files is a no-op, and restoreLinks correctly restores all symlinks.

The useful part of this suggestion is enriching the error message to tell operators which disks committed vs. which failed - which is what I did

Hazanel:
Test plan ans log's results are attached to the ticket

pkg/virt-v2v/conversion/overlay.go
```diff
@@ -0,0 +1,204 @@
...
+			return nil, fmt.Errorf("failed to remove old symlink %s: %w", disk.Link, err)
+		}
+
+		// Step 3: Create new symlink pointing at the overlay
+		if err := os.Symlink(overlayPath, disk.Link); err != nil {
+			_ = os.Remove(overlayPath)
+			_ = os.Symlink(originalLink, disk.Link)
+			c.rollbackOverlays(overlays)
+			return nil, fmt.Errorf("failed to create overlay symlink %s -> %s: %w", disk.Link, overlayPath, err)
+		}
```
mnecas, line 64: Be careful we are using the disk struct now that you are editing the symlink you will need to update the disk struct links aswell

pkg/virt-v2v/conversion/overlay.go
```diff
@@ -0,0 +1,204 @@
...
+func (c *Conversion) DiscardOverlays(overlays []*Overlay) {
+	for _, o := range overlays {
+		if err := os.Remove(o.Path); err != nil && !os.IsNotExist(err) {
+			fmt.Fprintf(os.Stderr, "WARNING: failed to remove overlay %s: %v\n", o.Path, err)
+		} else {
+			fmt.Printf("Discarded overlay %s (base disk unchanged)\n", filepath.Base(o.Path))
+		}
+	}
+	c.restoreLinks(overlays)
+}
```
mnecas, line 123: what's the differece between `DiscardOverlays` and `rollbackOverlays`?

pkg/virt-v2v/conversion/overlay.go
```diff
@@ -0,0 +1,204 @@
...
+	return info.Format, nil
+}
+
+func (c *Conversion) CreateOverlays() ([]*Overlay, error) {
+	var overlays []*Overlay
+	for _, disk := range c.Disks {
+		overlayPath := disk.Link + ".qcow2"
+
+		// Detect the backing disk format instead of assuming raw
+		backingFmt, err := c.detectDiskFormat(disk.Path)
```
mnecas: raw, the backing fmt will always be raw

pkg/virt-v2v/conversion/overlay.go
```diff
@@ -0,0 +1,204 @@
...
+		cmd := c.CommandBuilder.New("qemu-img").
+			AddPositional("commit").
+			AddPositional(o.Path).
+			Build()
+		cmd.SetStdout(os.Stdout)
+		cmd.SetStderr(os.Stderr)
+		if err := cmd.Run(); err != nil {
+			c.rollbackOverlays(overlays)
+			if len(committed) > 0 {
+				return fmt.Errorf("partial commit: disks %v already committed, failed on %s — VM may be in inconsistent state: %w",
```
mnecas, line 95: ohhh yeah this is bit scary, in case where we would partially commit, think we will need to fail in this case.

mnecas (changes requested):
Please add feature flag for this overlay so we could toggle it in case something breaks.

pkg/virt-v2v/conversion/overlay.go
Hazanel: We do fail , the function returns an error describing exactly which disks were already committed and which one failed. The problem is qemu-img commit is irreversible per disk, so there's no way to undo an earlier commit if a later one fails. The remaining uncommitted overlays are discarded , and the error message clearly flags the inconsistent state.

Hazanel:
Hey @coderabbitai, please follow up on @mnecas comments

Hazanel:
@coderabbitai - I already addressed please check

mnecas (approved):

PR #6845, opened 2026-06-01
MTV-5240 | Annotate AAP hooks with job template name for kubectl

gwencasey96:
Resolve AAP job template names on Hook reconcile and store them as annotations so kubectl users see the playbook name, not only the template ID. Add Hook CRD printer columns for terminal visibility.

Ref: https://redhat.atlassian.net/browse/MTV-5240
Resolves: MTV-5240

### DIFF
--- a/pkg/controller/hook/controller.go
+++ b/pkg/controller/hook/controller.go
@@ -18,14 +18,18 @@ package hook
 
 import (
 	"context"
+	"strconv"
+	"strings"
 
 	api "github.com/kubev2v/forklift/pkg/apis/forklift/v1beta1"
 	"github.com/kubev2v/forklift/pkg/controller/base"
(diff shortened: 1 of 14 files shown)


pkg/controller/hook/controller.go
```diff
@@ -124,6 +132,19 @@ func (r Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (r
...
+	updated, aapErr := r.ensureAAPJobTemplateAnnotations(ctx, hook)
+	if aapErr != nil {
+		err = aapErr
+		return
+	}
+	if updated {
+		// Requeue so we reconcile again with refreshed object data.
+		result.Requeue = true
+		return
+	}
```
mnecas: Why are we updating the job, can't we add the annotations when the job is getting created?

PR #6846, opened 2026-06-01
MTV-5529 | Rework reusing of DI conversion CRs

solenoci:
Issue: When a plan tries to reuse a deep inspection CR, it doesn't scope it's search to the plan owned/standalone and is reusing old/wring CRs.

Fix: change the search mechanism to first look for plan DI CRs, if not found then look for standalone CRs (with excluding other plans CRs), and if not found create a new plan owned one.

Resolves: MTV-5529
Ref: https://redhat.atlassian.net/browse/MTV-5529

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved deep inspection conversion resource reuse and lifecycle management during preflight checks.

* **Performance**
  * Optimized container image fetching by changing pull policy to fetch images only when not already present locally.

* **Documentation**
  * Updated Conversion CR documentation with clearer API endpoint descriptions and warm-migration preflight workflow details.

### DIFF
--- a/pkg/controller/conversion/builder.go
+++ b/pkg/controller/conversion/builder.go
@@ -182,7 +182,7 @@ func (b *Builder) GetVirtV2vPodSpec(vm *plan.VMStatus, volumes []core.Volume, vo
 			Containers: []core.Container{
 				{
 					Env:             nil, // set by type-specific builder
-					ImagePullPolicy: core.PullAlways,
+					ImagePullPolicy: core.PullIfNotPresent,
 					Resources: core.ResourceRequirements{
 						Requests: core.ResourceList{
 							core.ResourceCPU:    resource.MustParse(Settings.Migration.VirtV2vContainerRequestsCpu),
@@ -354,7 +354,7 @@ func (b *Builder) GetDeepInspectionPodSpec(volumes []core.Volume, volumeMounts [
 				{
 					Name:            "deep-inspection",
 					Image:           img,
-					ImagePullPolicy: core.PullAlways,
+					ImagePullPolicy: core.PullIfNotPresent,
 					Resources: core.ResourceRequirements{
 						Requests: core.ResourceList{
 							core.ResourceCPU:    resource.MustParse("100m"),
--- a/pkg/controller/plan/kubevirt.go
+++ b/pkg/controller/plan/kubevirt.go
@@ -582,20 +582,11 @@ func (r *KubeVirt) buildConversion(planName, vmID string, labels map[string]stri
 }
 
 // ensureConversion creates or updates the Conversion CR on the cluster.
-// An existing CR is found by matching all labels except LabelPlan (which
-// changes when a plan is re-created). When found its Spec and Labels are
-// refreshed; otherwise the provided CR is created verbatim.
 func (r *KubeVirt) ensureConversion(cr *api.Conversion) (*api.Conversion, error) {
-	lookupLabels := make(map[string]string, len(cr.Labels))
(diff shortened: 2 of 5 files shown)


solenoci:
@coderabbitai pls review

pkg/controller/plan/kubevirt.go
Hazanel: Isn't this a dead code now?

pkg/controller/plan/kubevirt.go
```diff
@@ -671,6 +672,111 @@ func (r *KubeVirt) GetDeepInspectionConversion(vm *plan.VMStatus) (*api.Conversi
...
+	}
+	if vm.LUKS.Name != "" {
+		return api.DiskEncryptionTypeLUKS
+	}
+	return ""
+}
+
+// deepInspectionMatchesPlan reports whether the Conversion CR is compatible with
+// what the plan would create for the given VM. Two fields affect this:
+//   - DiskEncryption.Type: a CR without dsik encryption can't be used for CR with disk encryption set and vice versa,
```
Hazanel: Typo: should be disk

pkg/controller/plan/migration.go
```diff
@@ -1599,6 +1532,121 @@ func (r *Migration) execute(vm *plan.VMStatus) (err error) {
...
+//	          Failed/Canceled - fail the migration step (no retry).
+//	          Running/Pending - wait (requeue).
+//
+// Returns true when the reconcile should requeue (CR still in progress or just
+// created). Returns false when the step has been settled - either advanced to
+// the next phase via NextPhase or failed via step.AddError.
+// All errors are consumed internally and recorded on the step, the caller never
+// receives a non-nil error from this method.
+func (r *Migration) runPreflightDeepInspection(vm *plan.VMStatus, step *plan.Step) (requeue bool) {
+	snapshotMoref := vm.Warm.Precopies[0].Snapshot
```
Hazanel, line 1598: Please add a nil guard here

pkg/controller/plan/kubevirt.go
solenoci: address from coderabbit review

pkg/controller/plan/kubevirt.go
solenoci: address from coderabbit review

Hazanel:
Hey @coderabbitai ensureConversion strips LabelPlan from the lookup labels before matching. Can this cause Phase 3 of one plan to match and overwrite a different plan's CR for the same VM, and if so, should LabelPlan be included in the lookup?

solenoci:
Addressed the logic flow issue. Now the flow looks like this
Screenshot From 2026-06-03 12-02-54

Hazanel:
lgtm

solenoci:
@coderabbitai PTAL

pkg/controller/conversion/builder.go
```diff
@@ -182,7 +182,7 @@ func (b *Builder) GetVirtV2vPodSpec(vm *plan.VMStatus, volumes []core.Volume, vo
 			Containers: []core.Container{
 				{
 					Env:             nil, // set by type-specific builder
-					ImagePullPolicy: core.PullAlways,
+					ImagePullPolicy: core.PullIfNotPresent,
```
mnecas, line 185: why do we need this?

solenoci: we don't, I used pullAlways for testing purposes and thought that I should adhere to the rest of pod definitions, will revert.

PR #6851, opened 2026-06-02
don't merge, just testing konflux trigger

solenoci:


solenoci:
ah... merged by mistake... so also reverted

PR #6901, opened 2026-06-03
MTV-5241 | Add MTV coding conventions for contributors and AI

gwencasey96:
Document implementation patterns from maintainer review feedback: logging, context, HTTP, reconcilers, APIs, tests, and security. PR scope and review process remain in code-review.md.

Ref: https://redhat.atlassian.net/browse/MTV-5241
Resolves: MTV-5241

### DIFF
--- a/community/coding-conventions.md
+++ b/community/coding-conventions.md
@@ -1,0 +1,185 @@
+# Coding Conventions
+
+Practical conventions for contributing to **Migration Toolkit for Virtualization (MTV)** / the [Forklift](https://github.com/kubev2v/forklift) codebase. These rules come from maintainer review feedback and merged PRs. They supplement—not replace—[code quality](code-quality.md), [code review](code-review.md), and [AI contribution policy](ai-contribution-policy.md).
+
+**Audience:** human contributors and AI assistants. When generating or reviewing code, apply these conventions by default unless a maintainer approves an exception.
+
+PR scope, design approval, and review process are in [code review](code-review.md)—not repeated here.
+
+---
+
+## Logging & observability
+
+### Log `name` and `namespace` as separate fields
+
+Pass resource identity as structured key/value pairs, not a single combined string:
+
+```go
+// Good
+log.Info("reconciling", "name", resource.GetName(), "namespace", resource.GetNamespace())
+
+// Bad — debug tooling cannot parse a combined label reliably
+log.Info("reconciling", "resource", resource.GetNamespace()+"/"+resource.GetName())
+```
+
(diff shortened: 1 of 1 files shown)


mrnold:
This looks okay, although I am not always sure what to think when I open the example pull request links. Maybe some of them could have direct links to the intended line numbers in the pull request? And maybe also say "yes, like this" or "not like this".

community/coding-conventions.md
```diff
@@ -0,0 +1,185 @@
...
+log.Info("reconciling", "resource", resource.GetNamespace()+"/"+resource.GetName())
+```
+
+Debug and observability tooling expects `name` and `namespace` as individual fields. ([PR #5555](https://github.com/kubev2v/forklift/pull/5555))
+
+---
+
+## Error & context handling
+
+### Use the reconciler/request `ctx`, not `context.Background()`
```
yaacov, line 31: @coderabbitai hi, this rule sounds like a good advice in general, not specific to our project, can you look at this doc and prepare a list of conventions that are specific to forklift, like separate name and namespace and conventions that are true in general, like good usage of context in go?

community/coding-conventions.md
```diff
@@ -0,0 +1,185 @@
...
+
+```go
+// Good
+log.Info("reconciling", "name", resource.GetName(), "namespace", resource.GetNamespace())
+
+// Bad — debug tooling cannot parse a combined label reliably
+log.Info("reconciling", "resource", resource.GetNamespace()+"/"+resource.GetName())
+```
+
+Debug and observability tooling expects `name` and `namespace` as individual fields. ([PR #5555](https://github.com/kubev2v/forklift/pull/5555))
```
yaacov, line 25: while it's nice to know where we picked up on this rule, i am not sure what are the pros and cons of documenting a specific place we implemented this rule in the past, the idea of a convention is that we implement it everywhere, so a specific place we used it does not add to the validity, for example, we may have implemented a bad rule in one place and we should not make it a rule.

@coderabbitai, IMHO we should not add the specific commit we picked the rule from because it does not add to the validity of the rule, a rule may be good or bad, but an incidental implementation does not add or subtract from it's validity, what are the pros and cons of adding links to incidental implementation in your opeinion?

yaacov: @gwencasey96 hi, IMHO we don't need the specific link, the rule is good stand on it's own even without a link, WDYT?

community/coding-conventions.md
yaacov: @gwencasey96 can you take a look ^^ ?

gwencasey96:
closing due to overlap with another project

PR #6870, opened 2026-06-02
MTV-5655 | Fix populator pod cleanup deleting pods of other VMs

borod108:
## Summary
- `getPopulatorPods()` filtered only by migration UID, returning pods for all VMs in the migration. When a new VM entered `PhaseStarted`, its cleanup killed other VMs' active populator pods, causing "PVC is missing; disk transfer cannot continue" errors.
- Add `vmID` label to populator pods at creation time (controller.go) and filter `getPopulatorPods()` by vmID (kubevirt.go), matching the fix already applied to `getPopulatorCrList` in PR #6628.
- Handle empty progress string in `PopulatorTransferredBytes` to avoid parse errors for pods not yet running.

## Test plan
- [x] Existing unit tests updated and passing (165/165)
- [ ] Deploy with concurrent VM migrations and verify populator pods are only cleaned up for their own VM
- [ ] Verify "PVC is missing" errors no longer occur when multiple VMs migrate concurrently

Jira: [MTV-5655](https://redhat.atlassian.net/browse/MTV-5655)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[MTV-5655]: https://redhat.atlassian.net/browse/MTV-5655?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #6907, opened 2026-06-03
MTV-5548 | Guard populator-controller PVC deletion with retain flag

yaacov:
The previous commit (81ce8b79) only guarded explicit pod deletion in the Forklift plan controller's cleanup(). However, during an active migration failure the populator-controller independently detects PodFailed and calls deleteFailedPVC(), which deletes the target PVC. Since the populator pod has an OwnerReference on that PVC, Kubernetes GC cascades the deletion to the pod — defeating the retain flag.

This fix reads FEATURE_RETAIN_POPULATOR_PODS in the populator- controller and skips deleteFailedPVC() on permanent failure when the flag is set. The operator template is updated to propagate the env var to the populator-controller deployment. Retry-path pod deletion (oVirt/OpenStack, up to 3 attempts) is intentionally left unchanged since those pods are immediately recreated.

Ref: https://redhat.atlassian.net/browse/MTV-5548
Resolves: MTV-5548

### DIFF
--- a/pkg/lib-volume-populator/populator-machinery/controller.go
+++ b/pkg/lib-volume-populator/populator-machinery/controller.go
@@ -32,6 +32,7 @@ import (
 	"time"
 
 	api "github.com/kubev2v/forklift/pkg/apis/forklift/v1beta1"
+	"github.com/kubev2v/forklift/pkg/settings"
 	corev1 "k8s.io/api/core/v1"
 	storagev1 "k8s.io/api/storage/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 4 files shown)


yaacov:
@coderabbitai, please re-revivew and close fixed review comments

pkg/lib-volume-populator/populator-machinery/controller.go
```diff
@@ -773,6 +780,11 @@ func (c *controller) syncPvc(ctx context.Context, key, pvcNamespace, pvcName str
 					vendor, _, _ := unstructured.NestedString(crInstance.Object, "spec", "storageVendorProduct")
 					c.metrics.recordCompletionFromScrape(pvc.UID, "failure", migration, string(pvc.UID), vendor, "", "", "", 0, 0, 0)
 					c.recorder.Eventf(pvc, corev1.EventTypeWarning, reasonPodFailed, "VSphere xcopy populator failed (no retry): Please check the logs of the populator pod, %s/%s", populatorNamespace, pod.Name)
+					if c.retainPopulatorPods {
+						klog.Infof("Retaining failed populator pod %s/%s and PVC %s/%s (FEATURE_RETAIN_POPULATOR_PODS enabled)",
+							populatorNamespace, pod.Name, pvc.Namespace, pvc.Name)
+						return nil
+					}
 					return c.deleteFailedPVC(ctx, pvc)
```
mnecas, line 788: After more thinking, i think the volume populator should not do the cleanup but the forklift-controller. Now we are faicing problem. Becuase the PVC is now owner of the populator pod causing every PVC delete also delete the logs why did the populaotr failed. If we will move this to the MTV archive we should be good.

mnecas (changes requested):

mnecas:
i don't like to have toggle around this as we should not ask to rerun every time when something fail, either have it always or never.

yaacov:
> i don't like to have toggle around this as we should not ask to rerun every time when something fail, either have it always or never.

what do you suggest ?

pkg/lib-volume-populator/populator-machinery/controller.go
yaacov: refactoring the way we do cleanup is a good idea, do you think we should refactor the cleanup ?
last time we had several PRs doing cleanup in different places in the code for different concerns, leaving chunks of cleanup in different places, do you want to clean the cleanup code ?

yaacov:
After offline discussion with @mnecas closing this PR in favor of AI generated code https://github.com/kubev2v/forklift/pull/6870

cc:// @borod108

PR #6908, opened 2026-06-03
MTV-5673 | Propagate customizationScripts volume through Conversion CR

Hazanel:
The customization scripts ConfigMap volume was added to the local volumes/mounts arrays in podVolumeMounts() but not to extraVolumes/ extraMounts. When the Conversion CR path is active, only extraVolumes/extraMounts are serialized into the CR spec and later mounted in the virt-v2v pod. This caused the scripts ConfigMap to be silently dropped, so virt-customize never found or executed the user-provided scripts.

Add the scripts volume and mount to extraVolumes/extraMounts so they propagate through the Conversion CR to the conversion pod.

Ref: https://redhat.atlassian.net/browse/MTV-5673
Resolves: MTV-5673

### DIFF
--- a/pkg/controller/plan/kubevirt.go
+++ b/pkg/controller/plan/kubevirt.go
@@ -3231,7 +3231,7 @@ func (r *KubeVirt) podVolumeMounts(vmVolumes []cnv.Volume, vddkConfigmap *core.C
 					configMapName, configMapNamespace))
 			return
 		}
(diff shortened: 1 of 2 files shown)


Hazanel:

Hazanel:
@coderabbitai - please check the  tests that were added

mnecas (approved):

solenoci (approved):

PR #6916, opened 2026-06-04
MTV-5674 | Add per-network guestIPConfig to NetworkMap

Hazanel:
Introduce a guestIPConfig field (preserve/dhcp/none) on NetworkPair entries in the NetworkMap CRD, allowing per-NIC control over whether static IPs are preserved during migration. This overrides the plan-level preserveStaticIPs flag on a per-network basis.

Changes:
- Add GuestIPConfig type and field to NetworkPair in mapping.go
- Add validation rejecting guestIPConfig on ignored destinations
- Add ResolveNICModes() and HasPreserveMode() helpers in base package
- Modify vSphere and HyperV builders to filter mapMacStaticIps by mode

Ref: https://redhat.atlassian.net/browse/MTV-5674
Resolves: MTV-5674

### DIFF
--- a/operator/config/crd/bases/forklift.konveyor.io_networkmaps.yaml
+++ b/operator/config/crd/bases/forklift.konveyor.io_networkmaps.yaml
@@ -74,6 +74,14 @@ spec:
                       required:
                       - type
                       type: object
+                    networkIPMode:
+                      description: Network IP mode for this network, overrides plan-level
+                        preserveStaticIPs.
+                      enum:
+                      - preserve
+                      - dhcp
+                      - none
+                      type: string
                     source:
                       description: Source network.
                       properties:
--- a/pkg/apis/forklift/v1beta1/mapping.go
+++ b/pkg/apis/forklift/v1beta1/mapping.go
@@ -54,12 +54,29 @@ type NetworkSourceRef struct {
 	Vlan string `json:"vlan,omitempty"`
 }
 
+// NetworkIPMode defines the guest OS network configuration behavior
+// for interfaces mapped through this network pair.
+// +kubebuilder:validation:Enum=preserve;dhcp;none
+type NetworkIPMode string
+
+const (
+	// NetworkIPModePreserve migrates static IP configuration from the source VM.
+	NetworkIPModePreserve NetworkIPMode = "preserve"
+	// NetworkIPModeDHCP signals the NIC should use DHCP on the target.
+	NetworkIPModeDHCP NetworkIPMode = "dhcp"
+	// NetworkIPModeNone leaves the NIC untouched — no IP configuration is applied.
+	NetworkIPModeNone NetworkIPMode = "none"
+)
+
 // Mapped network.
 type NetworkPair struct {
 	// Source network.
 	Source NetworkSourceRef `json:"source"`
 	// Destination network.
 	Destination DestinationNetwork `json:"destination"`
+	// Network IP mode for this network, overrides plan-level preserveStaticIPs.
+	// +optional
+	NetworkIPMode NetworkIPMode `json:"networkIPMode,omitempty"`
 }
 
 // OffloadPlugin is a storage plugin that acts on the storage allocation and copying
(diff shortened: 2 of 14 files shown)


Hazanel:

Hazanel:
thanks @coderabbitai, what about the nadpool logic, wdyt?

Hazanel:
@coderabbitai, please take a look

Hazanel:
@coderabbitai, the PR has changed , changes were amended to the PR

Hazanel:
Thanks @coderabbitai, I just addressed your final minor note

Hazanel:
Test plan and conversion pod's  log   are attached to the ticket

operator/config/crd/bases/forklift.konveyor.io_networkmaps.yaml
```diff
@@ -74,6 +74,14 @@ spec:
                       required:
                       - type
                       type: object
+                    guestIPConfig:
```
mnecas: @coderabbitai can you suggest some other names for this variable?

pkg/apis/forklift/v1beta1/mapping.go
```diff
@@ -45,12 +45,29 @@ type DestinationNetwork struct {
...
+// +kubebuilder:validation:Enum=preserve;dhcp;none
+type GuestIPConfig string
+
+const (
+	// GuestIPConfigPreserve migrates static IP configuration from the source VM.
+	GuestIPConfigPreserve GuestIPConfig = "preserve"
+	// GuestIPConfigDHCP signals the NIC should use DHCP on the target.
+	GuestIPConfigDHCP GuestIPConfig = "dhcp"
+	// GuestIPConfigNone leaves the NIC untouched — no IP configuration is applied.
+	GuestIPConfigNone GuestIPConfig = "none"
```
mnecas: @Hazanel reading the code i don't see the difference between DHCP and none behaviour 
both of them end up at `mapMacStaticIps(vm *model.VM, modeByMAC map[string]string) string {` and just ignore the mapping in `mode != string(api.GuestIPConfigPreserve)`

mnecas (changes requested):

operator/config/crd/bases/forklift.konveyor.io_networkmaps.yaml
mnecas: hmm kinda like the first two `networkIPMode`/`ipAssignmentMode` over the guestIpConfig wdyt @Hazanel ?

pkg/apis/forklift/v1beta1/mapping.go
Hazanel: On windows there is no difference, that was a preparation for linux which does have a difference

operator/config/crd/bases/forklift.konveyor.io_networkmaps.yaml
Hazanel: Let's go with  **networkIPMode**

pkg/apis/forklift/v1beta1/mapping.go
Hazanel: @coderabbitai - wdyt?

Hazanel:
Hey @coderabbitai - can you create a concise description(but don't miss anything) how to use this feature ?

pkg/controller/plan/adapter/base/network.go
```diff
@@ -5,6 +5,70 @@ import (
...
+func ResolveNICModes(nics []NICRef, networkMap *api.NetworkMap, preserveStaticIPs bool) map[string]string {
+	modes := map[string]string{}
+	if networkMap == nil {
+		// No NetworkMap available — fall back to plan-level preserveStaticIPs for all NICs.
+		if preserveStaticIPs {
+			for _, nic := range nics {
+				modes[nic.MAC] = string(api.NetworkIPModePreserve)
+			}
+		}
+		return modes
```
mnecas, line 65: if we retun empty map which i guess means to not preserve the NIC could we explicitly set? 
```
				mode = string(api.NetworkIPModeNone)
```

mnecas (approved):
small np, but lgmt

PR #6921, opened 2026-06-04
MTV-5652 | Propagate detailed SCC error for guest conversion pod failure

Hazanel:
When UseConversionCR is enabled (default in 2.12), pod creation errors such as SCC validation failures were lost. The Conversion CR condition only stored a generic "The conversion has failed." message, and the plan controller reported "Guest conversion pod not found" without checking the CR for details.

Store the pipeline error in the ConversionFailed condition message so it includes the actual API rejection reason. In the plan controller, read the Conversion CR status when the pod is missing and surface the detailed error in the migration step reasons.

Ref: https://redhat.atlassian.net/browse/MTV-5652
Resolves: MTV-5652

### DIFF
--- a/pkg/apis/forklift/v1beta1/conversion.go
+++ b/pkg/apis/forklift/v1beta1/conversion.go
@@ -37,6 +37,12 @@ const (
 	PhaseCanceled  ConversionPhase = "Canceled"
 )
 
+// Condition types set on Conversion status.
+const (
+	ConversionFailed   = "ConversionFailed"
+	ConversionCanceled = "ConversionCanceled"
+)
+
 // ConversionStage is the fine-grained pipeline position within the Running phase.
 type ConversionStage string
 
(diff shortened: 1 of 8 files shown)


Hazanel:

Hazanel:
Hey @coderabbitai, please check the fixes

Hazanel:
@coderabbitai - Do you think there is a better way to solve this?

pkg/controller/plan/kubevirt.go
```diff
@@ -664,6 +664,29 @@ func (r *KubeVirt) ensureConversionSecret(cl client.Client, secret *core.Secret)
...
+		return nil, liberr.Wrap(err)
+	}
+	for i := range list.Items {
+		t := api.ConversionType(list.Items[i].Labels[convctx.LabelConversionType])
+		if t == api.Remote || t == api.InPlace {
+			return &list.Items[i], nil
+		}
+	}
+	return nil, nil //nolint:nilnil
+}
```
solenoci: Can we please use "k8s.io/apimachinery/pkg/labels".NewRequirement to do this job so it's more concise?

solenoci:
Also thinking about adding a new field to the conversion.status, like "conversion.status.message" so we don't abuse the conditions as coderabbit suggested, wdyt?

pkg/apis/forklift/v1beta1/conversion.go
```diff
@@ -301,6 +307,11 @@ type ConversionStatus struct {
 	// +optional
 	// +kubebuilder:validation:Enum=CreatingPod;PodRunning;CreatingSnapshot;WaitingForSnapshot;RemovingSnapshot;WaitingForSnapshotRemoval;FetchingResults;Finished
 	Stage ConversionStage `json:"stage,omitempty"`
+	// Message is a human-readable explanation of the current phase, populated
+	// by the conversion controller when the conversion fails or is canceled.
```
solenoci: NP: I would like to reuse this field also for other states (succeeded, running...) in the future

solenoci (approved):
one more NP but otherwise lgtm

pkg/apis/forklift/v1beta1/conversion.go
Hazanel: sure, added

PR #6935, opened 2026-06-05
MTV-5344 | Use dedicated RBAC for wait-for-reboot pod

solenoci:
Issue: Limit permission scope for the wait-for-reboot pod to only necessary RBAC

Fix: Only VMI/console GET and VMI GET are needed so only use those.

Ref: https://redhat.atlassian.net/browse/MTV-5344

### DIFF
--- a/pkg/controller/plan/migration.go
+++ b/pkg/controller/plan/migration.go
@@ -1584,11 +1584,6 @@ func (r *Migration) execute(vm *plan.VMStatus) (err error) {
 				err = nil
 			}
 		}
-		if vm.HasCondition("WaitForGuestReboots") {
-			r.Log.Info("Retaining failed Windows wait-for-reboot pod for debugging.", "vm", vm.String())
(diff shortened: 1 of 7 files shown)


yaacov:
@solenoci title , description , like to jira ?

pkg/controller/plan/migration.go
```diff
@@ -1570,7 +1584,9 @@ func (r *Migration) execute(vm *plan.VMStatus) (err error) {
 				err = nil
 			}
 		}
-		if ierr := r.kubevirt.DeleteWaitForRebootPod(vm); ierr != nil {
+		if vm.HasCondition("WaitForGuestReboots") {
+			r.Log.Info("Retaining failed Windows wait-for-reboot pod for debugging.", "vm", vm.String())
```
mnecas: not sure what this is doing

solenoci: it's a gate for keeping the wait-for-reboot pod when it fails instead of deleting but it doesn't cause migration to fail (what we talked about offline)

mnecas (approved):

Issue #6203, opened 2026-05-07
Intent to contribute: IP/MAC Preservation for Calico CNI


aaaaaaaalex:
Hi folks,

Reaching out from the Calico team to get some early feedback:
We're building out our VM support and I'd like to add corresponding Forklift support for Calico:
- When a NAD has a type: calico field, check if it also has a reference to a (projectcalico) `Network` CR and/or VLAN field.
- If it does, emit calico annotations to be set on the VM template, asking Calico CNI to honour the source VM's MAC/IP's for each link preserved.

Implementation-wise, my initial plan would be to:
 - Amend the `NetworkConfig` struct (or create a similar, adjacent struct) with Calico fields parsed from the NAD describing which segment/VLAN the NIC should be on,
 - Ensure each builder captures the sources' MAC/iface information where possible, for emitting Calico annotations for the pod template,
 - Validation to ensure the requested addresses make sense within the Calico environment,
 - Add a test-table of different migration cases,
 - Possible documentation file?

Any thoughts on this from the get-go, and/or is there an RFC procedure I should follow?

Thanks in adv.

PR #6936, opened 2026-06-05
Calico Address Preservation Support

aaaaaaaalex:
### Background

This PR allows VMs to be migrated onto clusters using Calico as the CNI, with full
identity preservation on every NIC.

Prior to this change, a VM would be happily scheduled on a Calico cluster, but Calico
CNI would not be aware of its network config. The effect of this was that at CNI-ADD,
Calico would allocate a fresh IP/MAC to the pod-veth that encapsulates the VM,
which would mismatch with the guest OS's addresses. Ultimately, VM identity is lost.

This PR introduces Calico annotations to VM templates which inform Calico CNI
of the guest OS's L2 and L3 addresses. This allows Calico to work in lockstep
with kubevirt and the guest-OS to retain the right addresses when pod veths
are set up.

Additionally, Calico is introducing support for L2-connected, and VRF-connected workloads,
via a `v3.projectcalico.org/Network` CRD, and this PR adds support and validation for that, too.

Upfront, I'd like to thank you in advance for any help you can offer. I'll be more than happy
to dig into any topic you'd like clarification on!

### Background on Calico APIs

Currently, Calico supports the manual assignment of a user-specified IP and hardware address
to a pod's primary NIC. This is done via annotations on the Pod:

 - `cni.projectcalico.org/hwAddr` - set the MAC address on the pod-ns end of the pod's veth
 - `cni.projectcalico.org/ipAddrs` - set the ipv4 and optionally, ipv6 address of the pod's veth.

Development is underway to release NIC-specific versions of these annotations, which
would allow the same functionality but for secondary network attachments. the shape
of those annotations is: `cni.projectcalico.org/<nic>.hwAddr` and `cni.projectcalico.org/<nic>.ipAddrs`.

**Upcoming functionality:** To tell Calico to attach a pod's **secondary** NIC to an L2/VRF network,
a Multus NAD config with `type: calico` can be given a reference to a Calico `Network` CR.
This has the effect of informing Calico CNI at ADD time about how the pod veth should be plumbed.

**Upcoming functionality:** Primary NICs can also be attached to L2 segments: instead of referencing
the `Network` from a NAD, it is referenced from an annotation on the Pod: `cni.projectcalico.org/network`.

### Forklift API change

For the secondary-NIC preservation, no Forklift API's need to change, since users can opt-in via Multus NAD config.
Forklift will detect that `Network` field in the NAD to discover if it should set NIC-specific annotations, and to spur it into validating the network is compatible with the VM.

For the **primary** NIC preservation, however, there is no NAD to serve as a home for the user's intent to
opt in to the Calico features, so:

One new optional  struct-type field on NetworkMap destinations: `calico`, valid on
`type: pod` entries. Its presence, as an empty struct, opts the VM's primary
NIC into identity preservation:

```
map:
  - source: {id: net-1234}
    destination:
    type: pod
      calico: {}
```

This leads to the migrated pod using Kubevirt bridge binding (with the pod-bridge live-migration
annotation set) and preserves the source MAC (stamps the Calico `hwAddr` annotation).

When the Plan sets `preserveStaticIPs: true`, the source IP is stamped too, via Calico's
`cni.projectcalico.org/ipAddrs` annotation; with `preserveStaticIPs: false`, Calico
assigns a fresh IP and a Warn condition explains the trade-off for
guests with static in-guest configuration.

When the destination ships the cluster-scoped `projectcalico.org/v3 Network` resource, the primary NIC can additionally attach to a named L2 network. A VLAN ID must also be set, since the Network CR can be home to
many subnets on different VLANs:

```
        calico:
          network: production-vlans
          vlan: 100
```

NADs (secondary NIC) referencing a VRF network (routed, no VLANs) are also valid
destinations. VRF networks are secondary-only, since a VRF-only
primary would cut the pod off from cluster DNS, the API server, and
kubelet probes.

###  Validation

Plans are validated against the destination's actual capabilities, and problems surface as Plan conditions:

Requests the install cannot satisfy are Critical, e.g.: 
- Calico absent entirely, the Network resource not shipped when an L2 attach isrequested,
- Per-VM address problems: a preserved IP not covered by a Calico `IPPool`.

Risky-but-workable configurations are Warnings, e.g.:
- preservation without preserveStaticIPs,
- NADs in Calico's L3-IPAM mode (no identity preservation possible).

The plain calico: {} shape keeps working on installs that do not ship
the Network resource; only the L2/VRF requests are rejected there.

Variants which do not support secondary-NIC Calico address annotations
can still opt for primary-NIC preservation.

### Scope and limits

vSphere sources only in this PR right now, other providers reject a calico field. 

IPv4 only; at most one IPv4 per NIC.

The controller gains read access to projectcalico.org `networks`, `ippools`, `felixconfigurations`, and `bgppeers`.

### Testing

Unit coverage across:
- the CRD types,
- the Calico client helpers,
- the vSphere validator (every rejection reason) and builder (bindings and
annotations per case),
- per-provider rejection tests.
- There is no automated end-to-end Calico case; behaviour was validated manually
against a live cluster.

### Open questions

I have not integrated the fetching of Calico resources into the Inventory model; I understand this may pose
an issue for you. Please let me know if you'd like to see this reformed.


mrnold:
Linking to #6203 

The core of this seems straightforward enough, setting some annotations for a particular network type. I have read only a little bit about Calico though, and couldn't quite figure out how it works with forklift. Is the idea to do a regular migration from VMware to a Kubernetes cluster with Calico installed, and then having these annotations enables Calico to do something special with the migrated VMs? Or does Calico work with VMware/Kubernetes in some other way?

aaaaaaaalex:
Hiya @mrnold, thanks for having a look, and sorry for the delay in replying.

Some of the features I referenced in this code (e.g. NIC-specific annotations) are still unreleased so documentation is scarce at the moment, sorry if there was any confusion! Yes this PR would aim to facilitate migrations into Calico clusters with Kubevirt as the virtualization orchestrator, and Multus for multi-NIC support.

Calico has support for requesting a particular IP address with our `cni.projectcalico.org/ipAddrs` (and a hwAddr counterpart also) and we're adding more features to support multiple NICs across Calico and Calico Enterprise (including the upcoming Network CRD that this PR is querying).

I'm still very much deep in WIP territory, but with this PR we'd like to detect the variant of Calico running, and depending on the answer, do one / many / all of the following:
 - Detect a VM's primary IP and MAC, instruct Calico to preserve and perform IPAM for them during migration.
 - Detect a VM's secondary NIC addresses and do the same.
 - Detect  `Network` CRs are present and valid in the case where a NIC should be on a particular VLAN and must be hooked up to the host's trunk by Calico CNI.

I haven't addressed all of these points in this PR yet (only naively programmed per-interface annotations), and will hold the PR in draft until they're all addressed.

aaaaaaaalex:
Hi again @mrnold , sorry for my radio silence on this one.

I've beefed up this PR a bit, I hope you don't mind. I've rewritten the PR description to be as comprehensive as possible regarding the Calico support. 

Since last time, I've also added Calico support to VM primary NICs. This has necessitated a small addition to the NetworkMap CRD - please let me know if you have any feedback about it. I've made sure to include unit testing for everything added. I've also been testing this code in a real environment, migrating with various configs, and will be continuing to do so for the duration of this PR.

I believe this PR is ready for review. Looking forward to hearing from you.

aaaaaaaalex:
Hi @mrnold , did you get a chance to review this? Let me know if you'd like any changes made. Thx

mrnold:
> Hi @mrnold , did you get a chance to review this? Let me know if you'd like any changes made. Thx

I haven't managed to go through the whole thing yet, only small pieces here and there. I am wondering though, would it be possible to make it more self-contained? Like so we don't have to have a bunch of no-op validators in every non-vsphere provider? Every time I see these interfaces get bigger I think "there must be a better way".

Maybe some new infrastructure like the CsiImportPlugin, but for networks, if that's possible - I haven't really thought it through. Do you have any thoughts on what you would need to accomplish something like that?

yaacov:
@aaaaaaaalex hello, welcome to forklift, we talked offline about your PR, we have some concerns.

a. we don't run calico on our dev clusters, and we are not familiar with the basic settings, we have a directory for enhancment - docs/enhancments - can you add a doc for calico support:
   a.1 yaml we can apply and see the result, in your case we will need examples of how to set up minimal calico env, and run a vm using the calico interfaces
   a.2 a mapping example we can actually try and use.
b. we will need you to provide yaml and logs of demo runs you run in your lab so we will see how it should look like
c. we will not be able to test this feature because our CI also does not install calico, we will need your help with that.

about this PR, it's big :-) and we are unfamiliar with the subject, can you start with a smaller PR adding a new inventory information for OCP providers, that add information about calico support in this cluster ? it will help to ease us into calico networking 🐈

aaaaaaaalex:
Thanks a mill @mrnold & @yaacov , yes this PR got pretty hefty :sweat_smile: sorry about that.

I'll get to creating an enhancement doc as priority - will start a fresh PR and go from the start in increments.
The L2 functionality is still in development but we're approaching release time now so I'll be able to provide concrete installation steps very soon.

Regarding the inventory: I was hesitant to pollute it with Calico resources for the first draft - currently fetching Calico resources directly with a dedicated client, but happy to integrate some Calico info into the inventory with your go-ahead.

I also totally agree with @mrnold about the interface bloat - and actually after revisiting my additions to the validation interfaces, I think I can & should relocate my code elsewhere, so that will stop the interface creep.

Regarding Calico-specific additions, if you would rather keep 3rd-party-specific code nicely isolated, I'm happy to write the Calico validation / building as a plugin of some sort? Right now, this PR adds a calico-specific option to the NetworkMap items, but if you prefer I can make it generic and extensible. LMK

yaacov:
> happy to integrate some Calico info into the inventory with your go-ahead.

My concerns are that we don't have resources to test / verify, and we don't know how things look when they work, and that we might un intentionally break another flow. 

Anything small we can easily verify, understand with as much "howto" will be helpful. Also we need to make sure we don't break other flows by introducing a new one, small contained patch will help us verify that.

aaaaaaaalex:
Hi folks, I'm closing this PR in favour of an enhancement doc proposal here: https://github.com/kubev2v/forklift/pull/8554

I've kept the first PR down to simply: the doc, and a basic inventory amendment which I think shouldn't have any effect on the functionality of the codebase.

Thanks :pray:

PR #6937, opened 2026-06-05
MTV-4165 | Detect boot disk for in-place warm guest conversion

Hazanel:
In-place virt-v2v does not produce KubeVirt YAML with bootOrder, so warm migrations left DetectedBootDisk unset and the vSphere builder defaulted bootOrder to the first disk. Detect the boot disk with guestfish after in-place conversion, expose it on a /bootdisk endpoint, and have the plan controller consume it when /vm returns 204.

- Add guestfish-based DetectBootDiskIndex and sort mount paths numerically
- Expose boot disk index from the virt-v2v pod HTTP server
- Set vm.DetectedBootDisk from /bootdisk during in-place conversion
- Initialize vSphere bootDisk to -1 to avoid wrong bootOrder on failure
- Surface BootDiskNotDetected on the VM when detection is unavailable
- Add unit tests for parsing, guestfish flow, and /bootdisk handler

Ref: https://redhat.atlassian.net/browse/MTV-4165
Resolves: MTV-4165

### DIFF
--- a/cmd/virt-v2v/entrypoint.go
+++ b/cmd/virt-v2v/entrypoint.go
@@ -96,13 +96,28 @@ func main() {
 				Message: warningMsg,
 			})
 		}
+		// For in-place conversions, detect the boot disk from the converted
+		// disk images so the controller can set the correct boot order on the
+		// target VM. Regular (non-in-place) conversions get this from the
+		// KubeVirt YAML that virt-v2v produces.
+		var bootDiskIndex *int
+		if env.IsInPlace {
+			if idx, detectErr := convert.DetectBootDiskIndex(); detectErr != nil {
+				fmt.Printf("WARNING: Failed to detect boot disk index: %v\n", detectErr)
+			} else {
+				bootDiskIndex = &idx
+				fmt.Printf("Detected boot disk index: %d\n", idx)
+			}
+		}
+
 		// In the remote migrations we can not connect to the conversion pod from the controller.
 		// This connection is needed for to get the additional configuration which is gathered either form virt-v2v or
 		// virt-v2v-inspector. We expose those parameters via server in this pod and once the controller gets the config
 		// the controller sends the request to terminate the pod.
 		if convert.IsLocalMigration {
 			s := server.Server{
-				AppConfig: env,
+				AppConfig:     env,
+				BootDiskIndex: bootDiskIndex,
 			}
 			err = s.Start()
 			if err != nil {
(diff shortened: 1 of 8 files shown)


Hazanel:
hey @coderabbitai - please take a look, also suggest a better way if you can think of any

Hazanel:
Thanks @coderabbitai for the review! please check the changes

Hazanel:
Test plan and evidence logs/yaml’s are attached to the ticket

mrnold (approved):

gwencasey96:
guestfish? Its mentioned in the PR description but the comment in the code says you're using virt-inspector instead

yaacov:
@mrnold @gwencasey96 , AFAIR virt-v2v should find the boot disk automatically [1]
they have a bug about this issue [2] fixed in virt-v2v-2.8.1-10

@Kyl-Bempah [2] bumped the version [3]
from: 1:2.8.1-22.el10_1
to: 1:2.10.0-13.el10_2

so the fix should be in our current main and 2.12 builds

@Hazanel can you check why the root disk does not show in the kubvirt yaml ?
maybe we need to reopen the virt-v2v bug ?

[1] https://libguestfs.org/virt-v2v.1.html
[2] https://redhat.atlassian.net/browse/RHEL-110742
[3] https://github.com/kubev2v/forklift/pull/6583
[4] https://github.com/kubev2v/forklift/pull/6583/changes#diff-7ebf47172c0fc8f107c665b645906134072b7ce447e1af032adddb6647eacccdL551

yaacov:
> @mrnold @gwencasey96 , AFAIR virt-v2v should find the boot disk automatically [1] they have a bug about this issue [2] fixed in virt-v2v-2.8.1-10
> 
> @Kyl-Bempah [2] bumped the version [3] from: 1:2.8.1-22.el10_1 to: 1:2.10.0-13.el10_2
> 
> so the fix should be in our current main and 2.12 builds
> 
> @Hazanel can you check why the root disk does not show in the kubvirt yaml ? maybe we need to reopen the virt-v2v bug ?
> 
> [1] https://libguestfs.org/virt-v2v.1.html [2] https://redhat.atlassian.net/browse/RHEL-110742 [3] #6583 [4] https://github.com/kubev2v/forklift/pull/6583/changes#diff-7ebf47172c0fc8f107c665b645906134072b7ce447e1af032adddb6647eacccdL551

Sorry the in-place flow does not use the `-o kubevirt` so the rhel bug is not connected to this issue.

LGTM

yaacov (approved):

mnecas (commented):
This should be done from virt-v2v output where they provide us informaiton about which disk is boot disk. ie function `GetDiskBootOrderFromYaml` if they don't provide it then we need to open bug on them

mnecas:
Additionally the PR goes strongly against the virt-v2v-wrapper standards by implementing it's own command execution. reverting

PR #6938, opened 2026-06-07
MTV-5702 | Copy customizationScripts ConfigMap to target namespace

Hazanel:
When CustomizationScripts references a ConfigMap in a namespace different from TargetNamespace, the conversion pod fails to mount it because ConfigMap volumes are namespace-local. The pod stays in Pending forever, causing the migration to hang.

Fix:
- Add EnsureCustomizationScriptsConfigMap() that copies the ConfigMap from the source (plan) namespace to TargetNamespace before migration.
- Update podVolumeMounts() to use the generated copy name when the source namespace differs from target.
- Add a configurable pending pod timeout (CONVERSION_POD_PENDING_TIMEOUT env var, default 5 minutes) in the conversion pipeline so that stuck pods fail with a clear error instead of hanging indefinitely.

Ref: https://redhat.atlassian.net/browse/MTV-5702
Resolves: MTV-5702

### DIFF
--- a/pkg/controller/plan/kubevirt.go
+++ b/pkg/controller/plan/kubevirt.go
@@ -162,8 +162,9 @@ const (
 
 // Vddk v2v conf
 const (
-	ExtraV2vConf = "extra-v2v-conf"
-	VddkConf     = "vddk-conf"
+	ExtraV2vConf         = "extra-v2v-conf"
+	VddkConf             = "vddk-conf"
+	CustomizationScripts = "customization-scripts"
 
 	VddkAioBufSizeDefault  = "16"
 	VddkAioBufCountDefault = "4"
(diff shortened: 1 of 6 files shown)


Hazanel:

Hazanel:
Thanks @coderabbitai , please check again

pkg/controller/plan/kubevirt.go
```diff
@@ -1297,6 +1298,48 @@ func genVddkConfConfigMapName(plan *api.Plan) string {
 	return fmt.Sprintf("%s-%s-", plan.Name, VddkConf)
 }
 
+func genCustomizationScriptsConfigMapName(plan *api.Plan) string {
+	return fmt.Sprintf("%s-%s", plan.Name, CustomizationScripts)
+}
```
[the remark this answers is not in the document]
Hazanel: The same pattern (plan.Name-<suffix>) is used by genExtraV2vConfConfigMapName and genVddkConfConfigMapName. Fixing only our function would create an inconsistency. If this needs to be addressed, it should be done as a separate PR covering all three generators.

pkg/controller/plan/kubevirt.go
```diff
@@ -1297,6 +1298,48 @@ func genVddkConfConfigMapName(plan *api.Plan) string {
...
+			Namespace: sourceNamespace,
+		},
+		configMap,
+	)
+	if err != nil {
+		return liberr.Wrap(err)
+	}
+
+	err = ensureConfigMap(configMap, genCustomizationScriptsConfigMapName, r.Plan, r.Destination.Client)
+	if err == nil {
```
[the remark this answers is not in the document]
Hazanel: ensureConfigMap is intentionally create-only ,same pattern used by EnsureExtraV2vConfConfigMap and EnsureVddkConfConfigMap.
Each migration run creates a fresh Migration object, so this runs once at begin()

pkg/controller/plan/kubevirt.go
```diff
@@ -1297,6 +1298,48 @@ func genVddkConfConfigMapName(plan *api.Plan) string {
...
+}
+
+// Ensure the ConfigMap referenced by CustomizationScripts exists in TargetNamespace.
+// If the ConfigMap lives in a different namespace, it is copied to the target.
+func (r *KubeVirt) EnsureCustomizationScriptsConfigMap() error {
+	if r.Plan.Spec.CustomizationScripts == nil {
+		return nil
+	}
+
+	sourceNamespace := r.Plan.Spec.CustomizationScripts.Namespace
```
yaacov: its the config map namespace, we use source to denote the source vm, so using source may be confusing with the source vm namespace (different then the config map)

Hazanel: Done - thanks for noticing

pkg/controller/plan/kubevirt.go
```diff
@@ -1297,6 +1298,48 @@ func genVddkConfConfigMapName(plan *api.Plan) string {
...
+			Name:      r.Plan.Spec.CustomizationScripts.Name,
+			Namespace: sourceNamespace,
+		},
+		configMap,
+	)
+	if err != nil {
+		return liberr.Wrap(err)
+	}
+
+	err = ensureConfigMap(configMap, genCustomizationScriptsConfigMapName, r.Plan, r.Destination.Client)
```
yaacov, line 1333: this call may create a new configmap that is used only during migration and should be deleted after migration is done, I don't see the cleanup code to delete it, what am i missing?

Hazanel: The copied ConfigMap is not explicitly cleaned up ,this follows the same pattern as EnsureExtraV2vConfConfigMap() and EnsureVddkConfConfigMap(), which also use ensureConfigMap and have no explicit deletion. 
Cleanup happens implicitly when the target namespace is torn down. If we want explicit cleanup, it should be addressed as a follow-up for all three Ensure*ConfigMap functions together.

yaacov: > Cleanup happens implicitly when the target namespace is torn down

AFAIK, target namespace is not cleaned up, what am i missing?

> should be addressed as a follow-up for all three Ensure*ConfigMap functions together.

I hope we don't leave configmaps after migration, that is not nice, if we do, we should open a jira ticket with explanation what happen and how to reproduce, @Hazanel can you check we don't leave leftover configmaps we don't need , if we do, we need to track it in jira?

Hazanel: Yes, we do need to file a separate jira for this

yaacov (approved):

yaacov:
lgtm - will need to open JIRA ticket for the cleanup

PR #6958, opened 2026-06-08
MTV-5723 | Delete copied ConfigMaps from target namespace after migration

Hazanel:
Add CleanupCopiedConfigMaps() to remove extra-v2v-conf, customization-scripts, and vddk-conf ConfigMaps that were copied to TargetNamespace during migration.
The cleanup runs in the migration end() path regardless of outcome. Deterministic-name CMs are deleted by name, vddk-conf CMs (which use GenerateName) are found by label selector.

Ref: https://redhat.atlassian.net/browse/MTV-5723
Resolves: MTV-5723

### DIFF
--- a/pkg/controller/plan/migration.go
+++ b/pkg/controller/plan/migration.go
@@ -309,6 +309,8 @@ func (r *Migration) Archive() {
 		}
 	}
 
+	r.kubevirt.CleanupCopiedConfigMaps()
+
 	for _, vm := range r.Plan.Status.Migration.VMs {
 		dontFailOnError := func(err error) bool {
 			if err != nil {
@@ -1668,6 +1670,7 @@ func (r *Migration) end() (completed bool, err error) {
 	} else if succeeded > 0 {
 		// if the migration didn't fail and at least one VM succeeded,
 		// then the migration succeeded.
+		r.kubevirt.CleanupCopiedConfigMaps()
 		r.Log.Info("Migration [SUCCEEDED]")
 		snapshot.SetCondition(
 			libcnd.Condition{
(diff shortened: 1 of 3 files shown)


Hazanel:

Hazanel:
Thanks @coderabbitai, addressed your comments, please take a look

pkg/controller/plan/migration.go
```diff
@@ -1628,6 +1628,7 @@ func (r *Migration) end() (completed bool, err error) {
 			succeeded++
 		}
 	}
+	r.kubevirt.CleanupCopiedConfigMaps()
```
mnecas: This should be triggered from the cleanup not at the end of migration.
Now it will cleanup no matther what. The cleanup is triggered manually or when the migration passes. If something fails it woudl be hard to debug

mnecas (changes requested):

pkg/controller/plan/migration.go
Hazanel: Good point, I was also hesitant about that.
I moved CleanupCopiedConfigMaps() out of the unconditional end() path. It now runs only on success (in the succeeded > 0 path) and during Archive() (manual cleanup). 

@coderabbitai, what do you think?

Hazanel: Hey @coderabbitai, IMHO the mixed result scenario is already handled correctly. The cleanup is inside else if succeeded > 0 , which only executes when failed == 0.
If any VM failed, the if failed > 0 branch runs instead and cleanup is skipped.

mnecas: @Hazanel @coderabbitai which succeeded > 0 do you mean?
we call cleanup after the migraiton finishes sucesfully, this is newly added https://github.com/kubev2v/forklift/blob/7014738fbb8345b4ec9d55398fab6a5054ad0906/pkg/controller/plan/migration.go#L1611-L1615

mnecas: @coderabbitai wouldn't it be better to move it to the cleanup function instead to have everything at one place?

mnecas (approved):

PR #6988, opened 2026-06-08
MTV-5738 | Add multi-array RDM disambiguation and configurable NAA prefixes

borod108:
## Summary
- **Multi-array disambiguation**: When multiple storage map entries match the same vendor (e.g. two PowerMax arrays), disambiguate by comparing the RDM's NAA with each candidate datastore's backing device NAAs using longest-common-prefix matching, instead of erroring on ambiguous matches
- **Comprehensive NAA prefixes**: Expand vendor prefix list to cover all known OUIs for supported vendors, including multiple OUIs per vendor (e.g. Dell EMC Symmetrix + VMAX for PowerMax) and add VML format support
- **ConfigMap override**: Deploy a `forklift-naa-oui-map` ConfigMap that admins can edit to add custom NAA→vendor mappings without rebuilding the controller

## Test plan
- [x] Unit tests pass for `vendorFromNAA`, `findStorageMapEntriesForVendor`, `disambiguateRDMByNAA`, `extractNAAHex`, `commonPrefixLen` (all new/updated)
- [x] End-to-end migration with mixed VMDK+RDM VM on brnocp cluster completed successfully — RDM disk resolved to Dell PowerStore via NAA prefix, VMDK boot disk to Pure FlashArray
- [ ] Verify ConfigMap override: create `forklift-naa-oui-map` with a custom prefix and confirm it's picked up at runtime


operator/roles/forkliftcontroller/templates/controller/configmap-naa-oui-map.yml.j2
```diff
@@ -0,0 +1,10 @@
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  labels:
+    app: {{ app_name }}
+    service: {{ controller_service_name }}
+  name: {{ naa_oui_map_configmap_name }}
+  namespace: {{ app_namespace }}
+data: {}
```
mnecas: just realised that if we deploy it with operator than it will get always overriden by the ansible operator

mnecas (changes requested):

pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -1350,7 +1351,49 @@ func (r *Builder) PopulatorVolumes(vmRef ref.Ref, annotations map[string]string,
...
+					"vendor", rdmVendor, "storageClass", effectiveMapped.Destination.StorageClass)
+			} else if disk.Datastore.ID == ds.ID {
+				effectiveMapped = mapped
+			} else {
+				continue
+			}
+
+			if effectiveMapped.OffloadPlugin == nil || effectiveMapped.OffloadPlugin.VSphereXcopyPluginConfig == nil {
+				continue
+			}
```
mnecas: can we move it into separate function?

operator/roles/forkliftcontroller/templates/controller/configmap-naa-oui-map.yml.j2
borod108: Done - removed the ConfigMap template (your commit f76f113e5). The controller handles the missing ConfigMap gracefully — admins can create it manually if they need custom NAA prefix overrides.

pkg/controller/plan/adapter/vsphere/builder.go
borod108: Done - extracted the NAA datastore lookup into `lookupDatastoreNAA()` method. Also fixed the bug where RDM disks were using the outer-loop datastore instead of the resolved `effectiveMapped.Source` for NAA label lookup.

mnecas (approved):

PR #7008, opened 2026-06-08
MTV-5529 | Remove the retry/reusing of DI conversion CRs

solenoci:
Issue: DI conversion CR reusing has a bug where it is reusing wrong conversions from a different plan. The fix could to properly reuse the conversions with a defined flow.
https://github.com/kubev2v/forklift/pull/6846 However that fix is complicated and could introduce more bugs.

Fix: Instead of implementing the proper reusing of resources https://github.com/kubev2v/forklift/pull/6846 simplify the logic to not reuse DI conversions at all for now. Later we can add the proper method defined in the other PR.

Resolves: MTV-5529

### DIFF
--- a/pkg/controller/conversion/context/doc.go
+++ b/pkg/controller/conversion/context/doc.go
@@ -27,11 +27,6 @@ const (
 	LabelPlanNamespace  = "plan-namespace"
 	LabelMigration      = "migration"
 	LabelVM             = "vmID"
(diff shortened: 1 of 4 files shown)


Hazanel:
LGTM, but one minor NP: Add a unit test for GetDeepInspectionConversion returning nil when a same VM CR exists under a different plan UID would be a good regression guard,  that's the exact bug this PR fixes.

Hazanel (approved):

PR #7011, opened 2026-06-08
chore(deps): Network Policy RBAC

Kyl-Bempah:
Headstart on a requirement for the CSV to explicitly define the network policy for OCP 5

### DIFF
--- a/operator/streams/downstream/mtv-operator.clusterserviceversion.yaml
+++ b/operator/streams/downstream/mtv-operator.clusterserviceversion.yaml
@@ -87,6 +87,14 @@ spec:
   install:
     spec:
       deployments: null
+      clusterPermissions:
+        - serviceAccountName: forklift-operator
+          rules:
+            - apiGroups:
+                - networking.k8s.io
+              resources:
+                - networkpolicies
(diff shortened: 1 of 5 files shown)


operator/streams/downstream/mtv-operator.clusterserviceversion.yaml
```diff
@@ -85,6 +85,14 @@ spec:
...
     spec:
       deployments: null
+      clusterPermissions:
+        - serviceAccountName: forklift-operator
+          rules:
+            - apiGroups:
+                - networking.k8s.io
+              resources:
+                - networkpolicies
+              verbs: ["*"]
```
mnecas: Shouldn't this be in the rbac not in the CSV?

Kyl-Bempah: So this is actually part of the OCP 5 changeover requirements https://redhat.atlassian.net/browse/OCPSTRAT-2593 so it needs to be in the CSV but I also added to the RBAC for standalone usage

solenoci (approved):

PR #7012, opened 2026-06-08
chore(deps): Updating renovate to fix PR creation

Kyl-Bempah:


mnecas:
@Kyl-Bempah what is the status of this PR? should we get it in or waiting for some other changes?

Kyl-Bempah:
Ideally it should group the golang dependencies together in one PR so the bot stops opening up PRs for every individual library. I'd like @solenoci to take a look as well if he's happy with it or for @coderabbitai to review it. When merged we'll have to have engineers familiar with what we're using the libraries for to validate the updates are warranted and won't break anything

solenoci (approved):

solenoci (approved):
not sure about removal `"automerge": true` and about inclusion of `$schema` (that should be implicit) but we can add it once our deps are being grouped if needed

Kyl-Bempah:
I don't know if the libraries it updates could cause breakages or need containerfile updates so I'd like a human to look at the first few iterations first

PR #7028, opened 2026-06-09
MTV-5736 | Restrict permissions for wait-for-reboot pod

solenoci:
Issue: Pod security was not adequate for the profile used and migration failed on the WFR pod

Fix: Use runtime policy, restrict capabilities, disable privesc, run as default user

Resolves: MTV-5736

Ref: https://redhat.atlassian.net/browse/MTV-5736

### DIFF
--- a/pkg/controller/plan/kubevirt.go
+++ b/pkg/controller/plan/kubevirt.go
@@ -1031,15 +1031,6 @@ func (r *KubeVirt) EnsureWaitForRebootPod(vm *plan.VMStatus) (err error) {
 	activeDeadline := int64(settings.Settings.WindowsRebootTimeout + 600)
 	automount := true
 
-	seccompProfile := core.SeccompProfile{Type: core.SeccompProfileTypeRuntimeDefault}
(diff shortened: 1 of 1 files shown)


pkg/controller/plan/kubevirt.go
```diff
@@ -1027,13 +1030,27 @@ func (r *KubeVirt) EnsureWaitForRebootPod(vm *plan.VMStatus) (err error) {
 
 	activeDeadline := int64(settings.Settings.WindowsRebootTimeout + 600)
 	automount := true
+
+	seccompProfile := core.SeccompProfile{Type: core.SeccompProfileTypeRuntimeDefault}
+	if settings.Settings.OpenShift {
+		unshare := "profiles/unshare.json"
```
mnecas: Think we should be okay just with defautl
```
		SeccompProfile: &core.SeccompProfile{
			Type: core.SeccompProfileTypeRuntimeDefault,
		},
```

mnecas (changes requested):

mnecas (approved):

PR #7036, opened 2026-06-09
MTV-5737 | Filter prime PVCs from getPVCs to fix OpenStack cold migration

Hazanel:
Prime PVCs gained migration/vmID labels, making getPVCs() return them alongside real disk PVCs. Since they lack the disk-source annotation, mapDisks() produced empty duplicate disk names rejected by KubeVirt. Skip PVCs with a "prime-" name prefix or missing AnnDiskSource.

Ref: https://redhat.atlassian.net/browse/MTV-5737
Resolves: MTV-5737

### DIFF
--- a/pkg/controller/plan/kubevirt.go
+++ b/pkg/controller/plan/kubevirt.go
@@ -2023,11 +2023,12 @@ func (r *KubeVirt) getPVCs(vmRef ref.Ref) (pvcs []*core.PersistentVolumeClaim, e
 		return
 	}
 
-	pvcs = make([]*core.PersistentVolumeClaim, len(pvcsList.Items))
-	for i, pvc := range pvcsList.Items {
-		// loopvar
-		pvc := pvc
-		pvcs[i] = &pvc
+	for i := range pvcsList.Items {
+		pvc := &pvcsList.Items[i]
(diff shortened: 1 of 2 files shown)


Hazanel:
Hey @coderabbitai, please take a look.

Do you think the solution is sufficient? is there a better way? 
I was hesitant perhaps only the "--prime" filter is enough

Hazanel:
Thanks @coderabbitai,  Both guards are cheap, so keeping both is a reasonable belt-and-suspenders approach

yaacov:
@coderabbitai hi,

AFAIK the prime- PVCs should not have the labels from labelselctor , can you identify were we set this labels?

yaacov (approved):
LGTM

np - the prime PVCs should not be in the list in the first place, we might have another bug that insert the labels prime PVCs incorectly

yaacov:
https://github.com/kubev2v/forklift/pull/6945 - the PR adding the labels to the populators

yaacov:
@mnecas hi, when we added labels to the prime pvc we created regression in the way we collect disks, this is mitigated by this PR, maybe we should revert the label change and fix the broken UI in a cleaner way ?

PR #7056, opened 2026-06-09
MTV-5746 | fix Ansible default for feature_windows_wait_for_reboot to match Go

AmenB:
The Go controller defaults FEATURE_WINDOWS_WAIT_FOR_REBOOT to true
  via getEnvBool(..., true). The Ansible operator default was false,
  so the Jinja2 template never set the env var on the controller
  deployment. The Go fallback made the feature always true at runtime,
  but silently, without the env var being present.

  This also means users cannot disable the feature via the CR, since
  the template only emits the env var when true (no else branch).

  Ref: https://redhat.atlassian.net/browse/MTV-5746
  Resolves: MTV-5746

### DIFF
--- a/operator/roles/forkliftcontroller/defaults/main.yml
+++ b/operator/roles/forkliftcontroller/defaults/main.yml
@@ -11,7 +11,7 @@ feature_ocp_live_migration: true
 feature_vmware_system_serial_number: true
 feature_vsphere_vmware_driver_removal: false
 feature_windows_registry_network_config: false
-feature_windows_wait_for_reboot: false
+feature_windows_wait_for_reboot: true
 feature_use_conversion_cr: true
 feature_cli_download: true
 feature_mcp_server: false
--- a/operator/.downstream_manifests
+++ b/operator/.downstream_manifests
@@ -3709,7 +3709,7 @@ spec:
                 type: string
               feature_windows_wait_for_reboot:
                 description: 'Enable automatic wait-for-reboot step for Windows VM
-                  migrations (default: false)'
+                  migrations (default: true)'
                 enum:
                 - "true"
                 - "false"
--- a/operator/.upstream_manifests
+++ b/operator/.upstream_manifests
@@ -3709,7 +3709,7 @@ spec:
                 type: string
               feature_windows_wait_for_reboot:
                 description: 'Enable automatic wait-for-reboot step for Windows VM
-                  migrations (default: false)'
+                  migrations (default: true)'
                 enum:
                 - "true"
                 - "false"
--- a/operator/config/crd/bases/forklift.konveyor.io_forkliftcontrollers.yaml
+++ b/operator/config/crd/bases/forklift.konveyor.io_forkliftcontrollers.yaml
@@ -94,7 +94,7 @@ spec:
               feature_windows_wait_for_reboot:
                 type: string
                 enum: ["true", "false"]
-                description: "Enable automatic wait-for-reboot step for Windows VM migrations (default: false)"
+                description: "Enable automatic wait-for-reboot step for Windows VM migrations (default: true)"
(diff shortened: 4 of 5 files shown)


operator/roles/forkliftcontroller/defaults/main.yml
```diff
@@ -11,7 +11,7 @@ feature_ocp_live_migration: true
 feature_vmware_system_serial_number: true
 feature_vsphere_vmware_driver_removal: false
 feature_windows_registry_network_config: false
-feature_windows_wait_for_reboot: false
+feature_windows_wait_for_reboot: true
```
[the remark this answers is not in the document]
AmenB: done

yaacov, line 14: @coderabbitai hi, when we set a feature flag we use it to create env vars on the affected pod,
can you check we do it currectly, e.g. will the environment variable will be added with the currect value when setting the feature to false and true ?

yaacov:
@AmenB welcome 🎉 

Congratulations on your first PR, looks good 👍 

a. Please sing the DCO
b. Please update the description in the PR top comment to use the PR description guidelines:
https://github.com/kubev2v/forklift/blob/main/.cursor/skills/git-commit/SKILL.md
c. Please re-base

Note:
I am missing 
- how to repruduce
- what is expected
- what actually happen
can you add this informtion here or in the jira ? ( disregard if you already added this info )

operator/roles/forkliftcontroller/defaults/main.yml
yaacov: @AmenB , I agree with @coderabbitai, can you re-check and fix if needed ?

AmenB:
Thanks @yaacov! 🙏

Addressed all your points:
- **DCO**: Signed-off-by added ✅
- **PR description**: Updated to follow the commit message guidelines ✅
- **Rebase**: Rebased on latest main ✅

**How to reproduce / Expected / Actual** — full details are in the Jira ticket [MTV-5746](https://redhat.atlassian.net/browse/MTV-5746), but in short:

**Steps to Reproduce:**
1. Check the Ansible default: `feature_windows_wait_for_reboot: false` in `operator/roles/forkliftcontroller/defaults/main.yml`
2. Check the Go default: `getEnvBool(FeatureWindowsWaitForReboot, true)` in `pkg/settings/features.go`
3. On a live cluster, verify the env var is absent from the controller deployment:
   `oc get deployment forklift-controller -n openshift-mtv -o jsonpath='{.spec.template.spec.containers[0].env}' | python3 -m json.tool | grep -A1 WINDOWS_WAIT`

**Actual:** Ansible default is `false` → Jinja2 template skips the env var → Go fallback to `true` → feature is silently always enabled, and users cannot disable it via the CR.

**Expected:** Ansible default should be `true` to match Go, and the env var should be explicitly set on the deployment.

[MTV-5746]: https://redhat.atlassian.net/browse/MTV-5746?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

operator/roles/forkliftcontroller/defaults/main.yml
AmenB: i added what proposed, and fixed

solenoci:
Hi, @AmenB  please run `make update-manifests` and then commit `operator/` dir. @yaacov I think we should start going over the configuration of MTV, (FC CRD in code, it's settings automatically generated from code, etc.. like we talked about)

solenoci (approved):
LGTM, once CI finishes we can merge (and backport)

PR #7060, opened 2026-06-10
MTV-5740 | Consider snapshot presence in CheckSnapshotRemove.

mrnold:
Issue: if the final snapshot is deleted before the WaitForFinalSnapshotRemoval phase, the plan fails even though all the work is actually done. For VMware this happens becase the removal task ID is no longer present, so checking the task status returns an error.

Fix: in CheckSnapshotRemove, if task status returns "not found", also check the inventory to see if there's a snapshot present. Wait until the snapshot is removed from the inventory before considering the snapshot removed. 

Reference:[MTV-5740]( https://redhat.atlassian.net/browse/MTV-5740)

Resolves: MTV-5740

[MTV-5740]: https://redhat.atlassian.net/browse/MTV-5740?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/controller/plan/adapter/vsphere/client.go
+++ b/pkg/controller/plan/adapter/vsphere/client.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	liburl "net/url"
+	"strings"
 
 	"github.com/kubev2v/forklift/pkg/apis/forklift/v1beta1"
 	planapi "github.com/kubev2v/forklift/pkg/apis/forklift/v1beta1/plan"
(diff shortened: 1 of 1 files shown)


mrnold:
I repeated Ameen’s test steps with a patched controller image, and the Plan succeeded. I did not do any extensive negative testing yet though.

pkg/controller/plan/adapter/vsphere/client.go
```diff
@@ -318,10 +319,27 @@ func (r *Client) CheckSnapshotRemove(vmRef ref.Ref, precopy planapi.Precopy, hos
 	r.Log.Info("Check Snapshot Remove", "vmRef", vmRef, "precopy", precopy)
 
 	taskInfo, err := r.getTaskById(vmRef, precopy.RemoveTaskId, hosts)
-	if err != nil {
+	if err == nil {
+		return r.checkTaskStatus(taskInfo)
+	}
+
+	notFound := strings.Contains(err.Error(), fmt.Sprintf("task %s not found", precopy.RemoveTaskId))
+	alreadyDeleted := strings.Contains(err.Error(), "has already been deleted")
```
yaacov: @coderabbitai hi, parsing err msg is fragile, can you suggest more robust check ?
what are the tradeoffs of using more robust check vs current approach?

yaacov: @mrnold WDYT ^^ ?

mnecas (approved):

pkg/controller/plan/adapter/vsphere/client.go
mrnold: Sure, I like errors.Is better than string checking. I changed it around in a second commit, how does it look now?

yaacov (approved):

PR #7096, opened 2026-06-11
MTV-4165 | Detect boot disk for warm migrations from inspection XML

Hazanel:
Use virt-v2v-in-place's -O flag to produce inspection XML during in-place conversion, then parse the root device and mountpoints to determine the boot disk index.

When the /vm endpoint returns 204 (in-place, no KubeVirt YAML), fall back to inspection XML for boot disk detection. If detection fails, emit a BootDiskNotDetected warning and skip bootOrder assignment instead of defaulting to disk 0.

Translate libvirt-order boot disk indices to VMware disk order in mapDisks so bootOrder is applied on the correct volume for both warm and cold migrations on mixed-bus VMs.

Ref: https://redhat.atlassian.net/browse/MTV-4165
Resolves: MTV-4165

### DIFF
--- a/cmd/virt-v2v/entrypoint.go
+++ b/cmd/virt-v2v/entrypoint.go
@@ -74,12 +74,16 @@ func main() {
 			os.Exit(1)
 		}
 
-		// virt-v2v-inspector
-		err = convert.RunVirtV2VInspection()
-		if err != nil {
-			fmt.Println("Failed to inspect the disk", err)
-			os.Exit(1)
+		// For non-in-place conversions, run virt-v2v-inspector separately.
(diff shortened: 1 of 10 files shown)


Hazanel:
Hey @coderabbitai, please take a look, also look specifically for @mnecas comments on [reverted_pr](https://github.com/kubev2v/forklift/pull/6937) 
1. This should be done from virt-v2v output where they provide us informaiton about which disk is boot disk. ie function GetDiskBootOrderFromYaml
2. the PR goes strongly against the virt-v2v-wrapper standards by implementing it's own command execution

PR #7100, opened 2026-06-12
Add qemu user to the wait for reboot

mnecas:
image


solenoci (approved):

solenoci:
Let's see if the qemu user has necessary permissions as the "default root"

PR #7106, opened 2026-06-12
chore(inventory): skip ghost VMs in vSphere inventory API

mnecas:
During VM clone operations, vSphere can create transient placeholder VM objects that get picked up by the PropertyCollector with an enter event but never receive modify or leave events. These ghost VMs have no UUID, no InstanceUUID, and no host association, making them invalid migration candidates. They cause confusing duplicate entries in the inventory and misleading critical concerns about permissions.

Filter out these ghost VMs in the REST API List handler alongside the existing IsTemplate filter.


solenoci (approved):
lgtm

PR #7128, opened 2026-06-14
fix(powermax): UnMap always cleans up storage group and masking view

rgolangh:
## Summary

**Original problem:** PowerMax \`UnMap\` relied on a \`CleanupXcopyInitiatorGroup\` flag set by the caller to decide whether to delete the masking view and storage group. On partial failure (before xcopy starts), this flag was never set — so the masking view and xcopy storage group were leaked on the array. Since PowerMax uses randomly generated names (\`xcopy-<random>-SG\`), every failed run accumulated new orphans.

**Fix: struct-field guards instead of external flag**

\`UnMap\` now uses the struct fields (\`maskingViewID\`, \`storageGroupID\`) as guards:
- \`maskingViewID\` is only set when \`Map\` successfully creates the masking view
- \`storageGroupID\` is only set when \`EnsureClonnerIgroup\` creates the xcopy SG
- Their presence is a reliable indicator that the object exists and needs cleanup, regardless of which code path got there

**Continue-on-error:** all three cleanup steps (delete masking view, remove volume from SG, delete SG) now run even if an earlier step fails. Errors are accumulated and returned via \`errors.Join\` so the caller is informed of partial failures without blocking subsequent steps.

**Remap guard:** On PowerMax, volumes are never removed from their original storage groups during xcopy — only the temporary xcopy SG is created/deleted. The post-cleanup remap loop in the caller calls \`Map\` for each original group, but \`Map\` would previously try to operate on the already-deleted xcopy SG. After \`UnMap\` clears \`p.storageGroupID\`, \`Map\` uses it as a guard (\`storageGroupID == ""\` → no-op) matching the same sentinel used by \`UnMap\` to decide whether to delete.

## Test plan
- [ ] Existing unit tests pass
- [ ] Verify on a PowerMax array that a failed xcopy run no longer leaves orphaned storage groups / masking views
- [ ] Verify successful xcopy runs still clean up properly

### DIFF
--- a/cmd/vsphere-copy-offload-populator/internal/populator/storage.go
+++ b/cmd/vsphere-copy-offload-populator/internal/populator/storage.go
@@ -4,11 +4,6 @@ import (
 	"github.com/kubev2v/forklift/cmd/vsphere-copy-offload-populator/internal/vmware"
 )
 
-const (
-	// CleanupXcopyInitiatorGroup is the key to signal cleanup of the initiator group.
-	CleanupXcopyInitiatorGroup = "cleanupXcopyInitiatorGroup"
-)
-
 //go:generate go run go.uber.org/mock/mockgen -destination=mocks/storage_mock_client.go -package=mocks . StorageApi
 type StorageApi interface {
 	VMDKCapable
--- a/cmd/vsphere-copy-offload-populator/internal/powermax/powermax.go
+++ b/cmd/vsphere-copy-offload-populator/internal/powermax/powermax.go
@@ -368,48 +368,36 @@ func (p *PowermaxClonner) UnMap(_ string, targetLUN populator.LUN, mappingContex
 
 	ctx := context.TODO()
 
-	cleanup, ok := mappingContext[populator.CleanupXcopyInitiatorGroup]
-	if ok && cleanup.(bool) {
-		p.log.V(2).Info("full cleanup requested, deleting masking view", "masking_view", p.maskingViewID)
+	if p.maskingViewID != "" {
+		p.log.V(2).Info("deleting masking view", "masking_view", p.maskingViewID)
 		err := retryOnTransient(ctx, p.log, "DeleteMaskingView", func() error {
 			return p.client.DeleteMaskingView(ctx, p.symmetrixID, p.maskingViewID)
 		})
 		if err != nil {
-			return fmt.Errorf("failed to delete masking view: %w", err)
+			p.log.Info("failed to delete masking view", "masking_view", p.maskingViewID, "err", err)
 		}
-
-		p.log.V(2).Info("removing volume from storage group", "volume", targetLUN.ProviderID, "storage_group", p.storageGroupID)
-		err = retryOnTransient(ctx, p.log, "RemoveVolumesFromStorageGroup", func() error {
-			_, e := p.client.RemoveVolumesFromStorageGroup(ctx, p.symmetrixID, p.storageGroupID, false, targetLUN.ProviderID)
-			return e
-		})
-		if err != nil {
-			return fmt.Errorf("failed removing volume from storage group:  %w", err)
-		}
-
-		p.log.V(2).Info("deleting storage group", "storage_group", p.storageGroupID)
-		err = retryOnTransient(ctx, p.log, "DeleteStorageGroup", func() error {
-			return p.client.DeleteStorageGroup(ctx, p.symmetrixID, p.storageGroupID)
-		})
-		if err != nil {
-			return fmt.Errorf("failed to delete storage group: %w", err)
-		}
-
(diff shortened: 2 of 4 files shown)


cmd/vsphere-copy-offload-populator/internal/populator/storage.go
```diff
@@ -4,11 +4,6 @@ import (
 	"github.com/kubev2v/forklift/cmd/vsphere-copy-offload-populator/internal/vmware"
 )
 
-const (
```
rgolangh, line 7: for the record, no storage backend uses this anymore, since we now rely on the internal state of the clonner, to track the objects it created.
Plus, the clonner that primarily creates objects is Powermax, the rest locate existing host or host group objects.

cmd/vsphere-copy-offload-populator/internal/powermax/powermax.go
```diff
@@ -143,281 +113,155 @@ func (p *PowermaxClonner) EnsureClonnerIgroup(_ string, clonnerIqn []string) (po
...
-	err := retryOnTransient(ctx, p.log, "RemoveVolumesFromStorageGroup", func() error {
-		_, e := p.client.RemoveVolumesFromStorageGroup(ctx, p.symmetrixID, p.storageGroupID, false, targetLUN.ProviderID)
-		return e
-	})
-	if err != nil {
-		return fmt.Errorf("failed removing volume from storage group:  %w", err)
 	}
 
-	p.log.Info("volume unmapped successfully", "volume", targetLUN.ProviderID)
+	klog.Infof("volume %s unmapped and cleaned up", targetLUN.ProviderID)
```
amitosw15: we need to return the error (if occurred)

rgolangh: you are right, I see this cross all the changes. will change

cmd/vsphere-copy-offload-populator/internal/powermax/powermax.go
```diff
@@ -363,54 +364,50 @@ func (p *PowermaxClonner) ResolvePVToLUN(pv populator.PersistentVolume) (populat
 }
 
 // UnMap implements populator.StorageApi.
+// Continues through all cleanup steps even on error, so a masking view deletion
+// failure does not prevent storage group cleanup. All errors are returned.
```
shrihn-dell: Please be noted that you cannot delete a Storage Group (SG) if it is still actively tied to a Masking View (MV). If the API call to delete the MV fails (due to a timeout or other transient errors), the subsequent call to delete the storage group will be rejected by Unisphere with a dependency error. If a delete masking view fails then it should be treated as a blocking condition.

cmd/vsphere-copy-offload-populator/internal/powermax/powermax.go
```diff
@@ -363,54 +364,50 @@ func (p *PowermaxClonner) ResolvePVToLUN(pv populator.PersistentVolume) (populat
...
-		p.log.V(2).Info("full cleanup requested, deleting masking view", "masking_view", p.maskingViewID)
+	if p.maskingViewID != "" {
+		p.log.V(2).Info("deleting masking view", "masking_view", p.maskingViewID)
 		err := retryOnTransient(ctx, p.log, "DeleteMaskingView", func() error {
 			return p.client.DeleteMaskingView(ctx, p.symmetrixID, p.maskingViewID)
 		})
 		if err != nil {
-			return fmt.Errorf("failed to delete masking view: %w", err)
+			p.log.Info("failed to delete masking view", "masking_view", p.maskingViewID, "err", err)
+			errs = append(errs, err)
```
shrihn-dell: If a previous cleanup attempt partially succeeded (e.g., it deleted the masking view but failed on storage group), a retry of Unmap will call DeleteMaskingView again. Trying to delete an already deleted object returns a not found error. We should explicitly check for this status code and treat it as a successful deletion instead of appending it to errors. This ensures partial retries don't report false failures.

amitosw15: ptal

cmd/vsphere-copy-offload-populator/internal/powermax/powermax.go
amitosw15: ptal

amitosw15:
@shrihn-dell 
lgtm?

shrihn-dell:
> @shrihn-dell lgtm?
LGTM.
PowerMax team is running a quick test with these changes. Can it wait until the sanity check?

shrihn-dell (approved):
LGTM. PowerMax team is running a quick test with these changes. Can it wait until the sanity check?

mnecas:
Yes we have till Friday before code freeze for 2.12.1

borod108:
\backport 2.12

PR #7132, opened 2026-06-15
MTV-1878 | Proactive vSphere privilege check on provider validation

### LINKED ISSUE #7131: MTV-1878: Proactively check vSphere service account privileges during provider validation
## Problem

When a VMware (vSphere) provider is created in MTV, the provider validation checks credentials and connectivity but does **not** check whether the service account has the vSphere privileges required for migration. Users only discover permission issues when migrations fail, leading to frustration and wasted time.

The required privileges are documented at the [MTV VMware prerequisites](https://docs.redhat.com/en/documentation/migration_toolkit_for_virtualization/2.7/html-single/installing_and_using_the_migration_toolkit_for_virtualization/index#vmware-prerequisites_mtv), but there is no runtime validation.

## Solution

Add a proactive privilege check during provider validation that:

- Runs after the connection test succeeds, using the vSphere `AuthorizationManager` API
- Checks the service account's effective privileges on the vCenter root folder
- Reports missing privileges as a **non-blocking Warning** condition (provider still becomes Ready)
- Groups missing privileges by functional area (snapshots, power management, VDDK disk access, datastore access)
- Prunes required privileges against what the vCenter version supports to avoid false positives on older versions
- Skips the check for direct ESXi connections (different privilege model)

The approach is modeled on the [OpenShift installer's permission checking](https://github.com/openshift/installer/blob/main/pkg/asset/installconfig/vsphere/permissions.go), adapted for MTV's architecture.

## JIRA Reference

[MTV-1878](https://redhat.atlassian.net/browse/MTV-1878)

[MTV-1878]: https://redhat.atlassian.net/browse/MTV-1878?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

snecklifter:
## Summary

Adds a proactive vSphere privilege check during provider validation. When a VMware provider is created, the system checks whether the service account has the required vSphere privileges for migration and reports any missing privileges as a **non-blocking warning** condition (`InsufficientPrivileges`).

## Approach

- Uses `AuthorizationManager.HasPrivilegeOnEntity()` with the session key to check privileges on the vCenter root folder
- Required privileges are derived from the actual govmomi operations MTV performs (snapshot, power, disk access, datastore)
- Privileges are pruned against the vCenter's reported `privilegeList` to avoid false positives on older vCenter versions
- ESXi direct connections skip the check (different privilege model)
- **Reuses the inventory collector's already-open govmomi client** rather than opening a disposable connection — the privilege check runs in `getUpdates()` right after the client connects and caches results thread-safely for the reconciler to read via `MissingPrivileges()`, following the `hasParity` pattern

## Files Changed

| File | Change |
|------|--------|
| `pkg/controller/provider/container/vsphere/permissions.go` | Privilege definitions, pruning, comparison helpers, `checkPermissionsWithClient()`, and `checkAndCachePermissions()` |
| `pkg/controller/provider/container/vsphere/permissions_test.go` | Ginkgo tests for pure functions + `MissingPrivileges()` accessor |
| `pkg/controller/provider/container/vsphere/collector.go` | Added cached privilege state fields, `MissingPrivileges()` accessor, call to `checkAndCachePermissions()` in `getUpdates()` |
| `pkg/controller/provider/validation.go` | `validateVSpherePrivileges()` reads cached state from the collector via type assertion |

## Test Plan

- [x] Unit tests for `pruneToAvailablePrivileges`, `comparePrivileges`, `flattenPrivileges`, `FormatMissing`
- [x] Unit tests for `requiredPrivileges` integrity (no duplicates, non-empty)
- [x] Unit tests for `MissingPrivileges()` accessor (not-checked, cached results, all-granted)
- [x] `go vet ./...` passes
- [x] `go build ./...` passes
- [x] Full test suite passes (51 packages)
- [ ] Manual integration test: create a vSphere provider with limited permissions and verify warning appears

## CodeRabbit review suggestions not implemented

1. **Unit tests for `CheckPermissions` method** — requires mocking the govmomi client (`AuthorizationManager`, `SessionManager`, property collector). Low value given the pure-function tests already cover the logic.
2. **Unit tests for `validateVSpherePrivileges` branches** — requires mocking the `Reconciler` and `container`. The function is now a thin read from cached state.

Resolves: https://github.com/kubev2v/forklift/issues/7131

### DIFF
--- a/pkg/controller/provider/container/vsphere/collector.go
+++ b/pkg/controller/provider/container/vsphere/collector.go
@@ -297,6 +297,9 @@ type Collector struct {
 	cancel func()
 	// has parity.
 	parity bool
+	// Cached missing privileges from the last check.
+	missingPrivs   *privilegeCheckResult
+	missingPrivsMu *sync.RWMutex
 }
 
 // New collector.
@@ -307,11 +310,12 @@ func New(db libmodel.DB, provider *api.Provider, secret *core.Secret) *Collector
 			provider.GetNamespace(),
 			provider.GetName()))
 	return &Collector{
-		url:      provider.Spec.URL,
-		provider: provider,
-		secret:   secret,
-		db:       db,
-		log:      nlog,
+		url:            provider.Spec.URL,
+		provider:       provider,
+		secret:         secret,
+		db:             db,
+		log:            nlog,
+		missingPrivsMu: &sync.RWMutex{},
 	}
 }
 
@@ -343,6 +347,17 @@ func (r *Collector) Reset() {
 // Reset.
 func (r *Collector) HasParity() bool {
 	return r.parity
+}
+
+// MissingPrivileges returns the cached privilege check results.
+// The bool indicates whether the check has completed.
+func (r *Collector) MissingPrivileges() ([]MissingPrivileges, bool) {
+	r.missingPrivsMu.RLock()
+	defer r.missingPrivsMu.RUnlock()
+	if r.missingPrivs == nil {
+		return nil, false
+	}
+	return r.missingPrivs.missing, true
 }
 
 // Follow
@@ -431,6 +446,24 @@ func (r *Collector) getUpdates(ctx context.Context) error {
 		return err
 	}
 	defer r.close()
+	r.checkAndCachePermissions(ctx)
+
+	permCtx, cancelPerm := context.WithCancel(ctx)
+	defer cancelPerm()
+
+	go func() {
+		ticker := time.NewTicker(PrivilegeCheckInterval)
+		defer ticker.Stop()
+		for {
+			select {
+			case <-permCtx.Done():
+				return
+			case <-ticker.C:
+				r.checkAndCachePermissions(permCtx)
+			}
+		}
+	}()
+
 	about := r.client.ServiceContent.About
 	err = r.db.Insert(
 		&model.About{
--- a/pkg/controller/provider/validation.go
+++ b/pkg/controller/provider/validation.go
@@ -16,6 +16,7 @@ import (
 	hvutil "github.com/kubev2v/forklift/pkg/controller/hyperv"
 	"github.com/kubev2v/forklift/pkg/controller/ova"
 	"github.com/kubev2v/forklift/pkg/controller/provider/container"
+	vsphereCollector "github.com/kubev2v/forklift/pkg/controller/provider/container/vsphere"
 	vsphere "github.com/kubev2v/forklift/pkg/controller/provider/model/vsphere"
 	libcnd "github.com/kubev2v/forklift/pkg/lib/condition"
 	liberr "github.com/kubev2v/forklift/pkg/lib/error"
@@ -47,6 +48,7 @@ const (
 	ConnectionInsecure      = "ConnectionInsecure"
 	SSHReady                = "SSHReady"
 	SSHNotReady             = "SSHNotReady"
+	InsufficientPrivileges  = "InsufficientPrivileges"
 	SMBCSIDriverNotReady    = "SMBCSIDriverNotReady"
 	SMBMountFailed          = "SMBMountFailed"
 	WaitingForService       = "WaitingForService"
@@ -118,6 +120,10 @@ func (r *Reconciler) validate(provider *api.Provider) error {
 		return liberr.Wrap(err)
 	}
 	err = r.testConnection(provider, secret)
+	if err != nil {
+		return liberr.Wrap(err)
+	}
+	err = r.validateVSpherePrivileges(provider)
 	if err != nil {
 		return liberr.Wrap(err)
 	}
@@ -496,6 +502,64 @@ func (r *Reconciler) testConnection(provider *api.Provider, secret *core.Secret)
 					err.Error()),
 			})
 	}
+
+	return nil
+}
+
+// Validate vSphere service account privileges.
+func (r *Reconciler) validateVSpherePrivileges(provider *api.Provider) error {
+	if provider.Type() != api.VSphere {
+		return nil
+	}
+	if provider.Status.HasBlockerCondition() {
+		return nil
+	}
+
+	log.Info(
+		"Validating vSphere privileges.",
+		"name", provider.Name,
+		"namespace", provider.Namespace,
+	)
+
+	collector, found := r.container.Get(provider)
+	if !found {
+		return nil
+	}
+	vsphColl, ok := collector.(*vsphereCollector.Collector)
+	if !ok {
+		return nil
+	}
+
+	missing, checked := vsphColl.MissingPrivileges()
+	if !checked {
+		return nil
+	}
+
+	if len(missing) == 0 {
+		provider.Status.DeleteCondition(InsufficientPrivileges)
+		return nil
+	}
+
+	totalMissing := 0
+	var items []string
+	for _, m := range missing {
+		totalMissing += len(m.Privileges)
+		items = append(items, m.Privileges...)
+	}
+
+	provider.Status.SetCondition(
+		libcnd.Condition{
+			Type:     InsufficientPrivileges,
+			Status:   True,
+			Reason:   "PrivilegesNotGranted",
+			Category: Warn,
+			Message: fmt.Sprintf(
+				"The vSphere service account is missing %d privilege(s) required for migration. "+
+					"See the suggestion field in the Provider's YAML for details.",
+				totalMissing),
+			Suggestion: vsphereCollector.FormatMissing(missing),
+			Items:      items,
+		})
 
 	return nil
 }
--- a/pkg/controller/provider/container/vsphere/permissions.go
+++ b/pkg/controller/provider/container/vsphere/permissions.go
@@ -1,0 +1,208 @@
+package vsphere
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"time"
+
+	api "github.com/kubev2v/forklift/pkg/apis/forklift/v1beta1"
+	liberr "github.com/kubev2v/forklift/pkg/lib/error"
+	"github.com/vmware/govmomi"
+	"github.com/vmware/govmomi/object"
+	"github.com/vmware/govmomi/session"
+	"github.com/vmware/govmomi/vim25/mo"
+)
+
+const PrivilegeCheckInterval = 10 * time.Minute
+
+type privilegeGroup struct {
+	Description string
+	Privileges  []string
+}
+
+// MissingPrivileges groups missing privilege IDs by their functional description.
+type MissingPrivileges struct {
+	Group      string
+	Privileges []string
+}
+
+type privilegeCheckResult struct {
+	missing []MissingPrivileges
+}
+
+// requiredPrivileges defines the vSphere privileges MTV needs for migration,
+// grouped by functional area. Derived from the govmomi operations in
+// pkg/controller/plan/adapter/vsphere/client.go and
+// pkg/controller/conversion/client.go.
+var requiredPrivileges = []privilegeGroup{
+	{
+		Description: "Virtual Machine Snapshot Operations",
+		Privileges: []string{
+			"VirtualMachine.State.CreateSnapshot",
+			"VirtualMachine.State.RemoveSnapshot",
+		},
+	},
+	{
+		Description: "Virtual Machine Power Management",
+		Privileges: []string{
+			"VirtualMachine.Interact.PowerOn",
+			"VirtualMachine.Interact.PowerOff",
+		},
+	},
+	{
+		Description: "Virtual Machine Disk Access (VDDK)",
+		Privileges: []string{
+			"VirtualMachine.Provisioning.DiskRandomRead",
+			"VirtualMachine.Provisioning.GetVmFiles",
+			"VirtualMachine.Config.ChangeTracking",
+		},
+	},
+	{
+		Description: "Datastore Access",
+		Privileges: []string{
+			"Datastore.Browse",
+			"Datastore.FileManagement",
+		},
+	},
+}
+
+// pruneToAvailablePrivileges intersects required privilege groups against the
+// set of privileges the vCenter actually supports. This prevents false positives
+// on older vCenter versions that lack certain privilege IDs.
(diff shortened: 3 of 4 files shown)


yaacov:
@snecklifter welcome 🎉 🍰

pkg/controller/provider/validation.go
```diff
@@ -500,6 +506,59 @@ func (r *Reconciler) testConnection(provider *api.Provider, secret *core.Secret)
 	return nil
 }
 
+// Validate vSphere service account privileges.
+func (r *Reconciler) validateVSpherePrivileges(provider *api.Provider, secret *core.Secret) error {
+	if provider.Type() != api.VSphere {
+		return nil
```
yaacov, line 512: np:
for ssh validation we added a log line, not sure it's needed, but it's nice to use the same template.

pkg/controller/provider/validation.go
```diff
@@ -500,6 +506,59 @@ func (r *Reconciler) testConnection(provider *api.Provider, secret *core.Secret)
...
+	}
+	if provider.Status.HasBlockerCondition() {
+		return nil
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
```
yaacov: np:
in hyper-v timeout in this file we created a var for the timeout, so 
a. it will be clear from the name what the timeout do
b. it will be easy to play with timeout shorter longer depending on enviorment

pkg/controller/provider/validation.go
```diff
@@ -121,6 +123,10 @@ func (r *Reconciler) validate(provider *api.Provider) error {
 	if err != nil {
 		return liberr.Wrap(err)
 	}
+	err = r.validateVSpherePrivileges(provider, secret)
```
yaacov: @coderabbitai hi,

in testConnection we open a disposable client to check connection
in validateVSpherePrivileges will open a disposable client again to check permissions

the validation is called for each reconcile cycle ( each 30s )

in the inventory we keep an open client we can use the check the connection and permissions, does it make sense to move the connection and the new permissions check to the inventory keep alive open connection and pass the information via the inventory, for example like we do in "hasParity" method ?

yaacov: @snecklifter hi, we have a pre existing issue with the tesConnection method that open a connection each 30s, your suggestion is making the existing issue even harder to fix, please consider using the already open client in the inventory and pass the information to the controller via a public method like we do for "hasParity"

yaacov: @coderabbitai hi, can we check the permissions inside the inventory but before parity is achieved ?

pkg/controller/provider/validation.go
snecklifter: Thanks for the suggestion! Added an entry log line following the SSH validation template:

```go
log.Info(
    "Validating vSphere privileges.",
    "name", provider.Name,
    "namespace", provider.Namespace,
)
```

pkg/controller/provider/validation.go
snecklifter: Good point about the named timeout pattern. In the updated implementation the timeout is no longer needed — `validateVSpherePrivileges` now reads cached state from the collector (no network call), so there is nothing to time out on.

pkg/controller/provider/validation.go
snecklifter: Refactored as suggested. The privilege check now:

1. Runs inside `getUpdates()` right after `r.connect(ctx)` — reuses the already-open `r.client`, no new connection
2. Caches results in a `*privilegeCheckResult` with `sync.RWMutex` protection
3. Exposes results via `MissingPrivileges()` public accessor (following the `hasParity` pattern)
4. `validateVSpherePrivileges` in the reconciler reads cached state via `r.container.Get(provider)` + type assertion — zero extra sessions per reconcile cycle

The check runs before parity is achieved, so users get early feedback on missing privileges.

snecklifter: Done — the updated implementation uses the inventory's open client and the `hasParity`-style cached accessor. No more disposable connections for the privilege check.

pkg/controller/provider/container/vsphere/collector.go
```diff
@@ -431,6 +446,7 @@ func (r *Collector) getUpdates(ctx context.Context) error {
 		return err
 	}
 	defer r.close()
+	r.checkAndCachePermissions(ctx)
```
yaacov, line 449: this implementation will alert users when connecting to a vsphere provider that lacks permissions, like we want.

caveat - if the permissions change, we will  not pick the change, e.g. if we connect without permissions and then add them, the cached data will not get updated.

@coderabbitai hi, WDYT about this concern, what are the tradeoffs if we want to watch/pool  permissions changes and update the condition ?

yaacov: @snecklifter hi, current implementation of test permissions once, is fine, the only concern is that after fixing permissions on vsphere side the user will need to re-connect manually.

another option is to create a goroutine that will periodically test permissions with a ctx that can cancel it on getupdate exit.

my 2 cents is to leave as is, and document that users will need to trigger a re-connect to re-check missing permissions.

@coderabbitai what is the complexity of adding a new goroutine to getupdate that test permissions   with a ctx  that will cancel it when the main loop exit ?

yaacov: @snecklifter up to you ^^ ? i am find with keeping as is and document the need to re-connect, or implementing the suggestion.

snecklifter: @yaacov agree with the bot FWIW, happy to update the PR, just let me know.

yaacov: lets do it

yaacov:
lgtm

@snecklifter - can you create a controller image with the fix, run it and document your findings in the Jira tickets ?

yaacov (approved):
lgtm

( code looks good to me, i did not compile and test the image )

snecklifter:
Hi @yaacov 

> @snecklifter - can you create a controller image with the fix, run it and document your findings in the Jira tickets ?

I have created and image and pushed to quay.io/chribrow/forklift-controller:mtv-1878

Unfortunately I'm out of bandwidth to do the run and document part so will have to revisit this in the future.

snecklifter:
# MTV-1878 Test Report: Proactive vSphere Privilege Check

**Date:** 2026-06-16
**PR:** https://github.com/kubev2v/forklift/pull/7132
**Image:** quay.io/chribrow/forklift-controller:mtv-1878
**Cluster:** OpenShift 4.17.4 / Kubernetes v1.30.5
**MTV Version:** Upgraded 2.8.7 → 2.9.8, plus CRDs from upstream main

---

## 1. Environment Setup

### Cluster
```
OpenShift 4.17.4, Kubernetes v1.30.5
MTV Operator: v2.9.8 (upgraded from v2.8.7 via OLM subscription)
```

### Pre-requisites for testing image from upstream main
The controller image was built from the current `main` branch which is ahead of any
released version. To run it on the 2.9.8 operator, these additional steps were needed:
1. Installed CRDs not yet in any release:
   `forklift.konveyor.io_hypervproviderservers.yaml`,
   `forklift.konveyor.io_ovaproviderservers.yaml`
2. Added missing env var `VIRT_V2V_IMAGE_XFS` to the `main` container
3. Scaled down operator to prevent reconciliation reverting the image

### Providers tested
| Provider | Target | Purpose |
|----------|--------|---------|
| `vmware` | `vcsrs00-vc.infra.demo.redhat.com` (real vCenter 8.0.3) | Error path, happy path verification |
| `vcsim-test` | govmomi vcsim (pod in cluster) | Happy path, warning path |

---

## 2. Test Results

### Test 1: Graceful Error Path (Real vCenter)

The service account `ocpvirt-9chnv@infra` on the real vCenter lacks `Authorization.*`
privileges needed to call `HasPrivilegeOnEntity`. This exercises the graceful failure path.

**Inventory container log:**
```json
{"level":"info","logger":"provider","msg":"Validating vSphere privileges.","name":"vmware","namespace":"openshift-mtv"}
{"level":"error","logger":"collector|vsphere","msg":"Privilege check failed.",
  "provider":"openshift-mtv/vmware",
  "error":"ServerFaultCode: Permission to perform this operation was denied."}
{"level":"info","logger":"collector|vsphere","msg":"Initial parity.","provider":"openshift-mtv/vmware","duration":0.844}
```

**Provider status:** `Ready`, no `InsufficientPrivileges` condition set.

| Check | Result |
|-------|--------|
| Error logged at error level | ✅ |
| Provider not blocked | ✅ Phase: Ready |
| No false InsufficientPrivileges condition | ✅ |
| Inventory collection proceeds | ✅ Initial parity achieved |
| Check repeats on subsequent reconcile cycles | ✅ |

---

### Test 2: Happy Path — All Privileges Granted (vcsim)

A vcsim pod was deployed in the cluster and a vSphere provider (`vcsim-test`) was created
pointing to it. vcsim's default user has admin privileges, so `HasPrivilegeOnEntity`
returns all privileges as granted.

**Inventory container log:**
```json
{"level":"info","logger":"provider","msg":"Validating vSphere privileges.","name":"vcsim-test","namespace":"openshift-mtv"}
{"level":"info","logger":"collector|vsphere","msg":"Initial parity.","provider":"openshift-mtv/vcsim-test","duration":0.093}
```
No `Privilege check failed` error — the check completed successfully.

**Provider status:** `Ready`, no `InsufficientPrivileges` condition.

| Check | Result |
|-------|--------|
| Privilege check succeeds | ✅ No error logged |
| No InsufficientPrivileges condition | ✅ |
| Provider reaches Ready | ✅ Phase: Ready |
| All standard conditions present | ✅ ConnectionInsecure, ConnectionTestSucceeded, Validated, InventoryCreated, Ready |

---

### Test 3: Warning Path — Missing Privileges (vcsim + simulated)

Since vcsim doesn't enforce role-based permission restrictions in `HasPrivilegeOnEntity`,
a test image was built with a hook that injects simulated missing privileges for providers
whose URL contains "vcsim". The simulated missing privileges:
- `VirtualMachine.State.CreateSnapshot`
- `VirtualMachine.State.RemoveSnapshot`
- `Datastore.Browse`
- `Datastore.FileManagement`

**Inventory container log:**
```json
{"level":"info","logger":"collector|vsphere","msg":"TEST: Simulating missing privileges for vcsim provider.",
  "provider":"openshift-mtv/vcsim-test"}
{"level":"info","logger":"provider","msg":"Condition added.",
  "provider":{"name":"vcsim-test","namespace":"openshift-mtv"},
  "condition":{
    "type":"InsufficientPrivileges",
    "status":"True",
    "reason":"PrivilegesNotGranted",
    "category":"Warn",
    "message":"The vSphere service account is missing 4 privilege(s) required for migration...",
    "suggestion":"MISSING VSPHERE PRIVILEGES:\n\nVirtual Machine Snapshot Operations:\n  - VirtualMachine.State.CreateSnapshot\n  - VirtualMachine.State.RemoveSnapshot\n\nDatastore Access:\n  - Datastore.Browse\n  - Datastore.FileManagement\n\nRESOLUTION:\nAssign a vSphere role with these privileges to the service account\non the vCenter root folder with propagation enabled.\nSee: https://docs.redhat.com/en/documentation/...",
    "items":["VirtualMachine.State.CreateSnapshot","VirtualMachine.State.RemoveSnapshot","Datastore.Browse","Datastore.FileManagement"]
  }}
```

**Provider status (full YAML):**
```yaml
status:
  conditions:
  - category: Warn
    type: ConnectionInsecure
    status: "True"
  - category: Required
    type: ConnectionTestSucceeded
    status: "True"
  - category: Advisory
    type: Validated
    status: "True"
  - category: Warn
    type: InsufficientPrivileges
    status: "True"
    reason: PrivilegesNotGranted
    message: "The vSphere service account is missing 4 privilege(s) required for migration.
      See the suggestion field in the Provider's YAML for details."
    items:
    - VirtualMachine.State.CreateSnapshot
    - VirtualMachine.State.RemoveSnapshot
    - Datastore.Browse
    - Datastore.FileManagement
  - category: Required
    type: InventoryCreated
    status: "True"
  - category: Required
    type: Ready
    status: "True"
  phase: Ready
```

| Check | Result |
|-------|--------|
| InsufficientPrivileges condition set | ✅ |
| Category is Warn (non-blocking) | ✅ |
| Missing privileges listed in items[] | ✅ 4 items |
| Human-readable message | ✅ |
| Detailed suggestion with resolution steps | ✅ Includes doc link |
| Provider still reaches Ready | ✅ Phase: Ready |
| Condition updates on subsequent reconciles | ✅ "Condition updated" in logs |

---

## 3. Summary

All three code paths validated on a live OpenShift cluster:

| Scenario | Image | Provider | Result |
|----------|-------|----------|--------|
| **Error path**: privilege check API denied | mtv-1878 | Real vCenter | ✅ Error logged, provider Ready, no false condition |
| **Happy path**: all privileges granted | mtv-1878 | vcsim | ✅ No error, provider Ready, no condition set |
| **Warning path**: some privileges missing | mtv-1878-test | vcsim (simulated) | ✅ InsufficientPrivileges Warn condition set, provider Ready |

### Architecture Verification

- Privilege check runs inside the inventory collector (`collector.go:449 → permissions.go`), **not** via disposable connections
- The check uses the collector's already-open `govmomi.Client` — no extra vSphere sessions created
- Results are cached via `missingPrivs` pointer with `sync.RWMutex` protection
- Periodic re-check goroutine fires every 10 minutes (verified via repeated log entries)
- Reconciler reads cached state via `MissingPrivileges()` accessor — no vSphere calls during reconcile

---

## 4. Cleanup

```bash
# Delete test resources
oc delete provider vcsim-test -n openshift-mtv
oc delete svc vcsim -n openshift-mtv
oc delete pod vcsim -n openshift-mtv
oc delete secret vcsim-credentials -n openshift-mtv

# Restore operator (will reconcile controller image back to release version)
oc scale deployment forklift-operator-ansible -n openshift-mtv --replicas=1

# Remove test CRDs if desired
oc delete crd hypervproviderservers.forklift.konveyor.io
oc delete crd ovaproviderservers.forklift.konveyor.io
```

PR #7137, opened 2026-06-16
MTV-5784 | Fix nil pointer panic in SupportsVolumePopulators for conversion plans

mnecas:
Conversion plans (spec.type: conversion) don't require a storage map, so Context.Map.Storage is legitimately nil. SupportsVolumePopulators dereferences it unconditionally, causing a panic when a conversion plan is canceled during cleanup.

Add a nil guard for Map.Storage before accessing its spec.

Resolves: MTV-5784


solenoci (commented):
lgtm, though make the linter happy, please

PR #7138, opened 2026-06-16
chore: Add 2.12 to operator description

mnecas:


solenoci (approved):

solenoci:
lgtm, however, I think we could remove this and only link to https://access.redhat.com/product-life-cycles?product=migration%20toolkit%20for%20virtualization

PR #7190, opened 2026-06-18
MTV-5780 | Same-array detection and destination-aware adapter selection for copy-offload

amitosw15:
## Summary

`GetDatastoreActiveAdapters` queries vSphere multipath topology to find which HBAs have active
paths to the source VM's datastore. These adapters are passed to the destination storage
provider's `EnsureClonnerIgroup`, which registers the ESXi host's identity with the destination
array so it can map the target LUN.

### Current (bad) behavior, concretely

Say the host has two HBAs: HBA1 and HBA2. HBA1 has an active multipath to the **source** VM's
datastore. HBA2 has no path to the source at all — but it's the one adapter physically zoned to
reach the **destination** array. HBA1 has no connectivity to the destination.

7190 drawio

`GetDatastoreActiveAdapters` only ever asks one question: "which HBA has an active path to the
source datastore's LUN?" That's HBA1 — so that's what it returns, every time, regardless of what
the destination actually needs. HBA2 is never considered, because nothing in the function looks
at the destination side at all.

That HBA1 identity is exactly what gets registered with the destination array via
`EnsureClonnerIgroup`. The registration call itself may well succeed — array-side
igroup-add/host-registration APIs typically just record a well-formed identifier, they don't
validate physical SAN reachability. So the failure doesn't surface until later, when the actual
data path never comes up, because the destination was only ever reachable via HBA2.

The function is asking the wrong question. It resolves connectivity from the *source* side and
assumes that's good enough for the *destination* — which only happens to be true when source and
destination are the same array.

Additionally, the existing `hasSANAdapter` fallback only recognized `fc.*` and `iqn.*` prefixes.
After the scini GUID override, PowerFlex adapter IDs become UUIDs — the fallback fired and
replaced the scini adapter with a random FC HBA.

### New behavior: ask the destination, not the source

We don't care which adapter the source uses. We only care which of the host's adapters can
actually reach the **destination** array — independent of anything about the source disk, its
datastore, or whether it happens to be the same array or a different one. That's the shift this
PR makes: instead of deriving the adapter list from source-side multipath, it's derived directly
from destination-side reachability.

**Commit 1 — Same-array detection (`ArrayIdentifier` interface):**

Each storage provider implements `MatchesDevice(deviceName string) (bool, error)` to check if a
vSphere device belongs to its array. At adapter selection time, the source datastore's backing
device NAA is checked against the destination provider:

- **Same array** (`MatchesDevice` returns true) → return only source active-path adapters.
  XCOPY possible, no cross-protocol adapters needed.
- **Different array** (`MatchesDevice` returns false) → investigate destination multipath.

**Commit 2 — Destination multipath investigation:**

Instead of blindly returning all host adapters for cross-array migrations, the code walks the
host's SCSI LUN table to find devices already present from the destination array. It collects
only the adapters that have proven multipath paths to those devices.

Three code paths:
1. **Same-array**: source active-path adapters only → XCOPY possible
2. **Cross-array, destination devices on host**: destination multipath adapters → precise,
   proven connectivity. Only adapters that can actually reach the destination.
3. **Cross-array, no destination devices**: SAN-filtered fallback (FC, iSCSI, scini only —
   excludes block controllers)

This approach:
- Preserves XCOPY for same-array migrations
- Returns only adapters with proven connectivity to the destination
- Excludes unzoned FC HBAs, block controllers, and unrelated adapters
- Replaces the fragile `hasSANAdapter` / local-datastore fallback entirely

**Commit 3 — Replace per-LUN `MatchesDevice` scanning with target-port matching:**

The two-commit approach above still calls `MatchesDevice` once per LUN on the host to detect
cross-array destination devices — on a host with thousands of LUNs, that's thousands of REST
calls to the destination array per disk copied. It also still derives "the source array" from
the source datastore, which breaks for RDM disks: an RDM's real backing LUN can sit on a
completely different array than whatever datastore hosts the RDM's own descriptor file.

vSphere already tracks, per host HBA, which SAN target ports it can see
(`HostStorageDeviceInfo.ScsiTopology`) independent of whether any LUN is mapped through that
target yet — FC port-login and iSCSI session establishment happen at the fabric/network level
before any LUN masking. Combined with a single "list my own target ports" call per destination
array instead of one call per LUN, this determines exactly which local HBAs can reach the
destination at O(1) array calls instead of O(LUNs-on-host), without needing to resolve anything
about the source device at all. `GetDatastoreActiveAdapters` is renamed to `GetAdaptersForArray`
and drops its `datastoreName` parameter entirely; PowerFlex/scini is handled by a dedicated
branch checked first, since its FC-shaped adapter identity is a driver-invented value with no
real array-side counterpart to compare against.

`ArrayIdentifier` gains a `TargetPorts()` method alongside the existing `MatchesDevice`, which is
left in place and unused rather than removed. All 9 providers implement `TargetPorts()`; ONTAP,
Pure, PowerStore, and Infinibox are confirmed end-to-end against real hardware (byte-for-byte
matches between array-reported ports and what ESXi's `ScsiTopology` independently reports for the
same array). PowerMax, 3PAR, FlashSystem, and Vantara are implemented against documented
APIs/SDKs but not yet validated live.

## Test plan

Tested on eco lab cluster (ocp-edge114) with image
`quay.io/amitos/vsphere-copy-offload-populator:dst-multipath-dev-amd64`:

| # | Test | Source DS | Dest | Decision | Adapters returned | Result |
|---|---|---|---|---|---|---|
| 1 | FC→ONTAP | eco-fc-ds4 (ONTAP FC) | ONTAP | Same array | 2: [FC, FC] | **PASS** |
| 2 | Local→ONTAP | ecoesxi05_ds1 (local) | ONTAP | Cross-array, 37 dst devices | 3: [iSCSI, FC, FC] | **PASS** |
| 3 | iSCSI→ONTAP | eco-iscsi-ds3 (ONTAP iSCSI) | ONTAP | Same array | 1: [iSCSI] | **PASS** |
| 4 | 3PAR→ONTAP | eco-iscsi-ds1 (3PAR) | ONTAP | Cross-array, 37 dst devices | 1: [iSCSI] | **PASS** |

Cross-validated via SSH into each ESXi host: zero false positives (no adapter returned without
ONTAP multipath paths), zero false negatives (no adapter with ONTAP paths was missed).

Key validation: host-12657 has 7 HBAs (block, 4×FC, iSCSI, scini). Test 2 returned only the 3
with proven ONTAP multipath paths (iSCSI + 2 zoned FC). The 2 unzoned FC HBAs (vmhba65/66) and
the block/scini adapters were correctly excluded.

Real-hardware validation for the target-port matching redesign (host-3078, eco-tlv2 lab):
`selectAdaptersForArray` run against real `ScsiTopology` from a 6-HBA host correctly isolated the
single HBA zoned to Pure out of FC-only and unzoned HBAs, using Pure's real IQN independently
confirmed via Pure's own `/api/1.19/port` endpoint — byte-for-byte match with what ESXi reported.
Same pattern repeated for ONTAP (4 FC WWPNs + 1 iSCSI IQN), PowerStore, and Infinibox.

- [ ] PowerFlex cross-array test (PF→ONTAP or ONTAP→PF) — in progress
- [x] Unit tests for adapter selection and MatchesDevice
- [x] `MatchesDevice` validated against real arrays: ONTAP, 3PAR, PowerFlex, Pure
- [x] `TargetPorts()` validated end-to-end against real hardware: ONTAP, Pure, PowerStore, Infinibox
- [ ] `TargetPorts()` validation pending: PowerMax, 3PAR (lab connectivity issue), FlashSystem, Vantara
- [x] Regression test for RDM disks whose real backing LUN differs from the array behind their own descriptor file


amitosw15:
Vantara approved, all providers are tested

@mnecas Waiting for you buddy

PR #7206, opened 2026-06-21
MTV-5806 | Collect VLAN ID from Hyper-V NICs for network disambiguation

Hazanel:
When multiple NICs on a Hyper-V VM share the same virtual switch but use different VLANs, the network mapping resolver cannot distinguish between them. This change:

- Updates the PowerShell inventory script to query Get-VMNetworkAdapterVlan and include VlanId in the NIC data collected from Hyper-V hosts.
- Propagates VlanId through driver, types, model, and REST API layers.
- Adds a 'vlan' field to the shared ref.Ref struct so NetworkMap entries can specify a VLAN qualifier on the source reference.
- Modifies buildNICResolver to use a composite key (networkID/vlanId) only when multiple NICs on the same VM share a network, preserving backward compatibility for the common single-NIC-per-network case.

Ref: https://redhat.atlassian.net/browse/MTV-5806
Resolves: MTV-5806


Hazanel:

Hazanel:
Hey @coderabbitai, thanks for the review! 
please check the follow-ups

Hazanel:
Hey @coderabbitai - I intentionally skipped it because the existing vSphere provider already uses VlanId

pkg/apis/forklift/v1beta1/ref/ref.go
```diff
@@ -18,6 +18,12 @@ type Ref struct {
 	Namespace string `json:"namespace,omitempty"`
 	// Type used to qualify the name.
 	Type string `json:"type,omitempty"`
+	// VLAN identifier (numeric string, valid range 1-4094) for network disambiguation.
+	// Only semantically meaningful in NetworkMap source references (Hyper-V).
+	// Present in other CRDs due to shared Ref struct but unused there.
+	// +optional
+	// +kubebuilder:validation:Pattern=`^([1-9]|[1-9][0-9]{1,2}|[1-3][0-9]{3}|40[0-8][0-9]|409[0-4])$`
+	Vlan string `json:"vlan,omitempty"`
```
mnecas: Think the `Ref` is too generic for this, could we move it just to the NetworkMap?

Hazanel: Thanks for bringing that up!
I wasn't sure cause there are other fields in Ref which are specific but still there.

I'll create a NetworkSourceRef in the mapping types that embeds ref.Ref and adds the Vlan field. That way it only appears in the NetworkMap CRD schema - wdyt?

Hazanel:
Hey @coderabbitai, please take a look on recent changes in respect to @mnecas request

pkg/controller/plan/adapter/hyperv/builder.go
```diff
@@ -209,21 +210,59 @@ func (r *Builder) mapNetworks(vm *model.VM, object *cnv.VirtualMachineSpec) {
...
+// NIC keys remain plain too, preserving the old 1:1 fallback behavior.
+func buildNICKeys(nics []hyperv.NIC, networkCount map[string]int, vlanQualifiedNetworks map[string]bool) []string {
+	keys := make([]string, len(nics))
 	for i, nic := range nics {
-		nicKeys[i] = nic.Network.ID
+		if networkCount[nic.Network.ID] > 1 && nic.VlanId > 0 && vlanQualifiedNetworks[nic.Network.ID] {
+			keys[i] = nic.Network.ID + "/" + strconv.Itoa(nic.VlanId)
+		} else {
+			keys[i] = nic.Network.ID
+		}
```
mnecas, line 254: trying to understand this part do we need the `networkCount[nic.Network.ID] > 1`? we can have 1 NIC with specific VLAN

mnecas: What are the `keys` used afterwards?

mnecas (changes requested):

pkg/controller/plan/adapter/hyperv/builder.go
mnecas: @coderabbitai wdyt ^^?

mnecas: ack

mnecas (approved):
lgtm

PR #7230, opened 2026-06-22
MTV-5829 | Add retry backoff and auth detection to Hyper-V collector

Hazanel:
The Hyper-V collector retried failed runs with no delay, spamming the host with WinRM requests when credentials were incorrect or basic auth was not enabled.

- Add RetryInterval (5s) sleep on run() failure in Start() loop, matching the pattern used by oVirt, OpenStack, and OVA collectors.

- Introduce ErrUnauthorized sentinel error in the driver package, wrapping WinRM HTTP 401/403 responses so callers use errors.Is() instead of fragile string matching.

- Update Test() to execute a real WinRM command (IsAlive) and return http.StatusUnauthorized on auth failure, triggering ConnectionAuthFailed in the provider controller.

- Propagate auth errors from SMB prefix discovery in Connect()

Ref: https://redhat.atlassian.net/browse/MTV-5829
Resolves: MTV-5829

### DIFF
--- a/pkg/controller/provider/container/hyperv/collector.go
+++ b/pkg/controller/provider/container/hyperv/collector.go
@@ -2,7 +2,9 @@ package hyperv
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"net/http"
 	liburl "net/url"
 	"os"
 	libpath "path"
@@ -12,6 +14,7 @@ import (
 	api "github.com/kubev2v/forklift/pkg/apis/forklift/v1beta1"
 	hvutil "github.com/kubev2v/forklift/pkg/controller/hyperv"
 	model "github.com/kubev2v/forklift/pkg/controller/provider/model/hyperv"
+	"github.com/kubev2v/forklift/pkg/lib/hyperv/driver"
 	libmodel "github.com/kubev2v/forklift/pkg/lib/inventory/model"
 	"github.com/kubev2v/forklift/pkg/lib/logging"
 	core "k8s.io/api/core/v1"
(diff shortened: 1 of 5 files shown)


Hazanel:

Hazanel:
Thanks @coderabbitai - please follow-up

pkg/controller/provider/container/hyperv/collector.go
```diff
@@ -161,9 +164,21 @@ func (r *Collector) HasParity() bool {
 	return r.parity
 }
 
-// Test connect/logout.
-func (r *Collector) Test() (_ int, err error) {
-	err = r.client.Connect(r.provider)
+// Test validates connectivity and credentials against the Hyper-V host.
+func (r *Collector) Test() (status int, err error) {
+	if err = r.client.Connect(r.provider); err != nil {
+		if errors.Is(err, driver.ErrUnauthorized) {
```
yaacov, line 170: @coderabbitai hi, in client.go we return a non typed error, does this check make sense, what am i missing?

pkg/lib/hyperv/driver/errors.go
```diff
@@ -0,0 +1,42 @@
...
+}
+
+// WrapCommandError inspects a WinRM command error.
+func WrapCommandError(err error) error {
+	if err == nil {
+		return nil
+	}
+	if code, ok := httpStatus(err); ok {
+		if code == http.StatusUnauthorized || code == http.StatusForbidden {
+			return fmt.Errorf("%w: %w", ErrUnauthorized, err)
```
yaacov, line 38: @coderabbitai hi, should we return a typed error ? maybe include that actual http err code ?

yaacov (approved):

PR #7240, opened 2026-06-23
MTV-5883 | Schedule shared-disk creator VMs before consumers

mnecas:
Hold back consumer VMs (shared disk + migrateSharedDisks=false) from
the pending queue while any creator VM (shared disk +
migrateSharedDisks=true) has not yet completed. Non-shared VMs are
unaffected.

Ref: https://redhat.atlassian.net/browse/MTV-5883
Resolves: MTV-5883


mnecas:
Tested with 3 VMs, 1 normal, 1 with shared disks (migrate shared disk true) 1 with shared disks (migrate shared disks false)
the one with the migrate shared disks false is waiting for the one to get finished 
image

solenoci (approved):
lgtm

PR #7282, opened 2026-06-29
MTV-5636 | fix vSphere provider to validate server cert against user-provided CA

AmenB:
VerifyTLSConnection() was fetching the server's own certificate with
InsecureSkipVerify=true, adding it to a new cert pool, and validating
against itself — circular self-trust. The user-provided ca.crt was
never used for validation, so any valid PEM was accepted regardless
of whether it matched the vCenter server's actual certificate chain.

This fix uses the ca.crt from the provider secret as the trust anchor
instead of the server's own certificate. If no ca.crt is provided,
it falls back to the system cert pool.

Additionally, the collector blocker check is moved before the
secretChanged branch so the inventory collector is never started
when the connection test fails, and the provider phase is set to
ConnectionFailed instead of remaining stuck at Staging.

Ref: https://redhat.atlassian.net/browse/MTV-5636
Resolves: MTV-5636

### DIFF
--- a/pkg/controller/provider/controller.go
+++ b/pkg/controller/provider/controller.go
@@ -337,6 +337,13 @@ func (r *Reconciler) updateContainer(provider *api.Provider) (err error) {
 	}
 	secretChanged := secret.ResourceVersion != provider.Status.SecretResourceVersion
 
+	if provider.Status.HasBlockerCondition() ||
+		!provider.Status.HasCondition(ConnectionTestSucceeded) {
+		r.Log.V(1).Info(
+			"Provider not ready, postponing.")
+		return
+	}
+
 	if !secretChanged {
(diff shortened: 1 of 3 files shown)


pkg/controller/provider/controller.go
```diff
@@ -337,6 +337,13 @@ func (r *Reconciler) updateContainer(provider *api.Provider) (err error) {
 	}
 	secretChanged := secret.ResourceVersion != provider.Status.SecretResourceVersion
 
+	if provider.Status.HasBlockerCondition() ||
```
yaacov, line 340: @coderabbitai hi, why did we move the collector creation above the secret change check ?

yaacov (approved):

PR #7293, opened 2026-06-30
MTV-5940 | fix OpenStack populator SIGSEGV when c.Log is nil

AmenB:
The OpenStack populator creates a bare `Client{}` without setting `c.Log`.
When `getTLSConfig()` tries to log "CA certificate was not provided" with
`insecureSkipVerify=false` and no CA cert, it dereferences nil and panics
with SIGSEGV.

Initialize a default logger in `Connect()` when `c.Log` is nil, and add
nil guards in `getTLSConfig()` as belt-and-suspenders protection.

Ref: https://redhat.atlassian.net/browse/MTV-5940
Resolves: MTV-5940

### DIFF
--- a/pkg/lib/client/openstack/client.go
+++ b/pkg/lib/client/openstack/client.go
@@ -175,7 +175,9 @@ func (c *Client) Authenticate() (err error) {
 func (c *Client) getTLSConfig() (tlsConfig *tls.Config, err error) {
 	identityUrl, err := url.Parse(c.URL)
 	if err != nil {
-		c.Log.Error(err, "error parsing the URL", "url", c.URL)
+		if c.Log != nil {
+			c.Log.Error(err, "error parsing the URL", "url", c.URL)
+		}
 		return
 	}
 	if identityUrl.Scheme == "https" {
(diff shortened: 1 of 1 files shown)


pkg/lib/client/openstack/client.go
```diff
@@ -175,7 +175,9 @@ func (c *Client) Authenticate() (err error) {
 func (c *Client) getTLSConfig() (tlsConfig *tls.Config, err error) {
 	identityUrl, err := url.Parse(c.URL)
 	if err != nil {
-		c.Log.Error(err, "error parsing the URL", "url", c.URL)
+		if c.Log != nil {
```
yaacov, line 178: @coderabbitai hi, doesn't it make more sense to init the log when we create the clien, instead of checking if it's not nil ?

yaacov: @coderabbitai ^^ ?

yaacov (approved):

yaacov:
@AmenB , please sign the commit to solve the DCO issue

PR #7297, opened 2026-07-01
fix(powermax): skip CurrentMappedGroups to reduce API calls during mi…

jagandasaroju:
During XCOPY-based volume migration, the populator queries all masking views to find the current mapped groups of a LUN. For PowerMax, this is unnecessary because a dynamic masking view is created per disk copy.

This optimization eliminates multiple expensive REST API calls per populator pod (GetVolumeByID, GetMaskingViewList, GetMaskingViewByID), significantly reducing API load during large-scale migrations (e.g., 100 VM migrations).

Testing: Verified with 100 VM migration runs. Masking view related API call count reduced from ~36,000 to ~5700 per migration run when tested with 100 migration plans, each plan having 1 VM with 5 disks. No functional impact observed.

Resolves: None

### DIFF
--- a/cmd/vsphere-copy-offload-populator/internal/powermax/powermax.go
+++ b/cmd/vsphere-copy-offload-populator/internal/powermax/powermax.go
@@ -51,76 +51,8 @@ func (p *PowermaxClonner) UnmapTarget(targetLUN populator.LUN, mappingContext po
 
 // CurrentMappedGroups implements populator.StorageApi.
 func (p *PowermaxClonner) CurrentMappedGroups(targetLUN populator.LUN, mappingContext populator.MappingContext) ([]string, error) {
-	p.log.V(2).Info("querying current mapped groups", "volume", targetLUN.ProviderID)
-
-	ctx := context.TODO()
-	var volume *pmxtypes.Volume
-	err := retryOnTransient(ctx, p.log, "GetVolumeByID", func() error {
-		var e error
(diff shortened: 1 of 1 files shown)


borod108 (approved):
great find! LGTM :)

cmd/vsphere-copy-offload-populator/internal/powermax/powermax.go
```diff
@@ -51,76 +51,8 @@ func (p *PowermaxClonner) UnmapTarget(targetLUN populator.LUN, mappingContext po
...
-		}
-	}
-
-	if len(foundHostGroups) > 0 {
-		p.log.V(2).Info("found mapped groups", "volume", targetLUN.ProviderID, "groups", foundHostGroups)
-	} else {
-		p.log.V(2).Info("no host groups found for volume", "volume", targetLUN.ProviderID)
-	}
-	return foundHostGroups, nil
+	p.log.Info("skipping CurrentMappedGroups for PowerMax as dynamic masking view created per disk copy")
```
mnecas, line 54: what is the dynamic masking view ? is it automatic on the powermax?

borod108: meaning we are creating a short lived masking view, and then delete it - we do not need to search for an existing one ever, and we do not remove any existing one to attach it, so we do not need to save the list of all existing ones to return to it during unmap like in some other cases.

PR #7318, opened 2026-07-02
MTV-5962 | Re-queue provider periodically after auth failure

Hazanel:
Issue:
Previously the reconciler set reQ:0 on ConnectionAuthFailed, permanently stopping reconciliation. Providers stayed stuck even after the host-side fix.

Fix:
Replace with a 3-minute re-queue interval that avoids account lockout while allowing recovery.

Also removes dead constants OvaTimeout and OvaReconcilerRetry.

Ref: https://redhat.atlassian.net/browse/MTV-5962
Resolves: MTV-5962

### DIFF
--- a/pkg/controller/provider/controller.go
+++ b/pkg/controller/provider/controller.go
@@ -60,10 +60,8 @@ import (
 )
 
 const (
-	// Name.
-	Name               = "provider"
-	OvaTimeout         = 10 * time.Minute
-	OvaReconcilerRetry = 5 * time.Second
+	Name         = "provider"
+	AuthRetryReQ = 3 * time.Minute
 )
 
 // Package logger.
@@ -194,7 +192,11 @@ func (r Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (r
 	}()
 
 	defer func() {
-		// Stop reconciliation when auth fails
+		if provider.Status.HasCondition(ConnectionAuthRetry) {
+			result.RequeueAfter = AuthRetryReQ
+			err = nil
+			return
+		}
 		if provider.Status.HasCondition(ConnectionAuthFailed) {
 			result.RequeueAfter = 0
 			err = nil
--- a/pkg/controller/provider/validation.go
+++ b/pkg/controller/provider/validation.go
@@ -41,6 +41,7 @@ const (
 	SettingsNotValid        = "SettingsNotValid"
 	Validated               = "Validated"
 	ConnectionAuthFailed    = "ConnectionAuthFailed"
+	ConnectionAuthRetry     = "ConnectionAuthRetry"
 	ConnectionTestSucceeded = "ConnectionTestSucceeded"
 	ConnectionTestFailed    = "ConnectionTestFailed"
 	InventoryCreated        = "InventoryCreated"
@@ -463,17 +464,7 @@ func (r *Reconciler) testConnection(provider *api.Provider, secret *core.Secret)
 		// When the status is unauthorized controller stops the reconciliation, so the user account does not get locked.
 		// Providing bad credentials when requesting the token results in 400, and not 401.
 		if status == http.StatusUnauthorized || status == http.StatusBadRequest {
-			provider.Status.Phase = ConnectionFailed
-			provider.Status.SetCondition(
-				libcnd.Condition{
-					Type:     ConnectionAuthFailed,
-					Status:   True,
-					Reason:   Tested,
-					Category: Critical,
-					Message: fmt.Sprintf(
-						"Connection auth failed, error: %s",
-						err.Error()),
-				})
+			setAuthFailureConditions(provider, err)
 			return nil
 		}
 		log.Info(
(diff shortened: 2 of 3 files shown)


Hazanel:
Hey @coderabbitai , please take a look.
I wonder whether this implicit fix for all providers is OK and safe or we should make explicit for hyperv, something like:
`if provider.Status.HasCondition(ConnectionAuthFailed) {
    if provider.Type() == api.HyperV {
        result.RequeueAfter = AuthRetryReQ
    } else {
        result.RequeueAfter = 0
    }
    err = nil
    return
}`

solenoci:
When credentials are invalid, this will loop forever

yaacov (approved):

yaacov:
> When credentials are invalid, this will loop forever

yes, but that will happen for any reconcile cycle, not just here

Hazanel:
It's a valid point @solenoci , I wonder what's better, The old code (RequeueAfter = 0) stopped reconciliation permanently, but the provider was stuck forever even after the host-side issue was fixed.
If a user on purpose keep using wrong credentials it will keep looping forever but once that are correct the secret reconcile fires immediately.
The periodic retry specifically covers server-side changes (like WinRM Basic auth being re-enabled) where the Secret doesn't change.

@coderabbitai - wdyt?

solenoci:
I just think that it would be worth to distinguish between invalid creds and invalid provider config

yaacov:
> I just think that it would be worth to distinguish between invalid creds and invalid provider config

make sense, so if it is something that can be fixed from the source target side, we should requeue, but if it can't be fixed, e.g. missing credentials, wait for user changing the credentials.

@coderabbitai can we distinguish between this cases and change the requeu policy depending on the fail type?

yaacov:
> So: "config problem" vs. "credentials problem" is already split by condition type,

@solenoci it looks like this is already done, WDYT ?

> (currently 3 * time.Minute in the code I fetched — worth double-checking against the PR description, which says 5 minutes; might be worth reconciling that discrepancy).

@Hazanel the commit msg now say 5m, WDYT ^^ ?

Hazanel:
@yaacov , yes , just changed it

solenoci:
> The periodic retry specifically covers server-side changes (like WinRM Basic auth being re-enabled) where the Secret doesn't change.

I am worried about this ^, e.g. server side config will be reported as invalid creds, and that we should distinguish, I am not talking about the whole MTV connection testing. That seems correct as coderabbit mentioned. But I don't want to hold this PR if it's needed so I just wanted to raise a concern.

mnecas (changes requested):
If the secrets gets changed the reconcile should be triggered by the watch.

It might be better to use `ConnectionTestFailed` if there is bad connection and the `ConnectionAuthFailed` only for the 401. So we should do first connection to the node.

Hazanel:
The problem here is HyperV specifically : when WinRM Basic auth is disabled on the Windows host, it returns 401, which is the same as wrong password. There's no way to tell them apart at the HTTP level. So a connectivity check first wouldn't help,the node is reachable, it just rejects the auth method.

With RequeueAfter = 0 the provider stays stuck forever until someone restarts the controller, even after the admin re-enables Basic auth.

I can scope this to HyperV only so other providers keep the current behavior. WDYT?

@coderabbitai, @mnecas  - WDYT?

Hazanel:
@coderabbitai - WDYT?

`	defer func() {
		if provider.Status.HasCondition(ConnectionAuthFailed) {

			// HyperV returns 401 when WinRM Basic authentication is disabled.
			// Hence, we need to slow down the reconciliation to avoid account lockout,
			// but keep re-queuing so the provider can recover once the host-side issue is resolved.
			if provider.Type() == api.HyperV {
				result.RequeueAfter = AuthRetryReQ
			} else {
				result.RequeueAfter = 0
			}
			err = nil
			return
		}
	}()`

Hazanel:
@coderabbitai - please follow-up the revised

pkg/controller/provider/controller.go
```diff
@@ -194,9 +202,8 @@ func (r Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (r
 	}()
 
 	defer func() {
-		// Stop reconciliation when auth fails
 		if provider.Status.HasCondition(ConnectionAuthFailed) {
```
mnecas, line 200: @coderabbitai Could we return different Conditon/error for this case it would on reset the result.RequeueAfter

mnecas: @coderabbitai what about disticnt Condition?

Hazanel: Thanks @coderabbitai - please take a look if you're happy with recent changes

Hazanel: check now @coderabbitai

Hazanel: you are a bit nitpick @coderabbitai  - please check

mnecas, line 200: Please add above this line
```
if provider.Status.HasCondition(ConnectionAuthRetry) {
err=nil
return
}
```

pkg/controller/provider/validation.go
```diff
@@ -447,52 +448,8 @@ func (r *Reconciler) testConnection(provider *api.Provider, secret *core.Secret)
...
-				Reason:   Tested,
-				Category: Critical,
-				Message: fmt.Sprintf(
-					"Connection test, failed: %s",
-					err.Error()),
-			})
-	}
-
+	status, connErr := rl.Test()
+	handleConnectionTestResult(provider, status, connErr)
```
mnecas: Please don't move this to separate function, harder to reveiw what are the changes

mnecas (changes requested):

mnecas (approved):

PR #6136, opened 2026-05-03
MTV-3931 | VIB validation and auto-installation via Host controller

amitosw15:
## Problem

VIB (vmkfstools-wrapper) validation was embedded in the populator and ran per-migration, causing
repeated vCenter API calls and race conditions when multiple migrations targeted the same ESXi
host.
Operators had no visibility into fleet-wide VIB state before starting migrations, and no way to
auto-install missing VIBs without manual SSH access.

## What this achieves

**Centralized VIB lifecycle management across three controllers:**

- **Host controller** checks VIB installation status per-host via the vCenter API and reports
  `VIBInstalled` conditions. When `spec.installVIB` is enabled on a Host CR, the controller SSHes
  to the ESXi host, installs the VIB, restarts hostd, and verifies the loaded version on the next
  reconcile.
- **Provider controller** aggregates per-host VIB conditions into `VIBReady`/`VIBNotReady` with
  human-readable host names, giving operators a single place to see fleet-wide VIB state.
- **Plan controller** validates that all ESXi hosts backing a plan's VMs have VIB installed before
  allowing migration to proceed.

**Host CR auto-creation moved to the Host controller's vsphere handler**, where it belongs — the
handler already watches inventory and receives callbacks when hosts appear or change.

## Manual Test Results

Tested on a live ESXi host (standalone, PowerStore iSCSI storage) with controller image
`quay.io/amitos/forklift-controller:host-controller-vib-0.2.0-amd64`.

| # | Scenario | Expected | Result |
|---|----------|----------|--------|
| 1 | **Wrong credentials** — Host CR with bad secret | `VIBInstallFailed` condition with `VCenterConnectionFailed`, 5-min retry | ✅ Pass |
| 2 | **installVIB: false** — Host CR without opt-in | No VIB activity, host reaches Ready | ✅ Pass |
| 3 | **installVIB: true, SSH enabled** — Full install flow | VIB uploaded, hostd restarted, `VIBInstalled` set after requeue | ✅ Pass |
| 4 | **VDDK plan (non-xcopy)** — Plan without VIB requirement | No VIB conditions on plan | ✅ Pass |
| 5 | **Delete Host CR** — Clean removal | CR deleted, no orphaned resources | ✅ Pass |
| 6 | **SSH disabled on ESXi** — VIB install with blocked port 22 | VIB uploaded via vCenter API, `VIBInstallFailed` with `HostdRestartFailed`, 5-min retry | ✅ Pass |
[...]

PR #7320, opened 2026-07-02
MTV-3931 | Add lightweight VIB version validation to populator

amitosw15:
## Problem

After removing the expensive \`esxcli software vib get\` check in 2.11.5 (PR #6254), populator
migrations that require the VIB proceed without any validation — silently failing deep in the
clone flow when the VIB is missing or outdated.

The full Host/Provider/Plan controller approach (PR #6136) was too heavy for the immediate need.

## What this achieves

Adds a \`vmkfstools.version\` esxcli command to the VIB and calls it early in \`Populate()\`
(right after host resolution, VIB method only). Missing or outdated VIBs log a warning with a
remediation hint and proceed — migrations are not blocked. The detected version is emitted as a
\`vib_version\` label on the \`copy_duration_seconds\` Prometheus metric.

VIB_VERSION bumped 0.3.2 → 0.3.3 (patch bump to keep backports simpler). SCRIPT_VERSION in the
wrapper source is kept in sync so the SSH-path version check stays accurate.

## Manual Test Results

Tested on live ESXi hosts with the updated populator image.

| # | Scenario | Host | VIB state | Expected | Result |
|---|----------|------|-----------|----------|--------|
| 1 | **Happy** — VIB 0.3.3 installed | host-135 | 0.3.3 | `VIB version validated` logged, clone completes | ✅ Pass |
| 2 | **Middle** — old VIB (no version cmd) | host-136 | 0.3.2 | Warning logged (`method 'version' not found`), migration proceeds | ✅ Pass |
| 3 | **Sad** — no VIB at all | host-135 | removed | Warning logged (`object not found`), clone fails at xcopy level | ✅ Pass |

### DIFF
--- a/cmd/vsphere-copy-offload-populator/vmkfstools-wrapper/version.mk
+++ b/cmd/vsphere-copy-offload-populator/vmkfstools-wrapper/version.mk
@@ -1,1 +1,1 @@
-VIB_VERSION := 0.3.2
+VIB_VERSION := 0.3.3
--- a/pkg/controller/plan/adapter/base/doc.go
+++ b/pkg/controller/plan/adapter/base/doc.go
@@ -186,8 +186,8 @@ type Builder interface {
 	PopulatorVolumes(vmRef ref.Ref, annotations map[string]string, secretName string) ([]*core.PersistentVolumeClaim, error)
 	// Transferred bytes
 	PopulatorTransferredBytes(persistentVolumeClaim *core.PersistentVolumeClaim) (transferredBytes int64, err error)
-	// Whether xcopy offload was used for populator copy
-	PopulatorXcopyUsed(pvc *core.PersistentVolumeClaim) (xcopyUsed string, found bool, err error)
+	// Storage offload metadata for the populator copy (e.g. "xcopyUsed"), surfaced as task annotations.
+	PopulatorOffloadInfo(pvc *core.PersistentVolumeClaim) (info map[string]string, err error)
 	// Set the populator PVC labels
 	SetPopulatorDataSourceLabels(vmRef ref.Ref, pvcs []*core.PersistentVolumeClaim) (err error)
 	// Get the populator task name associated to a PVC
--- a/pkg/controller/plan/migration.go
+++ b/pkg/controller/plan/migration.go
@@ -2167,13 +2167,15 @@ func (r *Migration) updatePopulatorCopyProgress(vm *plan.VMStatus, step *plan.St
 
 		taskSeen[taskName] = true
 
-		if xcopyUsed, xcopyFound, xcopyErr := r.builder.PopulatorXcopyUsed(pvc); xcopyErr != nil {
-			r.Log.Info("Failed to get xcopyUsed", "pvc", pvc.Name, "error", xcopyErr)
-		} else if xcopyFound {
+		if offloadInfo, offloadErr := r.builder.PopulatorOffloadInfo(pvc); offloadErr != nil {
+			r.Log.Info("Failed to get populator offload info", "pvc", pvc.Name, "error", offloadErr)
+		} else if len(offloadInfo) > 0 {
 			if task.Annotations == nil {
 				task.Annotations = make(map[string]string)
 			}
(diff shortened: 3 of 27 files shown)


cmd/vsphere-copy-offload-populator/vmkfstools-wrapper/version.mk
```diff
@@ -1 +1 @@
-VIB_VERSION := 0.3.2
+VIB_VERSION := 0.4.0
```
mnecas: Please make it `0.3.3` and make the logic so it will not fail with error. We can't backport if it would be `0.4.0` and require users to update the VIB version on the ESXi

amitosw15: we wanted the `z` changes will indicate functional change, while `y` changes indicate api change (signature) 
and what it should do if not fail? i.e how will we force the user to update the vib?

mnecas: If it fails to get the version API then don't show the version

pkg/lib-volume-populator/populator-machinery/controller.go
```diff
@@ -988,6 +988,18 @@ func (c *controller) updateProgress(pod *corev1.Pod, pvc *corev1.PersistentVolum
 				return err
 			}
 		}
+
+		vibRegExp := regexp.MustCompile(`vsphere_xcopy_volume_populator_copy_duration_seconds\{[^}]*owner_uid="` + string(pvc.UID) + `"[^}]*\}`)
```
borod108: please validate the exp

borod108: and if there is an issue, perhaps check if we could add a failing test.

cmd/vsphere-copy-offload-populator/vmkfstools-wrapper/version.mk
borod108: good idea! please also sync the SCRIPT_VERSION var

amitosw15: done

amitosw15:
@mnecas Waiting for you buddy

pkg/controller/plan/migration.go
```diff
@@ -2176,6 +2176,15 @@ func (r *Migration) updatePopulatorCopyProgress(vm *plan.VMStatus, step *plan.St
...
 
+		if vibVersion, vibFound, vibErr := r.builder.PopulatorVibVersion(pvc); vibErr != nil {
+			r.Log.Info("Failed to get vibVersion", "pvc", pvc.Name, "error", vibErr)
+		} else if vibFound {
+			if task.Annotations == nil {
+				task.Annotations = make(map[string]string)
+			}
+			task.Annotations["vibVersion"] = vibVersion
+		}
+
```
mnecas: Not sure if we should propagate the vibVersion to the plan, thinking that the populator CR status should be enough

pkg/controller/plan/adapter/base/doc.go
```diff
@@ -188,6 +188,8 @@ type Builder interface {
 	PopulatorTransferredBytes(persistentVolumeClaim *core.PersistentVolumeClaim) (transferredBytes int64, err error)
 	// Whether xcopy offload was used for populator copy
 	PopulatorXcopyUsed(pvc *core.PersistentVolumeClaim) (xcopyUsed string, found bool, err error)
```
mnecas: Alternatively we can make this more generic to contain information about the Storage offload

pkg/controller/plan/migration.go
amitosw15: took this internally, will add this as an offload propertie

pkg/controller/plan/adapter/base/doc.go
amitosw15: yesss

mnecas (approved):
lgtm

PR #7346, opened 2026-07-06
MTV-6045 | CSI import: fix WaitForFirstConsumer deadlock in pvcinit pod

amitosw15:
## Problem

When the destination StorageClass uses \`volumeBindingMode: WaitForFirstConsumer\`, CSI import PVCs stall forever in \`Pending\`. The existing pvcinit pod (created by \`EnsurePVCInitPod\`) only referenced xcopy PVCs — CSI import PVCs were never included, so no consumer pod ever triggered their WFC binding.

## What it achieves

CSI import migrations now complete normally on \`WaitForFirstConsumer\` storage classes. The pvcinit pod covers all pending PVCs for the VM — both xcopy and CSI import — using the same pod and cleanup lifecycle that xcopy already relies on.

## Non-obvious decisions

**CsiImportPVCs now creates PVCs eagerly**: Previously \`CsiImportPVCs\` returned locally-built specs, which the caller then created via the ensurer. The ensurer iterates with a value loop, so \`Client.Create\` updated a throw-away copy — the original slice elements kept zero \`Status.Phase\` and were filtered out by \`EnsurePVCInitPod\`'s \`ClaimPending\` check. The fix mirrors the xcopy \`PopulatorVolumes\` pattern: \`Client.Create\` is called on the same pointer that is appended to the return slice, so callers receive live objects with real \`Status.Phase\` and \`Name\` (including server-assigned names for \`GenerateName\` PVCs).

**EnsurePopulatorVolumes renamed to EnsurePVCInitPod**: The function's purpose is creating a pvcinit pod to trigger WFC binding — not ensuring populator volumes. Renamed to reflect that.

## How to test

Configure a StorageMap with \`csiVolumeImport\` pointing to an RDM or VVol disk, where the destination StorageClass has \`volumeBindingMode: WaitForFirstConsumer\`. Without this fix, the migration stalls in \`CopyDisks\` because CSI import PVCs are never included in the pvcinit pod. With it, the pvcinit pod triggers binding and the migration proceeds.

Tested with ONTAP same-array CSI import (RDM disk, \`ontap-san-wfc\` SC) — migration succeeded end-to-end.

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/pkg/controller/plan/migration.go
+++ b/pkg/controller/plan/migration.go
@@ -865,7 +865,7 @@ func (r *Migration) execute(vm *plan.VMStatus) (err error) {
 			csiPVCs, csiErr := r.kubevirt.CsiImportPVCs(vm)
 			if csiErr != nil {
 				if !errors.As(csiErr, &web.ProviderNotReadyError{}) {
-					r.Log.Error(csiErr, "error creating CSI import PVCs", "vm", vm.Name)
+					r.Log.Error(csiErr, "error building CSI import PVCs", "vm", vm.Name)
 					step.AddError(csiErr.Error())
 					err = nil
 					break
@@ -874,6 +874,7 @@ func (r *Migration) execute(vm *plan.VMStatus) (err error) {
 					return
 				}
 			}
+			var resolvedCsiPVCs []*core.PersistentVolumeClaim
 			if len(csiPVCs) > 0 {
 				if csiErr = r.ensurer.PersistentVolumeClaims(vm, csiPVCs); csiErr != nil {
 					if !errors.As(csiErr, &web.ProviderNotReadyError{}) {
@@ -885,11 +886,23 @@ func (r *Migration) execute(vm *plan.VMStatus) (err error) {
 						return
 					}
 				}
+				var migrationPVCs []*core.PersistentVolumeClaim
+				if migrationPVCs, csiErr = r.kubevirt.getPVCs(vm.Ref); csiErr != nil {
+					if !errors.As(csiErr, &web.ProviderNotReadyError{}) {
+						step.AddError(csiErr.Error())
+						err = nil
+						break
+					} else {
+						err = csiErr
+						return
+					}
+				}
+				resolvedCsiPVCs = resolveCsiPVCs(csiPVCs, migrationPVCs)
 			}
 
+			var populatorPVCs []*core.PersistentVolumeClaim
 			if r.builder.SupportsVolumePopulators() {
-				var pvcs []*core.PersistentVolumeClaim
-				if pvcs, err = r.kubevirt.PopulatorVolumes(vm.Ref); err != nil {
+				if populatorPVCs, err = r.kubevirt.PopulatorVolumes(vm.Ref); err != nil {
 					if !errors.As(err, &web.ProviderNotReadyError{}) {
 						r.Log.Error(err, "error creating volumes", "vm", vm.Name)
 						step.AddError(err.Error())
@@ -899,7 +912,11 @@ func (r *Migration) execute(vm *plan.VMStatus) (err error) {
 						return
 					}
 				}
-				err = r.kubevirt.EnsurePopulatorVolumes(vm, pvcs)
+			}
+
+			bindPVCs := append(populatorPVCs, resolvedCsiPVCs...)
+			if len(bindPVCs) > 0 {
+				err = r.kubevirt.EnsurePVCInitPod(vm, bindPVCs)
 				if err != nil {
 					if !errors.As(err, &web.ProviderNotReadyError{}) {
 						step.AddError(err.Error())
@@ -1667,6 +1684,23 @@ func (r *Migration) execute(vm *plan.VMStatus) (err error) {
 	}
 
 	return
+}
+
+// Matches by AnnDiskSource, not name — CSI import PVC names may be unset until creation.
+func resolveCsiPVCs(csiSpecs []core.PersistentVolumeClaim, migrationPVCs []*core.PersistentVolumeClaim) []*core.PersistentVolumeClaim {
+	wanted := make(map[string]bool, len(csiSpecs))
+	for i := range csiSpecs {
+		if src := csiSpecs[i].Annotations[base.AnnDiskSource]; src != "" {
+			wanted[src] = true
+		}
+	}
+	var matched []*core.PersistentVolumeClaim
+	for _, pvc := range migrationPVCs {
+		if src := pvc.Annotations[base.AnnDiskSource]; src != "" && wanted[src] {
+			matched = append(matched, pvc)
+		}
+	}
+	return matched
 }
 
 const schedulerQueuedReasonFmt = "Waiting to start migration: in-flight limit reached (max %d)."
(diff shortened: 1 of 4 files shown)


borod108 (changes requested):
## Review — MTV-6045: CSI import WaitForFirstConsumer deadlock fix

**Intent**: Correct. CSI import PVCs need a consumer pod to trigger WaitForFirstConsumer binding, and they were being left out of the pvcinit pod.

**Structure**: Clean. Moving `EnsurePopulatorVolumes` outside the `SupportsVolumePopulators()` block and adding the `len > 0` guard is the right shape for this fix.

### Correctness Bug: CSI import PVCs will never pass the `ClaimPending` filter

The fix as written **does not solve the deadlock**. Here's why:

`EnsurePopulatorVolumes` (`kubevirt.go:1998`) filters PVCs by `Status.Phase == core.ClaimPending`:

```go
for _, pvc := range pvcs {
    if pvc.Status.Phase == core.ClaimPending {
        pendingPvcNames = append(pendingPvcNames, pvc.Name)
    }
}
```

The `csiPVCs` variable holds locally-built PVC specs returned by `CsiImportPVCs` — these are struct values built by the builder, **never fetched from the cluster**. Their `Status` is the zero value (`Phase == ""`), not `"Pending"`.

The ensurer at line 877 does create them on the cluster, but `Ensurer.PersistentVolumeClaims` iterates with a **value range loop** (`for _, pvc := range pvcs`), so `Client.Create` updates a local copy — the original `csiPVCs` slice elements are never mutated.

On subsequent reconcile iterations, `CsiImportPVCs` rebuilds fresh specs (again without Status) and the ensurer skips creation since they already exist. The `csiPVCs` elements **never** have `Status.Phase == ClaimPending`.

**Why xcopy PVCs work**: `PopulatorVolumes` (vsphere builder, line 1586) calls `Client.Create(&pvc)` directly on the struct that gets appended to the return slice, so the returned pointers carry the API server's Status response.

### Suggested fix

Replace the locally-built `csiPVCs` with cluster-fetched PVCs in the bind list. The simplest approach — consistent with what the PR description actually says ("calls `getPVCs(vm.Ref)`"):

```go
// After the SupportsVolumePopulators block, fetch all PVCs from the
// cluster so we have accurate Status.Phase for the binding filter.
bindPVCs := populatorPVCs
if len(csiPVCs) > 0 {
    clusterPVCs, pvcErr := r.kubevirt.getPVCs(vm.Ref)
    if pvcErr != nil {
        // ... error handling ...
    }
    bindPVCs = append(bindPVCs, clusterPVCs...)
}
```

Or more surgically: after `ensurer.PersistentVolumeClaims`, re-fetch only the CSI PVCs by name. But `getPVCs` already filters by VM label and `EnsurePopulatorVolumes` filters by `ClaimPending`, so fetching all VM PVCs is safe and idempotent.

**Note**: `getPVCs` is unexported — it returns `[]*core.PersistentVolumeClaim` which matches `populatorPVCs`' type, so the append works directly.

### Minor observations

1. **No tests** — there are no unit tests for `EnsurePopulatorVolumes` or the `CopyDisks` phase. Worth adding at least a targeted test that verifies CSI PVCs appear in the pvcinit pod's volume list.

2. **PR description vs code mismatch** — the description says "calls `getPVCs(vm.Ref)` to find any additional pending PVCs" but the code uses the locally-built `csiPVCs`. This is likely the original intent that got lost in implementation.

3. **Function naming** — `EnsurePopulatorVolumes` now serves a broader purpose (all pending PVCs, not just populator). Minor, not blocking.

pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -2668,23 +2666,46 @@ func (r *Builder) CsiImportPVCs(vmRef ref.Ref, pvcLabels map[string]string) (pvc
...
+		if existing, ok := existsByDiskSource[pvc.Annotations[planbase.AnnDiskSource]]; ok {
+			pvcs = append(pvcs, existing)
+		} else {
+			if err = r.Destination.Client.Create(context.TODO(), pvc); err != nil {
+				err = liberr.Wrap(err)
+				return
+			}
+			pvcs = append(pvcs, pvc)
 		}
 	}
```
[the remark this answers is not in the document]
amitosw15: Great idea but out of scope of the pr, will work on that in the future.

amitosw15:
@borod108 thanks for the detailed review. Addressing your points:

**Correctness bug (ClaimPending filter)** — Fixed. `CsiImportPVCs` now creates PVCs directly via `Client.Create` on the same pointer that is appended to the return slice, mirroring the xcopy `PopulatorVolumes` pattern. Callers receive live objects with real `Status.Phase` and `Name` from the API server. The ensurer call is removed entirely — no more value-loop problem.

**PR description vs code mismatch** — Fixed. Description now accurately describes the implementation (eager creation in the builder, not a getPVCs fetch).

**Function naming** — Fixed. `EnsurePopulatorVolumes` renamed to `EnsurePVCInitPod`.

pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -2668,23 +2667,47 @@ func (r *Builder) CsiImportPVCs(vmRef ref.Ref, pvcLabels map[string]string) (pvc
...
+		}
+		if existing, ok := existsByDiskSource[pvc.Annotations[planbase.AnnDiskSource]]; ok {
+			pvcs = append(pvcs, existing)
+		} else {
+			if err = r.Destination.Create(context.TODO(), pvc); err != nil {
+				err = liberr.Wrap(err)
+				return
+			}
+			r.Log.Info("Created PersistentVolumeClaim.", "pvc", path.Join(pvc.Namespace, pvc.Name), "vm", vmRef.String())
+			pvcs = append(pvcs, pvc)
```
mnecas: Think we don't need this as it's part of ensurer https://github.com/kubev2v/forklift/blob/e3bd2baecc6bdcbad03b1d1767ae5d9ba598e8f6/pkg/controller/plan/ensurer/ensurer.go#L104

pkg/controller/plan/migration.go
```diff
@@ -885,11 +886,23 @@ func (r *Migration) execute(vm *plan.VMStatus) (err error) {
...
+				resolvedCsiPVCs = resolveCsiPVCs(csiPVCs, migrationPVCs)
 			}
 
+			var populatorPVCs []*core.PersistentVolumeClaim
 			if r.builder.SupportsVolumePopulators() {
-				var pvcs []*core.PersistentVolumeClaim
-				if pvcs, err = r.kubevirt.PopulatorVolumes(vm.Ref); err != nil {
+				if populatorPVCs, err = r.kubevirt.PopulatorVolumes(vm.Ref); err != nil {
 					if !errors.As(err, &web.ProviderNotReadyError{}) {
 						r.Log.Error(err, "error creating volumes", "vm", vm.Name)
```
mnecas, line 907: @coderabbitai do we need to use both PopulatorVolumes and getPVCs couldn't we just use one or the other?

mnecas (approved):
lgtm

PR #7347, opened 2026-07-06
MTV-4416 | Skip map-back when LUN is already mapped to original groups

amitosw15:
## Problem

On HPE Primera/3PAR, xcopy cleanup logged a spurious 404 error when
attempting to remap the source LUN back to its original initiator groups.
The error appeared despite the migration completing successfully.

## What it achieves

The cleanup now checks the live mapped groups on the array before
attempting to remap. Any group the LUN is still mapped to (which in
practice is all of them, since only the temporary xcopy group is ever
unmapped) is skipped silently. Groups that were genuinely lost are
still remapped.

## Why this approach

The original map-back assumed the LUN's existing mappings were removed
during xcopy, but they weren't — only the xcopy initiator group is
unmapped. The 404 was the array rejecting a duplicate mapping creation.
A live check at cleanup time eliminates the error without weakening the
safety net.

Verified by running a full migration of `tshefismall` on HPE Primera/3PAR:
xcopy completes successfully, no 404 in cleanup, VM boots on OCP.

## How to test

Migrate any VM from an HPE Primera/3PAR iSCSI datastore via xcopy.
Confirm no `failed to map volume back to original holder` warning in
the populator logs and that the migration succeeds.

### DIFF
--- a/cmd/vsphere-copy-offload-populator/internal/populator/remote_esxcli.go
+++ b/cmd/vsphere-copy-offload-populator/internal/populator/remote_esxcli.go
@@ -265,13 +265,7 @@ func (p *RemoteEsxcliPopulator) Populate(vmId string, migrationHostId, sourceVMD
 			cleanupLog.Info("failed to unmap during cleanup", "lun", lun.Name, "err", errUnmap)
 		}
 
-		cleanupLog.V(2).Info("mapping volume back to original initiator groups", "groups", originalInitiatorGroups)
-		for _, group := range originalInitiatorGroups {
-			_, errMap := p.StorageApi.Map(group, lun, mappingContext)
-			if errMap != nil {
-				cleanupLog.Info("failed to map volume back to original holder", "group", group, "err", errMap)
-			}
-		}
+		remapOriginalGroups(p.StorageApi, lun, mappingContext, originalInitiatorGroups, cleanupLog)
 		cleanupLog.V(2).Info("deleting dead devices after short delay")
 		time.Sleep(5 * time.Second)
 		deleteDeadDevices(cleanupCtx, p.VSphereClient, host, dsActiveAdapters)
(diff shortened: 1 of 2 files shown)


cmd/vsphere-copy-offload-populator/internal/populator/remote_esxcli.go
```diff
@@ -265,11 +265,21 @@ func (p *RemoteEsxcliPopulator) Populate(vmId string, migrationHostId, sourceVMD
...
-		cleanupLog.V(2).Info("mapping volume back to original initiator groups", "groups", originalInitiatorGroups)
-		for _, group := range originalInitiatorGroups {
-			_, errMap := p.StorageApi.Map(group, lun, mappingContext)
-			if errMap != nil {
-				cleanupLog.Info("failed to map volume back to original holder", "group", group, "err", errMap)
+		if len(originalInitiatorGroups) > 0 {
+			currentGroups, errCurrent := p.StorageApi.CurrentMappedGroups(lun, mappingContext)
+			if errCurrent != nil {
+				cleanupLog.Info("failed to fetch current mapped groups for remap check, skipping remap", "err", errCurrent)
+				currentGroups = originalInitiatorGroups
```
borod108: **Bug: error fallback silently skips ALL remaps**

When `CurrentMappedGroups` fails, setting `currentGroups = originalInitiatorGroups` causes `slices.Contains(currentGroups, group)` to match every entry in the loop below — so every `Map` call is skipped.

This is a regression from the pre-PR behavior, which always attempted remap. The failure modes are asymmetric:
- Skip remap when genuinely needed → LUN access lost (data loss risk)
- Attempt remap when not needed → harmless 404 in logs (the cosmetic error this PR exists to fix)

The safer default when current state is unknown is to attempt remap for all groups:

```go
if errCurrent != nil {
    cleanupLog.Info("failed to fetch current mapped groups, attempting remap for all original groups", "err", errCurrent)
    currentGroups = nil  // slices.Contains(nil, x) returns false → all groups get remapped
}
```

`slices.Contains(nil, x)` returns `false` in Go, so this correctly falls through to the `Map` call for every group.

borod108 (changes requested):
Good approach — the live check before remap is the right fix for the spurious 404s. One bug in the error fallback path needs fixing before merge. See inline comment.

borod108:
@amitosw15 can you please verify that there are no implications on other vendors - we sometimes store the current storage groups in context, just make sure we do not break anything there.

PR #7351, opened 2026-07-07
MTV-6050 | Set SCSI reservation and errorPolicy on shared RDM LUN disks

yaacov:
When rdmAsLun=true and a disk is shared, set LunTarget.Reservation=true and Disk.ErrorPolicy="report" on the target KubeVirt VM spec. This enables SCSI-3 Persistent Reservations (pr-helper) so clustered workloads (WSFC, Oracle RAC, Veritas) can coordinate disk ownership, and reports I/O errors to the guest instead of pausing the VM, allowing proper failover.

Ref: https://redhat.atlassian.net/browse/MTV-6050
Resolves: MTV-6050


yaacov:
@coderabbitai hi, can you verify the new fields added are safe, and that they are added only when needed ?

solenoci (approved):

PR #7356, opened 2026-07-07
MTV-5880 | Skip re-adding persistent route when DefaultGateway already provides it

Hazanel:
## Summary

After Hyper-V to KubeVirt migration, the duplicate persistent route cleanup
script removes all duplicate default gateways and then re-adds one instance
via `PersistentStore` or `route.exe -p`. However, when the target interface
already has the same gateway configured through the `DefaultGateway` registry
key (set by the network-configure firstboot script), this re-add creates a
new duplicate from a different source.

### Changes

- Check `Get-NetIPConfiguration` before re-adding a cleaned default gateway,
  skip the `PersistentStore` write when the interface already has that gateway
  via its IP configuration
- Rename loop variable `$routes` → `$dupRoutes` to avoid shadowing the outer
  `$routes` collection
- Apply the fix to both the standard and registry script variants

Ref: https://redhat.atlassian.net/browse/MTV-5880
Resolves: MTV-5880

### DIFF
--- a/pkg/virt-v2v/customize/scripts/windows/9999-remove_duplicate_persistent_routes-registry.ps1
+++ b/pkg/virt-v2v/customize/scripts/windows/9999-remove_duplicate_persistent_routes-registry.ps1
@@ -84,10 +84,10 @@ if (-not $groupedRoutes) {
     
     # First, handle duplicate default gateways
     foreach ($group in $gatewayDuplicates) {
-        $routes = $group.Group
+        $dupRoutes = $group.Group
 
         # Choose the route on a live interface with the lowest metric
-        $sortedRoutes = $routes | Sort-Object RouteMetric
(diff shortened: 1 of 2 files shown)


Hazanel:
@coderabbitai, please take a look

Hazanel:
> ## [Quality Gate Failed](https://sonarcloud.io/dashboard?id=kubev2v_forklift&pullRequest=7356) **Quality Gate failed**
> Failed conditions  [100.0% Duplication on New Code](https://sonarcloud.io/component_measures?id=kubev2v_forklift&pullRequest=7356&metric=new_duplicated_lines_density&view=list) (required ≤ 80%)
> 
> [See analysis details on SonarQube Cloud](https://sonarcloud.io/dashboard?id=kubev2v_forklift&pullRequest=7356)

We can ignore - 2 different files, we can't combine them

yaacov (approved):

PR #7357, opened 2026-07-07
MTV-6056 | CSI import cross-array fallback to xcopy

amitosw15:
## Summary

- CSI import fails silently when the source volume is on a different array than the destination CSI driver manages (cross-array). The PVC gets stuck in Pending with a cryptic CSI driver error.
- Changes `CsiImportPlugin.Resolve()` to return `(annotations, found, error)` so "volume not on this array" (`found=false`) is distinguished from actual failures (API down, auth error).
- When `found=false`, `buildCsiImportPVC` returns `(nil, nil)` and `PopulatorVolumes` picks up the disk for xcopy automatically.
- Adds `diskBackingResolver` interface to decouple vSphere client from storage array plugin for testability.
- 15 new tests covering cross-array VVol/RDM fallback, mixed disk types (VMDK+RDM), CsiImportPVCs→PopulatorVolumes coordination, and backward compatibility.

## Test plan

- [x] `go test ./pkg/storage/resolver/...` — resolver interface + HPE cross-array tests
- [x] `go test ./pkg/controller/plan/adapter/vsphere/...` — 15 new builder tests (unit + integration)
- [x] `go build ./pkg/controller/plan/...` — clean compile
- [ ] CI passes

## Summary by CodeRabbit

* **New Features**
  * Improved vSphere offload handling so storage can be split more intelligently between CSI import and xcopy based on disk availability.
  * Added support for returning “not found” when a disk can’t be matched, allowing fallback behavior instead of failing outright.

* **Bug Fixes**
  * Preserved CSI import annotations for supported disks while skipping unsupported or unmatched disks.
  * Tightened configuration validation so required vSphere xcopy settings are always included.


pkg/apis/forklift/v1beta1/mapping.go
```diff
@@ -66,7 +66,7 @@ type NetworkPair struct {
 // phase of the migration. There can be more than one available but currently only
 // one will be supported
 type OffloadPlugin struct {
-	VSphereXcopyPluginConfig *VSphereXcopyPluginConfig `json:"vsphereXcopyConfig,omitempty"`
+	VSphereXcopyPluginConfig *VSphereXcopyPluginConfig `json:"vsphereXcopyConfig"`
 	CsiVolumeImport          *CsiVolumeImport          `json:"csiVolumeImport,omitempty"`
```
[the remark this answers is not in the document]
amitosw15: comment of ai isnt true, closing thread

operator/.upstream_manifests
```diff
@@ -10250,6 +10250,8 @@ spec:
                           - secretRef
                           - storageVendorProduct
                           type: object
+                      required:
+                      - vsphereXcopyConfig
                       type: object
```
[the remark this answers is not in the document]
amitosw15: must have it for cross-vendor and vmdk

borod108 (approved):
## Review: MTV-6056 — CSI import cross-array fallback to xcopy

**Verdict: Approve** with one callout below.

### Design

The `(annotations, found, error)` return pattern on `CsiImportPlugin.Resolve()` is the right design — it cleanly distinguishes "not on this array" (`found=false`, no error → fall through to xcopy) from "API/infra failure" (error → fail migration). The two-phase coordination is correct: `CsiImportPVCs` runs first and persists PVCs, then `PopulatorVolumes` checks `diskHandledByCsiImport` to skip already-handled disks.

### Callout: CRD breaking change

The CRD now adds `required: [vsphereXcopyConfig]` inside `offloadPlugin`, and `mapping.go` drops `omitempty` from the struct tag. This means any existing StorageMap with `offloadPlugin` containing only `csiVolumeImport` (no xcopy config) will fail validation on update.

This is **intentionally safe** — without xcopy as a fallback, cross-array disks would be silently dropped from migration (the `continue` at `builder.go:1420-1422` skips disks with no xcopy config). But it's a CRD-level breaking change that should be documented in release notes.

### Positives

- Clean interface extraction (`diskBackingResolver`, `CsiImportPluginFunc`) for testability
- HPE `volumeNameByFilter` correctly returns `("", false, nil)` for 0 results instead of `fmt.Errorf` — models "not on this array" as a data condition, not an error
- 15 new tests with good coverage: cross-array VVol/RDM, same-array, mixed disk types (VMDK+RDM), error propagation, missing credentials, no-CSI-config, and `diskHandledByCsiImport` positive/negative
- The integration tests (`CsiImportPVCs and PopulatorVolumes integration`) are the most valuable — they exercise the full two-phase coordination including PVC persistence

### Pre-existing (not a regression)

Per-disk vSphere connection in `vsphereDiskBackingResolver` — each call creates/tears down a session. Same N connections for N disks as before the PR. The new interface makes it easy to optimize later by caching the connection across the `CsiImportPVCs` loop.

pkg/controller/plan/adapter/vsphere/builder_test.go
```diff
@@ -1291,6 +1292,553 @@ var _ = Describe("mapDisks SCSI reservation", func() {
...
+		}
+		testVM = &model.VM{
+			VM1: model.VM1{
+				VM0:   model.VM0{ID: "vm-1", Name: "test-vm"},
+				Disks: []vsphere.Disk{vvolDisk},
+			},
+		}
+	})
+
+	It("should return nil for warm migration without calling resolver or plugin", func() {
```
borod108: does it check what it says?

pkg/controller/plan/adapter/vsphere/builder_test.go
```diff
@@ -1291,6 +1292,553 @@ var _ = Describe("mapDisks SCSI reservation", func() {
...
+	)
+
+	BeforeEach(func() {
+		csiStoragePair = &v1beta1.StoragePair{
+			Source: ref.Ref{ID: "ds-1"},
+			Destination: v1beta1.DestinationStorage{
+				StorageClass: "hpe-sc",
+			},
+			OffloadPlugin: &v1beta1.OffloadPlugin{
+				CsiVolumeImport: &v1beta1.CsiVolumeImport{
```
borod108, line 1705: should this not have the xcopyPlugin?

pkg/controller/plan/adapter/vsphere/builder_test.go
```diff
@@ -1291,6 +1292,553 @@ var _ = Describe("mapDisks SCSI reservation", func() {
...
+		// Volumes that exist on the destination HPE array — disk 2 is on a different array
+		arrayVolumes := map[string]string{
+			"vvol:uuid-disk-1": "hpe-vol-disk-1",
+		}
+		builder.csiPlugin = resolver.CsiImportPluginFunc(func(backing *resolver.DiskBacking) (map[string]string, bool, error) {
+			name, found := arrayVolumes[backing.VVolID]
+			if !found {
+				return nil, false, nil
+			}
+			return map[string]string{"csi.hpe.com/importVolAsClone": name}, true, nil
```
borod108: could we import the string?

pkg/controller/plan/adapter/vsphere/builder_test.go
```diff
@@ -1291,6 +1292,553 @@ var _ = Describe("mapDisks SCSI reservation", func() {
...
+		})
+
+		pvc, err := builder.buildCsiImportPVC(
+			context.TODO(), ref.Ref{ID: "vm-1"}, testVM, vvolDisk, 0, csiStoragePair, nil)
+
+		Expect(err).NotTo(HaveOccurred())
+		Expect(pvc).To(BeNil())
+	})
+
+	It("should return nil for VMDK disk without calling plugin", func() {
```
borod108: without?

pkg/controller/plan/adapter/vsphere/builder_test.go
```diff
@@ -1291,6 +1292,553 @@ var _ = Describe("mapDisks SCSI reservation", func() {
...
+		pvc, err := builder.buildCsiImportPVC(
+			context.TODO(), ref.Ref{ID: "vm-1"}, testVM, vvolDisk, 0, csiStoragePair, nil)
+
+		Expect(err).NotTo(HaveOccurred())
+		Expect(pvc).To(BeNil())
+	})
+})
+
+var _ = Describe("CsiImportPVCs and PopulatorVolumes integration", func() {
+	It("should fall through all disks to xcopy when all are cross-array", func() {
```
borod108: can you explain what this means?

pkg/storage/resolver/resolver_test.go
```diff
@@ -9,15 +9,32 @@ var _ CsiImportPlugin = (*stubPlugin)(nil)
...
+		t.Fatal("expected found=true")
+	}
 	if len(annotations) == 0 {
 		t.Error("Resolve must return at least one annotation")
 	}
 }
 
+func TestCsiImportPluginNotFound(t *testing.T) {
+	var p CsiImportPlugin = &stubPlugin{}
+	annotations, found, err := p.Resolve(&DiskBacking{})
```
borod108: not sure what are we testing besides golang

pkg/controller/plan/adapter/vsphere/builder_test.go
amitosw15: changed

pkg/controller/plan/adapter/vsphere/builder_test.go
amitosw15: it should added also code validation

pkg/controller/plan/adapter/vsphere/builder_test.go
amitosw15: yes, tnx

pkg/storage/resolver/resolver_test.go
amitosw15: stupied test, deleting them

pkg/controller/plan/adapter/vsphere/builder_test.go
amitosw15: changing the name of the test

.gitignore
```diff
@@ -80,3 +80,4 @@ gha-creds-*.json
 
 # developer-private test scenarios (verification, bug reproduction, etc.)
 tests/scenarios/
+.pi/
```
borod108: ?

borod108 (approved):
LGTM @mnecas (except one small nit pick)

.gitignore
amitosw15: Claude Code session directory — created automatically by the AI coding tool when working in this directory. Completely unrelated to the project code. Added to .gitignore so it never gets accidentally committed.

amitosw15:
@mnecas Waiting for you buddy

operator/config/crd/bases/forklift.konveyor.io_storagemaps.yaml
```diff
@@ -130,6 +130,8 @@ spec:
                           - secretRef
                           - storageVendorProduct
                           type: object
+                      required:
+                      - vsphereXcopyConfig
```
mnecas: this should not be mandatory for all cases you should add validation for your case

mnecas (commented):
instead of returning bool if found we could have some specific error type, wdyt?

operator/config/crd/bases/forklift.konveyor.io_storagemaps.yaml
amitosw15: Its mandatory under `offload` not in the general case

amitosw15:
> instead of returning bool if found we could have some specific error type, wdyt?

np, just out of curiosity why do you prefer it?

pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -176,8 +176,9 @@ var sanitizeNameRx = regexp.MustCompile(`[^a-zA-Z0-9_.-]+`)
 // vSphere builder.
 type Builder struct {
 	*plancontext.Context
-	// Host CRs.
-	hosts map[string]*api.Host
+	hosts               map[string]*api.Host
+	diskBackingResolver diskBackingResolver
+	csiPlugin           resolver.CsiImportPlugin
```
mnecas: @coderabbitai do we need to set these parameters from some other place? afaik in the normal code it gets defaulted to the vsphereDiskBackingResolver lets just stick with it

operator/config/crd/bases/forklift.konveyor.io_storagemaps.yaml
amitosw15: fixed

pkg/controller/plan/adapter/vsphere/builder.go
amitosw15: good for testing, wdyt?

pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -1719,13 +1720,12 @@ func (r *Builder) buildCsiImportPVC(
...
 	}
 
-	vsphereClient := &Client{Context: r.Context}
-	if err := vsphereClient.connect(); err != nil {
-		return nil, liberr.Wrap(err, "failed to connect to vSphere for disk backing detection")
+	dbr := r.diskBackingResolver
+	if dbr == nil {
+		dbr = &vsphereDiskBackingResolver{planCtx: r.Context}
 	}
-	defer vsphereClient.Close()
```
mnecas: @coderabbitai i don't like the high number of abstractions here, do we need it or could there be simpler alternative?

operator/config/crd/bases/forklift.konveyor.io_storagemaps.yaml
amitosw15: added code validation

pkg/controller/plan/adapter/vsphere/builder.go
amitosw15: I changed most of this and the code is way more simpler now.

pkg/controller/map/storage/validation.go
```diff
@@ -164,3 +167,23 @@ func (r *Reconciler) validateDestination(mp *api.StorageMap) (err error) {
 
 	return
 }
+
+func (r *Reconciler) validateOffloadPlugin(mp *api.StorageMap) {
+	notValid := []string{}
+	for _, entry := range mp.Spec.Map {
+		plugin := entry.OffloadPlugin
+		if plugin != nil && plugin.CsiVolumeImport != nil && plugin.VSphereXcopyPluginConfig == nil {
```
amitosw15: always be true no? @coderabbitai wdyt?

amitosw15:
Prerequisites for this pr: #7972

PR #7407, opened 2026-07-08
MTV-6055 | Throttle populator pods on the runtime host

amitosw15:
## Summary

- When `dedicatedMigrationHosts` are configured, the populator controller now throttles concurrent pods based on the actual ESXi host running the XCOPY (`migrationHost`), not the VM's registered host (`sourceHost`)
- Adds a `throttleHost` label to populator pods so inflight counting targets the correct host
- Defines shared label constants (`LabelSourceHost`, `LabelThrottleHost`) in `v1beta1/doc.go` to avoid string duplication across packages

## Test plan

- [x] `go test ./pkg/lib-volume-populator/...` passes
- [x] `go test ./pkg/controller/plan/adapter/vsphere/...` passes
- [x] Deploy with `dedicatedMigrationHosts` configured; verify populator pods carry `throttleHost` label matching the migration host
- [x] Run concurrent migrations exceeding `controller_max_populator_inflight` per dedicated host; verify throttling kicks in on the migration host, not the source VM host

Resolves: MTV-6055

### DIFF
--- a/pkg/apis/forklift/v1beta1/doc.go
+++ b/pkg/apis/forklift/v1beta1/doc.go
@@ -109,6 +109,5 @@ const (
 
 // Labels.
 const (
-	LabelSourceHost   = "sourceHost"
-	LabelThrottleHost = "throttleHost"
+	LabelSourceHost = "sourceHost"
 )
--- a/pkg/apis/forklift/v1beta1/doc.go
+++ b/pkg/apis/forklift/v1beta1/doc.go
@@ -106,3 +106,9 @@ const (
 	AnnDiskSource = "forklift.konveyor.io/disk-source"
 	AnnSource     = "forklift.konveyor.io/source"
 )
+
+// Labels.
+const (
+	LabelSourceHost   = "sourceHost"
+	LabelThrottleHost = "throttleHost"
+)
--- a/pkg/lib-volume-populator/populator-machinery/controller.go
+++ b/pkg/lib-volume-populator/populator-machinery/controller.go
@@ -83,8 +83,7 @@ const (
 
 	qemuGroup = 107
 
-	labelSourceHost   = api.LabelSourceHost
-	labelThrottleHost = api.LabelThrottleHost
+	labelSourceHost = api.LabelSourceHost
 )
 
 type empty struct{}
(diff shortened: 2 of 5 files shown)


pkg/apis/forklift/v1beta1/doc.go
```diff
@@ -106,3 +106,9 @@ const (
 	AnnDiskSource = "forklift.konveyor.io/disk-source"
 	AnnSource     = "forklift.konveyor.io/source"
 )
+
+// Labels.
+const (
+	LabelSourceHost   = "sourceHost"
+	LabelThrottleHost = "throttleHost"
```
borod108: validate this label the same way as the SourceHost

borod108 (approved):
LGTM @mnecas

amitosw15:
@mnecas Waiting for you buddy

pkg/lib-volume-populator/populator-machinery/controller.go
```diff
@@ -648,6 +655,13 @@ func (c *controller) syncPvc(ctx context.Context, key, pvcNamespace, pvcName str
 			if sourceHost != "" {
 				labels[labelSourceHost] = sourceHost
 			}
+			if throttleHost != "" {
+				if errs := k8svalidation.IsValidLabelValue(throttleHost); len(errs) == 0 {
+					labels[labelThrottleHost] = throttleHost
```
mnecas: Do we need to have new label for this? Couldn't we just use the sourceHost label

amitosw15: I thought that the sourceHost maybe be useful, but if you think that is not needed for telemetry or something - np to change it

amitosw15:
@mnecas Good catch — updated to reuse the `sourceHost` label instead of adding a new one. The pod's `sourceHost` label is now overwritten with the resolved runtime host (migrationHost when set, else the VM's source host), and `countActivePopulatorPodsForHost` selects on `sourceHost` again. Renamed the internal variable to `runtimeHost` for clarity. Verified end-to-end on a live cluster with 2 VMs on different source hosts, both throttling correctly against the shared dedicated host. Ready for another look — thanks!

PR #7418, opened 2026-07-08
MTV-6074 | Fix OpenStack populator CA certificate regression

krcmarik:
Commit f24a1192f (MTV-5867) changed getTLSConfig() to read the CA certificate exclusively from c.secret via util.GetCACert(), bypassing c.Options. The OpenStack populator binary creates a bare Client without c.secret (it receives secret data as environment variables into c.Options), so the CA certificate was silently ignored and the populator always fell back to the system CA pool. This caused TLS failures for OpenStack endpoints using custom/self-signed certificates.

The nil pointer dereference on c.Log (fixed by 42e7aba7e / MTV-5940 with c.Log != nil guards) was a secondary symptom: the regression forced the populator into a code path it never reached before (the "no CA cert" branch), where c.Log was uninitialized.

Fix: use util.GetCACert(c.secret) when c.secret is available (controller path), fall back to c.Options["cacert"] when c.secret is nil (populator path). Remove the c.Log != nil guards added by 42e7aba7e — they are artifacts of the regression; Connect() initializes c.Log before Authenticate() is called, and Authenticate() itself already has unguarded c.Log calls.

Resolves: MTV-6074

### DIFF
--- a/pkg/lib/client/openstack/client.go
+++ b/pkg/lib/client/openstack/client.go
@@ -175,9 +175,7 @@ func (c *Client) Authenticate() (err error) {
 func (c *Client) getTLSConfig() (tlsConfig *tls.Config, err error) {
 	identityUrl, err := url.Parse(c.URL)
 	if err != nil {
-		if c.Log != nil {
-			c.Log.Error(err, "error parsing the URL", "url", c.URL)
-		}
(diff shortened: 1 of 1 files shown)


mrnold (approved):

mnecas (approved):
gltm

PR #7426, opened 2026-07-09
MTV-5805 | Surface wait status when warm migration is scheduler-queued

gwencasey96:
Set plan Ready/Executing messages and Initialize step reasons when VMs are waiting on controller_max_vm_inflight instead of leaving the plan showing only "ready" and "EXECUTING".

Ref: https://redhat.atlassian.net/browse/MTV-5805
Resolves: MTV-5805

### DIFF
--- a/pkg/controller/plan/migration.go
+++ b/pkg/controller/plan/migration.go
@@ -139,7 +139,7 @@ func (r *Migration) Run() (reQ time.Duration, err error) {
 				return
 			}
 		} else {
-			r.Log.Info("The scheduler does not have any additional VMs.")
+			r.markSchedulerQueuedVMs()
 			break
 		}
 	}
@@ -1666,6 +1666,51 @@ func (r *Migration) execute(vm *plan.VMStatus) (err error) {
 	}
 
 	return
+}
+
+const schedulerQueuedStepReasonFmt = "Waiting to start migration: in-flight limit reached (max %d)."
+
+func (r *Migration) markSchedulerQueuedVMs() {
+	snapshot := r.Plan.Status.Migration.ActiveSnapshot()
+	limit := settings.Settings.MaxInFlight
+	queued := 0
+
+	for i := range r.Plan.Status.Migration.VMs {
+		vm := r.Plan.Status.Migration.VMs[i]
+		if vm.MarkedStarted() || vm.MarkedCompleted() || vm.HasCondition(api.ConditionCanceled) {
+			continue
+		}
+		queued++
+		if step, found := vm.FindStep(planmigrbase.Initialize); found {
+			step.Phase = api.StepPending
+			step.Reason = fmt.Sprintf(schedulerQueuedStepReasonFmt, limit)
+		}
+	}
+
+	if queued == 0 {
+		return
+	}
+
+	r.Plan.Status.SetCondition(libcnd.Condition{
+		Type:     libcnd.Ready,
+		Status:   True,
+		Category: api.CategoryRequired,
+		Message: fmt.Sprintf(
+			"%d VM(s) waiting to start migration (in-flight limit: %d).",
+			queued,
+			limit),
+	})
+	r.Log.Info("VMs waiting for scheduler capacity.", "count", queued, "limit", limit)
+	snapshot.SetCondition(libcnd.Condition{
+		Type:     api.ConditionExecuting,
+		Status:   True,
+		Category: api.CategoryAdvisory,
+		Message: fmt.Sprintf(
+			"The plan is EXECUTING. %d VM(s) waiting to start (in-flight limit: %d).",
+			queued,
+			limit),
+		Durable: true,
+	})
 }
 
 func (r *Migration) resetPrecopyTasks(vm *plan.VMStatus, step *plan.Step) {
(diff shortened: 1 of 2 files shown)


pkg/controller/plan/migration.go
```diff
@@ -1668,6 +1668,51 @@ func (r *Migration) execute(vm *plan.VMStatus) (err error) {
 	return
 }
 
+const schedulerQueuedStepReasonFmt = "Waiting to start migration: in-flight limit reached (max %d)."
+
+func (r *Migration) markSchedulerQueuedVMs() {
+	snapshot := r.Plan.Status.Migration.ActiveSnapshot()
```
yaacov: not sure we need to set condition on the migration, migration CR is a trigger ( like setting "start plan" ) when we check plan conditions we usually only check the plan CR, setting only the paln will simplify this method, @coderabbitai WDYT ?

pkg/controller/plan/migration.go
```diff
@@ -1668,6 +1668,51 @@ func (r *Migration) execute(vm *plan.VMStatus) (err error) {
...
+	limit := settings.Settings.MaxInFlight
+	queued := 0
+
+	for i := range r.Plan.Status.Migration.VMs {
+		vm := r.Plan.Status.Migration.VMs[i]
+		if vm.MarkedStarted() || vm.MarkedCompleted() || vm.HasCondition(api.ConditionCanceled) {
+			continue
+		}
+		queued++
+		if step, found := vm.FindStep(planmigrbase.Initialize); found {
```
yaacov: vms have conditions, does it make sense to add a condition instead of setting the step, this will remove the need to find step, @coderabbitai WDYT ?

pkg/controller/plan/migration.go
yaacov: can you show how the code will look using vm instead of snapshot ?

pkg/controller/plan/migration.go
yaacov: i don't understand your answer, can you rephrase ?

pkg/controller/plan/migration.go
yaacov: if we have the condition and the plan global condition do we need to change the vm phase and reason ?
removing it will simply the method without removing information

pkg/controller/plan/migration.go
yaacov: does it make sense to duplicate information unnecessary, having two sources of truth,  because we may have a UI that may not look at the conditions, do you know the UI don't look at them ? it the UI don't look at the conditions why do we have them ?

pkg/controller/plan/migration.go
```diff
@@ -1668,6 +1668,51 @@ func (r *Migration) execute(vm *plan.VMStatus) (err error) {
...
+		Type:     libcnd.Ready,
+		Status:   True,
+		Category: api.CategoryRequired,
+		Message: fmt.Sprintf(
+			"%d VM(s) waiting to start migration (in-flight limit: %d).",
+			queued,
+			limit),
+	})
+	r.Log.Info("VMs waiting for scheduler capacity.", "count", queued, "limit", limit)
+	snapshot.SetCondition(libcnd.Condition{
```
yaacov: @coderabbitai hi, where do we drop this conditions once the the vm is removed from the waiting list ?

pkg/controller/plan/migration.go
yaacov: can you show me how the method will look if we only set the vm condition ? e.g. not phase change or global plan condition ?

pkg/controller/plan/migration.go
yaacov: i think in other places we just delete condition instead of changing it, can you check ?
where do you suggest to add the delete condition line ?

pkg/controller/plan/migration.go
yaacov: i see from the bug report we actually need to set a global plan condition too,
@coderabbitai, can you show the method with both vm and plan conditions set ?

pkg/controller/plan/migration.go
yaacov: @gwencasey96 hi, I think we don't remove the conditions when the vm is running again, see above conversation, can you re check and fix if needed ?

pkg/controller/plan/migration.go
yaacov: @gwencasey96 hi, in our code we usually do not add conditions on the `snapshot`, we interact with users via the `r.Plan.Status.Migration.VMs` directly , see above discussion, can you re check and use the vm var instead of the sanpshot ?

pkg/controller/plan/migration.go
yaacov: @coderabbitai hi, can you give code example of deleting the vm condition and the plan conditions we set up in https://github.com/kubev2v/forklift/pull/7426#discussion_r3549804990 code example ?

yaacov: @gwencasey96 note, I don't think the above examples from code rabbit are good, but I do think we need to delete the conditions we set, so re check and fix, just disregard code rabbits suggestions ( i think they are not good code suggestions )

pkg/controller/plan/migration.go
yaacov: @gwencasey96 hi,
p.s.
code rabbit suggested this line:
`Type:     "SchedulerQueued",`
in our code we have a global place where we keep all the type strings as consts,
please check code rabbit code suggestions, and fix as you think is best be carful not to copy verbatim

gwencasey96:
passed testing with 2 VMs when limit set to 1

yaacov (approved):

PR #7482, opened 2026-07-14
MTV-5955 | CSI Volume Import — NetApp ONTAP (Trident) vendor plugin

amitosw15:
## Problem

Migrating VMs with RDM disks backed by NetApp ONTAP arrays currently requires the xcopy populator — a pod that connects to the ONTAP REST API, creates a new LUN, and copies data block-by-block. For same-array migrations, the source and destination LUN are on the same ONTAP cluster, so a full data copy is unnecessary.

## What it achieves

This PR adds ONTAP (Trident) as a CSI import vendor, enabling zero-copy RDM migration: the Trident CSI driver adopts the existing ONTAP LUN as a PersistentVolume by annotation, with no data movement. Migration time for RDM disks drops from minutes (xcopy) to seconds (PVC bind).

## How it works

The ONTAP importer:
1. Extracts the LUN serial from the vSphere RDM device name (NAA or VML format)
2. Queries the ONTAP REST API to resolve the serial to a LUN path (`/vol/<flexvol>/<lun>`)
3. Determines the Trident driver type (`ontap-san` vs `ontap-san-economy`) from the destination StorageClass `backendType` parameter, falling back to `TRIDENT_DRIVER` in the storage secret
4. Produces the Trident PVC annotations: `importOriginalName`, `importBackendUUID`, `notManaged: true`

VMDK and VVol disks fall through to xcopy/other paths.

## Non-obvious decisions

- **Driver type detection**: `ontap-san` and `ontap-san-economy` need different `importOriginalName` formats (`<flexvol>` vs `<flexvol>/<lun>`). Rather than requiring users to set `TRIDENT_DRIVER`, the importer reads `backendType` from the StorageClass — the standard Trident parameter — so most users need no extra configuration.
- **`notManaged: true`**: Trident adopts the LUN without taking ownership, leaving the source storage intact during migration.
- **VVol fall-through**: ONTAP VVol import via Trident requires mapping vSphere VVol IDs to LUN paths through the VASA protocol — not yet verified against a real VASA environment. VVol disks fall through to xcopy.
- **`nil, nil` from `Resolve()`**: The `CsiImportPlugin` interface documents that returning `(nil, nil)` signals "this vendor cannot handle this disk" and the caller falls through to other migration paths.
- **`STORAGE_SKIP_SSL_VERIFICATION`**: TLS skip flag is now read from the storage secret (consistent with xcopy), not the vSphere provider secret.
- **Labels fix**: CSI import PVCs now carry the full forklift tracking labels (`plan`, `migration`, `resource`) passed from the caller. This was a pre-existing gap that also affected HPE.

## Testing

Manual end-to-end test on eco lab cluster (edge114, ONTAP 9.17.1):
- VM `ontaprdm` with 1 VMDK + 1 RDM on `eco-iscsi-ds3` (iSCSI, `ontap-san` backend)
- RDM PVC bound in <5s via Trident import; VMDK handled by xcopy populator
- Migration succeeded: VM running on OCP with correct disk layout

Resolves: MTV-5955

### DIFF
--- a/pkg/controller/plan/adapter/vsphere/builder.go
+++ b/pkg/controller/plan/adapter/vsphere/builder.go
@@ -1745,33 +1745,8 @@ func (r *Builder) buildCsiImportPVC(
 	}, storageSecret); err != nil {
 		return nil, liberr.Wrap(err, "secretRef", csiCfg.SecretRef)
 	}
-	host := string(storageSecret.Data["STORAGE_HOSTNAME"])
-	user := string(storageSecret.Data["STORAGE_USERNAME"])
-	pass := string(storageSecret.Data["STORAGE_PASSWORD"])
-	skipSSL, err := strconv.ParseBool(string(storageSecret.Data["STORAGE_SKIP_SSL_VERIFICATION"]))
-	if err != nil {
-		r.Log.Error(err, "CSI import: invalid STORAGE_SKIP_SSL_VERIFICATION value, defaulting to false", "secretRef", csiCfg.SecretRef)
-		skipSSL = false
-	}
-
-	var missing []string
-	if host == "" {
-		missing = append(missing, "hostname")
-	}
-	if user == "" {
-		missing = append(missing, "username")
-	}
-	if pass == "" {
-		missing = append(missing, "password")
-	}
-	if len(missing) > 0 {
-		return nil, liberr.New(
-			fmt.Sprintf("storage secret %q is missing required keys: %s", csiCfg.SecretRef, strings.Join(missing, ", ")),
-		)
-	}
-
 	// Instantiate vendor plugin (xcopy pattern: switch in newCsiImportPlugin creates concrete type)
-	plugin, err := newCsiImportPlugin(csiCfg.StorageVendorProduct, host, user, pass, skipSSL)
+	plugin, err := newCsiImportPlugin(csiCfg.StorageVendorProduct, storageSecret.Data, r.Destination.Client, mapped.Destination.StorageClass)
 	if err != nil {
 		return nil, liberr.Wrap(err, "vendor", string(csiCfg.StorageVendorProduct))
 	}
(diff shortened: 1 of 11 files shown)


pkg/controller/plan/adapter/vsphere/csi_import.go
```diff
@@ -2,22 +2,37 @@ package vsphere
...
 	"github.com/kubev2v/forklift/pkg/storage/resolver/hpe"
+	"github.com/kubev2v/forklift/pkg/storage/resolver/ontap"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 // newCsiImportPlugin instantiates the CsiImportPlugin for the given storage vendor.
 // This is the equivalent of xcopy's main() vendor switch — the only place that imports
 // vendor sub-packages. Adding a new vendor = new sub-package + one case here.
-func newCsiImportPlugin(product forklift.StorageVendorProduct, host, user, pass string, skipSSL bool) (resolver.CsiImportPlugin, error) {
+func newCsiImportPlugin(product forklift.StorageVendorProduct, host, user, pass string, skipSSL bool, secretData map[string][]byte, k8sClient client.Client, storageClass string) (resolver.CsiImportPlugin, error) {
```
mnecas: NP: Now that we pass the `secretData` can we get the host, user, pass directly from the secret instead so we don't need to pass so many args?

pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -1754,6 +1733,10 @@ func (r *Builder) buildCsiImportPVC(
 	if err != nil {
 		return nil, liberr.Wrap(err, "disk", disk.File)
 	}
+	if vendorAnnotations == nil {
+		r.Log.Info("CSI import: vendor cannot handle this disk, deferring to other migration paths", "disk", disk.File)
+		return nil, nil //nolint:nilnil
```
mnecas, line 1761: Will this fallback to vmkfrstool copy?

amitosw15: yes, xcopy

mnecas (approved):
lgtm

PR #7485, opened 2026-07-15
MTV-5798 | Save multiple importer pod failure messages.

mrnold:
Issue: when importer pods hit CrashLoopBackoff, the first failure message can be something different than the final termination message that gets shown in the UI. It is useful to be able to see more than just the last message.

Partial fix: on importer restart limit failure, search through the Events list on the failing importer pod's PVC and add any ErrImportFailed warning events to the error list. This is not a complete fix, but it should work enough to help the situation from MTV-5798 (VMware API session timeout leading to later "snapshot not found" errors).

Reference: [MTV-5798](https://redhat.atlassian.net/browse/MTV-5798)

Resolves: MTV-5798

[MTV-5798]: https://redhat.atlassian.net/browse/MTV-5798?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/controller/plan/migration.go
+++ b/pkg/controller/plan/migration.go
@@ -13,6 +13,7 @@ import (
 	"time"
 
 	k8serr "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/fields"
 
 	api "github.com/kubev2v/forklift/pkg/apis/forklift/v1beta1"
 	"github.com/kubev2v/forklift/pkg/apis/forklift/v1beta1/plan"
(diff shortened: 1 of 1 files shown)


pkg/controller/plan/migration.go
```diff
@@ -2276,6 +2296,23 @@ func (r *Migration) propagateInspectionConcerns(vm *plan.VMStatus, result *api.I
...
+	events := &core.EventList{}
+	err := r.Destination.List(context.TODO(), events, &client.ListOptions{
+		Namespace: r.Plan.Spec.TargetNamespace,
+		FieldSelector: fields.AndSelectors(
+			fields.OneTermEqualSelector("involvedObject.kind", "PersistentVolumeClaim"),
+			fields.OneTermEqualSelector("involvedObject.name", pvcName),
+			fields.OneTermEqualSelector("reason", "ErrImportFailed"),
+			fields.OneTermEqualSelector("type", "Warning"),
+		),
+	})
```
mnecas, line 2312: Could we show even other events? Might be more user friendly

mnecas (approved):
lgtm

pkg/controller/plan/migration.go
mrnold: We can show whatever other events we want, but I think the UI would get pretty cluttered with every kind of event thrown in to the transfer status. The specific request was to preserve import failure reasons that were getting buried, so I would rather not widen the scope just yet.

PR #7493, opened 2026-07-16
MTV-6162 | Add olm_managed field to ForkliftController CRD schema

Hazanel:
OLM injects olm_managed: true into the spec of every CR it manages.
The ForkliftController CRD schema does not declare this field, causing
ArgoCD structured-merge-diff to fail with "field not declared in
schema" on every OLM reconciliation cycle .
Add the field as an optional boolean to the Go struct, regenerate the CRD, and add
specDescriptors entries to all three CSV manifests.

Ref: https://redhat.atlassian.net/browse/MTV-6162
Resolves: MTV-6162


Hazanel:
Hey @coderabbitai, please tale a look

Hazanel:
Thanks @coderabbitai, please check again

Hazanel:
Ticket was eventually marked as **won't do**

PR #7495, opened 2026-07-16
MTV-5917 | Change disk count concern from Critical to Warning

Hazanel:
Downgrade the disk count validation concern from Critical to Warning . The CNV
disk limit of 165 may cause network connectivity issues but should not block migration entirely.

See: https://redhat.atlassian.net/browse/CNV-91554
Ref: https://redhat.atlassian.net/browse/MTV-5917
Resolves: MTV-5917


solenoci (approved):
lgtm

PR #7517, opened 2026-07-16
MTV-6137 | Fall back to datastore matching when RDM NAA vendor resolution fails

borod108:
## Summary
- RDM disks whose NAA prefix resolves to a vendor with no matching storage map entry (e.g. 3PAR RDMs on a cluster with only PowerFlex configured) caused a hard failure at `CreateDataVolumes`: `"no storage map entry with offload plugin found for vendor primera3par"`
- Convert the three `return nil, fmt.Errorf(...)` paths in `PopulatorVolumes()` into log warnings that fall through to regular datastore-based matching, so the xcopy populator can still handle the disk (e.g. via `DISABLE_RDM_METHOD`)
- No changes to `rdm_storage.go` — the NAA resolution functions are unchanged; only the error handling in `builder.go` is softened

## Test plan
- [x] All 173 existing unit tests pass
- [x] Reproduced the bug on SNO cluster (tshefisno4) with VM `amit-3par-csi-test` (vm-170953, has a 3PAR RDM disk)
- [x] Old migration `rdm-3par-test` fails at `CreateDataVolumes` with `"no storage map entry with offload plugin found for vendor primera3par"`
- [x] New migration `rdm-3par-fix-test` with the fix passes `CreateDataVolumes`, creates PVCs, and reaches `CopyDisks`
- [x] Controller logs confirm the fallback: `"No storage map entry found for RDM vendor, falling back to datastore matching"` → `"RDM disk matched by datastore fallback"`

Resolves: MTV-6137

### DIFF
--- a/pkg/controller/plan/adapter/vsphere/builder.go
+++ b/pkg/controller/plan/adapter/vsphere/builder.go
@@ -1366,6 +1366,46 @@ func (r *Builder) PopulatorVolumes(vmRef ref.Ref, annotations map[string]string,
 	dsMapIn := r.Context.Map.Storage.Spec.Map
 	naaPrefixes := loadNAAPrefixes(r.Client)
 	dsNaaMap := make(map[string]string)
+
+	// Pre-pass: resolve RDM disks by NAA vendor prefix.
+	// Builds a map from disk index to the storage map entry whose offload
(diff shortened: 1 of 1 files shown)


mnecas (approved):
lgtm

PR #7544, opened 2026-07-20
MTV-5756 | Fix signal-conversion-done.ps1 for PowerShell Constrained Language Mode

Hazanel:
Replace .NET-based serial port communication (New-Object System.IO.Ports.SerialPort and [System.IO.File]::WriteAllText) with cmd /c echo, which is not subject to CLM restrictions. Windows VMs with Constrained Language Mode enabled blocked the original approach because CLM prevents instantiation of non-core .NET types and static method invocation.

Ref: https://redhat.atlassian.net/browse/MTV-5756
Resolves: MTV-5756


Hazanel:

yaacov (approved):

PR #7553, opened 2026-07-20
MTV-2419 | Preserve DataVolumeTemplates for OCP→OCP cold migration

gwencasey96:
## Summary
- Preserve `dataVolumeTemplates` on the target VM for OCP→OCP cold migration (previously cleared, so the VM only referenced the PVC).
- Sanitize copied templates: align names with Forklift-created DataVolumes/PVCs, clear `spec.source` (invalid on destination), clear namespace, and keep volume DataVolume refs in sync.
- Skip clearing `DataVolumeTemplates` in `vmTemplate` when the source provider is OCP so the OCP builder can keep them.

Resolves: MTV-2419

## Notes
Rebased successor to closed PR #4290 (stale). Same approach with sanitization from review feedback.

## Test plan
- [ ] Cold-migrate an OCP→OCP VM that has `dataVolumeTemplates`
- [ ] Confirm target VM retains sanitized `dataVolumeTemplates` (not only PVC volume refs)
- [ ] Confirm no KubeVirt/CDI attempt to re-import from the source URL/namespace
- [ ] Confirm volume DataVolume names match template names and destination PVCs
- [ ] Cold-migrate a VM without DVTs (no regression)
- [ ] Non-OCP source plan still clears DVTs as before

### DIFF
--- a/pkg/controller/plan/adapter/ocp/builder.go
+++ b/pkg/controller/plan/adapter/ocp/builder.go
@@ -300,10 +300,137 @@ func (r *Builder) VirtualMachine(vmRef ref.Ref, object *cnv.VirtualMachineSpec, 
 	// Preserve DataVolumeTemplates from source VM to maintain user workflows
 	// that may expect the VM's DataVolume to be present
 	object.DataVolumeTemplates = targetVmSpec.DataVolumeTemplates
+
+	// Sanitize DataVolumeTemplates to prevent conflicts with Forklift-created DataVolumes:
+	// 1. Match template names to the PVC names (Forklift creates DataVolumes with Name = pvc.Name)
+	// 2. Clear spec.source to prevent KubeVirt from trying to create new DataVolumes with invalid sources
+	// 3. Ensure namespace is set correctly (will be set by KubeVirt to match VM namespace)
+	// 4. Update volume references to match renamed DataVolumeTemplates
+	r.sanitizeDataVolumeTemplates(vmRef, object.DataVolumeTemplates, object)
+
 	r.mapNetworks(sourceVm, targetVmSpec)
 	r.mapVolumes(sourceVm, targetVmSpec, persistentVolumeClaims)
 
 	return nil
+}
+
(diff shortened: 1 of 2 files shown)


pkg/controller/plan/adapter/ocp/builder.go
```diff
@@ -297,12 +297,142 @@ func (r *Builder) VirtualMachine(vmRef ref.Ref, object *cnv.VirtualMachineSpec,
 
 	targetVmSpec := sourceVm.Spec.DeepCopy()
 	object.Template = targetVmSpec.Template
+	// Preserve DataVolumeTemplates from source VM to maintain user workflows
+	// that may expect the VM's DataVolume to be present
+	object.DataVolumeTemplates = targetVmSpec.DataVolumeTemplates
```
mnecas, line 302: @gwencasey96 i was always curious about this behaviour
will CNV reinicilize the datavolumetemplate causing the disk getting overriden as it's new VM resource or will it ignore it?

pkg/controller/plan/adapter/ocp/builder.go
```diff
@@ -297,12 +297,142 @@ func (r *Builder) VirtualMachine(vmRef ref.Ref, object *cnv.VirtualMachineSpec,
...
+
+	// Sanitize DataVolumeTemplates to prevent conflicts with Forklift-created DataVolumes:
+	// 1. Match template names to the PVC names (Forklift creates DataVolumes with Name = pvc.Name)
+	// 2. Clear spec.source to prevent KubeVirt from trying to create new DataVolumes with invalid sources
```
mnecas: what invalid source?

pkg/controller/plan/adapter/ocp/builder.go
gwencasey96: Neither. virt-controller will reconcile templates, and the DV is already created and the spec source is cleared so there’s not a second import and the already migrated disks won't be overridden.

pkg/controller/plan/adapter/ocp/builder.go
gwencasey96: sorry this was poor wording, updated

PR #7563, opened 2026-07-21
MTV-2838 | Preserve source PVC access/volume mode on cold OCP DVs

gwencasey96:
## Summary
- Cold OCP→OCP DataVolumes previously set only storage class and size, so CDI fell back to StorageProfile defaults and could flip RWO→RWX (seen on TPM persistent-state PVCs).
- When the plan storage mapping omits AccessMode/VolumeMode, preserve the source PVC values; mapping values still win when set.
- Adds unit tests for `applyDestinationStorageModes`.

Supersedes #4517 (that PR targeted live migration; this bug is cold OCP→OCP).

Resolves: MTV-2838

## Test plan
- [ ] Cold OCP→OCP migrate a VM whose source PVC is RWO on a storage class whose StorageProfile defaults to RWX; confirm destination DV/PVC stays RWO when the storage map omits AccessMode
- [ ] Confirm an explicit AccessMode/VolumeMode on the storage map is still honored
- [ ] Optional: TPM VM — confirm migrated `persistent-state-for-*` stays RWO; note if CNV still creates a second state PVC (expected leftover for CNV)

### DIFF
--- a/pkg/controller/plan/adapter/ocp/builder.go
+++ b/pkg/controller/plan/adapter/ocp/builder.go
@@ -140,8 +140,9 @@ func (r *Builder) DataVolumes(vmRef ref.Ref, secret *v1.Secret, configMap *v1.Co
 		if url == "" {
 			return nil, liberr.Wrap(fmt.Errorf("failed to get export URL, available formats: %v", volume.Formats))
 		}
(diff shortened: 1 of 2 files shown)


mnecas (commented):
were you able to reproduce the issue? could you please share sceenshots before and after the changes?

PR #7576, opened 2026-07-21
MTV-6176 | Require target credentials for EC2 providers

AmenB:
## Summary
- Make `targetAccessKeyId` and `targetSecretAccessKey` required fields in the EC2 provider secret. Without them, the provider fails validation with a Critical condition.
- This prevents cross-account migrations from silently creating EBS volumes in the wrong AWS account, which causes `InvalidVolume.NotFound` and migrations hanging forever.
- Introduces a `ProviderValidator` that delegates provider-specific validations to provider packages. Moves existing vSphere, HyperV, and EC2 validations into the dispatch.

## Design
- **General code** (`pkg/controller/provider/validation.go`): one line — `providervalidation.Build(r, r.Client).Validate(provider, secret)`
- **Dispatch** (`pkg/controller/provider/validation/doc.go`): `ProviderValidator` struct with a `Runner` interface. `Build()` takes the reconciler and k8s client. `Validate()` switches on provider type and calls the appropriate validations.
- **Runner interface**: exposes `ValidateVSpherePrivileges`, `ValidateSSHReadiness`, `ValidateSMBCSI`, `ValidateHyperVSettings` — implemented by the existing `Reconciler`, avoiding import cycles.
- **EC2 validator** (`pkg/provider/ec2/controller/validation/crossaccount.go`): checks two secret fields exist, sets Critical condition if missing.

## What moved into the dispatch
| Provider | Validations |
|----------|------------|
| EC2 | Target credentials check (new) |
| vSphere | `ValidateVSpherePrivileges`, `ValidateSSHReadiness` |
| HyperV | `ValidateSMBCSI`, `ValidateHyperVSettings` |

## Provider secret
Without target credentials (provider fails validation):
```yaml
data:
  accessKeyId: <source-account-key>
  secretAccessKey: <source-account-secret>
  region: us-east-2
```

With target credentials (provider passes validation):
```yaml
data:
  accessKeyId: <source-account-key>
  secretAccessKey: <source-account-secret>
  targetAccessKeyId: <cluster-account-key>
  targetSecretAccessKey: <cluster-account-secret>
  region: us-east-2
```

## Test plan
- [ ] Create EC2 provider without `--auto-target-credentials` → provider NOT Ready with `TargetCredentialsMissing` condition
- [ ] Create EC2 provider with `--auto-target-credentials` → provider Ready
- [ ] Verify vSphere provider validations still work (privileges, SSH readiness)
- [ ] Verify HyperV provider validations still work (SMB CSI, settings)
- [ ] Run unit tests: `go test ./pkg/controller/provider/...`

Resolves: MTV-6176

### DIFF
--- a/pkg/controller/provider/validation.go
+++ b/pkg/controller/provider/validation.go
@@ -18,6 +18,7 @@ import (
 	"github.com/kubev2v/forklift/pkg/controller/provider/container"
 	vsphereCollector "github.com/kubev2v/forklift/pkg/controller/provider/container/vsphere"
 	vsphere "github.com/kubev2v/forklift/pkg/controller/provider/model/vsphere"
+	providervalidation "github.com/kubev2v/forklift/pkg/controller/provider/validation"
 	libcnd "github.com/kubev2v/forklift/pkg/lib/condition"
 	liberr "github.com/kubev2v/forklift/pkg/lib/error"
 	"github.com/kubev2v/forklift/pkg/lib/inventory/model"
@@ -123,6 +124,7 @@ func (r *Reconciler) validate(provider *api.Provider) error {
 	if err != nil {
 		return liberr.Wrap(err)
 	}
+	providervalidation.Build(provider).Validate(provider, secret, r.Client)
 	err = r.validateVSpherePrivileges(provider)
 	if err != nil {
 		return liberr.Wrap(err)
--- a/pkg/controller/provider/validation.go
+++ b/pkg/controller/provider/validation.go
@@ -124,30 +124,11 @@ func (r *Reconciler) validate(provider *api.Provider) error {
 	if err != nil {
 		return liberr.Wrap(err)
 	}
-	providervalidation.Build(provider).Validate(provider, secret, r.Client)
-	err = r.validateVSpherePrivileges(provider)
-	if err != nil {
-		return liberr.Wrap(err)
-	}
 	err = r.inventoryCreated(provider)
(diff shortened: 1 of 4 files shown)


pkg/controller/provider/validation.go
```diff
@@ -1399,3 +1411,97 @@ func (r *Reconciler) validateHyperVSettings(provider *api.Provider) error {
 	provider.Status.DeleteCondition(SettingsNotValid)
 	return nil
 }
+
+func (r *Reconciler) validateEC2CrossAccount(provider *api.Provider, secret *core.Secret) error {
```
yaacov: we already have a plan validator in the pkg/providers/ec2
lets plan a mechanism to implement ec2 and other providers that will use /pkg/operators template that will delegate provider specific validations to the provider pkg

pkg/controller/provider/validation.go
```diff
@@ -123,6 +124,7 @@ func (r *Reconciler) validate(provider *api.Provider) error {
 	if err != nil {
 		return liberr.Wrap(err)
 	}
+	providervalidation.Build(provider).Validate(provider, secret, r.Client)
```
yaacov: instead of using a validator pattern, lets just use a switch case here that will run the validation method for each provider,

if we can more the check for inventoryCreated above this call, we can move the provider specific validation below into the switch case pattern

WDYT?

yaacov (approved):

PR #7595, opened 2026-07-21
MTV-5780 | Fix same-array PowerFlex xcopy adapter detection

amitosw15:
## Summary

- `hasSANAdapter` in the vsphere-copy-offload-populator only recognized `iqn.`/`fc.`-prefixed initiator IDs. PowerFlex's scini adapter gets its `Id` overridden with the raw SDC GUID (replacing the ESXi-masqueraded `fc.` identity it reports by default), so it never matched — incorrectly triggering the "local datastore" fallback, which then substituted a random unrelated FC/iSCSI adapter instead of the real SDC identity PowerFlex needs to register the host.
- Fix: treat a scini-driver adapter as a SAN adapter when the destination is confirmed PowerFlex, detected via the existing `SciniAware` interface at the one call site in `remote_esxcli.go` — no new plumbing through `main`/`factory`/`CopyContext` needed.
- Scoped narrowly per review guidance — does not touch the broader `ArrayIdentifier`/destination-aware adapter selection framework (tracked separately under MTV-5910).

## Verification

Verified end-to-end on real PowerFlex hardware (not just unit tests):
- Deployed the Dell CSM PowerFlex CSI driver on a test cluster to get a genuine PowerFlex-backed destination
- Confirmed the bug reproduces on a host where ESXi's scini adapter masquerades as an `fc.`-formatted virtual FC HBA
- Ran the fixed binary locally against real vCenter + PowerFlex (fast iteration loop via `make build` + direct invocation)
- Confirmed via live migration through the full K8s-orchestrated flow: `DiskTransfer` phase completes successfully, xcopy clone reaches 100%, and the volume is unmapped/cleaned up correctly

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./...` (full populator suite)
- [x] Live same-array PowerFlex xcopy migration completed successfully on real hardware


cmd/vsphere-copy-offload-populator/internal/populator/remote_esxcli.go
```diff
@@ -160,9 +160,13 @@ func (p *RemoteEsxcliPopulator) Populate(vmId string, migrationHostId, sourceVMD
...
 	setupLog.Info("initiator group", "group", xcopyInitiatorGroup)
 
+	// TODO(workaround): revisit — see MTV-5780.
+	_, isPowerFlexDestination := p.StorageApi.(SciniAware)
+	setupLog.Info("destination vendor check", "isPowerFlexDestination", isPowerFlexDestination)
+
 	// Filter HBA UIDs based on datastore active adapters
 	var dsActiveAdapters []vmware.HostAdapter
-	dsActiveAdapters, err = p.VSphereClient.GetDatastoreActiveAdapters(context.Background(), host, vmDisk.Datastore)
+	dsActiveAdapters, err = p.VSphereClient.GetDatastoreActiveAdapters(context.Background(), host, vmDisk.Datastore, isPowerFlexDestination)
```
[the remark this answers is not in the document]
amitosw15: fixed

cmd/vsphere-copy-offload-populator/internal/populator/remote_esxcli.go
```diff
@@ -160,9 +160,16 @@ func (p *RemoteEsxcliPopulator) Populate(vmId string, migrationHostId, sourceVMD
...
```
mnecas: I don't like passing vendor specific flag ot the generic interface

mnecas: @coderabbitai is there some alternative?

amitosw15: This suggested solution will not solve this:
After this line will execute:
`activeAdapters, err := p.VSphereClient.GetDatastoreActiveAdapters(
    context.Background(), host, vmDisk.Datastore,
)`
`activeAdapters` might not have the right adapters, as it will use the fallback that randomly picks one of the adapters.  If you use `a.Driver == scini` as the code does now, then you lose the local disk migration in case that PowerFlex is installed on the src host.

amitosw15: @coderabbitai  wdyt?

mnecas (approved):
lgtm

PR #7657, opened 2026-07-22
MTV-6194 | Nutanix provider base and inventory

mrnold:
Initial Nutanix provider with working inventory, continued from @jonner's starting work. This was mostly generated-by Cursor, I have reviewed all of the output (not all of it closely) and I have run a few dozen manual tests to check inventory against Prism Element and Prism Central. This includes some manual test tooling and a good amount of unit tests, and enough partially-filled-in other provider stuff that there shouldn't be any errors in the forklift controller or inventory logs. Hopefully this is enough to start actual migration and UI work.

Reference: [MTV-6194](https://redhat.atlassian.net/browse/MTV-6194)

Resolves: MTV-6194

[MTV-6194]: https://redhat.atlassian.net/browse/MTV-6194?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/controller/provider/container/nutanix/testdata/test_mock_server.sh
```diff
@@ -0,0 +1,119 @@
...
+    "$SERVER/api/nutanix/v3/clusters/list" \
+    -u wrong:credentials \
+    -H "Content-Type: application/json" \
+    -d '{}')
+
+if [ "$HTTP_CODE" = "401" ]; then
+    echo "✅ Authentication correctly rejected with 401"
+else
+    echo "❌ Expected 401, got $HTTP_CODE"
+fi
```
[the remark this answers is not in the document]
mrnold: Removed for now

pkg/controller/provider/container/nutanix/testdata/test_mock_server.sh
```diff
@@ -0,0 +1,119 @@
...
+echo "════════════════════════════════════════════════════════"
+echo "  Testing Mock Nutanix Prism API Server"
+echo "════════════════════════════════════════════════════════"
+echo
+
+# Check if server is running
+echo "Checking if server is running at $SERVER..."
+if ! curl $CURL_OPTS -s -f -u $USER:$PASSWORD "$SERVER/api/nutanix/v3" > /dev/null 2>&1; then
+    echo "❌ Server not responding. Please start the mock server first:"
+    echo "   go run mock_server.go"
```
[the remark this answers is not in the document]
mrnold: Removed for now

pkg/controller/provider/container/nutanix/testdata/test_mock_server.sh
```diff
@@ -0,0 +1,119 @@
+#!/bin/bash
+
+# Test script for the Mock Nutanix Prism API Server
+
+set -e
+
+SERVER="https://localhost:9440"
+USER="admin"
+PASSWORD="password"
+CURL_OPTS="-k"  # Skip certificate verification for self-signed cert
```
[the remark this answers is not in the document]
mrnold: Removed for now

pkg/controller/plan/adapter/nutanix/builder.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+func (r *Builder) DataVolumes(_ ref.Ref, _ *core.Secret, _ *core.ConfigMap, _ *cdi.DataVolume, _ *core.ConfigMap) (dvs []cdi.DataVolume, err error) {
+	// TODO: build CDI HTTP import DataVolumes from catalog image file URLs
+	return nil, nil
+}
+
+func (r *Builder) Tasks(_ ref.Ref) (tasks []*plan.Task, err error) {
+	// TODO: build per-disk progress tasks from inventory (skip CD-ROM)
+	return nil, nil
```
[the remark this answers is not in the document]
mrnold: This pull request is specifically for inventory, it is not expected to do any real migration work. The change set is massive enough as it is, these concerns will be addressed later.

pkg/controller/provider/container/nutanix/model.go
```diff
@@ -0,0 +1,684 @@
...
+				return false
+			}
+		} else {
+			return false
+		}
+	}
+
+	return false
+}
+
```
mnecas: These helper functions looks strange, do we need those?

pkg/controller/provider/container/nutanix/model.go
```diff
@@ -0,0 +1,684 @@
...
+// Apply cluster data to model.
+func applyCluster(entity map[string]interface{}, m *model.Cluster) {
+	// metadata
+	metadata, _ := entity["metadata"].(map[string]interface{})
```
mnecas: Can't we have a struct and cast it to it?

pkg/controller/provider/container/nutanix/normalize.go
```diff
@@ -0,0 +1,288 @@
+package nutanix
+
+import (
+	"fmt"
+	"strconv"
+
+	liberr "github.com/kubev2v/forklift/pkg/lib/error"
+)
+
+func storageContainerEntityFromV2(raw map[string]interface{}) map[string]interface{} {
```
mnecas: if i understand it correctly this generates the same data for v2 and v4 generates similar `map[string]interface{}` could we make it more version specific and have models and structs?
i'm generally not biggest fan of the `interface{}` as it allows us to do lot of hacks

mnecas (changes requested):

pkg/controller/provider/container/nutanix/model.go
mrnold: They aren't totally useless (I ripped out a few that were), but it looks like there are some existing methods from unstructured that would work instead of writing all new stuff. I will try it and see how it looks.

pkg/controller/provider/container/nutanix/normalize.go
jonner: I agree with this. I can look into it unless @mrnold  is already looking at it.

mrnold: @jonner Feel free to dig in to this, there is some weirdness about which stuff is available in which API, even down to minor dot versions.

mnecas (approved):
lgtm

PR #7667, opened 2026-07-23
MTV-6212 | Scope populator CR cleanup to completing VM in oVirt and OpenStack

Hazanel:
getPopulatorCrList() filtered only by migration UID, so when one VM completed its cleanup deleted populator CRs for ALL VMs in the plan. Remaining VMs failed with "OvirtVolumePopulator not found".

Add vmID to the label selector when called from DeletePopulatorDataSource(), matching the approach already used by the vSphere adapter. SetPopulatorCrOwnership() still lists all CRs (empty vmID) as intended.

Resolves: MTV-6212


Hazanel:

PR #7761, opened 2026-07-27
MTV-6251 | Adopt orphaned PowerMax xcopy SGs on retry instead of failing

borod108:
## Summary
- When descheduler evicts a populator pod, orphaned xcopy Storage Groups and Masking Views remain on the PowerMax array (no SIGTERM handler → deferred cleanup never runs). On retry, PowerMax rejects with "device already masked within a view."
- Instead of tearing down orphans, the retry pod now adopts the existing orphan SG/MV by swapping its identifiers. Zero extra API calls on normal path; 1 call (DeleteStorageGroup for empty SG) on adoption.
- Fixes a Map early-return bug where volume-already-in-SG skipped MV check/creation, leaving orphan masking views permanently.

## Test plan
- [x] Unit tests: 6 new `TestAdoptOrphanSG` subtests (no orphans, own SG ignored, single adoption, delete fails non-critical, multiple orphans, suffix filtering)
- [x] Unit tests: 3 new `TestMapWithOrphanAdoption` integration tests (orphan with MV, orphan without MV, no orphan normal flow)
- [x] Updated existing test for volume-already-in-SG to verify MV check still happens
- [x] All 36 PowerMax tests pass
- [ ] Partner lab validation on PowerMax array

Resolves: MTV-6251

### DIFF
--- a/cmd/vsphere-copy-offload-populator/vsphere-copy-offload-populator.go
+++ b/cmd/vsphere-copy-offload-populator/vsphere-copy-offload-populator.go
@@ -8,10 +8,12 @@ import (
 	"fmt"
 	"net/http"
 	"os"
+	"os/signal"
 	"path"
 	"strconv"
 	"strings"
 	"sync/atomic"
+	"syscall"
 	"time"
 
 	"github.com/prometheus/client_golang/prometheus/promhttp"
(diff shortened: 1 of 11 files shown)


cmd/populator-controller/populator-controller.go
```diff
@@ -53,7 +52,7 @@ var populators = map[string]populator{
 		kind:            "VSphereXcopyVolumePopulator",
 		resource:        "vspherexcopyvolumepopulators",
 		controllerFunc:  getVXPopulatorPodArgs,
-		imageVar:        "VSPHERE_COPY_OFFLOAD_POPULATOR_IMAGE",
+		imageVar:        "VSPHERE_XCOPY_VOLUME_POPULATOR_IMAGE",
```
amitosw15: why xcopy?

cmd/vsphere-copy-offload-populator/vsphere-copy-offload-populator_test.go
```diff
@@ -88,7 +88,7 @@ var _ = Describe("Populator", func() {
 
 			go func() {
```
rgolangh, line 89: can we have a test that catches the termination signal and verify some handling?

cmd/vsphere-copy-offload-populator/vsphere-copy-offload-populator.go
```diff
@@ -233,12 +235,23 @@ func main() {
...
 	copyStartTime = time.Now()
-	go p.Populate(sourceVmId, migrationHost, sourceVMDKFile, pv, hll, progressCh, xCopyUsedCh, quitCh)
+	go p.Populate(ctx, sourceVmId, migrationHost, sourceVMDKFile, pv, hll, progressCh, xCopyUsedCh, quitCh)
 
 	for {
 		select {
+		case sig := <-sigCh:
+			log.Info("received termination signal, cancelling operation and waiting for cleanup", "signal", sig)
+			cancel()
+			continue
```
rgolangh, line 254: Since at this point we are waiting for the cleanup, and the cleanup its self could either 1. panic 2. take time (api call to powermax....) - then we should carefull prepare for that. log, comment, whatever you think. 
After all, k8s will forcefully kill the pod after grace period 
https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/

anyhow, approved

PR #7790, opened 2026-07-28
MTV-6021 | vSphere scheduler with Storage offload

solenoci:
copy does not let other VMs to start when in conversion.

Fix: reduce cost when in phases after disk copy for storage offload

Resolves: MTV-6021
Ref: https://redhat.atlassian.net/browse/MTV-6021


mnecas (approved):

mnecas (commented):
lgtm, @solenoci please rebase

PR #7791, opened 2026-07-28
MTV-6213 | Cache the NetApp Shift SC resolution

solenoci:
Issue: the netapp shift resolution was called multiple times per VM and per plan reconcile which was very wasteful. Cache the result and only call the resolution once per plan reconcile.

Resolves: MTV-6213


solenoci:
@coderabbitai ptal

mnecas (approved):
lgtm

PR #7792, opened 2026-07-28
MTV-5316 | Fix static IP parsing to accept single DNS server

Hazanel:
The complementary static IP parser in injectComplementryStaticIPTemplate required at least 5 comma-separated fields (IP,GW,PREFIX,DNS1,DNS2), silently discarding entries with only one DNS server. The downstream slice `ipParts[3:]` already handles one-or-more DNS entries correctly.

Change `len(ipParts) < 5` to `len(ipParts) < 4` so that configurations with a single DNS are no longer skipped, which was causing an empty $networkConfigs array in the generated PowerShell script.

Ref: https://redhat.atlassian.net/browse/MTV-5316
Resolves: MTV-5316


Hazanel:

yaacov (approved):

PR #7799, opened 2026-07-28
MTV-6279 | Validate Hyper-V disk existence on SMB mount

Hazanel:
Previously, mapWindowsPathToSMB only performed string-based path mapping. In multi-node Hyper-V clusters a disk path can map syntactically but the file may not exist on the mounted SMB share (e.g. when the disk is local to a different cluster node).

validateDisksOnSMB now calls the hyperv-provider-server's /validate-disks endpoint to verify each mapped path with os.Stat. Disks that are missing have their SMBPath cleared, which triggers the existing OPA policy (hyperv.disk.smb_path.missing) to surface a warning in the UI.

The Rego assessment message is updated to mention multi-node clusters and recommend moving disks to shared storage.

Ref: https://redhat.atlassian.net/browse/MTV-6279
Resolves: MTV-6279


Hazanel:

yaacov (approved):

PR #7810, opened 2026-07-29
MTV-6203 | Implement conversion resume for migrations with completed disk copy

Hazanel:
When a migration fails during guest conversion (virt-v2v) after disks have been fully copied, users can now resume from the conversion phase without re-copying all disks. Applies to any plan where disk copy and conversion are separate phases (warm, cold with populators/CDI).

Changes:
- Add DisksCopied flag to VMStatus, set when entering conversion phase
- Add ResumeConversion field to Migration spec
- Add ConversionResumable condition on plans with resumable VMs
- Introduce ResumeConversion itinerary that skips disk copy and power-off phases (source already off, disks already on PVCs)
- Skip warm snapshot removal during resume (already cleaned at cutover)
- Guard checkConversionResumable to run on terminal plans and mark condition as durable
- Add unit tests for context, controller, validation, and migrator
- Update EC2 migrator to support resume-conversion itinerary

Ref: https://redhat.atlassian.net/browse/MTV-6203
Resolves: MTV-6203

***Blocked-By**:* https://github.com/kubev2v/forklift/pull/6838

### DIFF
--- a/pkg/controller/plan/migration.go
+++ b/pkg/controller/plan/migration.go
@@ -22,6 +22,7 @@ import (
 	"github.com/kubev2v/forklift/pkg/controller/plan/migrator"
 	planmigrbase "github.com/kubev2v/forklift/pkg/controller/plan/migrator/base"
 	"github.com/kubev2v/forklift/pkg/controller/plan/scheduler"
+	util "github.com/kubev2v/forklift/pkg/controller/plan/util"
 	"github.com/kubev2v/forklift/pkg/controller/provider/web"
 
 	libcnd "github.com/kubev2v/forklift/pkg/lib/condition"
@@ -97,6 +98,17 @@ func (r *Migration) Run() (reQ time.Duration, err error) {
 	err = r.begin()
 	if err != nil {
 		err = liberr.Wrap(err)
+		return
+	}
+
+	// begin() may reject the migration early (setting Failed/Canceled
+	// without Executing). Return now to prevent end() from overwriting
+	// the specific error with a generic status.
+	snapshot := r.Plan.Status.Migration.ActiveSnapshot()
+	if !snapshot.HasCondition(api.ConditionExecuting) {
+		if snapshot.HasAnyCondition(api.ConditionFailed, api.ConditionCanceled) {
+			reQ = NoReQ
+		}
 		return
 	}
 
@@ -198,6 +210,41 @@ func (r *Migration) begin() (err error) {
 	if snapshot.HasAnyCondition(api.ConditionExecuting, api.ConditionSucceeded, api.ConditionFailed, api.ConditionCanceled) {
 		return
 	}
+
+	// Validate resume-conversion prerequisites
+	if r.Migration.Spec.ResumeConversion {
+		hasResumable := false
+		for _, specVM := range r.Plan.Spec.VMs {
+			if !util.HasSeparateCopyAndConversion(r.Plan, specVM.Ref, r.Destination.Client) {
+				snapshot.SetCondition(libcnd.Condition{
+					Type:     api.ConditionFailed,
+					Status:   True,
+					Category: api.CategoryCritical,
+					Message: fmt.Sprintf(
+						"Resume conversion is not supported for VM %q: disk copy and conversion are not separate phases.",
+						specVM.ID),
+					Durable: true,
+				})
+				return
+			}
+			if status, found := r.Plan.Status.Migration.FindVM(specVM.Ref); found &&
+				status.DisksCopied && status.HasCondition(api.ConditionFailed) {
+				hasResumable = true
+			}
+		}
+		if !hasResumable {
+			snapshot.SetCondition(libcnd.Condition{
+				Type:     api.ConditionFailed,
+				Status:   True,
+				Category: api.CategoryCritical,
+				Message:  "Resume conversion requires VMs with completed disk copy (DisksCopied=true).",
+				Durable:  true,
+			})
+			return
+		}
+		r.Log.Info("Resume-conversion mode: skipping disk copy, running conversion only.")
+	}
+
 	r.Plan.Status.Migration.MarkReset()
 	r.Plan.Status.Migration.MarkStarted()
 	snapshot.SetCondition(
@@ -246,6 +293,17 @@ func (r *Migration) begin() (err error) {
 	for _, vm := range r.Plan.Spec.VMs {
 		status := r.migrator.Status(vm)
 		if status.Phase != api.PhaseCompleted || status.HasAnyCondition(api.ConditionCanceled, api.ConditionFailed) {
+			// For resume-conversion, only reset VMs that actually have their disks copied.
+			// VMs that failed before disk copy completed cannot be resumed — exclude them
+			// from this migration's VM list so stale failures don't cause end() to
+			// mark the overall migration as failed.
+			if r.IsResumeConversion() && !status.DisksCopied {
+				log.Info(
+					"Skipping VM without completed disk copy in resume-conversion.",
+					"vm",
+					vm.String())
+				continue
+			}
 			pipeline, pErr := r.migrator.Pipeline(vm)
 			if pErr != nil {
 				err = liberr.Wrap(pErr)
@@ -415,9 +473,12 @@ func markStartedStepsCompleted(vm *plan.VMStatus) {
 func (r *Migration) cleanup(vm *plan.VMStatus, failOnErr func(error) bool, forceDeleteGuestConversionPod bool) error {
 	r.Log.Info("Starting cleanup of migration resources.", "vm", vm.String())
 
-	// If the migration fails and the DeleteVmOnFailMigration is enabled, clean up the VM.
+	// If the migration fails, DeleteVmOnFailMigration is enabled and the disks were NOT fully copied, clean up the VM.
 	// When DeleteVmOnFailMigration is disabled, VM resources are preserved on failure.
-	if !vm.HasCondition(api.ConditionSucceeded) && (r.Plan.Spec.DeleteVmOnFailMigration || vm.DeleteVmOnFailMigration) && r.Plan.Spec.Type != api.MigrationOnlyConversion {
+	// During archive/cancel (forceDeleteGuestConversionPod=true), allow deletion even when
+	// DisksCopied is set — there's no opportunity to resume from those contexts.
+	if !vm.HasCondition(api.ConditionSucceeded) && (r.Plan.Spec.DeleteVmOnFailMigration || vm.DeleteVmOnFailMigration) &&
+		r.Plan.Spec.Type != api.MigrationOnlyConversion && (!vm.DisksCopied || forceDeleteGuestConversionPod) {
 		r.Log.Info("Deleting VM (failed migration with deleteVmOnFailMigration enabled).", "vm", vm.String())
 		if err := r.kubevirt.DeleteVM(vm); failOnErr(err) {
 			return err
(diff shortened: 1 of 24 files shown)


Hazanel:
Hey @coderabbitai, please:
1. create a flow diagram of the feature
2. review the code

pkg/controller/plan/migration.go
```diff
@@ -755,7 +810,8 @@ func (r *Migration) execute(vm *plan.VMStatus) (err error) {
 			vm.MarkStarted()
 			step.MarkStarted()
 			step.Phase = api.StepRunning
-			err = r.cleanup(vm, func(err error) bool { return err != nil }, true)
+			forceCleanup := !r.Context.IsResumeConversion()
+			err = r.cleanup(vm, func(err error) bool { return err != nil }, forceCleanup)
```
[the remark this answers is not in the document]
Hazanel: forceDeleteGuestConversionPod also gates PVC deletion when DisksCopied=true (line 473). Setting it to true during resume-conversion would delete the PVCs we need to reuse. The stale conversion CR is already invisible to EnsureConversion because it matches on LabelMigration=migration.UID, which differs between the old and new migrations.

Hazanel:
Hey @coderabbitai, please follow up on the fixes.
Overall - Do you think there is a better way to implement that?

Hazanel:
Hey @coderabbitai

Why we're keeping the current approach (resumeConversion: bool + DisksCopied) over a durable checkpoint:

The vmUUID label is the VM's immutable UUID from the source provider — it's not a generated correlation ID that could drift. PVC discovery via this label is deterministic across migration boundaries.

resumeConversion: bool honestly reflects the intent — resume is a binary decision ("skip disk copy, retry conversion"). A checkpoint implies resumability at arbitrary points, which we don't need and shouldn't promise.

The Ensure methods already provide artifact isolation — EnsureConversion/EnsureGuestConversionPod match on the current migration UID via conversionLabels(vm.Ref, false), so stale workloads from a prior migration are invisible to the new one. Exact PVC identity tracking would add redundancy, not safety.

A checkpoint introduces significant new surface area — new CRD or status subresource, lifecycle/garbage-collection policy, migration path for existing plans, UI for browsing/selecting checkpoints, and documentation. That's a feature in itself.

The validation gap is narrow and additive — rejecting a resume when storage/network mappings changed since the original copy is a worthwhile guard, but it's a validation check we can add to begin() without restructuring the API.

The current design is consistent with existing codebase patterns — conditions (ConversionResumable), boolean flags on spec/status, and label-based resource discovery are established conventions throughout the controller.

Incremental evolution is possible — if the feature expands later (multi-attempt tracking, partial disk resume, cross-plan resume), that's when the checkpoint abstraction earns its complexity. The boolean-based API can be deprecated in favor of resumeFrom: without breaking existing users.

All the safety-critical paths are covered — begin() validates DisksCopied=true + ConditionFailed before accepting a resume; cleanup() preserves PVCs when DisksCopied is set; deleteStaleConversionWorkloads removes old pods/CRs; Reset() preserves DisksCopied during resume; the r.Destination.Client fix ensures correct cluster targeting.

yaacov:
I have a concern about the split between skipping copy in the itinerary and discovering copied disks in getPVCs. The PR adds resumeConversionItinerary in pkg/controller/plan/migrator/base/migrator.go:245-259 (selected at 249-255) to skip power-off and disk-copy phases, while getPVCs in pkg/controller/plan/kubevirt.go:2078-2086 widens PVC discovery to vmID + vmUUID instead of the current migration UID. We may try to resume conversion while other code paths still create or list disks by migration UID, so copy skip and disk reuse could diverge depending on transfer path. What do you think?

I have a concern about the ensurer still listing and creating DataVolumes/PVCs with MigrationVMLabels (current migration UID) in pkg/controller/plan/ensurer/ensurer.go:106-124 and 141-158. On resume, getPVCs can find prior-run PVCs via vmUUID, but the ensurer may not see them and could create duplicates for the new migration UID. What do you think?

I have a concern about PopulatorVolumes still deduplicating against PVCs labeled with the current migration UID in pkg/controller/plan/adapter/vsphere/builder.go:1342-1353, while CSI import PVCs get LabelVMUUID in buildCsiImportPVC at 1781-1787 but populator PVCs do not. On resume we may try to reuse disks found by vmUUID in getPVCs while the populator path still treats prior PVCs as absent and creates new ones. What do you think?

I have a concern about stale conversion workload cleanup being described as fire-and-forget in pkg/controller/plan/migration.go:572-576, yet PhaseStarted at 835-850 calls deleteStaleConversionWorkloads after cleanup(..., forceCleanup=false) and treats errors as step failures. If a stale guest-conversion pod or CR is still terminating, we may block the new migration at start rather than wait/retry cleanly. What do you think?

I have a concern about DisksCopied being a VM-level flag set only when entering guest conversion in pkg/controller/plan/migration.go:1535-1538, while begin() resume validation at 215-236 requires prior plan status with DisksCopied=true on failed VMs. We may mark the whole VM resumable even if only some disks finished copy, or miss resume when copy completed but conversion was skipped. What do you think?

I have a concern about getPVCs returning every PVC matching vmID + vmUUID without deduplicating by disk identity in pkg/controller/plan/kubevirt.go:2100-2106. After a partial failed resume or repeated attempts, we may accumulate multiple PVCs per disk and pass them all downstream sorted by disk index. What do you think?

I have a concern about checkConversionResumable in pkg/controller/plan/validation.go:2553-2584 setting ConversionResumable from plan status (DisksCopied + failed VM) without verifying that reusable PVCs/DVs still exist in the target namespace. We may advertise that conversion can be resumed when disks were deleted by manual cleanup, deleteVmOnFailMigration, or a later migration. What do you think?

I have a concern about resume-conversion not being rejected upfront for plans where copy and conversion are not separate phases. begin() at pkg/controller/plan/migration.go:215-237 validates prior DisksCopied status but does not check util.HasSeparateCopyAndConversion (pkg/controller/plan/util/utils.go:239-253), which returns false for virt-v2v-for-transfer. We may allow Migration.spec.resumeConversion=true on paths where disk copy and conversion are a single step and resume cannot work as intended. What do you think?

I have a concern about itinerary proliferation: base migrator adds resumeConversionItinerary (pkg/controller/plan/migrator/base/migrator.go:245-259) and EC2 adds conversionOnlyItinerary (pkg/provider/ec2/controller/migrator/itinerary.go:65-83), each hardcoding which phases to skip. As more providers or warm/cold variants need resume, we may duplicate skip logic in multiple itineraries. What do you think?

I have an idea that might address several of the discovery and duplication concerns above. this PR widens PVC discovery in getPVCs on resume (plan + vmID + vmUUID at pkg/controller/plan/kubevirt.go:2078-2086) and skips copy phases via resumeConversionItinerary, but ensurer (pkg/controller/plan/ensurer/ensurer.go:106-124), PopulatorVolumes (pkg/controller/plan/adapter/vsphere/builder.go:1342-1353), and DV ensure paths still list by current migration UID. What do you think about adding a dedicated phase, PhaseReuseCopiedDisks, after PhaseStarted and before PhaseCreateDataVolumes, only when Migration.spec.resumeConversion=true? That phase would be the single place to discover prior-run disks using the same widened selector the PR already uses in getPVCs, validate them (Bound/Succeeded, matching AnnDiskSource, storage class/capacity), pick one PVC/DV per disk when multiple matches exist (kubevirt.go:2100-2106), record which disks are reused on VMStatus, and mark copy pipeline tasks complete so later phases do not start a new copy. The same selector helper would then be wired into ensurer, EnsureDataVolumes, and PopulatorVolumes dedup — not just getPVCs — so create and discovery paths agree. The PR’s resumeConversionItinerary would still handle what a phase cannot: skipping power-off, warm snapshot work, and EC2 provisioning. Stale conversion pod cleanup, begin() validation, and rejecting virt-v2v-for-transfer would remain separate follow-ups, @coderabbitai WDYT?

Hazanel:
Hey @coderabbitai, please follow up on the fixes from your discussion with @yaacov

Hazanel:
@coderabbitai - check now

pkg/controller/plan/migrator/base/migrator.go
```diff
@@ -251,7 +255,9 @@ func (r *BaseMigrator) Pipeline(vm plan.VM) (pipeline []*plan.Step, err error) {
 
 func (r *BaseMigrator) Itinerary(vm plan.VM) (itinerary *libitr.Itinerary) {
 	// Plan.Spec.Type supersedes the deprecated Warm boolean.
-	if r.Context.Plan.Spec.Type == api.MigrationOnlyConversion {
+	if r.IsResumeConversion() {
```
yaacov: @coderabbitai hi, does it make sense to make this a switch case instead of if else?

pkg/controller/plan/migration.go
```diff
@@ -1443,6 +1592,13 @@ func (r *Migration) execute(vm *plan.VMStatus) (err error) {
 				}
 			}
 		case api.PhaseCreateGuestConversionPod:
+
+			if !vm.DisksCopied && !r.IsResumeConversion() &&
```
yaacov, line 1596: The PR adds a single boolean VMStatus.DisksCopied that gets set atomically when the VM enters PhaseCreateGuestConversionPod

Problems with this approach:

All-or-nothing resume — If a VM has 5 disks and 4 copied successfully before a failure in the copy phase itself (not conversion), the VM is not marked DisksCopied=true, so the user must re-copy all 5 disks. The flag only fires at the conversion boundary, missing partial-copy scenarios entirely.

Redundant state — The PVCs already carry all the information needed. Each PVC has labels (vmID, vmUUID) and annotations (forklift.konveyor.io/diskSource, forklift.konveyor.io/diskIndex), and the ValidateResumePVCs function already checks their Bound state. The DisksCopied flag is essentially a cached query result that can become stale if PVCs are manually deleted.

No partial recovery — A per-disk approach (checking each PVC individually at resume time) would allow:

Picking up a VM where 3/5 disks were copied (re-copy only the missing 2)
No need for a dedicated DisksCopied field in the API at all
Natural idempotency: if the PVC exists and is Bound, skip it; otherwise, copy it

Suggested alternative: Instead of the DisksCopied boolean and the ResumeConversion itinerary that skips the entire copy phase, the normal itinerary's disk-copy phase should check each disk: "Does a Bound PVC with matching vmUUID + diskSource already exist? If yes, skip this disk."

pkg/provider/ec2/controller/migrator/itinerary.go
```diff
@@ -15,14 +15,21 @@ const (
...
-// Includes: Initialize→PreHook→PowerOff→CreateSnapshots→WaitSnapshots→[ShareSnapshots]→CreateVolumes→WaitForVolumes→CreatePVsAndPVCs→CreateGuestConversionPod→ConvertGuest→Finalize→CreateVM→RemoveSnapshots→PostHook→Completed.
-// Pre/post hooks are conditionally included based on VM hook configuration.
-// ShareSnapshots phase is conditionally included for cross-account migrations.
+// Itinerary builds the EC2 migration workflow sequence defining phase order.
+// For resume-conversion, skips EBS provisioning and runs only conversion phases.
 func (r *Migrator) Itinerary(vm planapi.VM) *libitr.Itinerary {
 	r.vm = &vm
 
-	itinerary := &libitr.Itinerary{
+	if r.IsResumeConversion() {
```
yaacov: EC2 is a no-copy provider. Its "disk copy" phase (PhaseCreateVolumes / PhaseWaitForVolumes / PhaseCreatePVsAndPVCs) doesn't transfer data — it creates EBS volumes from snapshots (a pointer/metadata operation) and creates PVCs pointing to them. This is fast and essentially free to repeat.

Yet the PR adds substantial EC2-specific resume machinery

Why this is problematic:

No real benefit — Since EC2 "copy" is just an API call to create a volume from a snapshot (seconds, not hours), there's no meaningful time saving from skipping it on resume.

Increased risk — The EC2 conversionOnlyItinerary still includes PhaseRemoveSnapshots, but skips volume creation. If the original volumes were somehow deleted between the failed run and the resume, the conversion pod would have no disks to work with — but the code never validates this because ValidateResumePVCs is only called from the base migrator path.

Divergent lifecycle — The EC2 Status() function was changed to return existing tracked state.
This is a subtle behavioral change for all EC2 migrations, not just resume ones. If a non-resume migration somehow has stale state in Plan.Status.Migration, this would now return that stale state.

Suggested alternative: In HasSeparateCopyAndConversion, add a check for the provider type. EC2 should return false (no separate copy/conversion phases for resume purposes), which would cause the validation in begin() to reject resume-conversion for EC2 plans with a clear error message.

yaacov:
In this pull request we introduces helper methods and consts but doesn't consolidate the existing inline implementations that pre-date them, my suggestions:

Replace getVmPVCs in validation.go to use planbase.LabelVMUUID directly
Deprecate or remove GetGuestConversionPod in favor of GetConversionPod
Refactor the pod/CR deletion in cleanup() to share logic with deleteStaleConversionWorkloads
Use HasSeparateCopyAndConversion in migration.go:2085 instead of the raw ShouldUseV2vForTransfer call

Hazanel:
@yaacov about your last comment:

- **getVmPVCs / kVmUuid in validation.go** — Already consolidated. kVmUuid is a package-level constant aliased to planbase.LabelVMUUID (defined in kubevirt.go:125). Since validation.go is in the same package, it uses the same shared constant. No divergence.

- **GetGuestConversionPod vs GetConversionPod** — These are pre-existing functions not introduced by this PR. They serve different callers and have different signatures. Consolidating them would touch unrelated code paths and is better as a separate cleanup PR.

- **Sharing cleanup() and deleteStaleConversionWorkloads** — Different semantics: cleanup() is fire-and-forget general teardown for all migration types. deleteStaleConversionWorkloads has wait/requeue logic with force-delete timeout, specific to the resume flow. Coupling them risks breaking non-resume cleanup behavior for a marginal DRY gain.

- **HasSeparateCopyAndConversion at migration.go:2241** — That call site asks "is this a v2v-for-transfer pod?" to select the correct progress monitor. HasSeparateCopyAndConversion answers a different question ("is this plan resumable?") and now also excludes EC2. Using it here would break progress monitoring for EC2 conversions.


@coderabbitai - WDYT?

pkg/controller/plan/migration.go
Hazanel: The current PR handles the most common and highest-value case: disk copy succeeded completely, conversion failed. This is the scenario where users lose the most time (hours of disk transfer wasted). Partial copy resume handles a rarer case (failure during copy itself), and the time savings are smaller (re-copy only the missing disks, but copy was already interrupted partway through).

The current DisksCopied flag and ValidateResumePVCs are forward-compatible — a future per-disk approach could build on top of the PVC validation infrastructure without breaking the existing API

@coderabbitai - WDYT?

yaacov (approved):

PR #7812, opened 2026-07-29
fix(populator): restore resource requests on populator worker pods

amitosw15:
## Problem

Populator worker pods (created by the populator-controller for Ovirt/OpenStack/VSphereXcopy volume population) came up with CPU and memory requests and limits explicitly set to zero. Kubernetes' QoS classifier treats a resource entry pinned at zero identically to having no entry at all, so these pods landed in the `BestEffort` QoS class. That made them a preferred target for descheduler eviction under normal node-balancing pressure, and for the vsphere-xcopy populator specifically, every eviction-triggered pod recreation repeated an expensive storage array masking-view setup/teardown cycle — reproduced on a live cluster during a multi-VM migration.

## What this achieves

Populator worker pods now come up `Burstable` with real, operator-configured resource requests/limits, removing them as a preferred descheduler eviction target and avoiding the repeated masking-view churn under eviction pressure.

## Non-obvious decision

The fix loads settings via `Migration.Load()` directly rather than the umbrella `Settings.Load()` used by every other `cmd/` binary in this repo. That's intentional: the populator-controller Deployment doesn't set `ROLE` or the virt-v2v/ovirt/vsphere env vars, so the umbrella loader would default to the main-controller role and then fail on those unrelated required vars. Calling `Migration.Load()` directly avoids that while still loading everything this binary actually needs.

## How to test

Verified on a live cluster: after deploying the fix, a freshly created populator worker pod showed `resources: {"limits":{"cpu":"1","memory":"1Gi"},"requests":{"cpu":"100m","memory":"512Mi"}}` and `qosClass: Burstable`, versus explicit `"0"`/`BestEffort` before the fix.

### DIFF
--- a/cmd/populator-controller/populator-controller.go
+++ b/cmd/populator-controller/populator-controller.go
@@ -73,10 +73,9 @@ func main() {
 	klog.InitFlags(nil)
 	flag.Parse()
 
-	resources, err := getResources()
-	if err != nil {
-		klog.Fatalf("Failed to parse resources: %v", err)
-	}
+	var populatorSettings settings.Populator
+	populatorSettings.Load()
+	resources := buildResources(&populatorSettings)
 
 	sigs := make(chan os.Signal, 1)
 	signal.Notify(sigs, syscall.SIGINT, syscall.SIGTERM)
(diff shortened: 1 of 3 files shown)


rgolangh:
for the sake of completeness, can you run a query on our CI runs and get a mean of the cpu/mem consumption and see it makes sense?

cmd/populator-controller/populator-controller.go
```diff
@@ -73,11 +73,13 @@ func main() {
 	klog.InitFlags(nil)
 	flag.Parse()
 
-	resources, err := getResources()
-	if err != nil {
-		klog.Fatalf("Failed to parse resources: %v", err)
+	// Migration.Load() only, not Settings.Load(): this Deployment sets no ROLE, so the umbrella loader would default MainRole=true and fail on missing virt-v2v/ovirt/vsphere env vars.
+	if err := settings.Settings.Migration.Load(); err != nil {
```
rgolangh: isn't this change bit risky now that migration.Load is being used? (risk of having unexpected/untested new settings in the pod spec?)

mnecas: I agree with Roy, this is ticking time bomb as this now it's using function from different controller which was not made for this one.

amitosw15: good catch thanks u

pkg/settings/populator.go
```diff
@@ -0,0 +1,76 @@
...
+	PopulatorContainerLimitsMemory   = "POPULATOR_CONTAINER_LIMITS_MEMORY"
+	PopulatorContainerRequestsCpu    = "POPULATOR_CONTAINER_REQUESTS_CPU"
+	PopulatorContainerRequestsMemory = "POPULATOR_CONTAINER_REQUESTS_MEMORY"
+)
+
+var (
+	DefaultPopulatorContainerLimitsCpu      = resource.NewQuantity(1000, resource.DecimalSI)
+	DefaultPopulatorContainerLimitsMemory   = resource.NewQuantity(1024, resource.BinarySI)
+	DefaultPopulatorContainerRequestsCpu    = resource.NewQuantity(100, resource.DecimalSI)
+	DefaultPopulatorContainerRequestsMemory = resource.NewQuantity(512, resource.BinarySI)
```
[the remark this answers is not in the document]
rgolangh: valid point - @amitosw15 .

PR #7443, opened 2026-07-12
MTV-5871 | Fix xcopyUsed metric scraping regex

amitosw15:
## Summary
- Fix the `xcopyUsed` metric scraping regex in the populator-machinery controller to use the correct full metric name (`vsphere_xcopy_volume_populator_xcopy_used`) and snake_case label (`owner_uid`) with wildcard matching for multi-label support
- Without this fix, the `xcopyUsed` field is never written to the populator CR status, so the plan task annotations never get the `xcopyUsed` annotation

## Root Cause
The original regex `xcopy_used{ownerUID="..."}` didn't match the actual Prometheus metric line because:
1. The metric name is `vsphere_xcopy_volume_populator_xcopy_used` (full name, not substring)
2. The label is `owner_uid` (snake_case), not `ownerUID` (camelCase)
3. The gauge has multiple labels (`owner_uid`, `storage_vendor`, `clone_method`), requiring `[^}]*` wildcard matching

## Test plan
- [x] E2E verified on ocp-edge114: 2 VMs in same plan, one on shared storage (xcopyUsed="1") and one on local disk (xcopyUsed="0")
- [x] Populator CR status correctly shows `xcopyUsed` field
- [x] Plan task annotations correctly show `xcopyUsed` per-disk

Resolves: MTV-5871

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #7838, opened 2026-07-30
MTV-5871 | Decouple xcopyUsed/vibVersion scraping from progress

amitosw15:
## Summary
- PR #7443 fixed the `xcopyUsed` metric regex, but the scrape remained structurally coupled to progress: any miss/error in the progress regex, parse, or CR-field update returned early and skipped the `xcopyUsed`/`vibVersion` scrapes for that cycle. Conversely, a failure writing `xcopyUsed` skipped the final `Update()` call, dropping an already-scraped progress value too.
- Makes progress, `xcopyUsed`, and `vibVersion` independent, best-effort scrapes in `updateProgress`: each is logged and skipped on failure rather than aborting the whole function, so a hiccup in one never blocks the others. A `dirty` flag ensures the CR is only updated when something new was actually scraped, and any field not touched this cycle naturally keeps its last persisted value.

## Test plan
- [x] `go build ./pkg/lib-volume-populator/...`
- [x] `go test ./pkg/lib-volume-populator/...`
- [x] Verified live on a cluster: ran a fresh XCOPY-offload migration (`vsphere` → `host`, ONTAP/Trident target) with the built image; observed the `VSphereXcopyVolumePopulator` CR reach `status.xcopyUsed=1` on a scrape cycle where `status.progress` was still empty, confirming the two scrapes are no longer coupled. Migration completed `DiskTransfer` successfully with `xcopyUsed: "1"` propagated onto the Plan's task annotations.

### DIFF
--- a/pkg/lib-volume-populator/populator-machinery/controller.go
+++ b/pkg/lib-volume-populator/populator-machinery/controller.go
@@ -951,26 +951,6 @@ func (c *controller) updateProgress(pod *corev1.Pod, pvc *corev1.PersistentVolum
 		c.parseAndRecordCompletion(bodyStr, pvc, cr)
 	}
 
-	// Pick the right progress regex for the populator type.
-	var importRegExp *regexp.Regexp
-	if populatorKind == api.VSphereXcopyVolumePopulatorKind {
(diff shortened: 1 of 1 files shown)


pkg/lib-volume-populator/populator-machinery/controller.go
amitosw15, line 960: @borod108 do you think we need to add retries? for stale objects

amitosw15: dont need it

mnecas (approved):

PR #7889, opened 2026-08-03
MTV-5700 | Warn when ESXi Disk.MaxLUN is too low for XCOPY

gwencasey96:
Add a non-blocking plan preflight warning when copy-offload migrations may exceed the default Disk.MaxLUN (1024) based on existing SCSI disks plus planned XCOPY disks per host.

Ref: https://redhat.atlassian.net/browse/MTV-5700
Resolves: MTV-5700


pkg/controller/plan/maxlun.go
```diff
@@ -0,0 +1,108 @@
+package plan
```
mnecas: Not biggest fan of one function files

pkg/controller/plan/maxlun.go
```diff
@@ -0,0 +1,108 @@
...
+		Type:     DiskMaxLUNTooLow,
+		Status:   True,
+		Reason:   NotValid,
+		Category: Warn,
+		Message: fmt.Sprintf(
+			"ESXi Disk.MaxLUN may be too low for this copy-offload migration on: %s. "+
+				"Increase Disk.MaxLUN to at least 2048 on those hosts (default is 1024).",
+			strings.Join(details, ", ")),
+		Items: items,
+	})
```
mnecas: What is issue with hte Disk.MaxLUN? 
Why do we need this check?

mnecas (changes requested):

pkg/controller/plan/maxlun.go
gwencasey96: refactored

pkg/controller/plan/maxlun.go
gwencasey96: If you don't want this check is important we can close the pr, but previously I asked you if you would be interested in more warnings for storage offload and you seemed interested-so I thought this is something we can check. I'm just hoping it will be more descriptive in case their setting is too low to discover certain LUNs

gwencasey96:
closing, never gained traction

PR #7892, opened 2026-08-03
MTV-6308 | Centralize ResolveTargetVmName across all adapters

Hazanel:
ResolveTargetVmName in base/template.go only checked status.migration.vms[].newName, so each adapter re-implemented spec.vms[].targetName lookup with divergent logic. The Hyper-V validator used vmRef.ID instead of the inventory vm.ID, which could miss the targetName match when IDs differed. The OCP validator ignored spec.targetName entirely, causing the validator and builder to evaluate PVC name templates with different TargetVmName values.

- ResolveTargetVmName now resolves in order: spec.targetName, status.newName, DNS1123 sanitization, vmName as-is.
- Plan lookup extracted into planOverrideName helper.
- All-invalid VM names produce a deterministic fallback from vmID.
- Hyper-V validator uses inventory vm.ID instead of vmRef.ID.
- OCP validator replaced local resolution with ResolveTargetVmName.
- vSphere validator/builder removed duplicated getPlanVMTargetName and inline DNS1123 check.

Ref: https://redhat.atlassian.net/browse/MTV-6308
Resolves: MTV-6308

### DIFF
--- a/pkg/controller/plan/adapter/base/template.go
+++ b/pkg/controller/plan/adapter/base/template.go
@@ -5,6 +5,7 @@ import (
 	"strings"
 
 	api "github.com/kubev2v/forklift/pkg/apis/forklift/v1beta1"
+	"github.com/kubev2v/forklift/pkg/controller/plan/util"
 	liberr "github.com/kubev2v/forklift/pkg/lib/error"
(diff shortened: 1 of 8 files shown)


Hazanel:

yaacov (approved):

PR #7943, opened 2026-08-04
MTV-6335 | Fix extra-v2v-conf and vddk-conf not mounted in Conversion CR pods

nijinashok:
The extra-v2v-conf and vddk-conf ConfigMap volumes were added to volumes/mounts but not to extraVolumes/extraMounts in podVolumeMounts. When UseConversionCR is enabled (default), EnsureConversion filters
 non-PVC volumes while populating the the Conversion CR. Since these ConfigMap volumes were missing from extraVolumes, they were silently dropped from the final pod spec, causing:
    

- extra-v2v-conf: "No such file or directory" when referenced via  VIRT_V2V_EXTRA_ARGS (-io vddk-config=/mnt/extra-v2v-conf/...)
-  vddk-conf: VDDK AIO optimization silently not applied even when useVddkAioOptimization is enabled on the provider
    
Resolves: https://redhat.atlassian.net/browse/MTV-6335

### DIFF
--- a/pkg/controller/plan/kubevirt.go
+++ b/pkg/controller/plan/kubevirt.go
@@ -3535,7 +3535,7 @@ func (r *KubeVirt) podVolumeMounts(vmVolumes []cnv.Volume, vddkConfigmap *core.C
 
 	extraConfigMapExists := len(Settings.Migration.VirtV2vExtraConfConfigMap) > 0
 	if extraConfigMapExists {
-		volumes = append(volumes, core.Volume{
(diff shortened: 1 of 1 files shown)


mnecas:
@solenoci could you ptal?

solenoci (approved):

nijinashok:
Looks like some of the tests are failing? I don't have the permission to view the  'rh-mtv-1-tenant' namespace to see the logs.

solenoci:
the failures are not related to this PR, merging

PR #7949, opened 2026-08-05
fix(hyperv): query VM power state on owner node in cluster mode

Hazanel:
PowerState() was always querying the provider entry-point host via LookupDomainByName, which fails in cluster mode when the VM resides on a different node. This caused Get-VM to return "VM not found", falling back to the stale inventory cache and leaving migrations stuck at WaitForPowerOff indefinitely.

Route the Get-VM query to vm.Host via RunOnNode when IsHyperVCluster() is true, mirroring what PowerOff() already does. Single-host mode is untouched — the original LookupDomainByName path remains the default.

Resolves: none

### DIFF
--- a/pkg/controller/plan/adapter/hyperv/client.go
+++ b/pkg/controller/plan/adapter/hyperv/client.go
@@ -3,6 +3,7 @@ package hyperv
 import (
 	"context"
 	"fmt"
+	"strings"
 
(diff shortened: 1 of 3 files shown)


Hazanel:

yaacov (approved):

PR #7972, opened 2026-08-05
MTV-6322 | Fix StorageMap entry order affecting CSI/xcopy resolution

amitosw15:
## Summary

- `CsiImportPVCs` resolved each disk's storage-map entry via `buildDatastoreMap`, which keyed a single map by source datastore ID and silently overwrote earlier entries sharing that ID.
- When a `StorageMap` legitimately lists the same datastore twice — one entry with `csiVolumeImport` for RDM disks, one with `vsphereXcopyConfig` for VMDKs — the outcome for CSI-eligible disks depended on which entry was listed last in `spec.map`. `csiVolumeImport` before `vsphereXcopyConfig` lost the CSI entry and the migration failed (MTV-6322).
- `buildDatastoreMap` now takes a `copyMethod` (`vddk`, `xcopy`, or `csi`), classified from each entry's `OffloadPlugin`, and only returns entries of that method. `CsiImportPVCs` and `PopulatorVolumes` each build their own method-scoped map from an independent pass over `spec.map`, so a `csiVolumeImport` entry and a `vsphereXcopyConfig` entry sharing a source datastore land in separate maps regardless of order — the reported order-dependence is gone by construction.
- Also bundled in while touching this path:
  - Extended the existing vendor/NAA-based RDM resolution (previously xcopy-only) to CSI via a shared, kind-aware `resolvePassthroughDisksByVendor`, so RDMs resolve consistently between the two copy methods.
  - Narrowed `CsiImportPVCs`' prior datastore-ID fallback — which matched *any* disk on a CSI-mapped datastore, including plain VMDKs that should never get a CSI PVC — to VVol disks specifically, via `resolveVVolDisksByDatastore`.
- An RDM disk whose NAA can't be matched to a known vendor is **not** guessed at via CSI datastore-ID matching: a `csiVolumeImport` entry is tied to a specific vendor's plugin/secret, so blindly matching by datastore risks picking the wrong vendor's target when a datastore hosts RDMs from more than one array. Such disks are left unresolved by `CsiImportPVCs` (logged) and fall through to `PopulatorVolumes`, which still has its own xcopy datastore-ID fallback.
- `DataVolumes` and `NetAppShiftPVCs` both request `vddk`-classified entries and keep filtering shift vs. plain VDDK downstream via the destination `StorageClass`, so shift behavior is unchanged.

## Out of scope / deferred

Two entries of the *same* copy method mapping to the *same* source datastore (e.g. two `xcopy` entries both sourced from the same datastore) are intentionally **not** rejected by this PR. Deciding whether that shape should be a hard validation error, or handled some other way, is a separate ambiguity-handling question still under discussion — deferred to a follow-up. Behavior for that specific case (last-write-wins) is unchanged from before this fix.

## Test plan

- [x] `go build ./...`
- [x] `go test ./pkg/controller/plan/...`
- [x] Unit test covering the exact reported scenario: a `StorageMap` with one `csiVolumeImport` and one `vsphereXcopyConfig` entry sourced from the same datastore resolves both correctly, independent of list order.
- [x] Unit test covering an RDM disk with an unrecognized vendor: `CsiImportPVCs` defers it (no PVC, no error) instead of guessing via datastore matching.

### DIFF
--- a/pkg/controller/plan/adapter/vsphere/builder.go
+++ b/pkg/controller/plan/adapter/vsphere/builder.go
@@ -472,13 +472,37 @@ func (r *Builder) Secret(vmRef ref.Ref, in, object *core.Secret) (err error) {
 	return
 }
 
-// buildDatastoreMap builds a map of storage mappings keyed by source datastore ID
-func (r *Builder) buildDatastoreMap() (map[string]*api.StoragePair, error) {
+type copyMethod string
+
+const (
+	copyMethodVDDK  copyMethod = "vddk"
+	copyMethodXcopy copyMethod = "xcopy"
+	copyMethodCsi   copyMethod = "csi"
+)
+
+// Shift destinations classify as vddk since Shift is identified by the destination StorageClass, not an offload plugin.
+func storagePairCopyMethod(pair *api.StoragePair) copyMethod {
+	if pair.OffloadPlugin != nil {
+		if pair.OffloadPlugin.VSphereXcopyPluginConfig != nil {
+			return copyMethodXcopy
+		}
+		if pair.OffloadPlugin.CsiVolumeImport != nil {
+			return copyMethodCsi
+		}
+	}
+	return copyMethodVDDK
+}
+
+// buildDatastoreMap builds a map of storage mappings keyed by source datastore ID, restricted to entries matching method.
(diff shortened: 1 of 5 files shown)


mnecas:
@amitosw15 do you require review from my side?

amitosw15:
@mnecas eventually yes, but i asked @borod108  to review it first

PR #7995, opened 2026-08-07
MTV-5701 | Warn when vmkfstools may exceed hostd-tmp memory

gwencasey96:
## Summary
- Adds a non-blocking plan Warn (`HostdTmpMemoryLow`) for VSphere copy-offload (XCOPY) plans when an ESXi host has ≥10 planned XCOPY disks.
- Points operators at raising `hostd-tmp` (default ~500MB) via `vmkfstools-wrapper/tweak-esxi-mem.sh`, matching the populator troubleshooting guidance.
- Same style as the MTV-5700 Disk.MaxLUN preflight warning.

## Test plan
- [ ] Unit: `go test ./pkg/controller/plan/ -run TestHostdTmp`
- [ ] Create an XCOPY plan with a host that has ≥10 XCOPY-mapped disks → expect `HostdTmpMemoryLow` Warn on the Plan
- [ ] Same plan with fewer than 10 disks on each host → no `HostdTmpMemoryLow` condition
- [ ] Non-XCOPY / non-vSphere plan → no condition

Resolves: MTV-5701

### DIFF
--- a/pkg/controller/plan/hostdtmp.go
+++ b/pkg/controller/plan/hostdtmp.go
@@ -6,7 +6,6 @@ import (
 	"strings"
 
 	api "github.com/kubev2v/forklift/pkg/apis/forklift/v1beta1"
-	"github.com/kubev2v/forklift/pkg/apis/forklift/v1beta1/ref"
 	"github.com/kubev2v/forklift/pkg/controller/provider/web"
 	vsphere "github.com/kubev2v/forklift/pkg/controller/provider/web/vsphere"
 	libcnd "github.com/kubev2v/forklift/pkg/lib/condition"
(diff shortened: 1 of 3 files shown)


pkg/controller/plan/hostdtmp.go
```diff
@@ -0,0 +1,95 @@
...
+	vsphere "github.com/kubev2v/forklift/pkg/controller/provider/web/vsphere"
+	libcnd "github.com/kubev2v/forklift/pkg/lib/condition"
+	"github.com/kubev2v/forklift/pkg/settings"
+)
+
+const (
+	HostdTmpMemoryLow = "HostdTmpMemoryLow"
+
+	// Default ESXi hostd-tmp limit (~500MB). See tweak-esxi-mem.sh in the populator.
+	defaultHostdTmpMB         = 500
```
mnecas: We can and do increase the memory limit so we can't have it hardcoded

pkg/controller/plan/hostdtmp.go
```diff
@@ -0,0 +1,95 @@
...
+			continue
+		}
+		vm, ok := v.(*vsphere.VM)
+		if !ok || vm.Host == "" {
+			continue
+		}
+		for _, disk := range vm.Disks {
+			m, found := plan.Map.Storage.FindStorage(disk.Datastore.ID)
+			if found && m.OffloadPlugin != nil && m.OffloadPlugin.VSphereXcopyPluginConfig != nil {
+				disksByHost[vm.Host]++
```
mnecas: We have storage offload dedicated migration hosts this uses only vm host

pkg/controller/plan/hostdtmp.go
```diff
@@ -0,0 +1,95 @@
...
+	libcnd "github.com/kubev2v/forklift/pkg/lib/condition"
+	"github.com/kubev2v/forklift/pkg/settings"
+)
+
+const (
+	HostdTmpMemoryLow = "HostdTmpMemoryLow"
+
+	// Default ESXi hostd-tmp limit (~500MB). See tweak-esxi-mem.sh in the populator.
+	defaultHostdTmpMB         = 500
+	hostdTmpDiskWarnThreshold = 10
```
mnecas: We thottle the number of disks running at once which can change

mnecas (changes requested):

gwencasey96:
closing, never gained traction

PR #8020, opened 2026-08-10
MTV-6301 | fix(metrics): re-emit xcopy per-disk metrics from forklift-controller

borod108:
## Summary
- Prometheus only scrapes the forklift-controller (:2112), not the populator-controller (:8082). Copy-offload completion metrics were invisible.
- Added 7 status fields to VSphereXcopyVolumePopulatorStatus to carry completion data (duration, result, vendor, method, protocol, bytes) from populator-controller to the CR.
- Forklift-controller now polls VSphereXcopyVolumePopulator CRs every 10s and emits per-disk `mtv_vsphere_xcopy_volume_populator_copy_duration_seconds` and `mtv_vsphere_xcopy_volume_populator_source_disk_bytes` gauges for all terminal states.

## Test plan
- [x] Unit test for `PopulatorOffloadInfo` covering all new status fields
- [x] `go build ./...` passes
- [x] End-to-end validated on cluster: manually patched CR status, confirmed metrics appear on `:2112/metrics`

Resolve: MTV-6301

### DIFF
--- a/pkg/monitoring/metrics/forklift-controller/metrics.go
+++ b/pkg/monitoring/metrics/forklift-controller/metrics.go
@@ -168,4 +168,54 @@ var (
 			"target",
 		},
 	)
+
+	// 'result' - [success, failure]
+	// 'migration' - [Migration UID]
+	// 'owner_uid' - [PVC UID]
+	// 'storage_vendor' - [ontap, powermax, ...]
+	// 'clone_method' - [vib, vddk]
+	// 'xcopy_used' - [0, 1]
+	// 'storage_protocol' - [iscsi, fc, nfs]
+	// 'vib_version' - [VIB version string, "unknown"]
+	xcopyDurationGauge = promauto.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "mtv_vsphere_xcopy_volume_populator_copy_duration_seconds",
+		Help: "Duration of copy-offload operation in seconds per disk",
+	},
+		[]string{
+			"result",
+			"migration",
+			"owner_uid",
+			"storage_vendor",
+			"clone_method",
+			"xcopy_used",
+			"storage_protocol",
+			"vib_version",
+		},
+	)
+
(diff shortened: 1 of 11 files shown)


pkg/monitoring/metrics/forklift-controller/metrics.go
```diff
@@ -168,4 +168,37 @@ var (
...
+			"migration",
+			"owner_uid",
+			"storage_vendor",
+			"clone_method",
+			"xcopy_used",
+			"storage_protocol",
+			"vib_version",
+			"type",
+		},
+	)
```
solenoci, line 220: please add examples of metrics/possible values similarly as they are in other metrics

pkg/monitoring/metrics/forklift-controller/xcopy_metrics.go
```diff
@@ -0,0 +1,105 @@
...
+
+var processedXcopyCompletions = make(map[string]struct{})
+
+func RecordXcopyMetrics(c client.Client) {
+	go func() {
+		for {
+			time.Sleep(10 * time.Second)
+
+			populators := api.VSphereXcopyVolumePopulatorList{}
+			err := c.List(context.TODO(), &populators)
```
solenoci: worried about scalability here, for plans/migrations, the list call shouldn't be a problem, but if we have 10 plans with 10 vms with 2 disks each running, this could take longer, do you think we need to address this or would this be fine?

solenoci (approved):
lgtm

solenoci:
@borod108 Hi Boris, there is a new check (verify-vendor) for storage offload and now it fails on this PR, can you please resolve it (running `go mod vendor`)? Thanks!

pkg/monitoring/metrics/forklift-controller/xcopy_metrics.go
borod108: done

PR #8060, opened 2026-08-11
fix(ci): verify vendored deps for vsphere-copy-offload-populator

amitosw15:
## Summary
- The vsphere-copy-offload-populator is a separate Go module that vendors `github.com/kubev2v/forklift` via a local `replace` directive, so changes to `pkg/apis`, `pkg/settings`, or `pkg/storage` outside the module are easy to forget to re-vendor, silently leaving it building against stale code.
- Adds a `verify-vendor-offload` job (alongside `golangci-lint`) that runs `go mod vendor` in the module and fails the check if it produces any diff.
- Syncs the module's vendor directory, which was already stale (pkg/apis/forklift/v1beta1, pkg/settings/migration.go, pkg/storage/resolver/resolver.go were out of date, and pkg/settings/populator.go was missing entirely).

## Test plan
- [x] `go build ./...` and `go vet ./...` in `cmd/vsphere-copy-offload-populator`
- [x] `go test ./...` (excluding e2e) in `cmd/vsphere-copy-offload-populator`
- [x] Verified the new CI check logic locally in an isolated git worktree: fails on stale vendor, passes once synced, and correctly flags a targeted edit to a vendored file


solenoci (approved):

mnecas (approved):
lgtm

PR #8072, opened 2026-08-11
MTV-6225 | Propagate NAA affinity label to intermediate prime PVC

amitosw15:
## Summary
- The `volume.csi.k8s.io/affinity-source-naa` label set on the target PVC for xcopy-offload migrations (`pkg/controller/plan/adapter/vsphere/builder.go`) was not being copied to the intermediate "prime" PVC created by the volume-populator machinery, so CSI drivers that use it for source/destination co-location (e.g. Dell CSM, used by PowerStore) couldn't see it on the prime PVC that actually triggers volume creation.
- Extends the existing allow-list of labels copied from the target PVC to the prime PVC (`migration`, `plan`, `vmID`) to also include `volume.csi.k8s.io/affinity-source-naa`, copied only when present on the source PVC.

Jira: https://redhat.atlassian.net/browse/MTV-6225

## Test plan
- [x] Deployed a build with the fix to a live OCP cluster's `forklift-volume-populator-controller` deployment
- [x] Ran an xcopy-offload migration and confirmed the prime PVC (`prime-<uid>`) now carries `volume.csi.k8s.io/affinity-source-naa` matching the target disk PVC
- [ ] CI

### DIFF
--- a/pkg/lib-volume-populator/populator-machinery/controller.go
+++ b/pkg/lib-volume-populator/populator-machinery/controller.go
@@ -83,7 +83,8 @@ const (
 
 	qemuGroup = 107
 
-	labelSourceHost = api.LabelSourceHost
+	labelSourceHost  = api.LabelSourceHost
+	labelTemplateNAA = "volume.csi.k8s.io/affinity-source-naa"
 )
 
 type empty struct{}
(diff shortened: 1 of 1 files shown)


pkg/lib-volume-populator/populator-machinery/controller.go
```diff
@@ -83,7 +83,8 @@ const (
 
 	qemuGroup = 107
 
-	labelSourceHost = api.LabelSourceHost
+	labelSourceHost  = api.LabelSourceHost
+	labelTemplateNAA = "volume.csi.k8s.io/affinity-source-naa"
```
mnecas, line 87: this is valid k8s label?

amitosw15: Yep validated with `content.IsLabelKey` ("k8s.io/apimachinery/pkg/api/validate/content")

amitosw15:
@mnecas Please review :)

mnecas:
lgtm but what is difference between https://github.com/kubev2v/forklift/pull/8245?

PR #8088, opened 2026-08-12
MTV-6321 | Fix primera3par rdm naa parsing

amitosw15:
## Summary
- `extractSerialFromNAA` (xcopy populator, primera3par plugin) only stripped a literal `naa.` prefix, so an ESXi 7.0+ `vml.` runtime device name was never recognized as a 3PAR device, logging `"does not appear to be a 3PAR device"` and falling back to a fragile substring search instead of resolving the volume directly.
- Adds inline `vml.` parsing: skip the fixed 10-hex-char ESXi 7.0+ preamble and take the next 32 hex chars as the NAA payload, mirroring the existing `naa.` path's OUI-prefix handling. Single-file, self-contained fix (no new shared package) per current scope.

Reproduces exactly with the ticket's device:
```
device="vml.020002000060002ac0000000000000628200021f6b565620202020"
err="NAA vml.020002000060002ac0000000000000628200021f6b565620202020 does not appear to be a 3PAR device (expected prefix 60002ac)"
```

## Test plan
- [x] Added a permanent regression test case in `clonner_test.go` using the exact ticket device.
- [x] Verified against all 19 real RDM devices currently in the eco-vsphere lab (3PAR/NetApp/Pure) — the 4 real 3PAR devices resolve correctly, the 15 non-3PAR devices are still correctly rejected.
- [x] `go test ./cmd/vsphere-copy-offload-populator/internal/primera3par/...` passes.

Jira: https://redhat.atlassian.net/browse/MTV-6321

### DIFF
--- a/cmd/vsphere-copy-offload-populator/internal/primera3par/clonner.go
+++ b/cmd/vsphere-copy-offload-populator/internal/primera3par/clonner.go
@@ -248,7 +248,7 @@ func (c *Primera3ParClonner) RDMCopy(vsphereClient vmware.Client, vmId string, s
 func (c *Primera3ParClonner) resolveRDMToLUN(deviceName string) (populator.LUN, error) {
 	c.log.V(2).Info("resolving RDM device to LUN", "device", deviceName)
 
-	serial, err := extractSerialFromNAA(deviceName)
+	serial, err := extractSerialFromVML(deviceName)
 	if err != nil {
 		c.log.Info("could not extract serial from NAA, trying to find by listing volumes", "device", deviceName, "err", err)
 		return c.findVolumeByDeviceName(deviceName)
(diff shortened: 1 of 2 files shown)


cmd/vsphere-copy-offload-populator/internal/primera3par/clonner.go
```diff
@@ -277,7 +277,27 @@ func (c *Primera3ParClonner) resolveRDMToLUN(deviceName string) (populator.LUN,
...
+		const naaHexLen = 32
+		rest := strings.TrimPrefix(naa, "vml.")
+		if len(rest) <= vmlHeaderLen {
+			return "", fmt.Errorf("NAA %s does not appear to be a 3PAR device (vml. name too short)", naa)
+		}
+		payload := rest[vmlHeaderLen:]
+		if len(payload) > naaHexLen {
+			payload = payload[:naaHexLen]
+		}
+		naa = payload
```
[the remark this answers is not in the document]
amitosw15: Will fail downstream with no volume with this serial found...

cmd/vsphere-copy-offload-populator/internal/primera3par/clonner.go
```diff
@@ -277,7 +278,19 @@ func (c *Primera3ParClonner) resolveRDMToLUN(deviceName string) (populator.LUN,
 
 func extractSerialFromNAA(naa string) (string, error) {
 	naa = strings.ToLower(naa)
-	naa = strings.TrimPrefix(naa, "naa.")
+
+	switch {
+	case strings.HasPrefix(naa, "naa."):
```
borod108: Hmmm, why is it extract serial from NAA if it could be vml? What in the flow is wrong?

amitosw15: it always assumes the serial will start with "naa.", but as we saw it might start with "vml."

amitosw15: checking this

amitosw15: tnx, changed

amitosw15:
@mnecas Please review :)

PR #8124, opened 2026-08-13
MTV-6364 | Extend and fix vmware removal scripts

solenoci:
added missing parts:
- check and remove MSI entries
- check and validate MSU entries
- check and remove class installer products (backup to MSI removal)
- check and remove broadcom renamed entries/drivers in a safe way
- check and remove control sets
- check and remove class registrations
- push current control set enum removal to a fallback option so we don't corrupt the database
- check and remove CLSID and TypeLib stale COM registrations
- check and remove stale autostart entries
- check and remove current control set video entry for vm3dmp
- expand removal of vmware directories and files across users
- reorder the phases so we hit less locks
- expand file checks with wildcards
- use windows get driver first as that is language agnostic, failover to pnputil
- check and remove ghost devices 
- add dism corruption health check
- fix 64 vs 32 bit resolution problem
- don't fail on ACL access errors

Resolves: MTV-6364

### DIFF
--- a/docs/9100_cleanup_vmware.md
+++ b/docs/9100_cleanup_vmware.md
@@ -256,12 +256,21 @@ HKLM\SOFTWARE\Microsoft\Windows\CurrentVersion\App Paths\vmtoolsd.exe
 ### Directories (`$VMwareDirs`)
 
 ```
-%ProgramFiles%\VMware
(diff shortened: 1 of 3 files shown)


solenoci:
@coderabbitai ptal

PR #8149, opened 2026-08-15
MTV-6405 | Cache Hyper-V inventory per cycle, drop per-VM WinRM calls

krcmarik:
ListVMs and ListNetworks were re-executed on every call within a single inventory cycle, and the standalone (non-cluster) path additionally issued a separate WinRM round trip per VM via getVMFromDomain/extractDisks/ extractNICs to enrich disks and NICs one VM at a time. On hosts with many VMs this produced O(n) extra WinRM calls per Load/Refresh cycle, which is the dominant cost of Hyper-V inventory collection.

Changes:
- Add per-cycle in-memory caches (vmCached/vmCache, netCached/netCache) on Client so ListVMs/ListNetworks only hit the driver once per cycle instead of once per caller.
- Remove the standalone-only getVMFromDomain/extractDisks/extractNICs path; all VMs now go through getVMBaseFromDomain plus the existing batch PowerShell enrichment (applyBatchDetails), which already covers both standalone and cluster mode with a single batched WinRM call.
- Rename InvalidateClusterCache to InvalidateCycleCache and broaden it to also clear the new VM/network caches; the collector calls it once at the start of each Load and Refresh phase so results are still fresh per cycle.
- Add unit tests asserting ListVMs/ListNetworks call the driver exactly once before invalidation and again after InvalidateCycleCache.

Ref: https://redhat.atlassian.net/browse/MTV-6405
Resolves: MTV-6405

### DIFF
--- a/pkg/controller/provider/container/hyperv/client.go
+++ b/pkg/controller/provider/container/hyperv/client.go
@@ -93,6 +93,10 @@ type Client struct {
 	smbMountPath     string
 	smbWindowsPrefix string
 	cache            *clusterCache
+	vmCached         bool
+	vmCache          []types.VM
+	netCached        bool
+	netCache         []types.Network
 }
 
 // Connect establishes a WinRM connection to the HyperV host using Secret credentials.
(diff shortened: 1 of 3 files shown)


pkg/controller/provider/container/hyperv/client.go
```diff
@@ -241,12 +253,7 @@ func (r *Client) ListVMs() ([]types.VM, error) {
 
 	var vms []types.VM
 	for _, domain := range domains {
-		var vm *types.VM
-		if isCluster {
-			vm, err = r.getVMBaseFromDomain(domain)
-		} else {
-			vm, err = r.getVMFromDomain(domain, networks, r.smbWindowsPrefix)
-		}
+		vm, err := r.getVMBaseFromDomain(domain)
```
yaacov, line 256: @coderabbitai hi, i am concerned about the drop of out of cluster special flow, what am i missing?

yaacov (approved):

PR #8163, opened 2026-08-16
MTV-6420 | allow Host Ready unless ESXi overallStatus is red

yaacov:
Treat yellow/gray host overallStatus as a Warn concern instead of blocking migration-network Host CRs. Keep NotHealthy Critical only for red. Reconcile Hosts when inventory overallStatus changes.

Ref: https://redhat.atlassian.net/browse/MTV-6420
Resolves: MTV-6420

Epic: https://redhat.atlassian.net/browse/MTV-6415
RFE: https://redhat.atlassian.net/browse/MTV-5820

### DIFF
--- a/pkg/controller/host/validation.go
+++ b/pkg/controller/host/validation.go
@@ -30,6 +30,7 @@ const (
 	ConnectionTestFailed    = "ConnectionTestFailed"
 	InMaintenance           = "InMaintenance"
 	NotHealthy              = "NotHealthy"
+	HostStatusNotGreen      = "HostStatusNotGreen"
 )
 
 // Categories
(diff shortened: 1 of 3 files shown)


Hazanel (approved):

pkg/controller/host/validation.go
```diff
@@ -351,3 +342,31 @@ func (r *Reconciler) testConnection(host *api.Host) (err error) {
...
+		host.Status.SetCondition(
+			libcnd.Condition{
+				Type:     NotHealthy,
+				Status:   True,
+				Reason:   StateEvaluated,
+				Category: Critical,
+				Message:  "Host overallStatus is red.",
+			},
+		)
+	}
```
mnecas, line 371: So if the status is red it gets both conditions?

PR #8193, opened 2026-08-19
MTV-6400 | Assign disk serial number on every migrated VM

Hazanel:
Add a shared DiskSerial helper that truncates the provider's disk
  UUID to 20 characters (virtio-blk limit) or falls back to a
  deterministic sha256-based serial from vmID+diskKey.  Hyphens are
  preserved so the truncated value matches what QEMU derives from the
  raw UUID, keeping /dev/disk/by-id/ references stable.
  
  vSphere: remove the DiskEnableUuid && SCSI guard so every disk gets
  a serial from backing.Uuid, not just SCSI disks with EnableUUID=TRUE.
  Hyper-V, OpenStack, OVF/OVA: set serial from the disk/volume ID.
  oVirt: unchanged (already sets Serial: da.Disk.ID).
  
  Ref: https://redhat.atlassian.net/browse/MTV-6400
  Resolves: MTV-6400

### DIFF
--- a/pkg/controller/plan/adapter/base/disk_serial.go
+++ b/pkg/controller/plan/adapter/base/disk_serial.go
@@ -1,0 +1,45 @@
+package base
+
+import (
+	"crypto/sha256"
+	"encoding/hex"
+	"fmt"
+	"strings"
+)
+
+const MaxDiskSerialLength = 20
+
+// DiskSerial returns a serial number for a KubeVirt disk device.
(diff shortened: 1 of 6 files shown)


Hazanel:
Hey @coderabbitai - ptal

Hazanel:
@coderabbitai - Are you satisfied with the suggested solution ?

yaacov (approved):

PR #8208, opened 2026-08-20
MTV-6462 | Append udev rules to existing 70-persistent-net.rules

nijinashok:
- Stop exiting early when `70-persistent-net.rules` already exists; instead filter generated rules per MAC/name and append only new ones
- Add `filter_existing_udev_rules()` to skip rules whose MAC address or interface name is already in the file

**Before fix:**

`File /etc/udev/rules.d/70-persistent-net.rules already exists and is not empty. Exiting.
`

**After fix:**

```
Info: MAC 00:50:56:86:78:3b already present in /etc/udev/rules.d/70-persistent-net.rules, skipping rule.
Wrote udev rules to /etc/udev/rules.d/70-persistent-net.rules
New udev rule:
SUBSYSTEM=="net", ACTION=="add", DRIVERS=="?*", ATTR{address}=="00:50:56:86:78:3b", NAME="ens33"
SUBSYSTEM=="net",ACTION=="add",ATTR{address}=="00:50:56:86:87:a4",NAME="ens35"
Created /etc/systemd/network/10-v2v-ens33.link: 00:50:56:86:78:3b -> ens33
Created /etc/systemd/network/10-v2v-ens35.link: 00:50:56:86:87:a4 -> ens35

```

Ref: https://redhat.atlassian.net/browse/MTV-6462
Resolves: MTV-6462


pkg/virt-v2v/customize/scripts/rhel/run/network_config_util.sh
```diff
@@ -605,17 +599,59 @@ check_dupe_hws() {
...
+            udev_from_nm_dhcp_lease
+            udev_from_dhclient_lease
+            udev_from_netplan
+            udev_from_ifquery
+            udev_from_wicked
+        } | check_dupe_hws | filter_existing_udev_rules
+    )
+
+    if [ -n "$new_rules" ]; then
+        printf '%s\n' "$new_rules" >> "$UDEV_RULES_FILE"
```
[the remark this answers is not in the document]
nijinashok: Done, added a check for this.

pkg/virt-v2v/customize/scripts/rhel/run/network_config_util.sh
```diff
@@ -605,17 +599,62 @@ check_dupe_hws() {
...
+    fi
+
+    while IFS= read -r rule || [ -n "$rule" ]; do
+        [ -z "$rule" ] && continue
+
+        mac=$(echo "$rule" | sed -nE 's/.*ATTR\{address\}=="([^"]+)".*/\1/p')
+        name=$(echo "$rule" | sed -nE 's/.*NAME="([^"]+)".*/\1/p')
+        mac_upper=$(echo "$mac" | tr 'a-f' 'A-F')
+
+        if [ -n "$mac" ] && grep -ioE 'ATTR\{address\}=="[0-9A-Fa-f:]+"' "$UDEV_RULES_FILE" | tr 'a-f' 'A-F' | grep -qF "$mac_upper"; then
```
Hazanel: Simplify MAC matching: The grep -ioE ... | tr 'a-f' 'A-F' | grep -qF pipeline is redundant — grep -i already does case-insensitive matching. This can be simplified to:

`if [ -n "$mac" ] && grep -qi "ATTR{address}==\"${mac}\"" "$UDEV_RULES_FILE"; then`
Same result, one process instead of three per rule.

pkg/virt-v2v/customize/scripts/rhel/run/network_config_util.sh
```diff
@@ -605,17 +599,62 @@ check_dupe_hws() {
     echo "$deduped"
 }
 
+# Skip a generated rule when its MAC or interface name is already in
+# 70-persistent-net.rules.
+filter_existing_udev_rules() {
```
Hazanel, line 604: SonarCloud flagged this: since the script uses #!/bin/bash, prefer [[ ... ]] over [ ... ] for all the new conditionals (lines 605, 647, 651). Safer quoting and consistent with bash idiom.

pkg/virt-v2v/customize/scripts/rhel/run/network_config_util.sh
```diff
@@ -605,17 +599,62 @@ check_dupe_hws() {
...
+            udev_from_nm_dhcp_lease
+            udev_from_dhclient_lease
+            udev_from_netplan
+            udev_from_ifquery
+            udev_from_wicked
+        } | check_dupe_hws | filter_existing_udev_rules
+    )
+
+    if [ -n "$new_rules" ]; then
+        if [ -s "$UDEV_RULES_FILE" ] && [ "$(tail -c1 "$UDEV_RULES_FILE")" != "" ]; then
```
Hazanel, line 647: The trailing-newline guard looks correct, but please add a test fixture for it — an existing 70-persistent-net.rules file that does not end with \n. Without that fixture, this code path is untested. If the tail -c1 check has a subtle edge case (e.g., empty file vs. no-newline file), you'd want to catch it here.

pkg/virt-v2v/customize/scripts/rhel/run/network_config_util.sh
```diff
@@ -605,17 +599,62 @@ check_dupe_hws() {
...
+    new_rules=$(
+        {
+            udev_from_ifcfg
+            udev_from_nm
+            udev_from_nm_dhcp_lease
+            udev_from_dhclient_lease
+            udev_from_netplan
+            udev_from_ifquery
+            udev_from_wicked
+        } | check_dupe_hws | filter_existing_udev_rules
```
Hazanel, line 643: Minor: the old code had 2>/dev/null on the pipeline. The new code drops it. This is probably fine (better to see errors in logs), but worth confirming that check_dupe_hws or the udev_from_* functions don't emit expected/harmless warnings to stderr that would now clutter guest firstboot logs.

pkg/virt-v2v/customize/scripts/rhel/run/network_config_util.sh
nijinashok: Thanks, done.

pkg/virt-v2v/customize/scripts/rhel/run/network_config_util.sh
nijinashok: Thanks, done.

pkg/virt-v2v/customize/scripts/rhel/run/network_config_util.sh
nijinashok: Thanks, added new test.

pkg/virt-v2v/customize/scripts/rhel/run/network_config_util.sh
nijinashok: The error will be observed only in the virt-v2v pod logs and we shouldn't see them in guest firstboot logs. Since v2v is already verbose, I think this should be ok.

pkg/virt-v2v/customize/scripts/rhel/run/network_config_util.sh
nijinashok: I have reverted this since ifcfg-double-quotes-test.d is failing:

```
 ./pkg/virt-v2v/customize/scripts/rhel/run/test-network_config_util.sh: 611: /home/runner/work/forklift/forklift/pkg/virt-v2v/customize/scripts/rhel/run/network_config_util.sh: [[: not found
./pkg/virt-v2v/customize/scripts/rhel/run/test-network_config_util.sh: 616: /home/runner/work/forklift/forklift/pkg/virt-v2v/customize/scripts/rhel/run/network_config_util.sh: [[: not found
./pkg/virt-v2v/customize/scripts/rhel/run/test-network_config_util.sh: 621: /home/runner/work/forklift/forklift/pkg/virt-v2v/customize/scripts/rhel/run/network_config_util.sh: [[: not found
 ./pkg/virt-v2v/customize/scripts/rhel/run/test-network_config_util.sh: 610: /home/runner/work/forklift/forklift/pkg/virt-v2v/customize/scripts/rhel/run/network_config_util.sh: [[: not found
./pkg/virt-v2v/customize/scripts/rhel/run/test-network_config_util.sh: 646: /home/runner/work/forklift/forklift/pkg/virt-v2v/customize/scripts/rhel/run/network_config_util.sh: [[: not found

```

mnecas:
kinda curious how could this happen? i guess there is quite low chance of users having the same udev rule name so i guess the conversion war run 2x on the VM

nijinashok:
The 70-persistent-net.rules filename is commonly used to define interface naming and is also recommended in the Red Hat documentation. Refer https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/8/html/configuring_and_managing_networking/consistent-network-interface-device-naming_configuring-and-managing-networking

Therefore, it is possible that the user manually created this rule using the standard filename. In the environment where this was observed, the entry was not added by Forklift, as the rule contained DRIVERS=="?*", which is not present in the MTV template.

nijinashok:
@mnecas  @Hazanel  Please let me know if anything else is needed from my side. Thanks!

gojeaqui:
This issue is pretty common, thanks @nijinashok for providing a fix

PR #8211, opened 2026-08-20
MTV-6393 | Retry transient block device errors before virt-v2v conversion

Hazanel:
NVMe/TCP namespace rescans can race with PVC attach/detach, causing ENXIO when virt-v2v opens a block device immediately after pod start. Add waitForBlockDevice() with exponential backoff (500ms→30s, 10 attempts) that retries ENXIO/ENOENT/EIO and fails fast on permanent errors like EPERM.

https://redhat.atlassian.net/browse/MTV-6393

### DIFF
--- a/pkg/virt-v2v/conversion/disk.go
+++ b/pkg/virt-v2v/conversion/disk.go
@@ -1,10 +1,14 @@
 package conversion
 
 import (
+	"errors"
 	"fmt"
+	"os"
 	"path/filepath"
 	"regexp"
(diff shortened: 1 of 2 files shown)


Hazanel:
Hey @coderabbitai - ptal and tell me if you can think of a better way or it's good enough

Hazanel:
Thanks @coderabbitai - I addressed your suggestions!

yaacov (approved):

PR #8221, opened 2026-08-20
MTV-6234 | Nutanix migrations

mrnold:
Following on from #7657, create an actual workflow that gets a VM migrated from Nutanix. This creates a temporary Image from a VM disk on the Nutanix side, and transfers it over an HTTP DataVolume importer.

This is the least fancy option in that it copies every byte from the source whether or not it was actually used, but it should work whether the provider is Prism Central or Prism Element. Does not run a conversion yet as existing VirtIO drivers seem to work, this will be revisited in a future pull request.

Reference: [MTV-6234](https://redhat.atlassian.net/browse/MTV-6234)

Resolves: MTV-6234

[MTV-6234]: https://redhat.atlassian.net/browse/MTV-6234?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mrnold:
> ## [Quality Gate Failed](https://sonarcloud.io/dashboard?id=kubev2v_forklift&pullRequest=8221) **Quality Gate failed**
> 
> Failed conditions  [D Security Rating on New Code](https://sonarcloud.io/dashboard?id=kubev2v_forklift&pullRequest=8221) (required ≥ A)
> 
> [See analysis details on SonarQube Cloud](https://sonarcloud.io/dashboard?id=kubev2v_forklift&pullRequest=8221)
> 
>  Catch issues before they fail your Quality Gate with our IDE extension  [SonarQube for IDE](https://www.sonarsource.com/products/sonarlint/features/connected-mode/?referrer=pull-request)

Both these cases exist to handle the "InsecureSkipVerify", which needs to be intentionally enabled. Certificates are checked as requested when this option is turned off.

pkg/lib/client/nutanix/client.go
```diff
@@ -0,0 +1,380 @@
...
+	if r.client != nil {
+		return http.StatusOK, nil
+	}
+
+	if r.Log == nil {
+		r.Log = logging.WithName("client|nutanix")
+	}
+
+	if base.GetInsecureSkipVerifyFlag(r.Secret) {
+		tlsClientConfig = &tls.Config{InsecureSkipVerify: true}
```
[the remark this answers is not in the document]
mrnold: InsecureSkipVerify option condition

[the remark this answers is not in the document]
mrnold: InsecureSkipVerify option condition

pkg/controller/plan/adapter/nutanix/builder.go
```diff
@@ -1,41 +1,699 @@
...
+					Bus: diskBus(disk.AdapterType),
+				},
+			},
+			Serial: disk.UUID,
+		}
+		if disk.DeviceIndex == bootDiskIndex {
+			var bootOrder uint = 1
+			kDisk.BootOrder = &bootOrder
+		}
+		kDisks = append(kDisks, kDisk)
```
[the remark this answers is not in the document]
jonner: see my commit here for an approach to fixing this in a way that paves the way for libvirt domain xml generation: https://github.com/jonner/forklift/commit/aab0d50054f2a169458646258a4c32e71a57b8c3

pkg/controller/plan/adapter/nutanix/builder.go
```diff
@@ -1,41 +1,699 @@
...
+	baseURL := strings.TrimRight(r.Source.Provider.Spec.URL, "/")
+
+	for _, disk := range vm.Disks {
+		if disk.IsCdrom {
+			continue
+		}
+		destination, mapped := storageMap[disk.StorageContainerUUID]
+		if !mapped {
+			continue
+		}
```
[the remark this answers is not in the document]
mrnold: @jonner Is this preferable to fetching order from the inventory? I tried to do this in the latest change. I am curious where that adapter priority list comes from too, doesn't that still risk choosing the wrong boot device if a VM is configured just right?

mrnold:
@coderabbitai Have your concerns been addressed?

pkg/controller/plan/adapter/nutanix/builder.go
```diff
@@ -1,41 +1,802 @@
...
+	dv.Spec = dvSpec
+	if dv.Annotations == nil {
+		dv.Annotations = make(map[string]string)
+	}
+	dv.Annotations[planbase.AnnDiskSource] = disk.UUID
+	// Nutanix has no conversion pod to act as the first consumer of a
+	// WaitForFirstConsumer storage class during a cold migration to the
+	// local cluster (unlike vSphere), so request immediate binding here --
+	// otherwise the PVC, and thus the import, would never start.
+	dv.Annotations[planbase.AnnBindImmediate] = "true"
```
jonner, line 773: Just flagging this as something that may need to change (according to the comment) when we add a conversion step for installing virtio drivers, etc.

pkg/controller/plan/adapter/nutanix/builder.go
```diff
@@ -1,41 +1,802 @@
...
+func (r *Builder) ensureDownloadCookieSecret(vmRef ref.Ref, diskUUID, cookie string, labels map[string]string) (*core.Secret, error) {
+	secretLabels := map[string]string{
+		labelDownloadCookie:    "true",
+		planbase.AnnDiskSource: diskUUID,
+	}
+	for k, v := range labels {
+		secretLabels[k] = v
+	}
+
+	header := cookieHeaderValue(cookie)
```
jonner, line 538: `cookie` here is the raw value received from the nutanix server. Should we do any validation of this value before passing it on to CDI?

pkg/controller/plan/kubevirt.go
```diff
@@ -1911,10 +1911,34 @@ func (r *KubeVirt) EnsureDataVolumes(vm *plan.VMStatus, dataVolumes []cdi.DataVo
...
+// Nutanix is the only implementation today; see nutanix.Builder.
+func (r *KubeVirt) adoptDownloadCookieSecretOwner(dv *cdi.DataVolume) error {
+	type downloadCookieSecretAdopter interface {
+		AdoptDownloadCookieSecretOwner(dv *cdi.DataVolume) error
+	}
+	adopter, ok := r.Builder.(downloadCookieSecretAdopter)
+	if !ok {
+		return nil
+	}
+	return adopter.AdoptDownloadCookieSecretOwner(dv)
```
jonner: In terms of future code readability, I don't really like this pattern of putting an interface definition inside of a method and then implementing that interface in a single provider. In general, I much prefer the convention of adding a new method to the Builder interface and providing an empty implementation in the providers that don't support it. That makes it much easier to reason about what the capabilities of the different providers are without having to search for optional interfaces hidden around the codebase.

jonner (commented):
Mostly looks fine. just a few comments after looking through the PR.

pkg/controller/plan/kubevirt.go
mrnold: Okay, this is moved to the Builder now.

pkg/controller/plan/adapter/nutanix/builder.go
mrnold: Added a TODO

pkg/controller/plan/adapter/nutanix/builder.go
mrnold: I don't want to add much specific validation to it, because it comes from Nutanix and gets passed straight back to Nutanix. But there are some extra minor checks for length and invalid characters now, if that's good enough.

jonner: yeah, I wasn't sure either. I think the additions are fine.

mnecas (approved):

PR #8250, opened 2026-08-21
MTV-6511 | Split Custom Field Definitions from VM table in inventory to dedicated table

tylergmuir:
An issue was found in using forklift to move VMs from vSphere where there was a large amount of custom value definitions. It resulted in a very large inventory database as well as very high memory consumption of the inventory container when the plans were reconciling.

This was due to the entire custom value definition being added to each VM in the VM table. In the test case there were over 1200 custom value definitions where all were used, but most VMs only used a few each. In that case about 95% of the entire inventory DB was the custom value definitions.

Originally, I thought to make the changes entirely within the inventory so the consumer did not need to be changed where at the time of the data being committed to the database, the custom value definitions were for each VM would be pruned to only the values that VM had defined. But when an incremental change is made where a value is added to a VM, the definitions are not resent and it would have resulted in annotations being excluded until the next time the pod was restarted.

Instead, I opted to move the custom value definitions into a separate database table and endpoint and have the consumer merge the data together.

This allows for the data in the DB to be reduced from n (VMs) to 1 as well as the amount of data passed by each of the calls to the inventory will be reduced by the size of the custom value definition at the expense of 1 additional call per VM to collect the custom value definitions. 

Ref: https://redhat.atlassian.net/browse/MTV-6511

### DIFF
--- a/pkg/controller/plan/adapter/vsphere/builder.go
+++ b/pkg/controller/plan/adapter/vsphere/builder.go
@@ -2880,32 +2880,44 @@ func (r *Builder) SourceVMLabelsAndAnnotations(vmRef ref.Ref, tagMapping *api.Ta
 
 	// Custom attributes to annotations
 	annotationOriginalKeys := make(map[string]string)
+	customDefs := []model.CustomFieldDef{}
+	err = r.Source.Inventory.List(&customDefs, web.Param{
+		Key:   web.DetailParam,
+		Value: "all",
+	})
+	if err != nil {
+		err = liberr.Wrap(err, "custom field definitions")
+		return
+	}
+	customDefByKey := make(map[int32]model.CustomFieldDef, len(customDefs))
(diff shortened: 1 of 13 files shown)


yaacov:
@tylergmuir hi, thank you for the pull request ❤️ 

to pass `ci build` tests:
run `make update-manifests` 
make sure `make lint` and `make test` pass

to pass `validate commit message` tests:
commit msg rules:
1 commit per pull request ( squash commits )
msg should include title, description, Resolves, and a signiture:
```
Split Custom Field Definitions from VM table in inventory to dedicated table

An issue was found in using forklift ...

Resolves: MTV-6511

Signed-off-by: ...
```

also need rebase

tylergmuir:
@yaacov I think I have all those issues fixed now if you want to have the checks run again.

yaacov (approved):

PR #8363, opened 2026-08-27
MTV-5946 | Install MTV on remote clusters

Kyl-Bempah:
Adds validation on remote clusters to check for Forklift CRD

resolves: MTV-5946


Kyl-Bempah:
@mnecas Does this match what we previously discussed, I created it as a validation, although the ticket mentioned this as an install, which we could do, via OLM but that then goes into the other discussion we had regarding OLMv1 migration, that would require a bit of RBAC upheaval in our install process since it needs a serviceaccount.

mnecas (approved):

PR #8374, opened 2026-08-31
MTV-6551| RFE: Add nfc support to cold migration

mnecas:
The check still runs in virt-v2v so it does not assume the init image has a shell. An NFC image that copies nbdkit-nfc-plugin.so into /opt now passes the validation job.

Resolves: MTV-6551


pkg/controller/plan/validation.go
```diff
@@ -2145,7 +2145,9 @@ func createVddkCheckJob(plan *api.Plan) *batchv1.Job {
 								},
 							},
 							VolumeMounts: []core.VolumeMount{mount},
-							Command:      []string{"file", "-E", "/opt/vmware-vix-disklib-distrib/lib64/libvixDiskLib.so"},
+							Command: []string{"/bin/sh", "-c",
+								"file -E /opt/vmware-vix-disklib-distrib/lib64/libvixDiskLib.so" +
```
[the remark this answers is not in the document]
yaacov: we require the image to ship with /bin/sh

yaacov (approved):

pkg/virt-v2v/conversion/conversion_test.go
```diff
@@ -1217,6 +1261,33 @@ var _ = Describe("Conversion", func() {
...
+			appConfig.Fingerprint = "AA:BB:CC"
+			appConfig.NfcPluginPath = plugin.Name()
+
+			mockCommandBuilder.EXPECT().AddArg("-i", "libvirt").Return(mockCommandBuilder)
+			mockCommandBuilder.EXPECT().AddArg("-ic", appConfig.LibvirtUrl).Return(mockCommandBuilder)
+			mockCommandBuilder.EXPECT().AddArg("-ip", appConfig.SecretKey).Return(mockCommandBuilder)
+			mockCommandBuilder.EXPECT().AddArg("--hostname", appConfig.HostName).Return(mockCommandBuilder)
+			mockCommandBuilder.EXPECT().AddArg("--root", "first").Return(mockCommandBuilder)
+			mockCommandBuilder.EXPECT().AddArg("-it", "nfc").Return(mockCommandBuilder)
+			mockCommandBuilder.EXPECT().AddArg("-io", "nfc-thumbprint=AA:BB:CC").Return(mockCommandBuilder)
```
rwmjones, line 1283: The plugin doesn't actually require the thumbprint, but it doesn't hurt to have it.

pkg/controller/plan/validation.go
```diff
@@ -2175,7 +2145,10 @@ func createVddkCheckJob(plan *api.Plan) *batchv1.Job {
...
+									" || file -E /opt/nbdkit-nfc-plugin.so" +
```
rwmjones, line 2180: I didn't follow the logic of the `/opt/nbdkit-nfc-plugin.so` path.

For running nbdkit directly it is fine to use any path you like (`nbdkit /opt/nbdkit-nfc-plugin.so ...` will work).

But virt-v2v runs `nbdkit nfc ..` so the plugin has to be in `/usr/lib64/nbdkit/plugins/nbdkit-nfc-plugin.so` and any other path will not work.

mnecas: we need to use the `/opt` as destination directory for the copy, but i have added symlink btween the `/opt/nbdkit-nfc-plugin.so`  and `/usr/lib64/nbdkit/plugins/nbdkit-nfc-plugin.so`

pkg/virt-v2v/conversion/conversion_test.go
mnecas: ack

PR #8382, opened 2026-09-01
MTV-6620 | Add plan flag to defer SELinux relabeling to guest boot

yaacov:
Add `selinuxRelabelAtBoot` and `selinuxRelabelExclude` to Plan spec with
per-VM overrides. The controller resolves the effective values per VM
and passes them to the conversion pod as V2V_selinuxRelabelAtBoot and
V2V_selinuxRelabelExclude (JSON array). The virt-v2v wrapper forwards
them as `--selinux-relabel-at-boot` and `--selinux-relabel-exclude`.

Usage examples:
```yaml
# Defer relabeling until first guest boot
  spec:
    selinuxRelabelAtBoot: true
    vms:
      - id: vm-123
        name: rhel-vm
  ```
  
  ```yaml
  # Exclude directories from relabeling during conversion
  spec:
    selinuxRelabelExclude:
      - /foo
      - /bar
    vms:
      - id: vm-123
        name: rhel-vm
  ```
  
  ```yaml
  # Per-VM overrides
  spec:
    selinuxRelabelAtBoot: true
    selinuxRelabelExclude:
      - /foo
    vms:
      - id: vm-456
        name: rhel-vm-fast
        selinuxRelabelAtBoot: false
        selinuxRelabelExclude:
          - /baz
```

Ref: https://redhat.atlassian.net/browse/MTV-6620
Resolves: MTV-6620

### DIFF
--- a/cmd/vsphere-copy-offload-populator/vendor/github.com/kubev2v/forklift/pkg/apis/forklift/v1beta1/plan.go
+++ b/cmd/vsphere-copy-offload-populator/vendor/github.com/kubev2v/forklift/pkg/apis/forklift/v1beta1/plan.go
@@ -281,6 +281,16 @@ type PlanSpec struct {
 	// Can be overridden per VM via spec.vms[].scsiReservation.
 	// +optional
 	SCSIReservation bool `json:"scsiReservation,omitempty"`
+	// SelinuxRelabelAtBoot defers SELinux relabeling until the guest's first boot after conversion.
+	// Passed to virt-v2v as --selinux-relabel-at-boot. Can be overridden per VM via spec.vms[].selinuxRelabelAtBoot.
+	// +optional
+	SelinuxRelabelAtBoot bool `json:"selinuxRelabelAtBoot,omitempty"`
+	// SelinuxRelabelExclude lists guest directories excluded from SELinux relabeling during conversion.
+	// Passed to virt-v2v as --selinux-relabel-exclude. Can be overridden per VM via spec.vms[].selinuxRelabelExclude.
+	// Use carefully: changes made inside excluded directories during customization may retain incorrect
+	// SELinux labels and cause failures later. Only exclude directories that are known not to need relabeling.
+	// +optional
+	SelinuxRelabelExclude []string `json:"selinuxRelabelExclude,omitempty"`
 	// DeleteGuestConversionPod determines if the guest conversion pod should be deleted after successful migration.
 	// Note:
 	//   - If this option is enabled and migration succeeds then the pod will get deleted. However the VM could still not boot and the virt-v2v logs, with additional information, will be deleted alongside guest conversion pod.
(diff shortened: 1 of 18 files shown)


mnecas (changes requested):
please add also `--selinux-relabel-exclude` option
```
--selinux-relabel-exclude DIR
Exclude directories from being relabelled.

This advanced option lets you list directories in the guest which should not be relabelled, even when SELinux relabelling is enabled. Use this carefully, as any changes that are made inside these directories during customization will have incorrect SELinux labels, leading to potential failures later, so you must be sure that the directories do not need relabelling.

If in doubt, do not use this option.

You can pass the option multiple times, eg. --selinux-relabel-exclude=/foo --selinux-relabel-exclude=/bar
```
https://libguestfs.org/virt-v2v.1.html

yaacov:
> please add also --selinux-relabel-exclude option

Added the option, note: `--selinux-relabel-exclude` option can make a vm fail in unexpected ways, added the option with a comment about the side-affects of using it

yaacov:
@mnecas hi, added the new flag.
also added to the jra ticket tests and run logs

mnecas (approved):
lgtm

akalenyu:
Folks, cautious suggestion: when do we **not** want to defer? shouldn't it be default to defer on k8s?
Related - https://kubernetes.io/blog/2026/04/22/breaking-changes-in-selinux-volume-labeling/

rwmjones:
No, we should *not* defer usually, only in exceptional cases.  Deferring relabelling means you don't get to see any problems during conversion.

The manual already says this:
https://libguestfs.org/virt-v2v.1.html

```
--selinux-relabel-at-boot

    This advanced option lets you defer SELinux relabelling until the first time the guest boots after conversion.

    The advantage of doing this is that conversion is quicker and uses less memory.

    The disadvantages are:

        The guest will take much longer to boot up and become ready the first time it boots.

        The guest will reboot at least once.

        SELinux relabelling problems cannot be detected during conversion.

    If in doubt, do not use this option.
--selinux-relabel-exclude DIR

    Exclude directories from being relabelled.

    This advanced option lets you list directories in the guest which should not be relabelled, even when SELinux relabelling is enabled. Use this carefully, as any changes that are made inside these directories during customization will have incorrect SELinux labels, leading to potential failures later, so you must be sure that the directories do not need relabelling.

    If in doubt, do not use this option.

    You can pass the option multiple times, eg. --selinux-relabel-exclude=/foo --selinux-relabel-exclude=/bar
```

akalenyu:
> No, we should _not_ defer usually, only in exceptional cases. Deferring relabelling means you don't get to see any problems during conversion.
> 
> The manual already says this: https://libguestfs.org/virt-v2v.1.html
> 
> ```
> --selinux-relabel-at-boot
> 
>     This advanced option lets you defer SELinux relabelling until the first time the guest boots after conversion.
> 
>     The advantage of doing this is that conversion is quicker and uses less memory.
> 
>     The disadvantages are:
> 
>         The guest will take much longer to boot up and become ready the first time it boots.
> 
>         The guest will reboot at least once.
> 
>         SELinux relabelling problems cannot be detected during conversion.
> 
>     If in doubt, do not use this option.
> --selinux-relabel-exclude DIR
> 
>     Exclude directories from being relabelled.
> 
>     This advanced option lets you list directories in the guest which should not be relabelled, even when SELinux relabelling is enabled. Use this carefully, as any changes that are made inside these directories during customization will have incorrect SELinux labels, leading to potential failures later, so you must be sure that the directories do not need relabelling.
> 
>     If in doubt, do not use this option.
> 
>     You can pass the option multiple times, eg. --selinux-relabel-exclude=/foo --selinux-relabel-exclude=/bar
> ```

I confused the guest relabeling with k8s volume handling

PR #8388, opened 2026-09-02
[release-2.12] MTV-6511 | Fixes performance regression from #8250

yaacov:
Manual Backport: https://github.com/kubev2v/forklift/pull/8367

Replaces failed automatic backport - https://github.com/kubev2v/forklift/pull/8385

With the PR https://github.com/kubev2v/forklift/pull/8250 there was a performance regression that caused the initial inventory parsing to take 3-4 times longer due to heavier reliance on the database to process updates instead of only inserts.

This change caches the custom def fields into memory in the inventory application and then does a compare against that before attempting the database insert/update. This is especially impactful on the initial startup which attempts to insert/update the same custom def fields for each VM.

Resolves: MTV-6511

### DIFF
--- a/pkg/controller/plan/adapter/vsphere/builder.go
+++ b/pkg/controller/plan/adapter/vsphere/builder.go
@@ -2331,40 +2331,6 @@ func (r *Builder) getPVCNameTemplate(vm *model.VM) string {
 	return ""
 }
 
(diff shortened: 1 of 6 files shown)


yaacov:
closing, wrong target branch

PR #8393, opened 2026-09-02
MTV-6458 | Fall back to system CA for wrong VMExport cert

mnecas:
Probe the export proxy TLS endpoint before setting certConfigMap on OCP cold-migration DataVolumes. Omit the cert when VMExport's CA does not verify but system trust does.

Ref: https://redhat.atlassian.net/browse/MTV-6458
Resolves: MTV-6458

### DIFF
--- a/pkg/controller/plan/adapter/ocp/builder.go
+++ b/pkg/controller/plan/adapter/ocp/builder.go
@@ -20,6 +20,7 @@ import (
 	ocpclient "github.com/kubev2v/forklift/pkg/lib/client/openshift"
 	liberr "github.com/kubev2v/forklift/pkg/lib/error"
 	libitr "github.com/kubev2v/forklift/pkg/lib/itinerary"
+	"github.com/kubev2v/forklift/pkg/lib/logging"
 	core "k8s.io/api/core/v1"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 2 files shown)


pkg/controller/plan/adapter/ocp/builder.go
```diff
@@ -744,3 +759,53 @@ func (r *Builder) setPVCNameFromTemplate(objectMeta *metav1.ObjectMeta, vmRef re
...
+
+const exportTLSTimeout = 10 * time.Second
+
+// tlsCertForExport returns providedCert if it verifies exportURL, "" if system CA works instead, or an error if neither works.
+func tlsCertForExport(exportURL, providedCert string) (string, error) {
+	if providedCert == "" {
+		return "", nil
+	}
+	if exportURL == "" {
+		return providedCert, nil
```
awels: If the exportURL is blank that is probably a completely different issue. Since you have nothing to connect to. I would expect that to return an error with message indicating a missing URL.

pkg/controller/plan/adapter/ocp/builder.go
```diff
@@ -744,3 +759,53 @@ func (r *Builder) setPVCNameFromTemplate(objectMeta *metav1.ObjectMeta, vmRef re
...
+}
+
+func verifyExportTLS(exportURL, caPEM string) error {
+	u, err := url.Parse(exportURL)
+	if err != nil {
+		return err
+	}
+	addr := u.Host
+	if _, _, err := net.SplitHostPort(addr); err != nil {
+		addr = net.JoinHostPort(addr, "443")
```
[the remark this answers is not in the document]
awels: maybe to make the bot happy you can wirte something like this:

```golang
port := "443"
if u.port != "" {
  port = u.port
}
addr = net.JoinHostPort(addr, port)
```
This way if the Route/Ingress is on a different port for some reason, the `url.Parse` should capture it, and if it is not specified we fall back to 443.

pkg/controller/plan/adapter/ocp/builder.go
```diff
@@ -744,3 +759,53 @@ func (r *Builder) setPVCNameFromTemplate(objectMeta *metav1.ObjectMeta, vmRef re
...
+		if err != nil {
+			roots = x509.NewCertPool()
+		}
+	}
+	conn, err := tls.DialWithDialer(&net.Dialer{Timeout: exportTLSTimeout}, "tcp", addr, &tls.Config{
+		ServerName: u.Hostname(),
+		RootCAs:    roots,
+	})
+	if conn != nil {
+		conn.Close()
```
awels: Just check it like this or something 
```golang
  closeErr := conn.Close()
  if closeErr != nil {
     // log/print the error message so we have at least some trace of it. I don't think we want to return that particular error, can't do anything with it.
  }
```

awels (commented):
Looks good some minor nitpicks.

awels (approved):

awels (approved):

PR #8459, opened 2026-09-07
MTV-6551 | Mount nbdkit NFC plugin via VDDK sidecar subPath

mnecas:
When VDDKImage is set, overlay nbdkit-nfc-plugin.so from the vddk
emptyDir onto /usr/lib64/nbdkit/plugins/ so virt-v2v can use -it nfc
without replacing the image's built-in nbdkit plugins. Remove the
virt-v2v entrypoint symlink; non-root qemu cannot create it.

Ref: https://redhat.atlassian.net/browse/MTV-6551
Resolves: MTV-6551


solenoci (approved):
lgtm

PR #8561, opened 2026-09-11
MTV-5319 | Cache provider inventory during plan validate and begin

gwencasey96:
## Summary

- Add a per-reconcile `CachingClient` for provider inventory `Find`, `VM`, and `List` calls.
- Wrap source and destination inventory during plan validation and migration `begin()`.
- Log validation/begin timing and cache hit stats to profile cold migration start latency.

**Depends on:** #8563 (VM validation loop trim). Merge #8563 first, then rebase this branch onto `main` so only the cache commit remains in the PR diff.

## Test plan

- [ ] Run plan controller and `caching_client` unit tests
- [ ] Start a cold migration with many VMs and confirm reduced inventory API traffic in logs

Ref: https://redhat.atlassian.net/browse/MTV-5319
Resolves: MTV-5319

### DIFF
--- a/pkg/controller/plan/migration.go
+++ b/pkg/controller/plan/migration.go
@@ -215,6 +215,7 @@ func (r *Migration) init() (err error) {
 
 // Begin the migration.
 func (r *Migration) begin() (err error) {
+	start := time.Now()
 	snapshot := r.Plan.Status.Migration.ActiveSnapshot()
 	if snapshot.HasAnyCondition(api.ConditionExecuting, api.ConditionSucceeded, api.ConditionFailed, api.ConditionCanceled) {
 		return
@@ -264,6 +265,7 @@ func (r *Migration) begin() (err error) {
 			Message:  "The plan is EXECUTING.",
 			Durable:  true,
 		})
+	setupStart := time.Now()
 	err = r.kubevirt.EnsureNamespace()
 	if err != nil {
 		err = liberr.Wrap(err)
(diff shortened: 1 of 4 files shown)


pkg/controller/plan/migration.go
```diff
@@ -215,6 +215,7 @@ func (r *Migration) init() (err error) {
 
 // Begin the migration.
 func (r *Migration) begin() (err error) {
+	start := time.Now()
```
mrnold, line 218: Is this profiling stuff temporary? I guess it is harmless but it seems pretty visible for what is hopefully a bug that stays fixed long-term.

pkg/controller/plan/validation.go
```diff
@@ -1024,15 +1064,7 @@ func (r *Reconciler) validateVM(plan *api.Plan, ctx *plancontext.Context) error
 			continue
 		}
 		// Source.
-		provider := plan.Provider.Source
```
mrnold, line 1027: ✅

pkg/controller/plan/validation.go
```diff
@@ -1113,15 +1145,6 @@ func (r *Reconciler) validateVM(plan *api.Plan, ctx *plancontext.Context) error
 				}
 			}
 		}
-		pAdapter, err := adapter.New(provider)
```
mrnold, line 1116: ✅ Actually kind of a lot of work going on behind plancontext.New, this should be a big help.

pkg/controller/provider/web/caching_client.go
```diff
@@ -0,0 +1,202 @@
+package web
```
mrnold, line 1: Do we really need all this? I would rather not maintain our own caching client if at all possible, but I guess it would be okay if it's really the thing that makes it all work. Like if you tested with only the change to move plancontext.New out of the VM loop above, and still saw bad slowdowns. But even then I would prefer not to make our own from scratch if there are any better options.

PR #8565, opened 2026-09-14
MTV-6615 | Pre-inspect vSphere source before cold virt-v2v conversion

yaacov:
Refactor vSphere cold migration to run virt-v2v-inspector against the source guest (-i libvirt) before conversion, then run virt-v2v with inline customize flags (--run/--firstboot/--upload). This drops the post-conversion virt-v2v-inspector and virt-customize passes, reducing libguestfs sessions from three to two.

Ref: https://redhat.atlassian.net/browse/MTV-6615
Resolves: MTV-6615


yaacov:
@coderabbitai hi, can you re-review, and close your old comments if already addressed

PR #8566, opened 2026-09-14
Skip Route lookups during cleanup on clusters without the Route API

aaaaaaaalex:
The disabled-feature cleanup loops iterate forklift_resources, which includes Route. cleanup.yml passes the kind to k8s_info with no api_version, so on a cluster without route.openshift.io the dynamic client searches every group and __search recurses until it dies with RecursionError. Ansible surfaces only "MODULE FAILURE: No start of json char found", and although cleanup.yml rescues the task, the operator still records the module failure and pins the ForkliftController at Failure=True.

The install itself completes, so the effect is that the CR never reaches Successful on Kubernetes:  "kubectl wait --for=condition=Successful" always times out, and Failure/Successful stop being usable health signals there.

Filter Route out of the cleanup kinds when the API group is absent, using the same api_groups detection the role already performs.

Cherry-picked from tigera/forklift#2.


mrnold (approved):
Looks okay to me

PR #8572, opened 2026-09-14
chore(deps): CVE updates for 2.10.8

Kyl-Bempah:
Update dependencies for security compliance


solenoci:
Hi @Kyl-Bempah, please also update the tekton task versions as some are deprecated

PR #8659, opened 2026-09-16
MTV-6743 | Improve Nutanix certificate handling

mrnold:
Issue: Prism Central redirects to Prism Element for actual download endpoints, so a provider configured with just the Prism Central certificate will fail in CDI's HTTP importer. Also, insecureSkipVerify won't work on older CNV releases because CDI only just recently added it.

Fix: for insecureSkipVerify, automatically fetch a certificate from the endpoint and have the importer trust it, similar to what oVirt does. This does not do anything really different for insecureSkipVerify=false, but maybe it would be helpful to have a separate feature that collects all the Prism Element certificates and presents them for approval.

Ref: https://redhat.atlassian.net/browse/MTV-6743

Resolves: MTV-6743


mrnold:
@coderabbitai were your concerns addressed?

pkg/controller/plan/adapter/nutanix/builder.go
```diff
@@ -109,24 +97,200 @@ func setCDICACerts(object *core.ConfigMap, cacert []byte) {
...
+
+// importCertTLSHost returns the host portion of an HTTPS import URL.
+func importCertTLSHost(rawURL string) (string, error) {
+	parsed, err := url.Parse(rawURL)
+	if err != nil {
+		return "", liberr.Wrap(err, "failed to parse import URL", "url", rawURL)
+	}
+	if parsed.Scheme == "" {
+		parsed.Scheme = "https"
+	}
```
jonner: this doesn't seem to be used for anything

jonner (approved):
Looks generally good to me.

pkg/controller/plan/adapter/nutanix/builder.go
mrnold: It sure wasn't, removed

PR #8702, opened 2026-09-21
MTV-6869 | Add array-offload copy support for migrating NFS VMs

sahana-everpure:
Adds an array-side file-copy path for NFS-backed source disks, offloading the copy to the storage array.

This is a proof of concept covering only the happy path; error handling and other disk configuration scenarios will be addressed in follow-up work.

Resolves: MTV-6869


mnecas (commented):
Thanks for the PR!
Forklift has some integration with NetApp Shift which copies the disks via NFS. In those cases we add specific annotations to the PVC requesting specific volume from the CSI and they provision it for us. Do you think we could reuse similar design here aswell?

sahana-everpure:
> Thanks for the PR! Forklift has some integration with NetApp Shift which copies the disks via NFS. In those cases we add specific annotations to the PVC requesting specific volume from the CSI and they provision it for us. Do you think we could reuse similar design here aswell?

Portworx CSI supports cloning only for block PVCs, not for files. There's an annotation-based import (pure-volume-name), but it only adopts an existing directory/filesystem in place — it doesn't actually copy data like Trident does for NetApp Shift. So we don't have anything like that in PX-CSI to follow this design.

